PR #13, opened 2020-08-31
install: Add initial version

patrickelectric:
Configures a fresh image with companion-docker.
After it reboots, companion-docker will star automatically. 

`Curl` is necessary to run this script.
The following lines need to run as **root**.
How to test (if the repository is public):
```sh
# These variables are not necessary since the default value now is master
# and my remote is necessary since it's not merged.
export VERSION=install
export REMOTE=https://raw.githubusercontent.com/patrickelectric/companion-docker
curl -fsSL https://raw.githubusercontent.com/patrickelectric/companion-docker/install/install/install.sh | sh
```
Changes are necessary for non public repositories. 

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Set desired version to be installed
+VERSION="${VERSION:-master}"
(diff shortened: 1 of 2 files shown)


patrickelectric:
For previous comments, check #12

jaxxzer:
unblock wifi

PR #19, opened 2020-09-01
core: services: ardupilot_manager: Add FirmwareDownload functionality

patrickelectric:
pytest will by default test the class for us

This also moves mavlink_proxy library from the lib folder, this avoids any include problem with pytest.


core/services/ardupilot_manager/firmware_download/lib/FirmwareDownload.py
```diff
@@ -0,0 +1,206 @@
...
+
+class FirmwareDownload:
+    _manifest_remote = "https://firmware.ardupilot.org/manifest.json.gz"
+    _supported_firmware_format = "apj"
+
+    def __init__(self):
+        self._content = {}
+
+    @staticmethod
+    def _generate_random_filename(length=16) -> pathlib.Path:
```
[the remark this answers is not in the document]
patrickelectric: It's safer to create our own random file to avoid any conflict with others programs and services that are running, maybe in the future a service will download a different manifest file and etc, it also does not hurt to be safer. 
With that, we also don't need to track the name of the files or identify the names in the code, we generate names for them and we track this name.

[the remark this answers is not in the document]
Williangalvani: Do we even need to download these to the filesystem? At least for the manifest, you could just:
```
from urllib.request import urlopen

with urlopen(URL) as resource:
    versions = json.load(resource)
```

The firmware I assume is going to be loaded from the filesystem when flashing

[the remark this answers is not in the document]
patrickelectric: Done

patrickelectric:
@Williangalvani corrections done.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -0,0 +1,206 @@
...
+    def _generate_random_filename(length: int = 16) -> pathlib.Path:
+        """Generate a random name for a temporary file.
+
+        Args:
+            length (int, optional): Size of the name. Defaults to 16.
+
+        Returns:
+            pathlib.Path: Path of the temporary file.
+        """
+        import random
```
Williangalvani, line 38: what is the reasoning for these local imports? this seems pretty non-standard.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -0,0 +1,206 @@
...
+            List[Dict[str, Any]]: A list of firmware items that match the arguments.
+        """
+        if not self._manifest and not self.download_manifest():
+            return []
+
+        found_version_item = []
+
+        # Make sure that the item matches all args value
+        for item in self._manifest["firmware"]:
+            for key in args.keys():
```
Williangalvani, line 128:
```suggestion
            for key in args:
```
FirmwareDownload.py:128:23: C0201: Consider iterating the dictionary directly instead of calling .keys() (consider-iterating-dictionary)

core/services/ardupilot_manager/firmware_download/test_FirmwareDownload.py
```diff
@@ -0,0 +1,36 @@
+from firmware_download.FirmwareDownload import FirmwareDownload, Vehicle
+
+
+def test_static() -> None:
+    downloaded_file = FirmwareDownload._download(FirmwareDownload._manifest_remote)
+    assert downloaded_file.exists(), "Download file does not exist."
```
Williangalvani, line 6: this is interesting....
test_FirmwareDownload.py:6:11: E1101: Instance of 'PurePath' has no 'exists' member (no-member)
test_FirmwareDownload.py:9:11: E1101: Instance of 'PurePath' has no 'stat' member (no-member)

If this is true, shouldnt the test be failing?

Williangalvani: neverming, pylint is insane

patrickelectric: image

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
patrickelectric: It has the imports that are only used inside this scope to track the imports easily, otherwise it makes harder to detect when a include is not being used. But since we have pylint, this is not necessary anymore.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
patrickelectric: Cool

PR #20, opened 2020-09-01
Use python black code formatter  

patrickelectric:


Williangalvani:
what about a  `--line-length 120` ?

patrickelectric:
@Williangalvani I just updated

PR #21, opened 2020-09-01
Move mypy to strict

patrickelectric:


core/services/ardupilot_manager/mavlink_proxy/lib/AbstractRouter.py
```diff
@@ -10,9 +10,9 @@
 
 
 class AbstractRouter(metaclass=abc.ABCMeta):
-    def __init__(self):
+    def __init__(self) -> None:
         self._endpoints: List[Endpoint] = []
-        self._subprocess: Optional[subprocess.Popen] = None
+        self._subprocess: Optional[Any] = None
```
Williangalvani, line 15: it seems to be working fine here with subprocess.Popen

patrickelectric: Are you running in the correct path ?
`core/services/ardupilot_manager/mavlink_proxy/lib/AbstractRouter.py:15: error: Missing type parameters for generic type "Popen"
`

Williangalvani: oops, I had the right file but I missed the --strict :man_facepalming:

PR #24, opened 2020-09-02
pylint: Update configuration file

patrickelectric:
- Add new test for dictionaries
- Add description why no-member should be disabled: https://github.com/bluerobotics/companion-docker/pull/19#discussion_r482017356

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,3 +1,36 @@
+[FORMAT]
+max-line-length=120
+
 [MESSAGES CONTROL]
-disable=all
-enable=unused-import
+enable=all
+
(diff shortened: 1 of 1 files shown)


.pylintrc
```diff
@@ -1,3 +1,5 @@
 [MESSAGES CONTROL]
-disable=all
+disable=all,
+    no-member, # Some members are created in runtime, pylint fail to detect such cases
```
Williangalvani, line 3:
```suggestion
    no-member, # Some members are created in runtime, pylint fails to detect such cases
```

patrickelectric: Done.

PR #27, opened 2020-09-09
core: Add ardupilot_manager basic funcionality 

patrickelectric:
- fight_controller_detector: Remove lib folder
    - unnecessary
- mavlink_proxy: Add manager class
    - To abstract the usage of the library for the ardupilot manager and move to async in the near future
- pylintrc: Remove test for locally-disabled
    - To allow us to disable some tests per file, like including files to have the register of such for subclass call
- mavlink-router: Install it
- ardupilot_manager: Add basic service functionality

Image: patrickelectric/companion-core
Only works with pixhawk


core/services/ardupilot_manager/mavlink_proxy/Manager.py
```diff
@@ -0,0 +1,53 @@
...
+            raise RuntimeError("No available interface found.")
+
+    @staticmethod
+    def available_interfaces() -> List[Type[AbstractRouter]]:
+        return AbstractRouter.available_interfaces()
+
+    def use(self, interface: AbstractRouter) -> None:
+        self.tool = interface
+
+    def add_endpoint(self, endpoints: List[Endpoint]) -> None:
```
Williangalvani, line 26:
```suggestion
    def add_endpoints(self, endpoints: List[Endpoint]) -> None:
```

core/services/ardupilot_manager/mavlink_proxy/main.py
```diff
@@ -1,19 +1,19 @@
 #!/bin/env python
 import argparse
+import sys
 import time
+from pathlib import Path
+
+# Import local library
+sys.path.append(str(Path(__file__).absolute().parent.parent))
```
Williangalvani, line 8: if all things you are importing are in the mavlink_proxy folder, you shouldn't need this...

Williangalvani (changes requested):

core/services/ardupilot_manager/mavlink_proxy/main.py
patrickelectric: The problem is that besides the main.py, mavlink_proxy is a library, if you want to use it from another main, like ardupilot_manager, it'll fail to include itself.
Calling `ardupilot_manager/main.py` will result in:
```
Traceback (most recent call last):
  File "./main.py", line 7, in <module>
    from mavlink_proxy.Manager import Manager
  File "/home/patrick/git/blue/companion-docker/core/services/ardupilot_manager/mavlink_proxy/Manager.py", line 6, in <module>
    import MAVLinkRouter
ModuleNotFoundError: No module named 'MAVLinkRouter'
```

And pytest will also fail, since it'll be unable to locate the library, the problem is the same related to the use of the library from `ardupilot_manager/main.py`

Williangalvani: could it be installed as mavlink_proxy library?

patrickelectric: yes, but for that we'll need to install first and run the test later ?

Williangalvani: (we discussed this over a call) This file is a tool, and not a core part of the library, so I'm ok with it.

core/services/ardupilot_manager/main.py
```diff
@@ -0,0 +1,64 @@
...
+        self.mavlink_manager = MavlinkManager()
+        available_mavlink_interfaces = self.mavlink_manager.available_interfaces()
+        if not available_mavlink_interfaces:
+            raise RuntimeError("No MAVLink interface available.")
+
+    def run(self) -> None:
+        ArduPilotManager.check_running_as_root()
+
+        while not self.start_board(BoardDetector.detect()):
+            print("Flight controller board not detect, will try again.")
```
jaxxzer, line 23: not detected

jaxxzer (approved):

core/services/ardupilot_manager/flight_controller_detector/Detector.py
```diff
@@ -0,0 +1,72 @@
...
+        """Check if navigator is connected using the sensors on the I²C BUS
+
+        Returns:
+            (bool, str): True if a navigator is connected, false otherwise.
+                String is always empty
+        """
+        try:
+            bus = SMBus(1)
+            PCA9685_address = 0x40
+            ADS115_address = 0x48
```
jaxxzer, line 34: ADS1115

Williangalvani (approved):

PR #28, opened 2020-09-09
Create bootstrap docker

Williangalvani:
Bootstrap would be the guy responsible for setting up things for the first run and handling updates (configuration changes where the docker image changes).

Every time it launches, it will load the startup.json file and launch the docker described there. After launching it quits (but I would consider keeping it running indefinitely, too).

The update process for core this way consists of:
 - pulling new image
 - updating startup.json
 - launching bootstrap (if it was not running indefinitely)
 - stopping itself.

As the bootstrap is a docker itself, it can be updated by a docker pull

This would be used to restart and update the supervisor, but the implementation will begin within companion-core

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -1,0 +1,4 @@
+FROM python:buster
+COPY setup.py startup.json.default main.py /bootstrap/
+RUN python3 bootstrap/setup.py install
+ENTRYPOINT /bootstrap/main.py
--- a/bootstrap/README.md
+++ b/bootstrap/README.md
@@ -1,0 +1,19 @@
+Bootstrap is responsible for setting things up for the first run of Companion and applying updates.
+
+To build the docker image:
+
+```bash
+cd bootstrap
+docker build . --tag bluerobotics/bootstrap:master
+```
+
+
+Usage:
+
+Everytime it is launched, it will load the startup settings, wait until the core container is not running, and re-create and re-launch it.
+
+```bash
+docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v /config:/config  bluerobotics/bootstrap:master
+```
+
+This will automatically populate /config if there is no valid config file in there, fetch if necessary, and then launch Companion.
--- a/bootstrap/main.py
+++ b/bootstrap/main.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import pathlib
+import shutil
+import sys
+import time
+from typing import Any
+from warnings import warn
+
+import docker
+
+client = docker.from_env()
+low_level_api = docker.APIClient(base_url="unix://var/run/docker.sock")
+
+DEFAULT_FILE_PATH = pathlib.Path("/bootstrap/startup.json.default")
+DOCKER_CONFIG_PATH = pathlib.Path("/root/.config/companion/startup.json")
+
+HOST_CONFIG_PATH = os.environ.get("COMPANION_CONFIG_PATH", None)
+
+CORE_CONTAINER_NAME = "companion_core"
+
+
+def ensure_dir(file_path: str) -> None:
+    """Makes sure that the file path exists
+
+    Args:
+        file_path (str): path to check/make
+    """
+    directory = pathlib.Path(file_path).parent
+    os.makedirs(directory, exist_ok=True)
+
+
+def stop_old_container() -> None:
+    """ Looks for an old core container and stops it if found"""
+    try:
+        old_container = client.containers.get(CORE_CONTAINER_NAME)
+        old_container.stop()
+        old_container.remove()
+    except docker.errors.NotFound:
+        pass
+
+
+def overwrite_config_file_with_defaults() -> None:
+    """Overwrites the config file with the default configuration"""
+    shutil.copy(DEFAULT_FILE_PATH, DOCKER_CONFIG_PATH)
+
+
+def read_config_file() -> Any:
+    """Tries to read the config file
+
+    Returns:
+        Any: Json data of startup.json
+    """
+
+    # Tries to open the current file
+    config = {}
+    try:
+        with open(DOCKER_CONFIG_PATH) as config_file:
+            config = json.load(config_file)
+    except Exception as error:
+        warn(f"unable to read startup.json file ({type(error).__name__}:{error}), reverting to defaults...")
+        # Copy defaults over and read again
+        overwrite_config_file_with_defaults()
+        with open(DEFAULT_FILE_PATH) as config_file:
+            config = json.load(config_file)
+
+    config["core"]["binds"][str(HOST_CONFIG_PATH)] = {
+        "bind": str(DOCKER_CONFIG_PATH),
+        "mode": "rw",
+    }
+    return config
+
+
+def start_core() -> None:
+    """Loads core settings and launches the core docker. Loads default settings if no settings are found"""
+    core_version = "stable"
+
+    config = read_config_file()
+
+    try:
+        core = config["core"]
+        image = core["image"]
+        core_version = core["tag"]
+        binds = core["binds"]
+        privileged = core["privileged"]
+        network = core["network"]
+
+    except Exception as error:
+        warn(f"Error reading startup json data! {error}")
+        sys.exit(1)
+
+    print("Attempting to pull an updated image... This might take a while...")
+    try:
+        for line in low_level_api.pull(f"{image}:{core_version}", stream=True, decode=True):
+            print(line["status"])
+    except docker.errors.APIError as error:
+        warn("Error trying to pull an update image: {error}")
+
+    print("Starting core")
+    client.containers.run(
+        f"{image}:{core_version}",
+        name=CORE_CONTAINER_NAME,
+        volumes=binds,
+        privileged=privileged,
+        network=network,
+        detach=True,
+    )
+    print("Core started")
+
+
+def core_is_running() -> bool:
+    """
+    Returns:
+        bool: True if the core container is running
+    """
+    for container in client.containers.list():
+        if CORE_CONTAINER_NAME in container.name:
+            return True
+    return False
+
+
+def remove_core() -> None:
+    """Deletes the core container if it exists (needed for updating the running image)"""
+    try:
+        old_container = client.containers.get(CORE_CONTAINER_NAME)
+        old_container.remove()
+    except docker.errors.NotFound:
+        # This exception is raised if the container does not exist
+        pass
+
+
+if __name__ == "__main__":
+    if HOST_CONFIG_PATH is None:
+        print("Please supply the host path for the config files as the COMPANION_CONFIG_PATH environment variable.")
+        print("Example docker command line:")
+        print(
+            "docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/.config/companion:"
+            "/root/.config/companion -e COMPANION_CONFIG_PATH=$HOME/.config/companion  bluerobotics/bootstrap:master"
+        )
+        sys.exit(1)
+
+    while True:
+        if core_is_running():
+            print("core is already running, waiting for it to stop...")
+            time.sleep(1)
+        else:
+            try:
+                remove_core()
+                start_core()
+                print("Done")
+                sys.exit(0)
+            except Exception as error:
+                warn(f"error: {error}, retrying...")
+        time.sleep(1)
--- a/bootstrap/setup.py
+++ b/bootstrap/setup.py
@@ -1,0 +1,13 @@
+#!/usr/bin/env python3
+
+from setuptools import setup
+
+setup(
+    name="companion_bootstrap",
+    version="0.0.1",
+    description="Blue Robotics Ardusub Companion Docker System Bootstrap",
+    license="MIT",
+    install_requires=[
+        "docker",
+    ],
+)
--- a/bootstrap/startup.json.default
+++ b/bootstrap/startup.json.default
@@ -1,0 +1,24 @@
+{
+    "core": {
+        "tag": "master",
+        "image": "bluerobotics/companion-core",
+        "enabled": true,
+        "webui": false,
+        "network": "host",
+        "binds": {
+            "/dev/": {
+                "bind": "/dev/",
+                "mode": "rw"
+            },
+            "/var/run/wpa_supplicant/wlan0": {
+                "bind": "/var/run/wpa_supplicant/wlan0",
+                "mode": "rw"
+            },
+            "/tmp/wpa_playground": {
+                "bind": "/tmp/wpa_playground",
+                "mode": "rw"
+            }
+        },
+        "privileged": true
+    }
+}


bootstrap/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:buster
+COPY setup.py /setup.py
+COPY startup.json.default /
```
jaxxzer, line 3: do some consistent ref to /

bootstrap/setup.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+
+from setuptools import setup
+
+setup(
+    name="companion_bootstrap",
+    version="0.1.0",
+    description="Blue Robotics Ardusub Companion Docker System Bootstraper",
```
jaxxzer, line 8: bootstrapper

bootstrap/main.py
```diff
@@ -0,0 +1,128 @@
...
+        network = config["core"]["network"]
+
+    except Exception as error:
+        print(f"Error reading startup json data! {error}")
+        sys.exit(1)
+
+    print("Starting core... this could take a while if the image is not available offline...")
+    client.containers.run(
+        f"{image}:{core_version}",
+        name="companion_core",
```
jaxxzer, line 84: `companion_core` should be a constant somewhere

bootstrap/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:buster
+COPY setup.py /
+COPY startup.json.default /
+COPY main.py /
```
patrickelectric, line 4: you know that you can do `COPY setup.py main.py startup.json.default /` right ?

bootstrap/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+RUN python3 setup.py install && chmod +x /main.py
```
patrickelectric, line 5: can you save the file in the repository already as executable ?

bootstrap/README.md
```diff
@@ -0,0 +1,19 @@
+Bootstrap is responsible for setting things up for the first run of Companion and applying updates.
+
+To build the docker image:
+
+```
```
patrickelectric, line 5:
```suggestion
```bash
```

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+import shutil
+import sys
+import time
+from typing import Any
+
+import docker
+
+client = docker.from_env()
+
+default_file_path = pathlib.Path("/startup.json.default")
```
patrickelectric, line 15: global variables should be CAPITALIZED

bootstrap/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+ENTRYPOINT /main.py
```
patrickelectric, line 6: I'm not a huge fan to have all files under `/` this can result in a mess after some updates or new files, maybe creating your own `/bootstrap` or `/service` folder for organization ?

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def ensure_dir(file_path: str) -> None:
+    """Makes sure that the file path exists
+
+    Args:
+        file_path (str): path to check/make
+    """
+    directory = os.path.dirname(file_path)
+    if not os.path.exists(directory):
```
patrickelectric, line 28: The check is not necessary, you can do `os.makedirs(directory, exist_ok=True)`

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+CORE_CONTAINER_NAME = "companion_core"
+
+
+def ensure_dir(file_path: str) -> None:
+    """Makes sure that the file path exists
+
+    Args:
+        file_path (str): path to check/make
+    """
+    directory = os.path.dirname(file_path)
```
patrickelectric, line 27: since you are using pathlib, I would recommend to not use `os.path`.
E.g: `os.path.dirname` is `path.parent`, `os.path.exists` is `path.exists()` and `os.path.makedirs` is `path.mkdir`.
For reference: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def stop_old_container() -> None:
+    """ Looks for an old core container and stops it if found"""
+    try:
+        old_container = client.containers.get(CORE_CONTAINER_NAME)
+        old_container.stop()
+        old_container.remove()
+    except Exception:
+        pass
```
patrickelectric, line 39: you should at least warn that something is wrong, we don't fix our problem keeping it in silence.
`warnings.warn("Oh jezz, something is wrong rick: {exception}")`

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def overwrite_config_file_with_defaults() -> None:
+    """Overwrites the config file with the default configuration"""
+    shutil.copy(default_file_path, config_file_path)
+
+
+def read_config_file() -> Any:
```
patrickelectric, line 47: This should return a json, and from what I see in your json file is a `Dict[str, Any]` right ?

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+    Returns:
+        Any: Json data of startup.json
+    """
+
+    # Tries to open the current file
+    try:
+        with open("/config/startup.json") as config_file:
+            return json.load(config_file)
+    except Exception:
+        # Copy defaults over and read again
```
patrickelectric, line 59: warn ?

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+def read_config_file() -> Any:
+    """Tries to read the config file
+
+    Returns:
+        Any: Json data of startup.json
+    """
+
+    # Tries to open the current file
+    try:
+        with open("/config/startup.json") as config_file:
```
patrickelectric, line 56: `config_file_path`

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+            return json.load(config_file)
+    except Exception:
+        # Copy defaults over and read again
+        shutil.copy(default_file_path, config_file_path)
+        with open("/config/startup.json") as config_file:
```
patrickelectric, line 61: `config_file_path`

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+    try:
+        image = config["core"]["image"]
+        core_version = config["core"]["tag"]
+        binds = config["core"]["binds"]
+        ports = config["core"]["ports"]
+        privileged = config["core"]["privileged"]
+        network = config["core"]["network"]
+
+    except Exception as error:
+        print(f"Error reading startup json data! {error}")
```
patrickelectric, line 80: warn

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def start_core() -> None:
+    """Loads core settings and launches the core docker. Loads default settings if no settings are found"""
+    core_version = "stable"
+
+    config = read_config_file()
+
+    try:
+        image = config["core"]["image"]
```
patrickelectric, line 72: maybe for readability ?
```python
core = config["core"]
```

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+    client.containers.run(
+        f"{image}:{core_version}",
+        name=CORE_CONTAINER_NAME,
+        volumes=binds,
+        ports=ports,
+        privileged=privileged,
+        network=network,
+        detach=True,
+    )
+    print("Started, open http://localhost:8080")
```
patrickelectric, line 93: Can we configure this port somehow ? I believe that 8080 is really complicated since is a default port for a bunch of services, maybe we should pick different number.

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+        bool: True if the core container is running
+    """
+    for container in client.containers.list():
+        if CORE_CONTAINER_NAME in container.name:
+            return True
+    return False
+
+
+def remove_core() -> None:
+    """Deletes the core container if it exists (needed for updating the running image)"""
```
patrickelectric, line 108: `needed to update the running image` ?

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+    try:
+        old_container = client.containers.get(CORE_CONTAINER_NAME)
+        old_container.remove()
+    except Exception:
+        # This exception is raised if the container does not exist
```
patrickelectric, line 113: pass if it raises the exception if the container does not exist, and warn if a different exception is raised.

bootstrap/main.py
```diff
@@ -0,0 +1,130 @@
...
+            print("core is already running, waiting for it to stop...")
+            time.sleep(1)
+        else:
+            try:
+                remove_core()
+                start_core()
+                print("Done")
+                sys.exit(0)
+            except Exception as error:
+                print(f"error: {error}, restarting...")
```
patrickelectric, line 129: it'll not restart, it'll try to remove the core again, no ?

patrickelectric, line 129: warn

bootstrap/startup.json.default
```diff
@@ -0,0 +1,29 @@
+{
+    "core": {
+        "tag": "master",
```
patrickelectric, line 3: this will work if we update the tag but it's using the old master version ?

bootstrap/startup.json.default
```diff
@@ -0,0 +1,29 @@
...
+            },
+            "/var/run/wpa_supplicant/wlan0": {
+                "bind": "/var/run/wpa_supplicant/wlan0",
+                "mode": "rw"
+            },
+            "/tmp/wpa_playground": {
+                "bind": "/tmp/wpa_playground",
+                "mode": "rw"
+            },
+            "/config": {
```
patrickelectric, line 22: maybe `/companion/config` or better `~/.config/companion`?

bootstrap/startup.json.default
```diff
@@ -0,0 +1,29 @@
+{
+    "core": {
+        "tag": "master",
+        "image": "bluerobotics/companion-core",
+        "enabled": true,
+        "webui": false,
+        "network": "host",
+        "ports": {},
```
patrickelectric, line 8: ports are not necessary since we are going to bind with host directly

patrickelectric (changes requested):

bootstrap/Dockerfile
Williangalvani: I do now!

bootstrap/Dockerfile
Williangalvani: sure, as long as we don't go to deep I'm ok with it.

patrickelectric: image

bootstrap/main.py
Williangalvani: [show they](https://www.python.org/dev/peps/pep-0008/#global-variable-names)?

constants should, but I don't think I'd call these constants

bootstrap/main.py
Williangalvani: it is more of a Dict[str, Dict[str, Any]]. But after some tries and reading [this](https://github.com/python/typing/issues/182)  I thought it was just not worth the extra clutter. I'll give it another try.

bootstrap/main.py
Williangalvani: that is just a text string, that is the port the supervisor used, I'll take it out.

bootstrap/main.py
Williangalvani: I think they are both right, @jaxxzer ?

bootstrap/main.py
Williangalvani: it goes back to line 119, I could reword that. "retrying"?

bootstrap/startup.json.default
Williangalvani: the only possible issue here is that we push over the same tag and it uses the local one instead of updating. I don't think we should do that, btw

Williangalvani: [Common, yes](https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375#:~:text=Latest%20is%20just%20a%20tag%20with%20a%20special%20name.&text=%E2%80%9CLatest%E2%80%9D%20simply%20means%20%E2%80%9Cthe,build%20%2Dt%20marc%2Ftest%20.&text=sudo%20docker%20build%20%2Dt%20marc%2Ftest%20.,-sudo%20docker%20tag).
We could use the master tag to check if there are updates and then update it here so it points to the "latest" (the the version tag instead).
I put master in here because we don't have any other tags and I didn't feel like creating a new one for no reason.
I also still prefer the idea of having "curated" lists of versions/tags

Williangalvani: We discussed this on a call. I'll add a "docker pull" before running, as  that is harmless and will help handling cases of force-pushing tags (useful while in development, I don't endorse it)

bootstrap/main.py
patrickelectric: Wait, how `default_file_path` is not a constant ? Are you changing it somewhere ?

Williangalvani: we agreed that we will use CAPS as a "this should not be changed/mutated" indicator

bootstrap/startup.json.default
jaxxzer: we can also use the hash instead of tag

jaxxzer:
do we need to update /install for this?

Williangalvani:
> do we need to update /install for this?

yes, but It is not done yet

bootstrap/setup.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+
+from setuptools import setup
+
+setup(
+    name="companion_bootstrap",
+    version="0.1.0",
```
jaxxzer, line 7: we want this version? instead of 0.0.1?

bootstrap/main.py
patrickelectric: Yep, `Dict[str, Any]` will not be the most appropriate and correct way, but I believe that is much better than `Any`.

bootstrap/main.py
```diff
@@ -0,0 +1,156 @@
...
+import time
+from typing import Any
+from warnings import warn
+
+import docker
+
+client = docker.from_env()
+low_level_api = docker.APIClient(base_url="unix://var/run/docker.sock")
+
+DEFAULT_FILE_PATH = pathlib.Path("/bootstrap/startup.json.default").as_posix()
```
patrickelectric, line 17: I'm not sure if I follow, but shouldn't this be absolute ? `absolute()`

bootstrap/main.py
```diff
@@ -0,0 +1,156 @@
...
+        Any: Json data of startup.json
+    """
+
+    # Tries to open the current file
+    config = {}
+    try:
+        with open(DOCKER_CONFIG_PATH) as config_file:
+            config = json.load(config_file)
+    except Exception:
+        warn("unable to read startup.json file, reverting to defaults...")
```
patrickelectric, line 63: can we f-string the warning with the exception type ?

patrickelectric (changes requested):
Looks great, only some minor corrections

bootstrap/main.py
Williangalvani: this is already absolute

bootstrap/main.py
Williangalvani: sure, done

bootstrap/main.py
patrickelectric: so, why posix ? If you want a string you could use `str()` or `absolute()`, the posix call may get the attention of someone that is reading the code asking why did you use the posix function for that ? And that's my initial question while reading the code.
both str and absolute are a clear way to express what you want, or is there a really deep meaning behind the `as_posix` call that I'm not aware in your implementation ?

Williangalvani: absolute() is apparently undocumented/unsupported. str() works fine, I'll switch to that.

patrickelectric (approved):

PR #40, opened 2020-09-16
Improve coverage 

patrickelectric:


.coveragerc
```diff
@@ -0,0 +1,6 @@
+[run]
+omit =
+    */__init__.py
+
+[report]
+fail_under=81
```
Williangalvani, line 6: newline?

.hooks/pre-push
```diff
@@ -18,6 +18,6 @@ echo "Running mypy.."
 git ls-files '*.py' | xargs --max-lines=1 mypy --ignore-missing-imports --warn-unreachable --allow-untyped-decorators --strict
 
 echo "Running pytest.."
-pytest --cov-fail-under=78 --cov="$repository_path"
+pytest --cov="$repository_path" --cov-config="$repository_path/.coveragerc"
 
 exit 0
```
Williangalvani, line 23: newline?

.coveragerc
```diff
@@ -0,0 +1,6 @@
+[run]
+omit =
+    */__init__.py
```
Williangalvani, line 3: Is this here to avoid checking non-empty __init__.py where you are just importing things from the other modules to expose them at the package level?

Does this format support comments? if so it would be nice to have it as comment.

Williangalvani (changes requested):

.coveragerc
patrickelectric: Does not support

Williangalvani (approved):

PR #42, opened 2020-09-16
CI: Allow deploying multiple dockers

Williangalvani:
This makes our CI build both docker images


patrickelectric (approved):
Merge at will

PR #43, opened 2020-09-17
Install: update to use bootstrap

Williangalvani:
~depends on #42~
needs testing

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -39,20 +39,18 @@ echo "Going to install companion-docker version ${VERSION}."
 echo "Downloading and installing udev rules."
 curl -fsSL $REMOTE/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
 
-echo "Downloading companion-core"
-COMPANION_CORE="bluerobotics/companion-core:master" # We don't have others tags for now
-docker pull $COMPANION_CORE
-docker create \
+echo "Downloading bootstrap"
+COMPANION_BOOTSTRAP="bluerobotics/companion-bootstrap:master" # We don't have others tags for now
+docker pull $COMPANION_BOOTSTRAP
+# Start bootstrap for the first time to fetch the other images and allow docker to restart it after reboot
+docker run \
+    -it \
     --restart unless-stopped \
-    --name companion-core \
-    --privileged \
-    --network host \
-    -v /dev:/dev \
-    -v /tmp/wpa_playground:/tmp/wpa_playground \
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -40,19 +40,14 @@ echo "Downloading and installing udev rules."
...
-    --network host \
-    -v /dev:/dev \
-    -v /tmp/wpa_playground:/tmp/wpa_playground \
-    -v /var/run/wpa_supplicant/wlan0:/var/run/wpa_supplicant/wlan0 \
-    $COMPANION_CORE
-# Start companion-core for the first time to allow docker to restart it after reboot
-docker start companion-core
+    --name companion-bootstrap \
+    $COMPANION_BOOTSTRAP
+# Start companion-core for the first time to fetch the other images and allow docker to restart it after reboot
```
patrickelectric, line 49: bootstrap*

install/install.sh
```diff
@@ -40,19 +40,14 @@ echo "Downloading and installing udev rules."
 curl -fsSL $REMOTE/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
 
 echo "Downloading companion-core"
```
patrickelectric, line 42: bootstrap*

patrickelectric (changes requested):

patrickelectric (changes requested):
Corrections are still necessary in the base image.

install/install.sh
```diff
@@ -39,20 +39,17 @@ echo "Going to install companion-docker version ${VERSION}."
...
-    -v /dev:/dev \
-    -v /tmp/wpa_playground:/tmp/wpa_playground \
-    -v /var/run/wpa_supplicant/wlan0:/var/run/wpa_supplicant/wlan0 \
-    $COMPANION_CORE
-# Start companion-core for the first time to allow docker to restart it after reboot
-docker start companion-core
+    --name companion-bootstrap \
+    -v $HOME/.config/companion:/root/.config/companion \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -e COMPANION_CONFIG_PATH=$HOME/.config/companion \
```
patrickelectric, line 51: Will this variable be accessible in core ?

Williangalvani: no

jaxxzer: maybe it will be nice to add an exports list to the core config file. here I think we agree this doesn't need to hold

patrickelectric (approved):
It works

PR #46, opened 2020-09-18
install: Add initial hardware configuration for navigator 

patrickelectric:


install/install.sh
```diff
@@ -9,11 +9,47 @@ REMOTE="$REMOTE/$VERSION"
...
+    echo "Companion only supports ARM computers."
+    exit 1
+)
 
 # Check if the script is running as root
 [[ $EUID != 0 ]] && echo "Script must run as root."  && exit 1
 
+# Running in a BCM27* or BCM28, do necessary Raspberry configuration
+cat /proc/cpuinfo | grep -Eq "BCM(27|28)" && (
+    echo "Configuring board.."
```
Williangalvani, line 22:
```suggestion
    echo "Raspberry Pi detected, configuring board.."
```

install/install.sh
```diff
@@ -9,11 +9,47 @@ REMOTE="$REMOTE/$VERSION"
...
+
+    # Check for valid modules file to load kernel modules
+    if [ -f "/etc/modules" ]; then
+        MODULES_FILE="/etc/modules"
+    else
+        MODULES_FILE="/etc/modules-load.d/companion.conf"
+        touch "$MODULES_FILE" || true # Create if it does not exist
+    fi
+
+    echo "- Load kernel modules."
```
Williangalvani, line 41:
```suggestion
    echo "- Set up kernel modules."
```
They will only be loaded after the reboot, right?

Williangalvani (approved):
looks good. I'm guessing x86 support (without navigator) will come in another pr?

install/install.sh
patrickelectric: This should work with any BCM27 and 28 SBC, not necessary raspberry.

Williangalvani: anyway, I think we should tell the user which configuration is being applied

Williangalvani: well, maybe in the next pr.

patrickelectric: > anyway, I think we should tell the user which configuration is being applied

Check the following lines.

PR #51, opened 2020-09-21
Add initial navigator support 

patrickelectric:
How to test:
    Replace install script docker to: `COMPANION_CORE="patrickelectric/companion-core"`
Open QGC or mavproxy to connect in: `tcp:raspberry_pi_ip:5766`

### DIFF
--- a/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
+++ b/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
@@ -1,8 +1,10 @@
 import enum
 import gzip
 import json
+import os
 import pathlib
 import random
+import ssl
 import string
 import tempfile
 from typing import Any, Dict, List, Optional
@@ -11,6 +13,11 @@ from urllib.request import urlopen, urlretrieve
 from warnings import warn
 
 from packaging.version import Version
+
+# TODO: This should be not necessary # pylint: disable=fixme
+# Disable SSL verification
+if not os.environ.get("PYTHONHTTPSVERIFY", "") and getattr(ssl, "_create_unverified_context", None):
+    ssl._create_default_https_context = ssl._create_unverified_context
 
 
 class Vehicle(enum.Enum):
--- a/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
+++ b/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
@@ -66,6 +66,18 @@ class FirmwareDownload:
         return filename
 
     @staticmethod
+    def _download_navigator() -> Optional[pathlib.Path]:
+        """This is a temporary method should be removed after navigator builds are provided
+            by ArduPilot community.
+            TODO: This function should not exist and be removed asap
+
+        Returns:
+            Optional[pathlib.Path]: Path of the binary
+        """
+        url = "https://s3.amazonaws.com/downloads.bluerobotics.com/ardusub/navigator/ardusub"
+        return FirmwareDownload._download(url)
+
+    @staticmethod
     def _validate_firmware(firmware_path: pathlib.Path) -> bool:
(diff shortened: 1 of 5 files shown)


core/services/ardupilot_manager/main.py
```diff
@@ -16,6 +20,17 @@ def __init__(self) -> None:
...
             raise RuntimeError("No MAVLink interface available.")
 
+        self.settings_path = appdirs.user_config_dir(self.__class__.__name__)
+        self.firmware_path = os.path.join(self.settings_path, "firmware")
+        self.subprocess: Optional[Any] = None
+        self.firmware_download = FirmwareDownload()
+
+        # Create settings folder, ignore if already exists
+        try:
+            os.makedirs(self.firmware_path)
```
Williangalvani, line 30: doesn't  `exist_ok` work here?

core/services/ardupilot_manager/main.py
```diff
@@ -16,6 +20,17 @@ def __init__(self) -> None:
         if not available_mavlink_interfaces:
             raise RuntimeError("No MAVLink interface available.")
 
+        self.settings_path = appdirs.user_config_dir(self.__class__.__name__)
```
Williangalvani, line 23: does this end up being /root/.config/ArdupilotManager?

core/services/ardupilot_manager/main.py
```diff
@@ -28,13 +43,42 @@ def check_running_as_root() -> None:
...
 
-    @staticmethod
-    def start_navigator() -> None:
-        raise NotImplementedError("We only support ArduPilot running on serial devices.")
+    def start_navigator(self) -> None:
+        firmware = os.path.join(self.firmware_path, "ardusub")
+        if not os.path.isfile(firmware):
+            temporary_file = self.firmware_download.download(Vehicle.Sub, "Navigator")
+            assert temporary_file, "Failed to download navigator binary."
+            os.chmod(temporary_file, stat.S_IXOTH)
```
Williangalvani, line 51: I wish I could go back in time and hit whoever thought naming something `S_IXOTH` was a good idea.
Can you add a comment saying that it just adds execute permission?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -12,6 +14,10 @@
 
 from packaging.version import Version
 
+# Disable SSL verification
```
Williangalvani, line 17: add a TODO to get rid of this when we find the source of the issue

Williangalvani (changes requested):
Code looks good, I haven't tested it though.

core/services/ardupilot_manager/main.py
patrickelectric: Yes.

jaxxzer:
We should add i2c4 to the overlay

core/services/ardupilot_manager/main.py
```diff
@@ -28,13 +41,45 @@ def check_running_as_root() -> None:
         if os.geteuid() != 0:
             raise RuntimeError("ArduPilot manager needs to run with root privilege.")
 
-    @staticmethod
-    def start_navigator() -> None:
-        raise NotImplementedError("We only support ArduPilot running on serial devices.")
+    def start_navigator(self) -> None:
+        firmware = os.path.join(self.firmware_path, "ardusub")
```
jaxxzer, line 45: Can we todo to support other vehicles?

jaxxzer (approved):

core/services/ardupilot_manager/main.py
patrickelectric: This information will be provided by our endpoint, and the download will be done by manifest. Since we don't have both right now I'm hardcoding the values, but our python library already handle any vehicle and hardware.

PR #57, opened 2020-09-29
Bootstrap: Fix config file path

Williangalvani:
It was erroneously trying to mount the startup.json file as a volume instead of the parent directory it is in. You cannot mount a file as a volume.

This makes docker mount the correct volume in line 57

for testing, use williangalvani/bootstrap:master in install.sh

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -24,7 +24,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install wheel # wheel need to be installed before mavproxy
         pip install mavproxy
-        pip install pyfakefs pytest-cov pylint mypy isort black
+        pip install pyfakefs pytest-cov pytest-timeout pylint mypy isort black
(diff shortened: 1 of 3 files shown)


jaxxzer (approved):

Williangalvani:
I can't figure out what is going on with CI here...

patrickelectric:
@Williangalvani can you disable fakefs ? it appears to be stuck on that part

patrickelectric:
It really looks that one of the tests are just stuck, my PR did pass by CI without problems here: https://github.com/bluerobotics/companion-docker/pull/64

Williangalvani:
> It really looks that one of the tests are just stuck

You were right. Updated testes and added timeouts.

patrickelectric:
@Williangalvani is the latest version available in dockerhub ?

patrickelectric:
@Williangalvani tests are failing

patrickelectric (approved):

PR #72, opened 2020-10-05
Bootstrap: handle invalid image names

### LINKED ISSUE #69: bootstrap: Failing when bootstrap image changes and install script is called again
image

Williangalvani:
Fix #69 

to test, change the image name in startup.json and then run:

```
docker run \                                                                                             
    -it \
    --rm  \
    --name companion_bootstrap \
    -v $HOME/.config/companion:/root/.config/companion \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -e COMPANION_CONFIG_PATH=$HOME/.config/companion \
    williangalvani/bootstrap:missingimage
```

### DIFF
--- a/bootstrap/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap/bootstrap.py
@@ -145,14 +145,20 @@ class Bootstrapper:
             warn(f"Error trying to pull an update image: {error}")
 
         print("Starting core")
-        self.client.containers.run(
-            f"{image}:{core_version}",
-            name=Bootstrapper.CORE_CONTAINER_NAME,
-            volumes=binds,
-            privileged=privileged,
-            network=network,
-            detach=True,
-        )
+        try:
+            self.client.containers.run(
+                f"{image}:{core_version}",
+                name=Bootstrapper.CORE_CONTAINER_NAME,
+                volumes=binds,
+                privileged=privileged,
+                network=network,
+                detach=True,
+            )
+        except docker.errors.APIError as error:
+            warn(f"Error trying to start image: {error}, reverting to default...")
+            self.overwrite_config_file_with_defaults()
+            return False
+
         print("Core started")
         return True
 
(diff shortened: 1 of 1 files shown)


bootstrap/bootstrap/bootstrap.py
```diff
@@ -86,12 +93,14 @@ def pull_core(self) -> None:
         lines: int = 0
         # map each id to a line
         id_line: Dict[str, int] = {}
-        for line in self.low_level_api.pull(f"{image}:{tag}", stream=True, decode=True):
-            if len(line.keys()) == 1:
-                # in some cases there is only "status", print that on the last line
-                screen.addstr(lines, 0, line["status"])
-                continue
-            try:
+        try:
```
patrickelectric, line 96: why is there a try without exception and with a finally when a for is inside try ?

patrickelectric (changes requested):

bootstrap/bootstrap/bootstrap.py
```diff
@@ -132,19 +137,30 @@ def start_core(self) -> None:
         print("Attempting to pull an updated image... This might take a while...")
         try:
             self.pull_core()
+        except docker.errors.ImageNotFound:
+            warn("Image not found, reverting to default...")
+            self.overwrite_config_file_with_defaults()
+            return False
         except docker.errors.APIError as error:
             warn(f"Error trying to pull an update image: {error}")
```
patrickelectric, line 145: what happens if the image does not exist here ?

bootstrap/bootstrap/bootstrap.py
Williangalvani: The exception is raised by `low_level_api.pull()`. The try-finally blocks makes sure that the curses interface is properly finalized before the exception is raised. I intend to re-do that eventually #71 , but that is gonna take more time than I'd like to allocate to it right now

patrickelectric (approved):

PR #92, opened 2021-02-03
Improve pylint check

patrickelectric:


Williangalvani:
What does it do? Turns the FIXME into a note instead of error?

patrickelectric:
> What does it do? Turns the FIXME into a note instead of error?

No, by default the words that pylint check are TODO and FIXME, since we use TODO for our internal development, I removed it from the list (we are suppressing it everywhere)

Williangalvani:
cool

Williangalvani (approved):

PR #95, opened 2021-02-05
Add simple helper service

patrickelectric:
It'll show all available ports with a http server running:
image
patrickelectric/companion-core:latest

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+import http
+import urllib
+from typing import List
+from urllib.request import urlopen
+
+import bottle
+import psutil
+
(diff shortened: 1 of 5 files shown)


core/services/helper/main.py
```diff
@@ -0,0 +1,50 @@
...
+
+import http
+import urllib
+from typing import List
+from urllib.request import urlopen
+
+import bottle
+import psutil
+
+PORT = 8080
```
Williangalvani, line 11: shouldn't this be 80?

patrickelectric: Sorry, the README was wrong, the correct is 8080, I thought that 80 should be reserved for our main page.

Williangalvani (approved):

PR #105, opened 2021-03-02
Improve message if no MAVLink interface found

patrickelectric:


Williangalvani:
It is going to be more work, but what about `No Mavlink routers found. Make sure that at least one is installed. Checked for 'Mavproxy' and 'mavlink-router'` ?

Williangalvani (approved):

PR #109, opened 2021-03-11
Extend ArduPilotManager functionalities regarding endpoints

rafaellehmkuhl:


patrickelectric (changes requested):
CI is failing

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self) -> None:
+        # Create folder for ardupilot-manager service if it doesn't exist
+        self.settings_path = appdirs.user_config_dir(Settings.app_name)
+        os.makedirs(self.settings_path, exist_ok=True)
+
+        # Create JSON file for storing mavlink endpoints
+        self.endpoints_path = os.path.join(self.settings_path, "endpoints.json")
+        if not os.path.isfile(self.endpoints_path):
+            with open(self.endpoints_path, "w") as endpoints_file:
+                endpoints_file.write("[]")
```
patrickelectric, line 19: Avoid writing raw strings, in a document that will be json, I believe that is better to write an empty json object

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __init__(self) -> None:
+        # Create folder for ardupilot-manager service if it doesn't exist
+        self.settings_path = appdirs.user_config_dir(Settings.app_name)
+        os.makedirs(self.settings_path, exist_ok=True)
+
+        # Create JSON file for storing mavlink endpoints
+        self.endpoints_path = os.path.join(self.settings_path, "endpoints.json")
+        if not os.path.isfile(self.endpoints_path):
+            with open(self.endpoints_path, "w") as endpoints_file:
```
patrickelectric, line 18: There is no try catch block here

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,31 @@
...
+import appdirs
+
+
+class Settings:
+    app_name = "ardupilot-manager"
+
+    def __init__(self) -> None:
+        # Create folder for ardupilot-manager service if it doesn't exist
+        self.settings_path = appdirs.user_config_dir(Settings.app_name)
+        os.makedirs(self.settings_path, exist_ok=True)
```
patrickelectric, line 13: missing try catch

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,31 @@
...
+
+        # Create JSON file for storing mavlink endpoints
+        self.endpoints_path = os.path.join(self.settings_path, "endpoints.json")
+        if not os.path.isfile(self.endpoints_path):
+            with open(self.endpoints_path, "w") as endpoints_file:
+                endpoints_file.write("[]")
+
+        # Create folder for storing ArduPilot firmwares
+        self.firmware_path = os.path.join(self.settings_path, "firmware")
+        os.makedirs(self.firmware_path, exist_ok=True)
```
patrickelectric, line 23: missing try catch here

core/services/ardupilot_manager/mavlink_proxy/Manager.py
```diff
@@ -43,6 +43,9 @@ def set_master_endpoint(self, master: Endpoint) -> None:
     def start(self) -> None:
         self.tool.start(self.master)
 
+    def restart(self) -> None:
+        self.tool.restart()
```
patrickelectric, line 103: restart returns a bool to verify that the service was restarted, why are you ignoring it ?

core/services/ardupilot_manager/mavlink_proxy/Manager.py
```diff
@@ -32,11 +36,63 @@ def available_interfaces() -> List[Type[AbstractRouter]]:
     def use(self, interface: AbstractRouter) -> None:
         self.tool = interface
 
+    def _saved_endpoints(self) -> List[Endpoint]:
+        try:
+            with open(self.endpoints_path) as f:
+                return [Endpoint(dictionary=endpoint) for endpoint in json.load(f)]
```
patrickelectric, line 42: Why the manager is doing this ? This should be the settings job.

core/services/ardupilot_manager/mavlink_proxy/Manager.py
```diff
@@ -32,11 +36,63 @@ def available_interfaces() -> List[Type[AbstractRouter]]:
...
+            with open(self.endpoints_path, "w") as endpoints_file:
+                json.dump(endpoints, endpoints_file)
+        except IOError as e:
+            print(f"Error: Could not save endpoints on cache.\n{e}")
+
+    def load_endpoints(self) -> None:
+        saved_endpoints = self._saved_endpoints()
+        for endpoint in saved_endpoints:
+            if not self.tool.add_endpoint(endpoint):
+                warn(f"Endpoint {endpoint} is not valid and was not loaded.", RuntimeWarning)
```
patrickelectric, line 57: This entire load/save thing should be in settings

patrickelectric (changes requested):
Some things need to change, the settings class is the most critical

patrickelectric:
I recommend to add a git-hook pre push to run the tests before pushing to the PR

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Good catch. I will catch the situation where the file is empty.

rafaellehmkuhl: Edit: I will first check if the file is empty, and return an empty array in that case, otherwise it would throw a JSONEncodeError everytime the user has no endpoints.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -26,9 +26,11 @@ class Endpoint:
         (["file"], EndpointType.File),
     ]
 
-    def __init__(self, string: str = "") -> None:
+    def __init__(self, string: str = "", dictionary: dict = {}) -> None:
```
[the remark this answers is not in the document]
rafaellehmkuhl: Good catch.

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: In the case we cannot make a settings directory, do you think it would be better to say to the user he will have limited operation or raise the error and stop the service?

core/services/ardupilot_manager/mavlink_proxy/Manager.py
rafaellehmkuhl: I considered that the Settings job would be to provide the correct paths to the modules, not to perform the IO operations, but this can be revisited.

Do you think this is Setting's job or there should be a dedicated class for IO (it can be in the same file)?

core/services/ardupilot_manager/settings.py
patrickelectric: I think that is better to raise the error and continue the operation, with that will at least will have the log of what happen and the user will still be able to use the ROV

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -47,6 +49,13 @@ def from_str(self, pattern: str) -> "Endpoint":
 
         return self
 
+    def from_dict(self, pattern: dict) -> "Endpoint":
+        self.connType = pattern["connType"]
```
jaxxzer, line 53: validation?

rafaellehmkuhl: Done

core/services/ardupilot_manager/mavlink_proxy/Manager.py
rafaellehmkuhl: I moved it to settings and turned it into a generic loder/saver.

core/services/ardupilot_manager/mavlink_proxy/Manager.py
rafaellehmkuhl: Right now it's on the Settings class, but we can segregate it. I'm thinking if it makes sense to.

core/services/ardupilot_manager/mavlink_proxy/Manager.py
rafaellehmkuhl: Did you see any parrots around?

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Done

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Done

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Done

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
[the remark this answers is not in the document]
rafaellehmkuhl: Done

patrickelectric:
@rafaellehmkuhl CI is still failing

patrickelectric (changes requested):
CI is still failing

rafaellehmkuhl:
> CI is still failing

I'm checking that. It's on PyTest. I thought it was already, because it's failing on code I didn't touch, but the other PRs seem fine. I will take a better look.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -47,6 +49,22 @@ def from_str(self, pattern: str) -> "Endpoint":
 
         return self
 
+    def from_dict(self, pattern: Dict[str, str]) -> "Endpoint":
+        if not ("connType" in pattern and "place" in pattern):
```
patrickelectric, line 53: there is a cleaner way using sets:
```python
if not {"connType", "place"}.issubset(pattern.keys()):
```

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -47,6 +49,22 @@ def from_str(self, pattern: str) -> "Endpoint":
...
+            raise RuntimeError("Wrong format for endpoint creation.")
```
patrickelectric, line 54: `Wrong format for endpoint creation, missing the following: {{"connType", "place"}.difference(pattern.keys())}`

core/services/ardupilot_manager/mavlink_proxy/Manager.py
```diff
@@ -32,11 +32,29 @@ def available_interfaces() -> List[Type[AbstractRouter]]:
     def use(self, interface: AbstractRouter) -> None:
         self.tool = interface
 
+    def add_endpoint(self, endpoint: Endpoint) -> bool:
+        if not self.tool.add_endpoint(endpoint):
```
patrickelectric, line 36: why not:
```python
return self.tool.add_endpoint(endpoint)
```

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,35 @@
...
+        try:
+            # Create folder for ardupilot-manager service
+            self.settings_path = Path(appdirs.user_config_dir(Settings.app_name))
+            os.makedirs(self.settings_path, exist_ok=True)
+
+            # Create folder for storing ArduPilot firmwares
+            self.firmware_path = Path(os.path.join(self.settings_path, "firmware"))
+            os.makedirs(self.firmware_path, exist_ok=True)
+        except OSError as e:
+            print(f"Could not create settings directories: {e}\n Read/write operation will be limited..")
```
patrickelectric, line 22: this should be a warning

core/services/ardupilot_manager/settings.py
```diff
@@ -33,3 +35,47 @@ def endpoints_store(self) -> Path:
...
+        except json.decoder.JSONDecodeError as e:
+            print(f"Error: Could not get saved instances from disk. Empty or corrupted JSON file. \n{e}\n")
+            return []
+
+    @staticmethod
+    def save_instance(path: Path, instance: Any) -> bool:
+        saved_instances = Settings.get_saved_instances(path)
+        if instance.__dict__ not in saved_instances:
+            saved_instances.append(instance.__dict__)
+            try:
```
patrickelectric, line 60: this entire try part should be a function since the same is duplicated from unsave_instance

core/services/ardupilot_manager/settings.py
```diff
@@ -33,3 +35,47 @@ def endpoints_store(self) -> Path:
...
         return self.firmware_path
+
+    @staticmethod
+    def get_saved_instances(path: Path) -> List[Any]:
+        try:
+            with open(path) as f:
+                instances = json.load(f)
+                if isinstance(instances, list):
+                    return instances
+                print("Error: Loaded file is not a list of instances")
```
patrickelectric, line 46: warn

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -104,3 +105,20 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
 
     def restart(self) -> None:
         self.mavlink_manager.restart()
+
+    def _load_endpoints(self) -> None:
+        endpoints = self.settings.get_saved_instances(self.settings.endpoints_store)
```
patrickelectric, line 110: what is the idea behind  sending self.settings.endpoints_store to self.settings ? 
If self.settings has the endpoints_store, why is it necessary to use it as argument ?

patrickelectric (changes requested):
There are some parts that needs to be improved.
The only thing that worries me more is that the settings file does not contain any header and/or version in his scruct.
Without such information, we are stuck and unable to detect, identify and modify this files between version without safety in mind. In other words, it's unsafe and not easy to break the api if necessary for future versions.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Cool.

core/services/ardupilot_manager/mavlink_proxy/Manager.py
rafaellehmkuhl: That's a good question.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Because the get_saved_instances function is unaware of what you are trying to get in the first place.

rafaellehmkuhl:
> There are some parts that needs to be improved.
> The only thing that worries me more is that the settings file does not contain any header and/or version in his scruct.
> Without such information, we are stuck and unable to detect, identify and modify this files between version without safety in mind. In other words, it's unsafe and not easy to break the api if necessary for future versions.

You have a good point on the possibility of break between versions.
We can have a version system, like in cable_guy.

core/services/ardupilot_manager/mavlink_proxy/Manager.py
patrickelectric: what ?

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Not really. The first one appends the dict while the other removes it.

core/services/ardupilot_manager/ArduPilotManager.py
patrickelectric: why not a simple `self.settings.endpoints` ?
why ArdupiLotManager should keep track of the settings path in the system ?

core/services/ardupilot_manager/mavlink_proxy/Manager.py
rafaellehmkuhl: I mean I don't know why am I ignoring it and I'm deliberately shifting your focus from that return.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: That's actually better.

patrickelectric:
> > There are some parts that needs to be improved.
> > The only thing that worries me more is that the settings file does not contain any header and/or version in his scruct.
> > Without such information, we are stuck and unable to detect, identify and modify this files between version without safety in mind. In other words, it's unsafe and not easy to break the api if necessary for future versions.
> 
> You have a good point on the possibility of break between versions.
> We can have a version system, like in cable_guy.
> 
> @patrickelectric is that what you meant?

yes

core/services/ardupilot_manager/settings.py
patrickelectric: The try part is the same:
```
            try:
                with open(path, "w") as f:
                    json.dump(saved_instances, f)
                return True
            except IOError as e:
```

This should be something like `_write_json(Path, json) -> bool `

rafaellehmkuhl: Cool! For that part I agree.

jaxxzer (approved):
I think the settings version can be moved into an issue if you want to follow up PR.

I'm also curious about what the settings format looks like. I think the 'instance' language in settings is curious, it looks like 'instance' == 'endpoint'?

patrickelectric:
> I think the settings version can be moved into an issue if you want to follow up PR.

This will break the older settings if they exist, resulting in the problem that I first comment

rafaellehmkuhl:
> I think the settings version can be moved into an issue if you want to follow up PR.

I was already changing this, so I don't mind putting on this PR. It's also something that changes the endpoint's structure (which will reside in the settings file in fact).

> I'm also curious about what the settings format looks like. I think the 'instance' language in settings is curious, it looks like 'instance' == 'endpoint'?

The instance can be an endpoint or anything you want to store. It's a generic interface. Currently, we are only using it for the endpoints.

core/install_gst.sh
```diff
@@ -3,6 +3,9 @@
 # exit when any command fails
 set -e
 
+# Check if the script is running as root
+[[ $EUID != 0 ]] && echo "Script must run as root."  && exit 1
```
patrickelectric, line 7: This is not part of this PR, a new PR should be created for it

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -107,3 +107,6 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
             raise RuntimeError("Invalid board type: {boards}")
 
         return False
+
+    def restart(self) -> None:
+        self.mavlink_manager.restart()
```
patrickelectric, line 118: Why restart returns `None` if `manvlink_manager.restart` returns a boolean ? The function should also return a boolean no notify users about related problems.

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,92 @@
...
+        except OSError as e:
+            warn(f"Could not create settings files: {e}\n No settings will be loaded or saved during this session.")
+
+    @property
+    def content(self) -> Any:
+        return self.root["content"]
+
+    @property
+    def version(self) -> Any:
+        return self.root["version"]
```
patrickelectric, line 39: version is an int, since this is know, the function should also return an int, otherwise every time that someone wants to use the function a cast will be necessary, we should unwrap unsafe stuff in the minimal scope possible.

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,92 @@
...
+                with open(self.settings_file, "w+") as file:
+                    print(f"Creating settings file: {self.settings_file}")
+                    json.dump(self.root, file, sort_keys=True, indent=4)
+
+        except OSError as e:
+            warn(f"Could not create settings files: {e}\n No settings will be loaded or saved during this session.")
+
+    @property
+    def content(self) -> Any:
+        return self.root["content"]
```
patrickelectric, line 35: This should return a dict, no ?

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,92 @@
...
+        if self.root["content"] == content:
+            print("No new data. Not updating settings file.")
+            return
+
+        self.root["content"] = deepcopy(content)
+
+        if not Path.exists(self.settings_path):
+            Path.mkdir(self.settings_path)
+
+        with open(self.settings_file, "w+") as file:
```
patrickelectric, line 90: This should be inside a try block since it can fail

core/services/ardupilot_manager/settings.py
```diff
@@ -0,0 +1,92 @@
...
+            content (list): Configuration list
+        """
+        # We don't want to write in disk if there is nothing different to write
+        if self.root["content"] == content:
+            print("No new data. Not updating settings file.")
+            return
+
+        self.root["content"] = deepcopy(content)
+
+        if not Path.exists(self.settings_path):
```
patrickelectric, line 87: why the check if it exists if you can call mkdir directly with `exist_ok=True` ?

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -104,3 +116,31 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
...
+            self.configuration["endpoints"] = []
+        else:
+            endpoints = self.configuration["endpoints"]
+            self.mavlink_manager.add_endpoints([Endpoint(e) for e in endpoints])
+
+    def get_endpoints(self) -> List[Endpoint]:
+        return self.mavlink_manager.endpoints()
+
+    def add_endpoint(self, endpoint: Endpoint) -> bool:
+        if self.mavlink_manager.add_endpoint(endpoint):
```
patrickelectric, line 131: Just a tip, you can avoid the code in the extra indentation layer using the invert return logic.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -104,3 +116,31 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
...
+        return False
+
+    def remove_endpoint(self, endpoint: Endpoint) -> bool:
+        if self.mavlink_manager.remove_endpoint(endpoint):
+            try:
+                self.configuration.get("endpoints", []).remove(endpoint.__dict__)
+                self.settings.save(self.configuration)
+                return True
+            except Exception as e:
+                warn(f"Error: {e}")
```
patrickelectric, line 144: `Error while removing endpoints: {e}`
Also, try to avoid single letter variables, try `error` or `exception`

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -104,3 +116,31 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
...
+                return False
+        return False
```
patrickelectric, line 146: Remember that you can use the invert logic to do the early return and avoid the extra indentation level.

patrickelectric (changes requested):

core/install_gst.sh
rafaellehmkuhl: How did this even got here????

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Man, I'm now seeing that I selected a wrong stash during the rebase process...
That's why there's old code in there. I had already fixed that :/

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: This and content where respectively str (i assume we won't have just int versions?) and Dict[str, Any], but I had complaints from MyPy from what seems to be a misunderstand from the library.
I will try to find a workaround.

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: [previous comment explains]

patrickelectric (approved):

rafaellehmkuhl:
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=04854XqcfCY)

Issue #106, opened 2021-03-05
Install script fails at enabling Docker Daemon


rafaellehmkuhl:
This was noticed on a fresh Raspbian Lite installation, running on a RaspberryPi 4.

patrickelectric:
Can you provide the output of the installation script ?

patrickelectric:
@rafaellehmkuhl any update ?

rafaellehmkuhl:
Entire output of running /install/install.sh:

```bash
pi@raspberrypi:~ $ sudo curl -fsSL https://raw.githubusercontent.com/bluerobotics/companion-docker/master/install/install.sh | sudo bash
Starting hardware configuration.
Configuring BCM27XX board (Raspberry Pi 4)..
- Enable I2C, SPI and UART.
dtparam=i2c_arm=on
dtparam=spi=on
dtoverlay=spi1-3cs
dtoverlay=i2c4,pins_6_7
dtoverlay=uart1
- Set up kernel modules.
bcm2835-v4l2
i2c-bcm2835
i2c-dev
- Configure serial.
Checking for blocked wifi and bluetooth.
Checking for available space.
Checking for docker.
bash: line 38: docker: command not found
# Executing docker install script, commit: 3d8fe77c2c46c5b7571f94b42793905e5b3e42e4
+ sh -c apt-get update -qq >/dev/null
+ sh -c DEBIAN_FRONTEND=noninteractive apt-get install -y -qq apt-transport-https ca-certificates curl >/dev/null
+ sh -c curl -fsSL "https://download.docker.com/linux/raspbian/gpg" | apt-key add -qq - >/dev/null
Warning: apt-key output should not be parsed (stdout is not a terminal)
+ sh -c echo "deb [arch=armhf] https://download.docker.com/linux/raspbian buster stable" > /etc/apt/sources.list.d/docker.list
+ sh -c apt-get update -qq >/dev/null
+ [ -n  ]
+ sh -c apt-get install -y -qq --no-install-recommends docker-ce >/dev/null
E: Sub-process /usr/bin/dpkg returned an error code (1)
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Going to install companion-docker version master.
Downloading and installing udev rules.
Downloading bootstrap
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
pi@raspberrypi:~ $ 
```

patrickelectric:
[...]

PR #110, opened 2021-03-12
Add check for network access and docker installation 

patrickelectric:


rafaellehmkuhl:
Fix #106

install/install.sh
```diff
@@ -41,7 +41,12 @@ NECESSARY_SPACE_GB=4
 
 # Check for docker and install it if not found
 echo "Checking for docker."
-docker --version || curl -fsSL https://get.docker.com | sh && systemctl enable docker
+docker --version || curl -fsSL https://get.docker.com | sh || (
+    echo "Failed to start docker, something is wrong."
+    echo "Please report this problem."
+    exit 1
+)
+systemctl enable docker
```
rafaellehmkuhl, line 49: I assume you were not able to reproduce #106, right?

rafaellehmkuhl (approved):

install/install.sh
patrickelectric: Nops, I got a network problem where my avahi work messed up with devconf and the result was that get.docker was failing. 

The install script is working as intended

install/install.sh
```diff
@@ -17,6 +17,12 @@ set -e
 # Check if the script is running as root
 [[ $EUID != 0 ]] && echo "Script must run as root."  && exit 1
 
+echo "Check if network and remote are available."
+curl -fsSL --silent $REMOTE/install/install.sh &> /dev/null || (
```
jaxxzer, line 21: I think we should only redirect stdout, stderr would be useful here?

patrickelectric: done

PR #111, opened 2021-03-12
Add avahi configuration

patrickelectric:


rafaellehmkuhl (approved):

install/network/avahi.sh
```diff
@@ -0,0 +1,30 @@
...
+NEW_HOSTNAME="companion"
+# Overwrite with new name
+echo $NEW_HOSTNAME > /etc/hostname
+# Replace current name
+sed -i "s/127.0.1.1.*$OLD_HOSTNAME/127.0.1.1\t$NEW_HOSTNAME/g" /etc/hosts
+
+systemctl is-active --quiet avahi-daemon || (
+    echo "Avahi daemon is installed and running."
+    exit 1
+)
```
patrickelectric, line 22: This step should be done before configuring /etc/hosts and /etc/hostname

jaxxzer: Do we have to replace the hostname?
Can you have two hostnames for 127.0.1.1 (old + new)?
This will mess up a desktop development hostname?
Can avahi run inside of the docker?

patrickelectric: > Do we have to replace the hostname?
yes, otherwise it'll appear has raspberry.local, is there any reason for us to stay with raspberry ? I think that is better to have companion as hostname for clarity. 


> Can you have two hostnames for 127.0.1.1 (old + new)?
I'm not sure if I understood. 

> This will mess up a desktop development hostname?
No, this is the install script, it'll only be used to install companion in a computer.

> Can avahi run inside of the docker?
No, this is a systemctl service, this enables the avahi server, without it we can't find and provide services using mdns.

jaxxzer (approved):

PR #114, opened 2021-03-16
Ignore files created by pytest

rafaellehmkuhl:


.gitignore
```diff
@@ -1,2 +1,4 @@
 *__pycache__
 *.pytest_cache
+.coverage
+htmlcov/*
```
patrickelectric, line 4: should it be *.coverage and *htmlcov to remove such files in any subdirectory ?

rafaellehmkuhl: As far as I understand htmlcov/* will cover all folder called "htmlcov", no?

rafaellehmkuhl:
We can also jump in and copy a good part of Github's default .gitignore for Python. My sugestion would be:

```
# Byte-compiled / optimized / DLL files
__pycache__/
*.py[cod]
*$py.class

# C extensions
*.so

# Distribution / packaging
.Python
build/
develop-eggs/
dist/
downloads/
eggs/
.eggs/
lib/
lib64/
parts/
sdist/
var/
wheels/
share/python-wheels/
*.egg-info/
.installed.cfg
*.egg
MANIFEST

# Installer logs
pip-log.txt
pip-delete-this-directory.txt

# Unit test / coverage reports
htmlcov/
.tox/
.nox/
.coverage
.coverage.*
.cache
nosetests.xml
coverage.xml
*.cover
*.py,cover
.hypothesis/
.pytest_cache/
cover/

# Sphinx documentation
docs/_build/

# Environments
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/

# mkdocs documentation
/site

# mypy
.mypy_cache/
.dmypy.json
dmypy.json

# Cython debug symbols
cython_debug/
```

.gitignore
patrickelectric: Nops, it'll only hit root, for recursive you should start with *
image

patrickelectric:
> We can also jump in and copy a good part of Github's default .gitignore for Python. My suggestion would be:

I believe that we should with the minimum necessary and use the gitignore repository as a guide, otherwise we are going to have a 400 lines gitignore that'll be hard to track and with a bunch of unnecessary stuff, since we work with python, C++, js, ts, html and etc

jaxxzer:
I'm with @patrickelectric, I like a bespoke gitignore rather than a template with things that may or may not not exist in our project.

.gitignore
rafaellehmkuhl: https://git-scm.com/docs/gitignore#_examples

I think you're putting a slash before the folder's name. Without it, it works fine with subdirectories.

patrickelectric: Is that clear enough ?
image

patrickelectric: the asterisk means zero or more characters, if you expand what you wrote you'll understand.  
If it's still not clear, please provide an example to show that works, because that is not the case.

.gitignore
```diff
@@ -1,2 +1,4 @@
 *__pycache__
 *.pytest_cache
+.coverage
+htmlcov/
```
patrickelectric, line 4: Is there any reason for `.coverage` to not have the `/`and `htmlcov` to have ?

rafaellehmkuhl: .coverage is a file, not a folder.

patrickelectric (approved):

PR #115, opened 2021-03-17
Add camera manager service

### LINKED ISSUE #80: Video: Move from tmux to service
- [x] Create basic working progress
- [x] Handle UDP for now
- [x] Work with multiple videos or code structure for that

patrickelectric:
Fix #80 

Tested and working

### DIFF
--- a/core/tools/mavlink_camera_manager/bootstrap.sh
+++ b/core/tools/mavlink_camera_manager/bootstrap.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+LOCAL_BINARY_PATH="/usr/bin/mavlink-camera-manager"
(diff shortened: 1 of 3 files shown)


core/tools/mavlink_camera_manager/bootstrap.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
```
jaxxzer, line 2: i would call this an install.sh rather than bootstrap

jaxxzer (approved):

rafaellehmkuhl (approved):

core/tools/mavlink_camera_manager/bootstrap.sh
patrickelectric: I agree, sadly the name is also used by mavlink-routerd, I'll send a PR to fix both layer for consistency.

PR #118, opened 2021-03-19
ArduPilotManager: turn class into Singleton

rafaellehmkuhl:
With the current code structure, it only makes sense to have one instance of the manager running.

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -13,9 +13,10 @@ from flight_controller_detector.Detector import FlightControllerType
 from mavlink_proxy.Endpoint import Endpoint
 from mavlink_proxy.Manager import Manager as MavlinkManager
 from settings import Settings
+from Singleton import Singleton
 
 
-class ArduPilotManager:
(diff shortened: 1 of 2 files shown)


core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -15,7 +15,16 @@
...
 
 
-class ArduPilotManager:
+class Singleton(type):
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
+        return cls._instances[cls]
```
patrickelectric, line 24: I would recommend to move the singleton class to a new class to make it more flexible and useful, it also makes sense since you are using a dict of instances and not a single instance.

patrickelectric (approved):

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Makes sense.

patrickelectric (approved):

PR #119, opened 2021-03-19
Add REST API to ArduPilotManager

rafaellehmkuhl:
API is running as a uvicorn[uvicorn](https://www.uvicorn.org/)  process, started with core.


core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,15 +1,15 @@
...
-    TCP = 3
-    Serial = 4
-    File = 5
+class EndpointType(str, Enum):
+    Undefined = "undefined"
+    UDPServer = "udpin"
+    UDPClient = "udpout"
+    TCP = "tcp"
+    Serial = "serial"
+    File = "file"
```
patrickelectric, line 12: EndPoint class should be updated to use the strings of EddpointType over connTypeMap

core/services/ardupilot_manager/rest_api.py
```diff
@@ -0,0 +1,33 @@
...
+from ArduPilotManager import ArduPilotManager
+from fastapi import FastAPI
+from mavlink_proxy.Endpoint import Endpoint, EndpointType
+
+app = FastAPI()
+autopilot = ArduPilotManager()
+
+
+@app.get("/available_endpoints", response_model=List[Dict[str, Any]])
+def get_available_endpoints() -> Any:
```
patrickelectric, line 14: Can you explain why Any is used in return type and not List of Dict ?

core/services/ardupilot_manager/rest_api.py
```diff
@@ -0,0 +1,33 @@
...
+    return [endpoint.__dict__ for endpoint in autopilot.get_endpoints()]
+
+
+@app.post("/endpoint", response_model=bool)
+def create_new_endpoint(connection_type: EndpointType, place: str, argument: str = "") -> Any:
```
patrickelectric, line 19: Why Any over bool ?

core/services/ardupilot_manager/rest_api.py
```diff
@@ -0,0 +1,33 @@
...
+
+@app.delete("/endpoint", response_model=bool)
+def remove_endpoint(connection_type: EndpointType, place: str, argument: str = "") -> Any:
+    if autopilot.remove_endpoint(Endpoint({"connType": connection_type, "place": place, "argument": argument})):
+        return True
+    return False
+
+
+if __name__ == "__main__":
+    uvicorn.run("rest_api:app", host="0.0.0.0", port=8000)
```
patrickelectric, line 33: can we have a better name to identify the service ?

core/start-companion-core
```diff
@@ -7,6 +7,7 @@ SERVICES=(
     'cable_guy',"$SERVICES_PATH/cable_guy/main.py"
     'wifi',"$SERVICES_PATH/wifi/main.py"
     'autopilot',"$SERVICES_PATH/ardupilot_manager/main.py"
+    'autopilot-rest_api',"$SERVICES_PATH/ardupilot_manager/rest_api.py"
```
patrickelectric, line 10: How autopilot and autopilot-rest-_api are going to run concurrently ?

core/services/ardupilot_manager/setup.py
```diff
@@ -13,5 +13,12 @@
...
     packages=setuptools.find_packages(),
     python_requires=">=3.6",
-    install_requires=["appdirs==1.4.4", "packaging==20.4", "smbus2==0.3.0"],
+    install_requires=[
+        "appdirs==1.4.4",
+        "packaging==20.4",
+        "smbus2==0.3.0",
+        "fastapi==0.63.0",
+        "uvicorn==0.13.4",
+        "python-multipart==0.0.5",
```
patrickelectric, line 22: where is this being used ?

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: You mean using udp instead of udpout?

core/services/ardupilot_manager/rest_api.py
rafaellehmkuhl: This is [a suggestion of the library's author](https://github.com/tiangolo/fastapi/issues/101#issuecomment-613242344), in fact.

On FastAPI, the types of the responses are defined by the response_model, not the typehint, for reasons he explains in the link before, and the function itself has parsers for serializing/deserializing, so their types not exactly represent the HTTP responses.

[There are libraries](https://fastapi-utils.davidmontague.xyz/user-guide/inferring-router/) that offer this funcionality, thought.

core/services/ardupilot_manager/rest_api.py
rafaellehmkuhl: Same as before.

core/services/ardupilot_manager/rest_api.py
rafaellehmkuhl: Sure! Any ideia?

core/start-companion-core
rafaellehmkuhl: Humnnn, this is a flaw. In this implementation, they are not.

I think the best way to do it is to [make the singleton implementation thread-safe](https://stackoverflow.com/questions/50566934/why-is-this-singleton-implementation-not-thread-safe).

core/services/ardupilot_manager/setup.py
rafaellehmkuhl: This is for the firmware-upload part. I mistakenly committed it already.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
patrickelectric: I'm saying that `connTypeMap` should not exist, since this enum will do the job.

core/services/ardupilot_manager/rest_api.py
patrickelectric: ardupilot-manager ?

core/start-companion-core
patrickelectric: Can't you just remove the main.py line or replace main.py with rest_api ?
I'm not sure if I follow what is the problem here

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Oh, sure, this makes sense!

core/start-companion-core
rafaellehmkuhl: Oh, ok, now I got it. It makes much more sense, for sure.

core/services/ardupilot_manager/rest_api.py
rafaellehmkuhl: Got the ideia.

core/services/ardupilot_manager/setup.py
```diff
@@ -13,5 +13,11 @@
     long_description_content_type="text/markdown",
     packages=setuptools.find_packages(),
     python_requires=">=3.6",
-    install_requires=["appdirs==1.4.4", "packaging==20.4", "smbus2==0.3.0"],
+    install_requires=[
+        "appdirs==1.4.4",
+        "packaging==20.4",
+        "smbus2==0.3.0",
+        "fastapi==0.63.0",
+        "uvicorn==0.13.4",
```
patrickelectric, line 21: will this provide a swagger document webpage ?

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -74,5 +74,5 @@ def test_mavlink_router() -> None:
         endpoint_2,
     ], "Endpoint list does not match."
 
-    assert mavlink_router.start(Endpoint("udp:0.0.0.0:14550")), "Failed to start MAVLinkRouter"
+    assert mavlink_router.start(Endpoint("udpout:0.0.0.0:14550")), "Failed to start MAVLinkRouter"
```
patrickelectric, line 77: this means that we do not support "udp" anymore ? This is a pretty standard thing between mavlink-router and mavproxy

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
patrickelectric, line 12: Will this break the settings system when it loads with the old settings ?

core/services/ardupilot_manager/setup.py
```diff
@@ -13,5 +13,11 @@
     long_description_content_type="text/markdown",
     packages=setuptools.find_packages(),
     python_requires=">=3.6",
-    install_requires=["appdirs==1.4.4", "packaging==20.4", "smbus2==0.3.0"],
```
patrickelectric, line 16: The commit message does no help, which setup this commits points to ? please add the service name as prefix

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,15 +1,15 @@
 from dataclasses import dataclass
```
patrickelectric, line 1: same thing about the service name as a prefix for the commit

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -81,8 +81,6 @@ def start_navigator(self) -> None:
```
patrickelectric, line 81: same thing for service name as prefix

patrickelectric (changes requested):

core/services/ardupilot_manager/setup.py
rafaellehmkuhl: Yes, at port 8000, as specified on main.py

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Yes

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: It means, yes.
An honest workaround would be to have a parser from `udp` to `udpout` on endpoint creation.

patrickelectric: I'm a bit inclined to only support `udpout` over `udp` and 'udpout`, having a single name for a single functionality makes it more coerent, but I'm not sure since mostly mavlink routers support boths names, @jaxxzer what you think about it ?

core/services/ardupilot_manager/ArduPilotManager.py
patrickelectric: Can you follow the commit style of the previous commits ?

rafaellehmkuhl: Done

core/services/ardupilot_manager/main.py
```diff
@@ -1,6 +1,34 @@
...
+
+
+@app.get("/available_endpoints", response_model=List[Dict[str, Any]])
+def get_available_endpoints() -> Any:
+    return [endpoint.__dict__ for endpoint in autopilot.get_endpoints()]
+
+
+@app.post("/endpoint", response_model=bool)
+def create_new_endpoint(connection_type: EndpointType, place: str, argument: str = "") -> Any:
+    if autopilot.add_endpoint(Endpoint({"connType": connection_type, "place": place, "argument": argument})):
```
patrickelectric, line 20: Can we make it a payload over a query ?

core/services/ardupilot_manager/main.py
```diff
@@ -1,6 +1,34 @@
...
+from fastapi import FastAPI
+from mavlink_proxy.Endpoint import Endpoint, EndpointType
+
+app = FastAPI()
+autopilot = ArduPilotManager()
+
+
+@app.get("/available_endpoints", response_model=List[Dict[str, Any]])
+def get_available_endpoints() -> Any:
+    return [endpoint.__dict__ for endpoint in autopilot.get_endpoints()]
```
patrickelectric, line 15: Can return a pretty json over a plain text ?

core/services/ardupilot_manager/main.py
```diff
@@ -1,6 +1,34 @@
...
+
+
+@app.post("/endpoint", response_model=bool)
+def create_new_endpoint(connection_type: EndpointType, place: str, argument: str = "") -> Any:
```
patrickelectric, line 19: It's not validating place and argument:
```
[{"connType":"udpin","place":"0.0.0.0","argument":"14550"},{"connType":"udpin","place":"potato","argument":"-1234"},{"connType":"undefined","place":"5","argument":"what"}]
```

patrickelectric (changes requested):

core/services/ardupilot_manager/main.py
patrickelectric, line 20: We should also follow the api design document and move to snake case:
https://github.com/bluerobotics/companion-docker/wiki/REST-API-design

core/services/ardupilot_manager/main.py
rafaellehmkuhl: I think the validation is being done inside the Endpoint class, during instantiation. It is just not that good yet. We should improve it on the class itself, instead of passing the responsibility to the REST API, right?

In this case, I would suggest using a [BaseModel](https://pydantic-docs.helpmanual.io/usage/models/) instead of a dataclass for Endpoint, as BaseModel is much more powerful, especially with validations.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: You mean with connType?

core/services/ardupilot_manager/main.py
rafaellehmkuhl: It is already. FastAPI parse dicts into JSONs internally.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: We can, in fact. In fact, if we move the Endpoint class to BaseModel, as suggested in the other comment, the schema validation comes free of charge.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: I'm also a fan of having a single way to do it.

I always have to check the docs to see if `udp` means client or server.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: Edit: I think we can use a [pydantic dataclass](https://pydantic-docs.helpmanual.io/usage/dataclasses/) instead of BaseModel. It's a drop-in replacement for the standard dataclass, but comes with automatic parsing from FastAPI and enhanced validation.

core/services/ardupilot_manager/main.py
patrickelectric: yes

core/services/ardupilot_manager/main.py
patrickelectric: I'm talking about pretty json over plain json, as pretty I'm saying: `JSON.stringify(json, null 2)`

core/services/ardupilot_manager/mavlink_proxy/test_all.py
jaxxzer: only `udpout` and `udpin` is the best for me.

core/services/ardupilot_manager/main.py
```diff
@@ -1,6 +1,50 @@
...
+
+
+@app.get("/endpoints", response_model=List[Dict[str, Any]])
+def get_available_endpoints() -> Any:
+    return [endpoint.asdict() for endpoint in autopilot.get_endpoints()]
+
+
+@app.post("/endpoint", response_model=bool)
+def create_new_endpoint(endpoint: Endpoint = Body(...)) -> Any:
+    if autopilot.add_endpoint(endpoint):
```
patrickelectric, line 36: `return autopilot.add_endpoint(endpoint)`

core/services/ardupilot_manager/main.py
```diff
@@ -1,6 +1,50 @@
...
+        return True
+    return False
+
+
+@app.delete("/endpoint", response_model=bool)
+def remove_endpoint(endpoint: Endpoint = Body(...)) -> Any:
+    if autopilot.remove_endpoint(endpoint):
```
patrickelectric, line 43: `return autopilot.remove_endpoint(endpoint)`

patrickelectric (changes requested):
Just a small review, will do a deeper one tomorrow

core/services/ardupilot_manager/main.py
rafaellehmkuhl: I wonder how I still do that...

patrickelectric (commented):
I did run python setup.py install and getting:
```
./main.py                                                         
Traceback (most recent call last):
  File "/home/patrick/git/blue/companion-docker/core/services/ardupilot_manager/./main.py", line 7, in <module>
    from fastapi import Body, FastAPI
  File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/__init__.py", line 7, in <module>
    from .applications import FastAPI as FastAPI
  File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/applications.py", line 18, in <module>
    from fastapi.openapi.utils import get_openapi
  File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/openapi/utils.py", line 17, in <module>
    from fastapi.responses import Response
  File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/responses.py", line 10, in <module>
    from starlette.responses import UJSONResponse as UJSONResponse  # noqa
ImportError: cannot import name 'UJSONResponse' from 'starlette.responses' (/home/patrick/.local/lib/python3.9/site-packages/starlette/responses.py)
```

rafaellehmkuhl:
> I did run python setup.py install and getting:
> 
> ```
> ./main.py                                                         
> Traceback (most recent call last):
>   File "/home/patrick/git/blue/companion-docker/core/services/ardupilot_manager/./main.py", line 7, in <module>
>     from fastapi import Body, FastAPI
>   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/__init__.py", line 7, in <module>
>     from .applications import FastAPI as FastAPI
>   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/applications.py", line 18, in <module>
>     from fastapi.openapi.utils import get_openapi
>   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/openapi/utils.py", line 17, in <module>
>     from fastapi.responses import Response
>   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/responses.py", line 10, in <module>
>     from starlette.responses import UJSONResponse as UJSONResponse  # noqa
> ImportError: cannot import name 'UJSONResponse' from 'starlette.responses' (/home/patrick/.local/lib/python3.9/site-packages/starlette/responses.py)
> ```

Could you run `pip list` and send it here?

patrickelectric:
> > I did run python setup.py install and getting:
> > ```
> > ./main.py                                                         
> > Traceback (most recent call last):
> >   File "/home/patrick/git/blue/companion-docker/core/services/ardupilot_manager/./main.py", line 7, in <module>
> >     from fastapi import Body, FastAPI
> >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/__init__.py", line 7, in <module>
> >     from .applications import FastAPI as FastAPI
> >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/applications.py", line 18, in <module>
> >     from fastapi.openapi.utils import get_openapi
> >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/openapi/utils.py", line 17, in <module>
> >     from fastapi.responses import Response
> >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/responses.py", line 10, in <module>
> >     from starlette.responses import UJSONResponse as UJSONResponse  # noqa
> > ImportError: cannot import name 'UJSONResponse' from 'starlette.responses' (/home/patrick/.local/lib/python3.9/site-packages/starlette/responses.py)
> > ```
> 
> Could you run `pip list` and send it here?
> 
> Edit: there was a change in starlette and [they removed UJSONResponse in the newest version](https://github.com/tiangolo/fastapi/issues/2342), but this should not occur here, since FastAPI froze starlette's version on 0.13.6 for backwards compatibility.

image

rafaellehmkuhl:
> > > I did run python setup.py install and getting:
> > > ```
> > > ./main.py                                                         
> > > Traceback (most recent call last):
> > >   File "/home/patrick/git/blue/companion-docker/core/services/ardupilot_manager/./main.py", line 7, in <module>
> > >     from fastapi import Body, FastAPI
> > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/__init__.py", line 7, in <module>
> > >     from .applications import FastAPI as FastAPI
> > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/applications.py", line 18, in <module>
> > >     from fastapi.openapi.utils import get_openapi
> > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/openapi/utils.py", line 17, in <module>
> > >     from fastapi.responses import Response
> > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/responses.py", line 10, in <module>
> > >     from starlette.responses import UJSONResponse as UJSONResponse  # noqa
> > > ImportError: cannot import name 'UJSONResponse' from 'starlette.responses' (/home/patrick/.local/lib/python3.9/site-packages/starlette/responses.py)
> > > ```
> > 
> > 
> > Could you run `pip list` and send it here?
> > Edit: there was a change in starlette and [they removed UJSONResponse in the newest version](https://github.com/tiangolo/fastapi/issues/2342), but this should not occur here, since FastAPI froze starlette's version on 0.13.6 for backwards compatibility.
> 
> image

There's probably some other package in your environment taking starlette to an upper version. 
The github action is installing the correct one, so the build is fine:

```bash
Best match: starlette 0.13.6
Processing starlette-0.13.6-py3-none-any.whl
Installing starlette-0.13.6-py3-none-any.whl to /home/runner/.local/lib/python3.7/site-packages
Adding starlette 0.13.6 to easy-install.pth file
```

I suggest using a dedicated python virtual environment to isolate package's versions or downgrading your version of starlette to 0.13.6.

patrickelectric:
> > > > I did run python setup.py install and getting:
> > > > ```
> > > > ./main.py                                                         
> > > > Traceback (most recent call last):
> > > >   File "/home/patrick/git/blue/companion-docker/core/services/ardupilot_manager/./main.py", line 7, in <module>
> > > >     from fastapi import Body, FastAPI
> > > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/__init__.py", line 7, in <module>
> > > >     from .applications import FastAPI as FastAPI
> > > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/applications.py", line 18, in <module>
> > > >     from fastapi.openapi.utils import get_openapi
> > > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/openapi/utils.py", line 17, in <module>
> > > >     from fastapi.responses import Response
> > > >   File "/home/patrick/.local/lib/python3.9/site-packages/fastapi-0.63.0-py3.9.egg/fastapi/responses.py", line 10, in <module>
> > > >     from starlette.responses import UJSONResponse as UJSONResponse  # noqa
> > > > ImportError: cannot import name 'UJSONResponse' from 'starlette.responses' (/home/patrick/.local/lib/python3.9/site-packages/starlette/responses.py)
> > > > ```
> > > 
> > > 
> > > Could you run `pip list` and send it here?
> > > Edit: there was a change in starlette and [they removed UJSONResponse in the newest version](https://github.com/tiangolo/fastapi/issues/2342), but this should not occur here, since FastAPI froze starlette's version on 0.13.6 for backwards compatibility.
> > 
> > 
> > image
> 
> There's probably some other package in your environment taking starlette to an upper version.
> The github action is installing the correct one, so the build is fine:
> 
> ```shell
> Best match: starlette 0.13.6
> Processing starlette-0.13.6-py3-none-any.whl
> Installing starlette-0.13.6-py3-none-any.whl to /home/runner/.local/lib/python3.7/site-packages
> Adding starlette 0.13.6 to easy-install.pth file
> ```
> 
> I suggest using a dedicated python virtual environment to isolate package's versions or downgrading your version of starlette to 0.13.6.

Moved to a virtualenv solves the problem.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,56 @@
...
+class Endpoint:
+    connection_type: str
+    place: str
+    argument: int
+
+    @root_validator
+    @classmethod
+    def is_mavlink_endpoint(cls: Type["Endpoint"], values: Any) -> Any:
+        connection_type, place, argument = (values.get("connection_type"), values.get("place"), values.get("argument"))
+        if connection_type in [EndpointType.UDPServer, EndpointType.UDPClient, EndpointType.TCP]:
```
patrickelectric, line 30: Can we return the possible options when connection_type has the wrong argument ?

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,56 @@
...
+            if validators.domain(place) or validators.ipv4(place) or validators.ipv6(place):
+                if argument in range(1, 65536):
+                    return values
+                raise ValueError(f"Ports must be in the range 1:65535. Received {argument}.")
+            raise ValueError(f"Invalid network address: {place}")
+
+        if connection_type == EndpointType.Serial.value:
+            if is_valid_filepath(place):
+                if argument in range(0, 10000001):
```
patrickelectric, line 39: This should allow the following baudrates:
```
3000000
2000000,
1000000,
921600,
570600,
460800,
257600,
250000,
230400,
115200,
 57600,
 38400,
 19200,
  9600,
```
And 115200 when place is zero since this is the default baudrate

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,56 @@
...
+                    return values
+                raise ValueError(f"Serial baudrate must be in the range 0:10000000. Received {argument}.")
+            raise ValueError(f"Bad serial address: {place}.\nMake sure to use an absolute path.")
+
+        if connection_type == EndpointType.File.value:
+            if is_valid_filepath(place):
+                if argument in range(-1, 101):
```
patrickelectric, line 46: What is argument for file endpoint ?

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -13,17 +15,32 @@
...
+        Endpoint.is_mavlink_endpoint(endpoint_values)
+    endpoint_values = {"connection_type": "file", "place": "/path/to/file", "argument": -30}
+    with pytest.raises(ValueError):
+        Endpoint.is_mavlink_endpoint(endpoint_values)
+    endpoint_values = {"connection_type": "file", "place": "path/to/file", "argument": 100}
+    with pytest.raises(ValueError):
+        Endpoint.is_mavlink_endpoint(endpoint_values)
+    endpoint_values = {"connection_type": "potato", "place": "path/to/file", "argument": 100}
+    with pytest.raises(ValueError):
+        Endpoint.is_mavlink_endpoint(endpoint_values)
```
patrickelectric, line 43: Can you add tests for valid endpoint configuration ?

mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
+plugins = pydantic.mypy
```
patrickelectric, line 2: Why is it here ?

patrickelectric (changes requested):
When you post an endpoint that already exists, it returns false but with a number code of 200.
Is that the expected behaviour ? Should we return 200 or true/false at all ? Maybe 409 (Conflict) ?

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: I was using 100000000 as max because that's what mavproxy uses, but agree that it's better to have available options. I Will add this on another file thought, because there are other packages that would benefit from this.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Yes, that would be good.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Stream rate, or I got it wrong?

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: Sure.

mypy.ini
rafaellehmkuhl: Long-story-short: mypy opines on some specific points (in that case, class __init__) that causes it to misinterpret some of pydantic behaviours, causing errors that are not true. Pydantic has a mypy plugin to fix that.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
patrickelectric: a file endpoint is just a place where we are going to store any incoming message, stream rate is not an option that I can see to be used and how to be used.

mypy.ini
patrickelectric: This should global and not per project

rafaellehmkuhl: Didn't get it.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Mistook it with the idea of a log file again.

mypy.ini
patrickelectric: We should avoid personal specific files in the repository, such as IDE configuration files and etc.
If this file is related to mypy checks used in CI, it should exist in the global folder of this project and not in this service only.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,82 @@
...
-class EndpointType(IntEnum):
-    Undefined = 0
-    UDPServer = 1
-    UDPClient = 2
-    TCP = 3
-    Serial = 4
-    File = 5
 
+def is_valid_filepath(filepath: str) -> bool:
+    return filepath.startswith("/") and not filepath.endswith("/")
```
patrickelectric, line 9: We should save the fine in a predefine, having the filename as a global thing may be a bad idea, since this application will run as root, the user may overwrite important files for the operating system or any other file in the companion!
We should only accept filenames and save the files in a specific folder like `Documents/companion/ArduPilotManager/file_endpoints/file`

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,82 @@
...
+            raise ValueError(f"Bad serial address: {place}. Make sure to use an absolute path.")
+
+        if connection_type == EndpointType.File.value:
+            if is_valid_filepath(place):
+                if argument is None:
+                    return values
+                raise ValueError(f"File endpoint has should have no argument. Received {argument}.")
+            raise ValueError(f"Bad file path: {place}. Make sure to use an absolute path.")
+        raise ValueError(
+            f"Invalid connection_type: {connection_type}. Valid types are: udpin, udpout, tcp, serial and file."
```
patrickelectric, line 53: valid types should be populated from enum and not hardcoded, doing like that makes it harder to maintain and also duplicates the information unnecessary.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,82 @@
...
+                    return values
+                raise ValueError(f"Ports must be in the range 1:65535. Received {argument}.")
+            raise ValueError(f"Invalid network address: {place}")
+
+        if connection_type == EndpointType.Serial.value:
+            if is_valid_filepath(place):
+                if argument in VALID_SERIAL_BAUDRATES:
+                    return values
+                raise ValueError(
+                    f"Invalid serial baudrate. Valid option are{VALID_SERIAL_BAUDRATES}. Received {argument}."
```
patrickelectric, line 42: Space between are and {

patrickelectric (changes requested):

mypy.ini
rafaellehmkuhl: But it is in the global folder.

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
rafaellehmkuhl: Agreed!

rafaellehmkuhl: One question that I make myself, though, is "where should this be enforced?"

From my point of view, we can enforce in the Endpoint class itself, making in a way that the file endpoint should always have just a filename, not a complete path, and the ArduPilotManager takes care of completing the path on its way.

Also, for organization purposes, I'm thinking about storing just the endpoint filename in the settings file, so that ArduPilotManager always composes the complete path when loading the files (not when storing).

What do you think?

mypy.ini
patrickelectric: Right, sorry for that.
We still need to figure a better way to use that, maybe with the hook script ?
Right now mypy does not support pyproject, so I'm afraid that this project ends up with a bunch of files in the global folder for python tools configuration.
@Williangalvani any tip ?

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,82 @@
...
+                )
+            raise ValueError(f"Bad serial address: {place}. Make sure to use an absolute path.")
+
+        if connection_type == EndpointType.File.value:
```
patrickelectric, line 46: maybe inverse logic to avoid the nested ifs

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,82 @@
...
+    def is_mavlink_endpoint(cls: Type["Endpoint"], values: Any) -> Any:
+        connection_type, place, argument = (values.get("connection_type"), values.get("place"), values.get("argument"))
+        if connection_type in [EndpointType.UDPServer, EndpointType.UDPClient, EndpointType.TCP]:
+            if validators.domain(place) or validators.ipv4(place) or validators.ipv6(place):
+                if argument in range(1, 65536):
+                    return values
+                raise ValueError(f"Ports must be in the range 1:65535. Received {argument}.")
+            raise ValueError(f"Invalid network address: {place}")
+
+        if connection_type == EndpointType.Serial.value:
```
patrickelectric, line 37: maybe inverse logic to avoid the nested ifs

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,76 +1,82 @@
...
+class Endpoint:
+    connection_type: str
+    place: str
+    argument: Optional[int] = None
+
+    @root_validator
+    @classmethod
+    def is_mavlink_endpoint(cls: Type["Endpoint"], values: Any) -> Any:
+        connection_type, place, argument = (values.get("connection_type"), values.get("place"), values.get("argument"))
+        if connection_type in [EndpointType.UDPServer, EndpointType.UDPClient, EndpointType.TCP]:
```
patrickelectric, line 30: maybe inverse logic to avoid the nested ifs

mypy.ini
patrickelectric: Also, should we add such plugin in our CI to use it ?

rafaellehmkuhl: > Right, sorry for that.
> We still need to figure a better way to use that, maybe with the hook script ?
> Right now mypy does not support pyproject, so I'm afraid that this project ends up with a bunch of files in the global folder for python tools configuration.
> @Williangalvani any tip ?

No prob.
As I understand, mypy is probably one of the few, if not the only, tool that [does not support or plan to support pyproject.toml](https://github.com/python/mypy/issues/5205), so we can have only pyproject.toml and mypy.ini.

> Also, should we add such plugin in our CI to use it ?

No need. Pydantic is already a sub-dependency of fastapi.

patrickelectric (approved):

Williangalvani (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -118,23 +118,33 @@ def _load_endpoints(self) -> None:
...
-            self.mavlink_manager.add_endpoints([Endpoint(e) for e in endpoints])
+        for endpoint in endpoints:
+            if not self.add_endpoint(Endpoint(**endpoint)):
+                warn(f"Could not lot endpoint {endpoint}")
 
     def get_endpoints(self) -> List[Endpoint]:
         return self.mavlink_manager.endpoints()
 
     def add_endpoint(self, endpoint: Endpoint) -> bool:
+        if endpoint.connection_type == EndpointType.File.value:
```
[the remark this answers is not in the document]
rafaellehmkuhl: Basically, add_new_endpoint will append it to the settings file, while the other does not (adds just to mavlink_manager).

Will add some comments.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -81,8 +81,6 @@ def start_navigator(self) -> None:
 
         # Check if subprocess is running and wait until it finishes
```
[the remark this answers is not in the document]
rafaellehmkuhl: Good catch

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -118,23 +118,33 @@ def _load_endpoints(self) -> None:
             self.configuration["endpoints"] = []
         else:
             endpoints = self.configuration["endpoints"]
-            self.mavlink_manager.add_endpoints([Endpoint(e) for e in endpoints])
+        for endpoint in endpoints:
+            if not self.add_endpoint(Endpoint(**endpoint)):
+                warn(f"Could not lot endpoint {endpoint}")
```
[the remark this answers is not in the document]
rafaellehmkuhl: Load lol

core/services/ardupilot_manager/ArduPilotManager.py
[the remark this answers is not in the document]
rafaellehmkuhl: Basically the add_new_endpoint also appends the dict to the settings file. The other just adds to the mavlink manager.

Will add comments on that.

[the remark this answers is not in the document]
Williangalvani: maybe `add_and_save_endpoint()` would be a better fitting name, or perhaps `add_endpoint_and_save()`?

core/services/ardupilot_manager/mavlink_proxy/Endpoint.py
```diff
@@ -1,62 +1,78 @@
...
+            if not argument in VALID_SERIAL_BAUDRATES:
+                raise ValueError(f"Invalid serial baudrate: {argument}. Valid option are {VALID_SERIAL_BAUDRATES}.")
+            return values
 
-        return self
+        if connection_type == EndpointType.File.value:
+            if "/" in place:
+                raise ValueError(f"Bad filename: {place}. Valid filenames do not contain '/' characters.")
+            if argument is not None:
+                raise ValueError(f"File endpoint has should have no argument. Received {argument}.")
```
jaxxzer, line 45: File endpoint ~has~ should have no argument.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -79,14 +79,11 @@ def start_navigator(self) -> None:
         # does not accept TCP master endpoints
         # self.start_mavlink_manager(Endpoint(local_endpoint))
 
-        # Check if subprocess is running and wait until it finishes
-        # Necessary since we don't have mavlink_manager running for navigator yet
```
jaxxzer, line 83: minor detail, you don't need to fix it, but these comments should have come out in the previous commit :)

jaxxzer (approved):

PR #120, opened 2021-03-22
bootstrap: Fix bind to .config

patrickelectric:
This allows us to store and load companion-core settings

### DIFF
--- a/bootstrap/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap/bootstrap.py
@@ -14,7 +14,7 @@ class Bootstrapper:
 
     DEFAULT_FILE_PATH = pathlib.Path("/bootstrap/startup.json.default")
     DOCKER_CONFIG_PATH = pathlib.Path("/root/.config")
-    DOCKER_CONFIG_FILE_PATH = pathlib.Path(DOCKER_CONFIG_PATH, "startup.json")
+    DOCKER_CONFIG_FILE_PATH = DOCKER_CONFIG_PATH.joinpath("bootstrap/startup.json")
     HOST_CONFIG_PATH = os.environ.get("COMPANION_CONFIG_PATH", "/tmp/companion/.config")
     CORE_CONTAINER_NAME = "companion-core"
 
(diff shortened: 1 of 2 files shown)


bootstrap/startup.json.default
```diff
@@ -18,6 +18,10 @@
                 "bind": "/tmp/wpa_playground",
                 "mode": "rw"
             }
+            "/root/.config/companion-core": {
+                "bind": "/root/.config",
+                "mode": "rw"
+            }
```
rafaellehmkuhl, line 24: Just to be sure, this would be for companion-core general settings, right?
Because the settings for the different services are currently under `/roor/config/{service.name}`

Williangalvani:
There's already some code in bootstrap to set the config folder, check https://github.com/bluerobotics/companion-docker/blob/master/bootstrap/main.py#L16.
The config folder should be mounted the usual way AND passed as an environment variable so bootstrap can mount it for companion-core too (if we don't pass it as an environment variable, bootstrap doesn't know where the settings are in the host and can't mount them in the containers it launches).

patrickelectric:
> There's already some code in bootstrap to set the config folder, check https://github.com/bluerobotics/companion-docker/blob/master/bootstrap/main.py#L16.
> The config folder should be mounted the usual way AND passed as an environment variable so bootstrap can mount it for companion-core too (if we don't pass it as an environment variable, bootstrap doesn't know where the settings are in the host and can't mount them in the containers it launches).

Thanks for pointing,  it folder exists but there is nothing there besides `startup.jason`, I'll investigate it more tomorrow

bootstrap/startup.json.default
patrickelectric: yep, but this was the wrong approach as willian comment points:
https://github.com/bluerobotics/companion-docker/pull/120#issuecomment-804425202

patrickelectric:
@Williangalvani let me know if the new approaches is better, the final output is the following:
```
.config
└── companion
    ├── ardupilot-manager
    │   ├── firmware
    │   └── settings.json
    ├── bootstrap
    │   └── startup.json
    └── mavlink-camera-manager
        └── settings.json

```

jaxxzer (approved):

PR #125, opened 2021-03-25
core: start-companion-core: Use BlueROV default settings for  mavlink-camera-manager

patrickelectric:
- It'll generate a stream per camera that provides H264 stream, each stream will be a udp://192.168.2.1:(5600 + stream_index)


jaxxzer:
What about:
- if there is only one camera connected at boot
- and there is no prior configuration
- then set up a  default stream for that camera

patrickelectric:
ping @jaxxzer

Williangalvani (approved):
I like this solution

PR #129, opened 2021-03-26
Fixup

### LINKED ISSUE #128: install steps are creating bad urls
When using custom remote + version:

in [step 1](https://github.com/bluerobotics/companion-docker/blob/master/install/install.sh#L6) install.sh, we set `$REMOTE` to `github.com/branch`.
in [step 2](https://github.com/bluerobotics/companion-docker/blob/master/install/boards/configure_board.sh#L7) configure board, we set `$REMOTE` to `github.com/branch/branch`, this leads to curl 404

jaxxzer:
fix #128 
fix a problem with the spi0-dts download

### DIFF
--- a/install/boards/bcm_27xx.sh
+++ b/install/boards/bcm_27xx.sh
@@ -5,9 +5,10 @@ echo "Configuring BCM27XX board (Raspberry Pi 4).."
 # Download, compile, and install spi0 mosi-only device tree overlay for
 # neopixel LED on navigator board
 echo "- compile spi0 device tree overlay."
-DTS_PATH="$REMOTE/install/overlays"
-curl -fsSL $DTS_PATH/spi0-led.dts /tmp/
-dtc -@ -Hepapr -I dts -O dtb -o /boot/overlays/spi0-led.dtbo /tmp/spi0-led.dts
+DTS_PATH="$REMOTE/$VERSION/install/overlays"
+DTS_NAME="spi0-led.dts"
+curl -fsSL -o /tmp/$DTS_NAME $DTS_PATH/$DTS_NAME
+dtc -@ -Hepapr -I dts -O dtb -o /boot/overlays/spi0-led.dtbo /tmp/$DTS_NAME
 
 # Remove any configuration related to i2c and spi/spi1 and do the necessary changes for navigator
 echo "- Enable I2C, SPI and UART."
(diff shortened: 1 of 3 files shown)


jaxxzer:
In testing I ran into #130

install/boards/bcm_27xx.sh
```diff
@@ -5,9 +5,10 @@ echo "Configuring BCM27XX board (Raspberry Pi 4).."
 # Download, compile, and install spi0 mosi-only device tree overlay for
 # neopixel LED on navigator board
 echo "- compile spi0 device tree overlay."
-DTS_PATH="$REMOTE/install/overlays"
-curl -fsSL $DTS_PATH/spi0-led.dts /tmp/
-dtc -@ -Hepapr -I dts -O dtb -o /boot/overlays/spi0-led.dtbo /tmp/spi0-led.dts
+DTS_PATH="$REMOTE/$VERSION/install/overlays"
```
patrickelectric, line 8: This will not work once merged, these variables will not exist when running the install.sh since they don't exist in the environment variable.
Check #131 solution

jaxxzer: > This will not work once merged, these variables will not exist when running the install.sh, they don't exist in the environment variable by default.

Ok.

I can't test for a couple of hours, but to me it doesn't look like #131 will fix the problem where bcm27xx.sh is not downloaded. In this comment, the url it is trying for bcm27xx.sh download is wrong: https://github.com/bluerobotics/companion-docker/issues/128#issuecomment-807890363

patrickelectric: Yes, you are correct, but once merged this will not work since REMOTE will not be defined, as we found out, REMOTE and VERSION will only be available/updated if the variable is already defined in the environment variable.
```sh
TEST="POTATO"
REMOTE=$REMOTE/$REMOTE

echo "in 1: $REMOTE"
echo "in 2 test: $REMOTE"

sh test2.sh
```
```sh
#test2.sh
echo "in 2: $REMOTE"
echo "in 2 test: $TEST"
```

patrickelectric: You can cherry-pick it or merge the other PR to fix that and rebase this patch.

patrickelectric: We are going to merge this now and fix the unrelated problems in others PRs

Issue #144, opened 2021-04-02
udpout autopilot connections are not working


jaxxzer:
The configuration is loaded, but I am not getting any data on the other side

```
root@companion:/# /home/pi/services/ardupilot_manager/main.py
Loaded settings from /root/.config/ardupilot-manager/settings.json:
{'endpoints': [{'argument': 14550,
                'connection_type': 'udpout',
                'place': '192.168.2.1'}]}
```

jaxxzer:
The default tcp 5760 port is working, the default and settings configured udp ports are not showing as in use:
```
root@companion:/home/pi# netstat -tulpn
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN      972/python3         
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      555/sshd            
tcp        0      0 0.0.0.0:8000            0.0.0.0:*               LISTEN      959/python3         
tcp        0      0 0.0.0.0:5760            0.0.0.0:*               LISTEN      1158/ardusub        
tcp        0      0 0.0.0.0:9090            0.0.0.0:*               LISTEN      948/python3         
tcp        0      0 0.0.0.0:6020            0.0.0.0:*               LISTEN      1310/mavlink-camera 
tcp        0      0 0.0.0.0:9000            0.0.0.0:*               LISTEN      953/python3         
tcp6       0      0 :::22                   :::*                    LISTEN      555/sshd            
udp        0      0 0.0.0.0:68              0.0.0.0:*                           545/dhcpcd          
udp        0      0 0.0.0.0:50797           0.0.0.0:*                           1310/mavlink-camera 
udp        0      0 0.0.0.0:57554           0.0.0.0:*                           1310/mavlink-camera 
udp        0      0 0.0.0.0:5353            0.0.0.0:*                           397/avahi-daemon: r 
udp        0      0 0.0.0.0:60960           0.0.0.0:*                           397/avahi-daemon: r 
udp6       0      0 :::5353                 :::*                                397/avahi-daemon: r 
udp6       0      0 :::57648                :::*                                1310/mavlink-camera 
udp6       0      0 :::57371                :::*                                397/avahi-daemon: r 
root@companion:/home/pi# netstat -tupn
Active Internet connections (w/o servers)
[...]

PR #148, opened 2021-04-05
core: ardupilot-manager: Fix adding/removing endpoints

rafaellehmkuhl:
Restart router after adding or removing endpoints, so they can be loaded.


core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -138,7 +138,7 @@ def add_new_endpoint(self, endpoint: Endpoint) -> bool:
             self.configuration.get("endpoints", []).remove(endpoint.asdict())
             return False
         self.settings.save(self.configuration)
-        return True
+        return self.restart()
```
patrickelectric, line 141: I think that is not a good solution, the API should request restart, otherwise if someone wants to add/remove multiple endpoints will end up with multiple resets of the tool. We should restart only once and when is really required.

patrickelectric (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: We can go with one of two paths:
1. Expose an endpoint (on the REST API) for manager restarting (meaning that the frontend will need to know when to call the restart).
2. Modifying the method on the manager to allow multiple endpoints to be added once (as it was before), an also it's corresponding method on the API (meaning the API would allow for multiple endpoints to be added once).

I like the second one more, as I think the frontend should not know how the internals of the manager work. Seems more reasonable.

patrickelectric: The front end should not know, but if the user changes a bunch of endpoints, adding and removing, a simple "apply" button will do the restart after doing all deletes/creations

rafaellehmkuhl: Yes! I was thinking about storing them on the frontend while being added and as soon as the user hits "apply", the API will be called with multiple endpoints at once. This makes everything very clear and we will have just one interface.

core/services/ardupilot_manager/main.py
```diff
@@ -31,16 +31,20 @@ def get_available_endpoints() -> Any:
...
-@app.delete("/endpoint", status_code=status.HTTP_204_NO_CONTENT)
-def remove_endpoint(response: Response, endpoint: Endpoint = Body(...)) -> Any:
-    if not autopilot.remove_endpoint(endpoint):
-        response.status_code = status.HTTP_404_NOT_FOUND
+@app.post("/endpoints", status_code=status.HTTP_201_CREATED)
+def create_endpoints(response: Response, endpoints: List[Endpoint] = Body(...)) -> Any:
+    for endpoint in endpoints:
+        if not autopilot.add_new_endpoint(endpoint):
+            response.status_code = status.HTTP_409_CONFLICT
+    autopilot.restart()
```
patrickelectric, line 39: will this restart if status code is 409 ?

rafaellehmkuhl: Yes, because status code 409 means that there was a conflict adding one or more endpoints, but not all of them, so still necessary to restart.

The other option would be to add zero or all (if any add results in conflict, add none of them, and don't restart).

patrickelectric: - We should at least say which one fail
- If everything fails the system will also restart
- I'm a bit confuse if "partial accept" is a ok thing to do in an API

rafaellehmkuhl: I'm adding the following handling:

- If there's at least one failing endpoint operation, it will report and return the failing endpoints
- If there's more than one successful endpoint, it will restart

What do you think?

core/services/ardupilot_manager/ArduPilotManager.py
jaxxzer: I think that start/restart/stop routing is not something that needs to be exposed to the api/frontend. It is an implementation detail to be handled by the backend. We can eventually implement our own router that has no problem dynamically adding/removing endpoints and so a start/restart will not be required in the backend.

I think @rafaellehmkuhl suggestion to have the API to add multiple endpoints in a single request is the best solution to avoid unnecessary restarts of the backend tool.

core/services/ardupilot_manager/main.py
jaxxzer: @patrickelectric we talked, and think that it should be a 100% pass/fail. If there is a problem with a single part of the configuration, we won't apply any of it.

rafaellehmkuhl:
Although we now have a working version that can be merged, there's some work I still want to do on the manager:

- Endpoints can be made hashable, so we can use sets for storing and prevent duplicates by default
- Error handling right now is not ideal. If we improve it from the deeper classes, we can have better error messages for the user
- There's no clear source of truth for the endpoints right now. Is it the manager or the settings file?

This can be done now or after other priorities. @jaxxzer what do you think?

rafaellehmkuhl:
Fix #144

jaxxzer (approved):

jaxxzer:
I think we can do these things one patch at a time in next PRs and we will be able to address the outstanding problems and other priorities soon anyway. Please file issues for them.

patrickelectric (approved):

PR #152, opened 2021-04-07
Fix and extend endpoint types

rafaellehmkuhl:
- Allow for TCP client endpoints 
- Fix creation of serial endpoints on MAVProxy
- Fix validation for both MAVProxy and MavlinkRouter
- Simplify some syntax

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -116,8 +116,11 @@ class ArduPilotManager(metaclass=Singleton):
             self.configuration["endpoints"] = []
         endpoints = self.configuration["endpoints"]
         for endpoint in endpoints:
-            if not self.mavlink_manager.add_endpoint(Endpoint(**endpoint)):
-                warn(f"Could not load endpoint {endpoint}")
+            try:
+                if not self.mavlink_manager.add_endpoint(Endpoint(**endpoint)):
+                    raise ValueError("Cannot add endpoint.")
+            except Exception as error:
+                warn(f"Could not load endpoint {endpoint}. {error}")
 
     def _reset_endpoints(self, endpoints: List[Endpoint]) -> None:
         try:
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -10,7 +10,7 @@ from warnings import warn
 from firmware_download.FirmwareDownload import FirmwareDownload, Vehicle
 from flight_controller_detector.Detector import Detector as BoardDetector
 from flight_controller_detector.Detector import FlightControllerType
-from mavlink_proxy.Endpoint import Endpoint, EndpointType
+from mavlink_proxy.Endpoint import Endpoint
 from mavlink_proxy.Manager import Manager as MavlinkManager
 from settings import Settings
 from Singleton import Singleton
@@ -142,10 +142,6 @@ class ArduPilotManager(metaclass=Singleton):
 
     def add_new_endpoints(self, new_endpoints: List[Endpoint]) -> bool:
         """Add multiple endpoints to the mavlink manager and save them on the configuration file."""
-        for endpoint in new_endpoints:
-            if endpoint.connection_type == EndpointType.File.value:
-                endpoint.place = os.path.join(self.settings.file_endpoints_path, endpoint.place)
-
         saved_endpoints = deepcopy(self.configuration["endpoints"])
         loaded_endpoints = deepcopy(self.get_endpoints())
 
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
@@ -29,20 +29,25 @@ class MAVLinkRouter(AbstractRouter):
                 return f"--tcp-port {endpoint.argument}"
             if endpoint.connection_type == EndpointType.TCPClient:
                 return f"--tcp-endpoint {endpoint.place}:{endpoint.argument}"
-            return f"{endpoint.place}:{endpoint.argument}"
+            if endpoint.connection_type == EndpointType.UDPClient:
+                return f"--endpoint {endpoint.place}:{endpoint.argument}"
+            raise NotImplementedError(f"Endpoint of type {endpoint.connection_type} not supported on MavlinkRouter.")
 
-        endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
+        endpoints = " ".join([convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
         if master.connection_type not in [
             EndpointType.UDPServer,
-            EndpointType.UDPClient,
             EndpointType.Serial,
             EndpointType.TCPServer,
         ]:
             raise NotImplementedError
 
         log = f"--log {self.logdir().resolve()}"
-        return f"{self.binary()} {convert_endpoint(master)} {log} {endpoints}"
+
+        if master.connection_type == EndpointType.TCPServer:
+            return f"{self.binary()} --tcp-port {master.argument} {log} {endpoints}"
+
+        return f"{self.binary()} {master.place}:{master.argument} --tcp-port 0 {log} {endpoints}"
 
     @staticmethod
     def name() -> str:
@@ -55,9 +60,9 @@ class MAVLinkRouter(AbstractRouter):
     @staticmethod
     def _validate_endpoint(endpoint: Endpoint) -> bool:
         return endpoint.connection_type in [
-            EndpointType.UDPServer,
             EndpointType.UDPClient,
-            EndpointType.Serial,
+            EndpointType.TCPServer,
+            EndpointType.TCPClient,
         ]
 
     @staticmethod
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
@@ -26,11 +26,10 @@ class MAVLinkRouter(AbstractRouter):
         def convert_endpoint(endpoint: Endpoint) -> str:
             # TCP uses a special argument and only works with localhost
             if endpoint.connection_type == EndpointType.TCPServer:
-                port = str(endpoint).split(":")[2]
-                return f"-t {port}"
+                return f"--tcp-port {endpoint.argument}"
             if endpoint.connection_type == EndpointType.TCPClient:
                 return f"--tcp-endpoint {endpoint.place}:{endpoint.argument}"
-            return str(endpoint)[str(endpoint).find(":") + 1 :]
+            return f"{endpoint.place}:{endpoint.argument}"
 
         endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
@@ -24,11 +24,13 @@ class MAVLinkRouter(AbstractRouter):
     def assemble_command(self, master: Endpoint) -> str:
         # Convert endpoint format to mavlink-router format
         def convert_endpoint(endpoint: Endpoint) -> str:
-            if endpoint.connection_type != EndpointType.TCP:
-                return str(endpoint)[str(endpoint).find(":") + 1 :]
             # TCP uses a special argument and only works with localhost
-            port = str(endpoint).split(":")[2]
-            return f"-t {port}"
+            if endpoint.connection_type == EndpointType.TCPServer:
+                port = str(endpoint).split(":")[2]
+                return f"-t {port}"
+            if endpoint.connection_type == EndpointType.TCPClient:
+                return f"--tcp-endpoint {endpoint.place}:{endpoint.argument}"
+            return str(endpoint)[str(endpoint).find(":") + 1 :]
 
         endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
@@ -36,7 +38,7 @@ class MAVLinkRouter(AbstractRouter):
             EndpointType.UDPServer,
             EndpointType.UDPClient,
             EndpointType.Serial,
-            EndpointType.TCP,
+            EndpointType.TCPServer,
         ]:
             raise NotImplementedError
 
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
@@ -21,17 +21,22 @@ class MAVProxy(AbstractRouter):
         return None
 
     def assemble_command(self, master: Endpoint) -> str:
-        endpoints = " ".join([f"--out={str(endpoint)}" for endpoint in self.endpoints()])
+        endpoints_string = ""
+        for endpoint in self.endpoints():
+            if endpoint.connection_type != EndpointType.Serial:
+                endpoints_string += f" --out={str(endpoint)}"
+            else:
+                endpoints_string += f" --out={endpoint.place},{endpoint.argument}"
 
         master_string = str(master)
         if master.connection_type == EndpointType.Serial:
             if not master.argument:
                 master_string = f"{master.place}"
             else:
-                master_string = f"{master.place} --baudrate={master.argument}"
+                master_string = f"{master.place},{master.argument}"
 
         log = f"--state-basedir={self.logdir().resolve()}"
-        return f"{self.binary()} --master={master_string} {endpoints} {log} --non-interactive"
+        return f"{self.binary()} --master={master_string} {endpoints_string} {log} --non-interactive"
 
     @staticmethod
     def name() -> str:
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
@@ -48,7 +48,13 @@ class MAVProxy(AbstractRouter):
 
     @staticmethod
     def _validate_endpoint(endpoint: Endpoint) -> bool:
-        return True
+        return endpoint.connection_type in [
+            EndpointType.Serial,
+            EndpointType.UDPServer,
+            EndpointType.UDPClient,
+            EndpointType.TCPServer,
+            EndpointType.TCPClient,
+        ]
 
     @staticmethod
     def is_ok() -> bool:
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
@@ -21,15 +21,14 @@ class MAVProxy(AbstractRouter):
         return None
 
     def assemble_command(self, master: Endpoint) -> str:
-        endpoints = " ".join(["--out=" + endpoint.__str__() for endpoint in self.endpoints()])
+        endpoints = " ".join([f"--out={str(endpoint)}" for endpoint in self.endpoints()])
 
         master_string = str(master)
         if master.connection_type == EndpointType.Serial:
-            master_args = str(master).split(":")
-            if len(master_args) == 2:
-                master_string = f"{master_args[1]}"
+            if not master.argument:
+                master_string = f"{master.place}"
             else:
-                master_string = f"{master_args[1]} --baudrate={master_args[2]}"
+                master_string = f"{master.place} --baudrate={master.argument}"
 
         log = f"--state-basedir={self.logdir().resolve()}"
         return f"{self.binary()} --master={master_string} {endpoints} {log} --non-interactive"
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
@@ -1,6 +1,6 @@
 import re
 import subprocess
-from typing import Optional
+from typing import Callable, Optional
 
 from mavlink_proxy.AbstractRouter import AbstractRouter
 from mavlink_proxy.Endpoint import Endpoint, EndpointType
@@ -21,22 +21,25 @@ class MAVProxy(AbstractRouter):
         return None
 
     def assemble_command(self, master: Endpoint) -> str:
-        endpoints_string = ""
-        for endpoint in self.endpoints():
+        serial_endpoint_as_input: Callable[[Endpoint], str] = lambda endpoint: f"{endpoint.place},{endpoint.argument}"
+
+        # Convert endpoint format to mavproxy format
+        def convert_endpoint(endpoint: Endpoint) -> str:
             if endpoint.connection_type != EndpointType.Serial:
-                endpoints_string += f" --out={str(endpoint)}"
-            else:
-                endpoints_string += f" --out={endpoint.place},{endpoint.argument}"
+                return f"--out={str(endpoint)}"
+            return f"--out={serial_endpoint_as_input(endpoint)}"
+
+        endpoints = " ".join([convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
         master_string = str(master)
         if master.connection_type == EndpointType.Serial:
             if not master.argument:
                 master_string = f"{master.place}"
             else:
-                master_string = f"{master.place},{master.argument}"
+                master_string = f"{serial_endpoint_as_input(master)}"
 
         log = f"--state-basedir={self.logdir().resolve()}"
-        return f"{self.binary()} --master={master_string} {endpoints_string} {log} --non-interactive"
+        return f"{self.binary()} --master={master_string} {endpoints} {log} --non-interactive"
 
     @staticmethod
     def name() -> str:
--- a/core/services/ardupilot_manager/mavlink_proxy/test_all.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/test_all.py
@@ -98,5 +98,5 @@ def test_mavlink_router() -> None:
         endpoint_2,
     ], "Endpoint list does not match."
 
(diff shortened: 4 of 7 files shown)


core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -29,21 +29,20 @@ def convert_endpoint(endpoint: Endpoint) -> str:
...
 
-        endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
+        endpoints = " ".join([convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
         if master.connection_type not in [
             EndpointType.UDPServer,
-            EndpointType.UDPClient,
             EndpointType.Serial,
-            EndpointType.TCPServer,
-            EndpointType.TCPClient,
```
patrickelectric: Master should also be TCP, check sitl.

rafaellehmkuhl:
Fixes #60

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -29,21 +29,20 @@ def convert_endpoint(endpoint: Endpoint) -> str:
...
+            raise NotImplementedError(f"Endpoint of type {endpoint.connection_type} not supported on MavlinkRouter.")
 
-        endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
+        endpoints = " ".join([convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
         if master.connection_type not in [
             EndpointType.UDPServer,
-            EndpointType.UDPClient,
             EndpointType.Serial,
-            EndpointType.TCPServer,
```
patrickelectric: Why are you removing something that was added in f3ad4371971c3d70183c14d04b16d1b058c994c7 ?

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -24,24 +24,25 @@ def _get_version(self) -> Optional[str]:
...
-                return str(endpoint)[str(endpoint).find(":") + 1 :]
             # TCP uses a special argument and only works with localhost
-            port = str(endpoint).split(":")[2]
-            return f"-t {port}"
+            if endpoint.connection_type == EndpointType.TCPServer:
+                return f"-t {endpoint.argument}"
+            if endpoint.connection_type == EndpointType.TCPClient:
+                return f"-p {endpoint.place}:{endpoint.argument}"
+            if endpoint.connection_type == EndpointType.UDPServer:
+                return f"-e {endpoint.place}:{endpoint.argument}"
```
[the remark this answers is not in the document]
rafaellehmkuhl: Makes sense.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: Should not have entered f3ad437 in the first place. Will fix.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: MavlinkRouter says explicitly on the helper that the master endpoints are Serial or UDP. Have you tried it with TCP?

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -24,24 +24,25 @@ def _get_version(self) -> Optional[str]:
...
+            raise NotImplementedError(f"Endpoint of type {endpoint.connection_type} not supported on MavlinkRouter.")
 
-        endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
+        endpoints = " ".join([convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
         if master.connection_type not in [
             EndpointType.UDPServer,
-            EndpointType.UDPClient,
             EndpointType.Serial,
-            EndpointType.TCP,
```
patrickelectric, line 39: It supports TCPServer as master

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -24,24 +24,25 @@ def _get_version(self) -> Optional[str]:
...
+                return f"--endpoint {endpoint.place}:{endpoint.argument}"
+            raise NotImplementedError(f"Endpoint of type {endpoint.connection_type} not supported on MavlinkRouter.")
 
-        endpoints = " ".join(["--endpoint " + convert_endpoint(endpoint) for endpoint in self.endpoints()])
+        endpoints = " ".join([convert_endpoint(endpoint) for endpoint in self.endpoints()])
 
         if master.connection_type not in [
             EndpointType.UDPServer,
-            EndpointType.UDPClient,
             EndpointType.Serial,
```
patrickelectric, line 40: for anything other than TCPServer we need to set tcp-port to 0 to avoid the bind behaviour

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -24,24 +24,30 @@ def _get_version(self) -> Optional[str]:
...
-            EndpointType.TCP,
```
patrickelectric, line 39: Why TCPClient is not here but server is ?
mavlink-routerd can have a server as master and endpoints as clients

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -55,8 +61,8 @@ def binary_name() -> str:
     def _validate_endpoint(endpoint: Endpoint) -> bool:
         return endpoint.connection_type in [
             EndpointType.UDPServer,
-            EndpointType.UDPClient,
```
patrickelectric, line 58: Why UDPClient is not here, it's the `-e`, you can check that with `nc`

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -55,8 +61,8 @@ def binary_name() -> str:
...
-            EndpointType.Serial,
```
patrickelectric, line 59: Why UDPServer is here and Serial not if both are valid master endpoints ?

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -86,14 +81,14 @@ def test_mavlink_router() -> None:
     assert mavlink_router.set_logdir(pathlib.Path(".")), "Local path as MAVLinkRouter log directory failed."
     assert re.search(r"\d+", str(mavlink_router.version())) is not None, "Version does not follow pattern."
 
-    endpoint_1 = Endpoint("udpout", "0.0.0.0", 14551)
-    endpoint_2 = Endpoint("udpout", "0.0.0.0", 14552)
+    endpoint_1 = Endpoint("udpin", "0.0.0.0", 14551)
```
patrickelectric, line 84: mavlink-routerd accepts udp clients and not server

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: Can it have a TCP client as master?

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: This validation is for output endpoints, not master.

patrickelectric: That's my point, UDPClient is the endpoint and not UDPServer, UDPServer is master.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: I forgot to change there since our last talk.
Just to confirm, we concluded that the `-e` creates an output UDPClient, and is not possible to use UDPServer as output, right?

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
patrickelectric: Oh this is only for master connection type validation.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
patrickelectric: yes, you can test with netcat.

rafaellehmkuhl: Yeah. I get somewhat confused with netcat syntax. Will take a better look at it's docs.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: True. I responded that one before the other conversation hehe

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
Williangalvani:
```
pacman -S tldr
tldr nc
```

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -24,19 +24,21 @@ def _get_version(self) -> Optional[str]:
...
         # Convert endpoint format to mavlink-router format
         def convert_endpoint(endpoint: Endpoint) -> str:
-            if endpoint.connection_type != EndpointType.TCP:
-                return str(endpoint)[str(endpoint).find(":") + 1 :]
             # TCP uses a special argument and only works with localhost
-            port = str(endpoint).split(":")[2]
-            return f"-t {port}"
+            if endpoint.connection_type == EndpointType.TCPServer:
+                port = str(endpoint).split(":")[2]
+                return f"-t {port}"
```
patrickelectric: Dejavu, this should use the full argument, always avoid single letter arguments.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -90,13 +90,28 @@ def test_mavlink_router() -> None:
...
+        endpoint_3,
     ], "Endpoint list does not match."
+    with pytest.raises(NotImplementedError):
+        mavlink_router.add_endpoint(Endpoint("udpin", "0.0.0.0", 14551))
+    with pytest.raises(NotImplementedError):
+        mavlink_router.add_endpoint(Endpoint("serial", "/dev/autopilot", 115200))
 
     assert mavlink_router.start(Endpoint("udpin", "0.0.0.0", 14550)), "Failed to start MAVLinkRouter"
     assert mavlink_router.is_running(), "MAVLinkRouter is not running after start."
+    assert mavlink_router.exit(), "MAVLinkRouter could not stop."
```
patrickelectric, line 127: You should add a test to see if is_running is false

core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
```diff
@@ -21,17 +21,22 @@ def _get_version(self) -> Optional[str]:
...
+            else:
+                endpoints_string += f" --out={endpoint.place},{endpoint.argument}"
 
         master_string = str(master)
         if master.connection_type == EndpointType.Serial:
             if not master.argument:
                 master_string = f"{master.place}"
             else:
-                master_string = f"{master.place} --baudrate={master.argument}"
+                master_string = f"{master.place},{master.argument}"
```
patrickelectric, line 36: you can create a lambda to abstract the endpoint -> string conversion between serial and others, avoiding the same logic in different places.

core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
```diff
@@ -48,7 +48,13 @@ def binary_name() -> str:
 
     @staticmethod
     def _validate_endpoint(endpoint: Endpoint) -> bool:
-        return True
+        return endpoint.connection_type in [
+            EndpointType.Serial,
+            EndpointType.UDPServer,
+            EndpointType.UDPClient,
+            EndpointType.TCPServer,
+            EndpointType.TCPClient,
```
patrickelectric, line 56: Are we going to support File endpoints at all ?
This commit is unnecessary if you put 74122 before.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -63,17 +63,39 @@ def test_mavproxy() -> None:
...
+    endpoint_1 = Endpoint("udpin", "0.0.0.0", 14551)
     endpoint_2 = Endpoint("udpout", "0.0.0.0", 14552)
+    endpoint_3 = Endpoint("tcpin", "0.0.0.0", 14553)
+    endpoint_4 = Endpoint("tcpout", "0.0.0.0", 14554)
+    endpoint_5 = Endpoint("serial", "/dev/radiolink", 57600)
     assert mavproxy.add_endpoint(endpoint_1), "Failed to add first endpoint"
     assert mavproxy.add_endpoint(endpoint_2), "Failed to add second endpoint"
+    assert mavproxy.add_endpoint(endpoint_3), "Failed to add third endpoint"
+    assert mavproxy.add_endpoint(endpoint_4), "Failed to add fourth endpoint"
+    assert mavproxy.add_endpoint(endpoint_5), "Failed to add fifth endpoint"
```
patrickelectric, line 70: Well, it's time to use for loops here and avoid all duplicated code.

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: This is changed in the following commit. I just separated the changes because logically they are different changes.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: Makes sense

core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
rafaellehmkuhl: This was more to make the validation explicitly. Imagine we add a new EndpointType, say, potato. We cannot assume prior that the potato type is supported by MavProxy.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: This is one of those funny moments when I look at it and think "why the hell is that like this?"

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: It's failing, but this is not the scope of this PR, so I won't add it right now and will open an issue for that one.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -116,8 +116,10 @@ def _load_endpoints(self) -> None:
             self.configuration["endpoints"] = []
         endpoints = self.configuration["endpoints"]
         for endpoint in endpoints:
-            if not self.mavlink_manager.add_endpoint(Endpoint(**endpoint)):
-                warn(f"Could not load endpoint {endpoint}")
+            try:
+                self.mavlink_manager.add_endpoint(Endpoint(**endpoint))
```
patrickelectric, line 120: Is there no way to us to identify via pylint which function can raise ?
Why there return type is not being used now ? Should we don't warn when a endpoint already exists in the list ?

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -28,11 +45,8 @@ def test_endpoint() -> None:
 
 
 def test_endpoint_validators() -> None:
-    Endpoint.is_mavlink_endpoint({"connection_type": "tcpin", "place": "0.0.0.0", "argument": 14550})
-    Endpoint.is_mavlink_endpoint({"connection_type": "tcpout", "place": "0.0.0.0", "argument": 14550})
-    Endpoint.is_mavlink_endpoint({"connection_type": "udpin", "place": "0.0.0.0", "argument": 14550})
-    Endpoint.is_mavlink_endpoint({"connection_type": "udpout", "place": "0.0.0.0", "argument": 14550})
-    Endpoint.is_mavlink_endpoint({"connection_type": "serial", "place": "/dev/autopilot", "argument": 115200})
+    for endpoint in OUTPUT_ENDPOINTS:
+        Endpoint.is_mavlink_endpoint(endpoint.asdict())
```
patrickelectric, line 49: Can't we have a function in Endpoint to just call: `endpoint.is_mavlink_endpoint()`, since the function is already inside the Endpoint class it makes sense.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -106,29 +102,28 @@ def test_mavlink_router() -> None:
...
-        mavlink_router.start(Endpoint("tcpout", "0.0.0.0", 14550))
+    allowed_output_types = [EndpointType.UDPClient, EndpointType.TCPServer, EndpointType.TCPClient]
+    allowed_master_types = [EndpointType.UDPServer, EndpointType.Serial, EndpointType.TCPServer]
+    run_common_routing_tests(mavlink_router, allowed_output_types, allowed_master_types)
+
+
+def run_common_routing_tests(
+    router: AbstractRouter, allowed_output_types: List[EndpointType], allowed_master_types: List[EndpointType]
+) -> None:
+    allowed_output_endpoints = [e for e in OUTPUT_ENDPOINTS if e.connection_type in allowed_output_types]
```
patrickelectric, line 113: It's clear to use the filter function, since this will make more clear for someone what are you doing.
```python
filter(lambda endpoint: endpoint.connection_type in allowed_output_types, OUTPUT_ENDPOINTS)
```
Also, be aware to avoid single letter variables

core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -106,29 +102,28 @@ def test_mavlink_router() -> None:
...
+    for endpoint in allowed_output_endpoints:
+        assert router.add_endpoint(endpoint), f"Failed to add endpoint {endpoint}."
+    assert router.endpoints() == allowed_output_endpoints, "Endpoint list does not match."
+
+    for endpoint in [e for e in OUTPUT_ENDPOINTS if not e.connection_type in allowed_output_types]:
```
patrickelectric, line 118: just use the `difference` function between between both sets.

patrickelectric (changes requested):

patrickelectric:
You can also merge 0d63e2a with 601bd90

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: I solved a problem there but created another one.

The problem was that, when the Endpoint was not valid, it raised a NotImplementedError, which was not being catched, so I have put it inside a try/except block. The problem is that we still have the situation of duplicated endpoints, which was catched before and is not anymore.

Will fix that.

rafaellehmkuhl: Still, I don't think that returning false is very explicitly about what is happening.

Let's say that I don't know how the logic of the endpoints are just a list appending. Maybe it was a database. How do I know, just by receiving a false response, what happened? Was that false a database store failing? Was a duplicated endpoint? Was a validation error? The validation error is the only one were I truly know what happened, because it raised an error, instead of a status.

rafaellehmkuhl: That's why I think raising errors and catching them explicitly on the proper interface is the correct way to do it:

```python
def add_endpoint(self, endpoint):
    if duplicated: raise ValueError("Duplicated endpoint")
    if not_validated: raise ValueError("Endpoint not valid for current router")
    if ...: raise ...
    self._endpoints.append(endpoint)
```

patrickelectric: To me, the clear solution is to use a different architecture to fix the issue:
https://pypi.org/project/result/

patrickelectric: But we can do that in a future PR, this is already getting long. 
Just make sure to create issues for all relevant points here.

rafaellehmkuhl: I *really* liked the idea of the result library. I will open an issue for that and we can discuss about turning it a standard for the repo.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: We can. Will open an issue for that.

patrickelectric: If that's true, the test does not make sense, because you are already creating the endpoints here:
https://github.com/bluerobotics/companion-docker/pull/152/files/2970d5511111328f26168f0d9e5ea0c87bb5643c..0d63e2aea44c7c215713c0426a5e95a2b0bce0b9#diff-7498f7ef4068e0981d03dbdc5e0c34411f5fd4df2727948d43587aefac10a4cdR17
So you are testing something that was already validated by the first constructor call.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: They are not sets yet, but since they are hashable now, they can be.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: > You can also merge [0d63e2a](https://github.com/bluerobotics/companion-docker/commit/0d63e2aea44c7c215713c0426a5e95a2b0bce0b9) with [601bd90](https://github.com/bluerobotics/companion-docker/commit/601bd90d544d98f7196692e696ee203cd7bbc44f)

They seen to be different, I think. The first one extends the test coverage, while the last just refactor the code, keeping the functionality. From a commiting point of view one wouldn't know if a problem raised from extending the tests or refactoring them.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: Oh, wait, those are different types. The first one are the endpoints itself, while the second is a list of types allowed. I will implement the filter/lambda solution here.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
patrickelectric: This just creates unnecessary commits, you can refactor the code first and add the extra endpoints next, not the other way around. But ok.

rafaellehmkuhl: > Also, be aware to avoid single letter variables

About this: I'm trying to use just in cases where it's really explicitly what the variable are and writing the entire name will end up breaking the 120 characters limit and end up with a two or even three-lines expression, which becomes ugly and less readable.

In your example, for example, it ends up like that:
```
allowed_output_endpoints = filter(
    lambda endpoint: endpoint.connection_type in allowed_output_types, OUTPUT_ENDPOINTS
)
```

rafaellehmkuhl: Just to ilustrate, this is the expression with the single letter variable:

```python
allowed_output_endpoints = filter(lambda e: e.connection_type in allowed_output_types, OUTPUT_ENDPOINTS)
```

What do you think it better (honest question)?

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: Got your point now.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
patrickelectric: The extra space is not a problem, there is plenty of space, the important thing in code is readability and maintainability, a single letter is not maintainable and clear. For the same reason the code is being worked from a for loop to a filter, since this shows what is the intended behavior in code.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: Making them set demands that we change it to sets on the AbstractRouter. I will open an issue for that and attack on a following PR.

patrickelectric: So, can't you just make a filter ?

rafaellehmkuhl: Yes yes, same way as the other.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: That's true.

patrickelectric (approved):
looks good

rafaellehmkuhl:
@patrickelectric the deploy images on armv7 action is taking a lot longer than usual

Do you know what could it be?

patrickelectric:
> @patrickelectric the deploy images on armv7 action is taking a lot longer than usual
> 
> Do you know what could it be?

It depends of some factors, but the most important one is that our default branch is not master for github, so our cache system is not working as we want.

PR #157, opened 2021-04-13
Bootstrap: don't pull a new image if there is a local version available

### LINKED ISSUE #156: Bootstrap ALWAYS attempts to pull a new image
This leads to loss of local changes and very long boot times when the local tag differs from the remote.

Williangalvani:
Fix #156

### DIFF
--- a/bootstrap/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap/bootstrap.py
@@ -128,6 +128,11 @@ class Bootstrapper:
             curses.endwin()
         print("Done")
 
+    def image_is_available_locally(self, image_name: str, tag: str) -> bool:
+        """Checks if the image is already available locally"""
+        images = self.client.images.list(image_name)
+        return any(f"{image_name}:{tag}" in image.tags for image in images)
+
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap/bootstrap.py
```diff
@@ -128,6 +128,13 @@ def pull_core(self) -> None:
             curses.endwin()
         print("Done")
 
+    def image_is_available_locally(self, image_name: str, tag: str) -> bool:
+        """Checks if the image is already available locally"""
+        for image in self.client.images.list(image_name):
+            if f"{image_name}:{tag}" in image.tags:
+                return True
+        return False
```
patrickelectric, line 136: Not necessary, but you can also do:
```python
images = mageself.client.images.list(image_name)
return any(f"{image_name}:{tag}" in image.tags for image in images)
```

patrickelectric (approved):

bootstrap/bootstrap/bootstrap.py
Williangalvani: beautiful. thanks!

PR #165, opened 2021-04-16
Use sets for endpoint aggregation logic

### LINKED ISSUE #164: Endpoints are aggregated as a list, not a set, on AbstractRouter
Since we cannot have two equal endpoints, `set` would be a better implementation, and it is possible since the endpoints are now hashable.

That would also remove some code from filtering and getting the difference between them, currently implemented by hand, when Python's set already delivers it for free.

rafaellehmkuhl:
To be approved after #152 
Fix #164

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -113,12 +113,10 @@ class ArduPilotManager(metaclass=Singleton):
 
     def _load_endpoints(self) -> None:
         """Load endpoints from the configuration file to the mavlink manager."""
-        if "endpoints" not in self.configuration:
-            self.configuration["endpoints"] = []
-        endpoints = self.configuration["endpoints"]
+        endpoints = [Endpoint(**endpoint) for endpoint in self.configuration.get("endpoints") or []]
         for endpoint in endpoints:
             try:
-                if not self.mavlink_manager.add_endpoint(Endpoint(**endpoint)):
+                if not self.mavlink_manager.add_endpoint(endpoint):
                     raise ValueError("Cannot add endpoint.")
             except Exception as error:
                 warn(f"Could not load endpoint {endpoint}. {error}")
@@ -131,9 +129,9 @@ class ArduPilotManager(metaclass=Singleton):
         except Exception as error:
             warn(f"Error reseting endpoints: {error}")
 
-    def _update_endpoints(self, updated_endpoints: List[Endpoint]) -> bool:
+    def _update_endpoints(self) -> bool:
         try:
-            self.configuration["endpoints"] = deepcopy(updated_endpoints)
+            self.configuration["endpoints"] = [endpoint.asdict() for endpoint in self.get_endpoints()]
             self.settings.save(self.configuration)
             return self.restart()
         except Exception as error:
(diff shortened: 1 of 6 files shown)


core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -112,74 +112,69 @@ def restart(self) -> bool:
 
     def _load_endpoints(self) -> None:
         """Load endpoints from the configuration file to the mavlink manager."""
-        if "endpoints" not in self.configuration:
-            self.configuration["endpoints"] = []
-        endpoints = self.configuration["endpoints"]
+        endpoints = {Endpoint(**endpoint) for endpoint in self.configuration.get("endpoints") or []}
```
patrickelectric, line 115: Isn't it better to have a `_get_configuration_endpoints()` call that gets and parse the configuration endpoint in a friendly way ?

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -112,74 +112,69 @@ def restart(self) -> bool:
...
             self.mavlink_manager.add_endpoints(endpoints)
             print("Reseting endpoints to previous state.")
         except Exception as error:
             warn(f"Error reseting endpoints: {error}")
 
-    def _update_endpoints(self, updated_endpoints: List[Endpoint]) -> bool:
+    def _update_endpoints(self) -> bool:
         try:
-            self.configuration["endpoints"] = deepcopy(updated_endpoints)
+            self.configuration["endpoints"] = [endpoint.asdict() for endpoint in self.get_endpoints()]
```
patrickelectric, line 133: Is there a problem to have the configuration as a set ?

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
```diff
@@ -117,12 +117,12 @@ def remove_endpoint(self, endpoint: Endpoint) -> bool:
...
         return False
 
-    def endpoints(self) -> List[Endpoint]:
+    def endpoints(self) -> Set[Endpoint]:
         return self._endpoints
 
     def clear_endpoints(self) -> None:
         """Remove all output endpoints."""
-        self._endpoints = []
+        self._endpoints = set()
```
patrickelectric, line 125: `{}` ?

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
patrickelectric: image

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: I didn't do it because that's the only interface between the configured endpoints and the APManager, but doing it does no harm, for sure.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Yes. JSON does not support sets.

patrickelectric (approved):

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -110,76 +110,76 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
...
-            if endpoint.asdict() in saved_endpoints:
-                warn(f"Endpoint {endpoint} already stored.")
-                self._reset_endpoints(loaded_endpoints)
-                return False
             try:
                 self.mavlink_manager.add_endpoint(endpoint)
-                saved_endpoints.append(endpoint.asdict())
                 print(f"Adding endpoint {endpoint} and saving it to the settings file.")
             except NotImplementedError:
                 print(f"Failed to add endpoint {endpoint}. Connection_type not compatible with current router.")
```
Williangalvani, line 163: it would be nice to print the name of the router instead of "current router"

Williangalvani (approved):

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -110,76 +110,76 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
...
-    def add_new_endpoints(self, new_endpoints: List[Endpoint]) -> bool:
+    def add_new_endpoints(self, new_endpoints: Set[Endpoint]) -> bool:
         """Add multiple endpoints to the mavlink manager and save them on the configuration file."""
-        saved_endpoints = deepcopy(self.configuration["endpoints"])
-        loaded_endpoints = deepcopy(self.get_endpoints())
+        loaded_endpoints = self.get_endpoints()
+
+        duplicated_endpoints = loaded_endpoints.intersection(new_endpoints)
+        if duplicated_endpoints:
+            warn(f"Endpoints {duplicated_endpoints} already loaded. Reseting operation.")
```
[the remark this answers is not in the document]
rafaellehmkuhl: We are. What happens is that we backup the original endpoints, keep adding/removing the new endpoints and if something goes wrong, we go back to the original.
I thought about "abort", but it sounds like we will keep the endpoints that were added/removed till the operation failed.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Yes

core/services/ardupilot_manager/ArduPilotManager.py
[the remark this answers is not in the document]
Williangalvani: are we?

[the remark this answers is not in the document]
rafaellehmkuhl: Rsrs
Not there

Williangalvani (approved):

Issue #153, opened 2021-04-08
Fancy Endpoints


rafaellehmkuhl:
About the new Endpoint structure, there are some things to define, and the first one is which fields we are going to have. I'm thinking about those:
- id
- name
- owner
- description
- persistent
- protected

Any thoughts @jaxxzer @Williangalvani @patrickelectric ?

rafaellehmkuhl:
The idea for them is:

- id: used to identify endpoints
- name: facilitate identification by users
- owner: facilitate identification of who create, by users (maybe ownership/protection? would add some work thought, because we would need a identification system) 
- description: remember why was it opened (maybe not necessary?)
- persistent: opposed to dynamically created endpoints, that would die on system restart
- protected: only the system can remove/modify

rafaellehmkuhl:
Logic:

- User generated endpoints receive available ids automatically (smallest available id)
- User generated endpoints are persistent by default, 
- Additional route for dynamic endpoints, which would open a tcp/udp (user chooses) socket (ip 0.0.0.0) with a random port

rafaellehmkuhl:
After our discussions today:

- id is not needed, as the endpoints are now hashable and can be uniquely identified by their string representation

PR #166, opened 2021-04-16
Fancy endpoints

rafaellehmkuhl:
To be approved after #165  
Implements #153

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -136,7 +136,8 @@ class ArduPilotManager(metaclass=Singleton):
 
     def _update_endpoints(self) -> bool:
         try:
-            self._save_endpoints_to_configuration(self.get_endpoints())
+            persistent_endpoints = set(filter(lambda endpoint: endpoint.persistent, self.get_endpoints()))
+            self._save_endpoints_to_configuration(persistent_endpoints)
             self.settings.save(self.configuration)
             return self.restart()
         except Exception as error:
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -180,6 +180,11 @@ class ArduPilotManager(metaclass=Singleton):
             warn(f"Endpoints {missing_endpoints} not found. Aborting operation.")
             return False
 
+        protected_endpoints = set(filter(lambda endpoint: endpoint.protected, endpoints_to_remove))
+        if protected_endpoints:
+            warn(f"Endpoints {protected_endpoints} are protected. Aborting operation.")
+            return False
+
         for endpoint in endpoints_to_remove:
             if not self.mavlink_manager.remove_endpoint(endpoint):
                 warn(f"Mavlink manager failed to remove endpoint {endpoint}.")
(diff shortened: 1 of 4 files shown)


core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -81,10 +81,10 @@ def start_navigator(self) -> None:
         # self.start_mavlink_manager(Endpoint(local_endpoint))
 
     def start_serial(self, device: str) -> None:
-        self.start_mavlink_manager(Endpoint("serial", device, 115200))
+        self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, False, True, "serial", device, 115200))
```
patrickelectric, line 84: This is really bad as API, can we have a scruct or anything that provides explicit names for the fields ?
"something", something, True, False is not what is each field.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -179,6 +180,11 @@ def remove_endpoints(self, endpoints_to_remove: Set[Endpoint]) -> bool:
             warn(f"Endpoints {missing_endpoints} not found. Aborting operation.")
             return False
 
+        protected_endpoints = set(filter(lambda endpoint: endpoint.protected, endpoints_to_remove))
```
patrickelectric, line 183: If protected is only used inside the if, it should be declared there.

patrickelectric (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
patrickelectric: I would recommend to read:
https://wiki.qt.io/API_Design_Principles
https://doc.qt.io/archives/qq/qq13-apis.html
https://harrymoreno.com/assets/the-little-manual-of-api-design.pdf

rafaellehmkuhl: You mean to write it explicitly there?

```python
self.start_mavlink_manager(Endpoint(
    name="Master",
    owner=self.settings.app_name, 
    persistent=False, 
    protected=True,
    connection_type="serial",
    place=device,
    argument=115200
))
```

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: It's a typo on the if. It should be `if protected_endpoints`. Fixing

core/services/ardupilot_manager/ArduPilotManager.py
patrickelectric: Yes, but as you can see that makes the use really hard, I would recommend to break some of the extra configurations in a second function/struct to avoid unnecessary fields, and take some of the values as default on the constructor for the common use case.

rafaellehmkuhl: There are some common use cases, like the not-persistent/protected/APManager endpoint, the not-persistent/protected/"external-app" endpoint, etc. What we can do is to create builder methods for this. What do you think?

jaxxzer: > take some of the values as default on the constructor for the common use case.

This is a good suggestion that rafael has just implemented. Beyond that, I don't think that the code can be simplified or reduced on the caller side. These are necessary parameters to construct an endpoint. They cannot be configured in a second step like `set_owner`.

@patrickelectric If you see a better way, please suggest a patch on top of this.

rafaellehmkuhl: Will open an issue to discuss a new idea for this API.

patrickelectric: https://github.com/bluerobotics/companion-docker/issues/183

PR #179, opened 2021-04-21
Core: migrate to Python 3.9

Williangalvani:
Use williangalvani/companion-core:trespontonove to test.

startup.json for convenience:
```
{
    "core": {
        "tag": "trespontonove",
        "enabled": true,
        "webui": false,
        "network": "host",
        "binds": {
            "/dev/": {
                "bind": "/dev/",
                "mode": "rw"
            },
                "mode": "rw"
            },
            "/tmp/wpa_playground": {
                "mode": "rw"
            }
        },
        "privileged": true
    }
}

```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9] # Python 3.7 is the version in our buster image
+        python-version: [3.9] # Our base image has Python 3.9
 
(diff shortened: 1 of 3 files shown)


rafaellehmkuhl (changes requested):
We should also make unit tests 3.9 only and ensure mavlink-router is using the same version in bootstrap and tests.

patrickelectric (approved):
Looks good

jaxxzer (approved):

PR #184, opened 2021-04-23
Helper: help more

Williangalvani:
This extends the helper a little, trying to fetch the title of the service page and checking if swagger is available at /v1.0/ui:

image
should I go ahead and format it with a table? :thinking: 

convenience startup.json:
```
{
    "core": {
        "tag": "morehelp",
        "enabled": true,
        "webui": false,
        "network": "host",
        "binds": {
            "/dev/": {
                "bind": "/dev/",
                "mode": "rw"
            },
                "mode": "rw"
            },
            "/tmp/wpa_playground": {
                "mode": "rw"
            }
        },
        "privileged": true
    }
}
```

### DIFF
--- a/core/services/cable_guy/html/index.html
+++ b/core/services/cable_guy/html/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
 
 <head>
+  <title>Cable-guy</title>
   <!-- Required meta tags -->
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -39,11 +39,11 @@ class Helper:
         info = ServiceInfo(valid=False, title="Unknown", ui="", port=port)
 
         try:
-            response = urlopen(f"http://127.0.0.1:{port}/", timeout=0.2)
-            info.valid = True
-            soup = BeautifulSoup(response.read(), features="html.parser")
-            title_element = soup.find("title")
-            info.title = title_element.text if title_element else "Unknown"
+            with urlopen(f"http://127.0.0.1:{port}/", timeout=0.2) as response:
+                info.valid = True
+                soup = BeautifulSoup(response.read(), features="html.parser")
+                title_element = soup.find("title")
+                info.title = title_element.text if title_element else "Unknown"
         except urllib.error.HTTPError as error:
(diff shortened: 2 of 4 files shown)


core/services/helper/main.py
```diff
@@ -35,16 +72,16 @@ def get_webpage_ports() -> List[int]:
         # And check if there is a webpage available that is not us
         ports = iter([connection.laddr.port for connection in connections])
         ports = filter(lambda port: port != PORT, ports)
-        ports = filter(Helper.check_webpage, ports)
-
-        return list(ports)
+        services = [Helper.check_port(port) for port in ports]
+        valid_services = [service for service in services if service.valid]
+        return valid_services
```
patrickelectric, line 77: You can do:
```python
services = map(Helper.check_port, ports)
valid_services = filter(lambda service: service.valid, services)
```

Also, check port is a terrible name, since it implies that will `check` or return a True and False and that is not the case.

patrickelectric (changes requested):

core/services/cable_guy/html/index.html
```diff
@@ -2,9 +2,10 @@
 <html lang="en">
 
 <head>
-  <!-- Required meta tags -->
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+   <title>Cable-guy</title>
+   <!-- Required meta tags -->
+   <meta charset="utf-8">
+   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
```
patrickelectric, line 8: wait, the linter was not checking here ?

Williangalvani: good catch, let me check

Williangalvani: it turns out that it catches indentation errors, but not always. I opened https://github.com/bluerobotics/companion-docker/issues/189

patrickelectric: How is that possible ? We have not enabled cache under github actions for this

Williangalvani: I'm guessing it is too permissive to what is "acceptable" indentation.
If you try the current config with https://github.com/bluerobotics/companion-docker/blob/5e4dd905560fbebcefc88c90b3c539065bf34ed3/core/services/cable_guy/html/index.html
it seems to fail properly.

patrickelectric (approved):

PR #188, opened 2021-04-26
core: services: aurdupilot_manager: suppress 'consider-using-with' warning

Williangalvani:
This "fixes" CI on master.

We should consider fixing the implementation instead.


core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -16,6 +16,8 @@
 from settings import Settings
 from Singleton import Singleton
 
+# pylint: disable=consider-using-with
+
```
patrickelectric, line 20: isn't it better to have it in place and not for the entire file ?

Williangalvani: yes

patrickelectric (approved):

PR #192, opened 2021-04-27
gitignore: Update to include npm and cable-guy stuff

rafaellehmkuhl:
The necessity to run eslint as part of the tests turned the npm install mandatory, and so what comes with it :sweat_smile: 

I'm also adding ignore for the static files generated by other services.

### DIFF
--- a/core/services/cable_guy/.gitignore
+++ b/core/services/cable_guy/.gitignore
@@ -1,0 +1,1 @@
+html/static/


.gitignore
```diff
@@ -6,3 +8,11 @@ env/
 build/
 *.egg-info
 dist/
+node_modules/
+package.json
+package-lock.json
```
patrickelectric, line 13: This is addressing the problem in the wrong way, we should use npm install in --user to avoid such packages locally.

.gitignore
```diff
@@ -6,3 +8,11 @@ env/
...
+
+## Service-specific ignores:
+
+### Cable-guy
+core/services/cable_guy/html/static/
```
patrickelectric, line 18: This should be done inside cable_guy as a gitignore.

patrickelectric (changes requested):

.gitignore
rafaellehmkuhl: You mean installing the packages globally?
That's really bad for development.

patrickelectric: We are installing tools, not libraries, what could be wrong ?

rafaellehmkuhl: > We are installing tools, not libraries, what could be wrong ?
Because you are enforcing version (latest) on user's system, which is simply not needed. Why not to allow segregated environment?

> There is also no `package.json` for this project.
You need to install the npm packages to be able to run the tests, thus creating the package.json...

patrickelectric: > thus creating the package.json

I was not aware that npm was that bad, well, yarn does not do that if you install global.
`yarn global add eslint`

rafaellehmkuhl: But it's the same problem. Your enforcing package installation/update in the user's system.
Yarn and Npm have the same behaviour here, and is not a problem. It's a dependency package that we should be using, in fact, since we have javascript dependencies on the project.

patrickelectric: > You're enforcing package installation/update in the user's system.

I don't see how is that a problem.

Our project actually do not use npm,yarn or any alternative, what we use is eslint, how eslint is installed is not our problem, the eslint version is not relevant.

As example, I can install eslint in my system via the package manager (brew, yaourt, pikaur, choco), without node_modules or anything in this .gitignore.

If we get `package.json` in this project and the entire project is node based, it'll make sense, but as you can see that's not the case here. Otherwise, node is not important. 

What you think @Williangalvani ?

rafaellehmkuhl: Ok, suggestion is:

`yarn global add eslint eslint-plugin-vue eslint-plugin-html --no-lock-file`

Will add this in development wiki

patrickelectric (approved):

.gitignore
Williangalvani: Honestly, I believe it makes more sense to keep it local to the project, so we don't enforce more stuff into the users environment.
If we add packages.json/lock and node_modules to gitignore, is there any harm?

PR #209, opened 2021-05-03
Move WiFi service to FastAPI

### LINKED ISSUE #198: Move WiFi service to FastAPI

rafaellehmkuhl:
Fix #198 

- Segregate WifiManager class to dedicated file
- Replace dynamically generated routes with OpenAPI docs (swagger/redoc)
- Add types for WifiManager data
- Replace error-string-return with warning+httpexception

### DIFF
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -1,4 +1,6 @@
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List
+
+from pydantic import BaseModel
 
 from wpa_supplicant import WPASupplicant
 
@@ -70,20 +72,20 @@ class WifiManager:
 
         return output
 
-    def get_wifi_available(self) -> Union[List[Dict[str, Any]], str]:
+    def get_wifi_available(self) -> List[Dict[str, Any]]:
         """Get a dict from the wifi signals available"""
         self.wpa.send_command_scan()
         data, result = self.wpa.send_command_scan_results()
         if not result:
-            return "Failed to fetch wifi list"
+            raise ValueError("Failed to fetch wifi list.")
 
         return WifiManager.__dict_from_table(data)
 
-    def get_saved_wifi_network(self) -> Union[List[Dict[str, Any]], str]:
+    def get_saved_wifi_network(self) -> List[Dict[str, Any]]:
         """Get a list of saved wifi networks"""
         data, result = self.wpa.send_command_list_networks()
         if not result:
-            return "Failed to fetch saved wifi list"
+            raise ValueError("Failed to fetch saved wifi list.")
 
         return WifiManager.__dict_from_table(data)
 
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -1,0 +1,125 @@
+from typing import Any, Dict, List, Union
+
+from wpa_supplicant import WPASupplicant
+
+
+class WifiManager:
+    wpa = WPASupplicant()
+
+    def connect(self, path: Any) -> None:
+        """Does the connection with wpa_supplicant service
+
+        Arguments:
+            path {[tuple/str]} -- Can be a tuple to connect (ip/port) or unix socket file
+        """
+        self.wpa.run(path)
+        # Request scan to update state
+        self.wpa.send_command_scan()
+
+    @staticmethod
+    def __decode_escaped(data: bytes) -> str:
+        """Decode escaped byte array
+        For more info: https://stackoverflow.com/questions/14820429/how-do-i-decodestring-escape-in-python3
+        """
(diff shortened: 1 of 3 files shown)


rafaellehmkuhl:
Isort and Black seem to be conflicting.
Isort thinks, on main.py, imports should be in this way:
```python
from WifiManager import (SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials,
                         WifiManager)
```
while black thinks that's better:
```python
from WifiManager import SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials, WifiManager
```

I think we should [add an Isort configuration file](https://pycqa.github.io/isort/docs/configuration/black_compatibility/) and make it compatible with black configs.

core/services/wifi/WifiManager.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python3
```
patrickelectric, line 1: This shebang is not necessary, this will not be an executable

rafaellehmkuhl: Mistakenly added. Good.

patrickelectric:
> Isort and Black seem to be conflicting.
> Isort thinks, on main.py, imports should be in this way:
> 
> ```python
> from WifiManager import (SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials,
>                          WifiManager)
> ```
> 
> while black thinks that's better:
> 
> ```python
> from WifiManager import SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials, WifiManager
> ```
> 
> I think we should [add an Isort configuration file](https://pycqa.github.io/isort/docs/configuration/black_compatibility/) and make it compatible with black configs.
> 
> EDIT: if we opt to not to have a configuration file, we can simply add `--profile black` to the isort call.

Maybe, we should try to create a `pyproject.toml` to make it compatible.

rafaellehmkuhl:
> > Isort and Black seem to be conflicting.
> > Isort thinks, on main.py, imports should be in this way:
> > ```python
> > from WifiManager import (SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials,
> >                          WifiManager)
> > ```
> > 
> > 
> > while black thinks that's better:
> > ```python
> > from WifiManager import SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials, WifiManager
> > ```
> > 
> > 
> > I think we should [add an Isort configuration file](https://pycqa.github.io/isort/docs/configuration/black_compatibility/) and make it compatible with black configs.
> > EDIT: if we opt to not to have a configuration file, we can simply add `--profile black` to the isort call.
> 
> Maybe, we should try to create a `pyproject.toml` to make it compatible.

This is the long the solution, for sure. I can take this.

For the short term I think the `--profile black` argument is ok, since, when moving to pyproject.toml, we will need to change all the test calls anyway.

Williangalvani:
I see no harm in adding the `--profile black` while we don't have a pyproject.toml. The migration to pyproject.toml wouldn't be part of this PR anyway, right?

core/services/wifi/WifiManager.py
```diff
@@ -70,20 +71,22 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
...
 
-    def get_wifi_available(self) -> Union[List[Dict[str, Any]], str]:
+    def get_wifi_available(self) -> List[Dict[str, Any]]:
         """Get a dict from the wifi signals available"""
         self.wpa.send_command_scan()
         data, result = self.wpa.send_command_scan_results()
         if not result:
-            return "Failed to fetch wifi list"
+            warnings.warn(RuntimeWarning("Failed to fetch wifi list."))
+            return []
```
Williangalvani, line 101: Does it mean there's no way to tell if the wifi is failing or there are not networks in range?

rafaellehmkuhl: The warn is meant to do it.

The other option would be to raise an error an catch it later. Discussing with @patrickelectric we though it would be better to warn so the user should be able to continue using the program.

What do you think?

rafaellehmkuhl:
> I see no harm in adding the `--profile black` while we don't have a pyproject.toml. The migration to pyproject.toml wouldn't be part of this PR anyway, right?

Will be solved with #213

core/services/wifi/WifiManager.py
patrickelectric: It's a bit weird IMO using a warning log system as a return mechanism.. 

So for every call, we need to check for warnings.. if that's the plan here I don't see how is that different from a raise with a try block

rafaellehmkuhl: I was thinking about this, and yeah, there's no reason not to raise there. If the library is used by a higher-level app, it should try/catch the errors and answer accordingly.

Will move to raises.

Williangalvani (approved):

patrickelectric (approved):

PR #212, opened 2021-05-03
Fix isort, run per python subproject 

patrickelectric:


rafaellehmkuhl (approved):
Should we also [add the `--profile black`](https://github.com/bluerobotics/companion-docker/issues/210)?

Williangalvani (approved):

Williangalvani:
> Should we also [add the `--profile black`](https://github.com/bluerobotics/companion-docker/issues/210)?

why isn't this causing an error in this PR?

patrickelectric:
> Should we also [add the `--profile black`](https://github.com/bluerobotics/companion-docker/issues/210)?

I believe that the way to go is to move at least the isort and black configuration to pyproject

PR #214, opened 2021-05-03
Config

### LINKED ISSUE #182: Cannot use i2c0 with latest raspbian release
Filed for visibility + FYI

https://github.com/raspberrypi/linux/issues/2981
This isn't a showstopper for us.

jaxxzer:
fix #182

### DIFF
--- a/install/boards/bcm_27xx.sh
+++ b/install/boards/bcm_27xx.sh
@@ -64,3 +64,11 @@ done
 # Remove any console serial configuration
 echo "- Configure serial."
 sudo sed -e 's/console=serial[0-9],[0-9]*\ //' -i /boot/cmdline.txt
+
+# Update raspberry pi firmware
+# this is required to avoid 'i2c transfer timed out' kernel errors
+# on older firmware versions
+if grep -q ID=raspbian < /etc/os-release; then
+    echo "- Run rpi update."
+    sudo rpi-update 1340be4
+fi
--- a/install/boards/bcm_27xx.sh
+++ b/install/boards/bcm_27xx.sh
@@ -19,7 +19,7 @@ echo "- Enable I2C, SPI and UART."
 for STRING in \
     "enable_uart=" \
     "dtoverlay=uart" \
-    "dtparam=i2c_arm=" \
+    "dtparam=i2c" \
     "dtoverlay=i2c" \
     "dtparam=spi=" \
     "dtoverlay=spi" \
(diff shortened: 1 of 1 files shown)


jaxxzer:
NOTE: run `sudo rpi-update` to fix 'i2c transfer timed out' kernel errors on latest raspbian release

install/boards/bcm_28xx.sh
```diff
@@ -28,4 +28,7 @@ done
 
 # Remove any console serial configuration
 echo "- Configure serial."
-sudo sed -e 's/console=serial[0-9],[0-9]*\ //' -i /boot/cmdline.txt
\ No newline at end of file
+sudo sed -e 's/console=serial[0-9],[0-9]*\ //' -i /boot/cmdline.txt
+
+echo "- Run rpi update."
+sudo rpi-update
```
patrickelectric, line 34: Since this is raspbian only, we should check the OS for that.
In this case, you can load `/etc/os-release` and compare `ID == raspbian`

Be also aware that rpi-update will update the system without an specific version in mind, but using the latest version available under the master branch of the raspberry project.

> WARNING: Pre-release versions of software are not guaranteed to work. You should not use rpi-update on any system unless recommended to do so by a Raspberry Pi engineer. It may leave your system unreliable or even completely broken. It should not be used as part of any regular update process.

> rpi-update will download the latest pre-release version of the linux kernel, its matching modules, device tree files, along with the latest versions of the VideoCore firmware. It will then install these files to relevant locations on the SD card, overwriting any previous versions.

> All the source data used by rpi-update comes from the GitHub repo https://github.com/Hexxeh/rpi-firmware. This repository simply contains a subset of the data from the official firmware repository, as not all the data from that repo is required.

ref: https://www.raspberrypi.org/documentation/raspbian/applications/rpi-update.md

install/overlays/config.txt
```diff
@@ -0,0 +1,92 @@
+[pi4]
```
patrickelectric, line 1: Maybe the wiki is a better place for it ?

install/boards/bcm_27xx.sh
```diff
@@ -30,18 +30,18 @@ done
 for STRING in \
     "enable_uart=1" \
     "dtoverlay=uart1" \
-    "dtoverlay=uart2" \
     "dtoverlay=uart3" \
     "dtoverlay=uart4" \
-    "dtparam=i2c_arm=on" \
-    "dtoverlay=i2c0" \
+    "dtoverlay=uart5" \
+    "dtparam=i2c_vc=on" \
```
patrickelectric, line 36: Can you be more clear of what all of this is doing ?
There is no explanation on the commit body.

jaxxzer: done

install/overlays/config.txt
jaxxzer: done

jaxxzer:
rebased

patrickelectric (approved):

PR #226, opened 2021-05-06
CI: use the dockerfile hash instead of commit sha to cache dockers

Williangalvani:
If we use a different key each time we run, we can't get any cache hits:

image

we should just cache based on the image name  and let buildx decide if it is a hit or a miss.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -67,7 +67,7 @@ jobs:
         id: cache
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ matrix.docker }}-${{ github.sha }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         id: cache
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ matrix.docker }}-${{ github.sha }}
+          key: ${{ runner.os }}-buildx-${{ matrix.docker }}
```
patrickelectric, line 70: This is redundant with restore-keys, the point here is that will try to match a cache from the same commit, and if it fails, it'll try to use the cache from restore-keys.
From the documentation:
> restore-keys - An ordered list of keys to use for restoring the cache if no cache hit occurred for key

ref: https://github.com/actions/cache

What we should do is to do a hash from both dockerfiles or a hash for each dockerfile based in the matrix.docker

patrickelectric (approved):
image

PR #227, opened 2021-05-06
Create a base image

### LINKED ISSUE #224: Add companion-base docker
This will help us to decrease companion time and improve future dockers based in over it.

Williangalvani:
The idea is to have a common image to build our stuff on top of, with the tools we deem essential. This should also decrease the build time of companion-core on eventual cache misses.

fix #224


base/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.9-slim-buster
+
+# Create default user folder
+RUN mkdir -p /home/pi
+
+# Install gstreamer
+COPY install_gst.sh /install_gst.sh
+RUN GST_VERSION=1.17.2 ./install_gst.sh && rm /install_gst.sh
+
+RUN apt install -y file locate tmux unzip nano htop iputils-ping watch wget iproute2
```
patrickelectric, line 10:
```suggestion
RUN apt install -y file htop iproute2 iputils-ping locate nano tmux unzip watch wget
```

core/Dockerfile
```diff
@@ -1,20 +1,11 @@
-FROM python:3.9-slim-buster
```
patrickelectric, line 1: Should we also base bootstrap over it ?

.github/workflows/deploy.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        docker: [bootstrap, core]
+        docker: [base, bootstrap, core]
```
patrickelectric, line 15: core needs base to build before it, how can we guarantee that ?

core/Dockerfile
Williangalvani: I'm considering this.. bootstrap will take longer to download. but as it will have the base layer, core will download faster the first time.

I guess we can do it for consistency.

patrickelectric: Good point, maybe base over the same slim should help the 3.9 image.

.github/workflows/deploy.yml
patrickelectric: That's the problem, if you update base the core of the same PR will not use the same image (or will not be available [https://github.com/bluerobotics/companion-docker/pull/227/checks?check_run_id=2521026033#step:7:191])

rafaellehmkuhl: Is it even necessary to include base here, since core is already bringing it into scope?

Williangalvani: core won't automatically build base. it needs to be built somewhere. I considered doing a [multi-stage dockerfile](https://docs.docker.com/develop/develop-images/multistage-build/),** but I'm afraid that might not fix the caching issues we're having.

Williangalvani:
after some discussion, we decided to take companion-base to its own repo. this was superseded bi #268

PR #233, opened 2021-05-13
Implement Ping service

Williangalvani:
- [x] Ping 1D
 - [x] Ping360
 - [x] Hotplugging

Convenience startup.json:
```
{
  "core": {
    "tag": "ping",
    "enabled": true,
    "webui": false,
    "network": "host",
    "binds": {
      "/dev/": {
        "bind": "/dev/",
        "mode": "rw"
      },
      "/sys/": {
        "bind": "/sys/",
        "mode": "ro"
      },
        "mode": "rw"
      },
      "/tmp/wpa_playground": {
        "mode": "rw"
      },
      "/var/run/docker.sock": {
        "bind": "/var/run/docker.sock",
        "mode": "rw"
      }
    },
    "privileged": true
  }
}
```

Helps #264

### DIFF
--- a/core/services/ping/bridges.py
+++ b/core/services/ping/bridges.py
@@ -1,0 +1,26 @@
+import logging
+import shlex
+from shutil import which
+from subprocess import Popen
+
+from serial.tools.list_ports_linux import SysFS
+
+from serialhelper import Baudrates
+
+
+class Bridges:
+    """Basic abstraction of Bridges. Used to bridge serial devices to UDP ports"""
+
+    def __init__(self, serial_port: SysFS, baud: Baudrates, ip: str, udp_port: int) -> None:
+        bridges = which("bridges")
+        command_line = f"{bridges} -u {ip}:{udp_port} -p {serial_port.device}:{baud}"
+        logging.info(f"Launching {command_line}")
+        # pylint: disable=consider-using-with
+        self.process = Popen(shlex.split(command_line))
+
+    def stop(self) -> None:
+        if self.process:
+            self.process.kill()
+
+    def __del__(self) -> None:
+        self.stop()
--- a/core/services/ping/ping1d_driver.py
+++ b/core/services/ping/ping1d_driver.py
@@ -1,0 +1,7 @@
+from pingdriver import PingDriver
+from pingutils import PingDeviceDescriptor
+
+
+class Ping1DDriver(PingDriver):
+    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+        super().__init__(ping, port)
--- a/core/services/ping/ping360_driver.py
+++ b/core/services/ping/ping360_driver.py
@@ -1,0 +1,7 @@
+from pingdriver import PingDriver
+from pingutils import PingDeviceDescriptor
+
+
+class Ping360Driver(PingDriver):
+    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+        super().__init__(ping, port)
--- a/core/services/ping/pingdriver.py
+++ b/core/services/ping/pingdriver.py
@@ -1,0 +1,57 @@
+import logging
+from typing import Optional
+
+from brping import PingDevice
+from brping.definitions import COMMON_DEVICE_INFORMATION
+
+from bridges import Bridges
+from pingutils import PingDeviceDescriptor
+from serialhelper import Baudrates, set_low_latency
+
+
+class PingDriver:
+    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+        self.ping = ping
+        self.port = port
+        self.bridge: Optional[Bridges] = None
+
+    def detect_highest_baud(self) -> Baudrates:
+        """Tries to communicate in increasingly high baudrates up to 4M
+        returns the highest one with at least 90% success rate.
+        """
+        failure_threshold = 0.1  # allow up to 10% failure rate
+        attempts = 10  # try up to 10 times per baudrate
+        max_failures = attempts * failure_threshold
+
+        last_valid_baud = Baudrates.b115200
+        for baud in Baudrates:
+            logging.debug(f"Trying baud {baud}...")
+            failures = 0
+            for _ in range(attempts):
+                ping = PingDevice()
+                ping.connect_serial(self.ping.port.device, baud)
+                device_info = ping.request(COMMON_DEVICE_INFORMATION)
+                if device_info is None:
+                    failures += 1
+                    if failures > max_failures:
+                        break  # there's no pointing in testing again if we already failed.
+            if failures <= max_failures:
+                last_valid_baud = baud
+            logging.debug(f"Baudrate {baud} is {'valid' if baud==last_valid_baud else 'invalid'}")
+        logging.info(f"Highest baudrate detected: {last_valid_baud}")
+        return last_valid_baud
+
+    def start(self) -> None:
+        """Starts the driver"""
+        baud = self.detect_highest_baud()
+        # Do a ping connection to set the baudrate
+        PingDevice().connect_serial(self.ping.port.device, baud)
+        set_low_latency(self.ping.port)
+        self.bridge = Bridges(self.ping.port, baud, "0.0.0.0", self.port)
+
+    def stop(self) -> None:
+        """Stops the driver"""
+        logging.info(f"Forcing Ping1d at port {self.port} to stop.")
+
+    def __del__(self) -> None:
+        self.stop()
--- a/core/services/ping/pingprober.py
+++ b/core/services/ping/pingprober.py
@@ -1,0 +1,102 @@
+import logging
+from typing import Callable, Optional
+
+from brping import PingDevice
+from brping.definitions import COMMON_DEVICE_INFORMATION, PING1D_FIRMWARE_VERSION
+from serial.tools.list_ports_linux import SysFS
+
+from pingutils import PingDeviceDescriptor, PingType
+
+
+class PingProber:
+    """PingProber is responsible for identifying Ping-enabled devices on serial ports"""
+
+    def __init__(self) -> None:
+        self.ping_found_callback: Callable[[PingDeviceDescriptor], None] = lambda x: None
+
+    def probe(self, port: SysFS) -> Optional[PingDeviceDescriptor]:
+        """Attempts to communicate via Ping Protocol at port "port",
+        calls on_ping_found when a ping device is found"""
+        logging.info(f"Probing {port}")
+        detected_device = self.detect_device(port)
+        if detected_device:
+            self.ping_found_callback(detected_device)
+        return detected_device
+
+    def on_ping_found(self, callback: Callable[[PingDeviceDescriptor], None]) -> None:
+        self.ping_found_callback = callback
+
+    @staticmethod
+    def legacy_detect_ping1d(port: SysFS) -> Optional[PingDeviceDescriptor]:
+        """
+        Detects Ping1D devices without DEVICE_INFORMATION implemented
+        """
+        ping = PingDevice()
+        ping.connect_serial(port.device, 115200)
+        firmware_version = ping.request(PING1D_FIRMWARE_VERSION)
+        if firmware_version is None:
+            return None
+        descriptor = PingDeviceDescriptor(
+            ping_type=PingType.PING1D,
+            device_id=firmware_version.src_device_id,
+            device_model=firmware_version.device_model,
+            device_revision=0,  # not available in this message
+            firmware_version_major=firmware_version.firmware_version_major,
+            firmware_version_minor=firmware_version.firmware_version_minor,
+            firmware_version_patch=0,
+            port=port,
+        )
+        logging.info("Identified ping device:")
+        logging.info(descriptor)
+        return descriptor
+
+    def detect_device(self, port: SysFS) -> Optional[PingDeviceDescriptor]:
+        """
+        Attempts to detect the Ping device attached to serial port 'dev'
+        Returns the new path with encoded name if detected, or None if the
+        device was not detected
+        """
+
+        try:
+            ping = PingDevice()
+            ping.connect_serial(port.device, 115200)
+
+        except Exception as exception:
+            if exception.args[0] and "Errno 16" in exception.args[0]:
+                logging.info(f"Device {port.hwid} is busy. Re-trying...")
+                return None
+
+            logging.info(
+                f"An exception has occurred while attempting to"
+                f"talk Ping to device {port.hwid}: {exception}"
+                f"If this is not a Ping device, this is expected."
+            )
+            return None
+
+        if not ping.initialize():
+            return None
+
+        device_info = ping.request(COMMON_DEVICE_INFORMATION)
+        if not device_info:
+            return self.legacy_detect_ping1d(port)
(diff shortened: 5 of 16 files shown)


core/services/ping/portwatcher.py
```diff
@@ -0,0 +1,91 @@
...
+MAX_ATTEMPTS = 10
+
+
+def pingDevicesFilter(port: SysFS) -> bool:
+    """Accepts ports with known chips in the Ping1D and Ping360. Internal ports
+    Are handled by Ardupilot and show None for interface, so we ignore those too.
+    """
+    if not port.interface:
+        return False
+    return "FT231X" in port.interface or "FT230" in port.interface
```
patrickelectric, line 20: That may be a bit problematic since the use can use different serial-usb chips, the user can use a ch341 or anything like it.  Maybe you want to use a blacklist over a whitelist ? It makes more sense.

core/services/ping/ping1d_driver.py
```diff
@@ -0,0 +1,25 @@
...
+        bridges = which("bridges")
+        command_line = f"{bridges} -u 0.0.0.0:{self.port} -p {self.ping.port.device}:115200"
+        logging.info(f"launching {command_line}")
+        # pylint: disable=consider-using-with
+        self.process = Popen(command_line.split(" "))
+
+    def stop(self) -> None:
+        logging.info(f"Forcing Ping1d at port {self.port} to stop.")
+        if self.process:
+            self.process.kill()
```
patrickelectric, line 25: `__del__` should call stop, otherwise we could have bridges running with a dead python code, resulting in a deadend of the sensor without a system reboot.

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,48 @@
...
+        self.process: Optional[Popen[bytes]] = None
+        self.baudrate: int = 2000000
+
+    def set_low_latency_and_high_baud(self) -> None:
+        """
+        sets the latency_timer for the device and attemps to set the ping360 to a 2M baudrate
+        """
+
+        # Open device with 2M baud to setup Ping360
+        logging.info("Setting baud to 2M...")
```
patrickelectric, line 24: `{self.baudrate}`

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,48 @@
...
+
+from pingdriver import PingDriver
+from pingutils import PingDeviceDescriptor
+
+
+class Ping360Driver(PingDriver):
+    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+        super().__init__(ping, port)
+        self.process: Optional[Popen[bytes]] = None
+        self.baudrate: int = 2000000
```
patrickelectric, line 16: This may not work in all scenarios

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+
+        # Open device with 2M baud to setup Ping360
+        logging.info("Setting baud to 2M...")
+        ser = serial.Serial(self.ping.port.device, self.baudrate)
+        ser.send_break()
+        ser.write("UUUUUUU".encode())
+        ser.close()
+        target_device = self.ping.port.device
+        device_name = target_device.strip().split("/")[-1]
```
patrickelectric, line 30: Path.name ?

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,48 @@
...
+    def set_low_latency_and_high_baud(self) -> None:
+        """
+        sets the latency_timer for the device and attemps to set the ping360 to a 2M baudrate
+        """
+
+        # Open device with 2M baud to setup Ping360
+        logging.info("Setting baud to 2M...")
+        ser = serial.Serial(self.ping.port.device, self.baudrate)
+        ser.send_break()
+        ser.write("UUUUUUU".encode())
```
patrickelectric, line 27: `7*"U".encode()`

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,48 @@
...
+        ser.close()
+        target_device = self.ping.port.device
+        device_name = target_device.strip().split("/")[-1]
+        latency_file = "/sys/bus/usb-serial/devices/{0}/latency_timer".format(device_name)
```
patrickelectric, line 31: fstring ?

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,48 @@
...
+            p.write("1")
+            p.flush()
+
+    def start(self) -> None:
+        self.set_low_latency_and_high_baud()
+        bridges = which("bridges")
+        command_line = f"{bridges} -u 0.0.0.0:{self.port} -p {self.ping.port.device}:{self.baudrate}"
+        logging.info(f"launching {command_line}")
+        # pylint: disable=consider-using-with
+        self.process = Popen(command_line.split(" "))
```
patrickelectric, line 44: You should use `shlex.split`, the result of split(" ") is different, since it may break with multiple arguments per flag

core/services/ping/ping1d_driver.py
patrickelectric: You should do it in `PingDriver` base class

core/services/ping/pingdriver.py
```diff
@@ -0,0 +1,13 @@
...
+class PingDriver:
+    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+        self.ping = ping
+        self.port = port
+
+    def start(self) -> None:
+        raise NotImplementedError
+
+    def stop(self) -> None:
+        raise NotImplementedError
```
patrickelectric, line 13: Do del stop here.

core/services/ping/portwatcher.py
```diff
@@ -0,0 +1,91 @@
...
+    def set_port_post_callback(self, callback: Callable[[SysFS], None]) -> None:
+        self.port_lost_callback = callback
+
+    def port_should_be_probed(self, port: SysFS) -> bool:
+        """A port should be probed if there hasn't been MAX_ATTEMPTS to probe it yet
+        and it is caught by our filters
+        """
+        if self.probe_attempts_counter.get(port, 0) >= MAX_ATTEMPTS:
+            return False
+        return any(filter(port) for filter in self.filters)
```
patrickelectric, line 47: this is a bit weird since filter is also an python built-in function. maybe `port_filter(port)` ?

patrickelectric (changes requested):
Just an initial review

core/services/ping/ping360_driver.py
Williangalvani: you mean 2M for baudrate? as in "some setups may fail to communicate with the ping360 at 2M"?

I can add some logic to raise the baudrate progressively until talking to the sensor fails, then go one step back. Thoughts?

core/services/ping/portwatcher.py
Williangalvani: maybe a user-configurable black/whitelist?
The issue with a blacklist, is that the user will need know that the probing will happen, and that his device will need to handle it or he'll need to blacklist it in the ping(or higher level in core) blacklist.

core/services/ping/ping360_driver.py
patrickelectric: Yep, we could raise or decrease the baudrate, your choice, I just recommend to do N times to make sure that is a good baudrate, we could also have a dynamic baudrate change while using the sensor and not changing only when starting the connection, but that's up to you.

core/services/ping/portwatcher.py
patrickelectric: But the same thing happens to the whitelist, like an user with an arduino that contains a ftdi. 
From what I saw already is that users may not use FT230/FT231X, there are people also that uses Ping360 with an RS485 to USB adapter.

Should we let it open for users to configure ? If so, should we have an REST API for that ? Should we support both white and blacklist or only one ?

core/services/ping/ping360_driver.py
patrickelectric: I think that you should start at the lowest baudrate, if it does not communicate we could just ignore it and start with the next candidate.

core/services/ping/portwatcher.py
Williangalvani: can we get a linter to catch these?
somethign like this?
https://pypi.org/project/flake8-builtins/

patrickelectric: We are already testing such thing, are you running or pre-push file ?
You should have: 
```
W0622: Redefining built-in 'filter' (redefined-builtin)
```

Williangalvani: yes, and it also passed CI...

patrickelectric: I have created an issue, it appears to be a bug.
https://github.com/PyCQA/pylint/issues/4472

core/services/ping/ping360_driver.py
patrickelectric: image

core/services/ping/portwatcher.py
Williangalvani: I'm wondering if maybe it doesn't count because its scope is just that line

patrickelectric: Ok, but we should just update the name to avoid such problem when possible, otherwise is a bit confuse.

Williangalvani: sure, done.

core/services/ping/portwatcher.py
Williangalvani: What I have in mind right now is to have one blacklist for the whole system.
This way the user can tell us "please don't touch this device". and we avoid it in all our services.
we would need to expose this blacklist eventually (maybe on a settings editor?).

For our own devices, I don't expect we'll have trouble making sure they handle being probed.

My concern is probing badly-written arduino peripherals done by our users and putting them in a state where the user's service is unable to work properly.

Williangalvani: After chatting to @patrickelectric :
Try to udev rules to link all not-blacklisted devices to /dev/companion. Our drivers should test the devices in /dev/companion only.

Williangalvani: I opened https://github.com/bluerobotics/companion-docker/issues/237

core/services/ping/ping1d_driver.py
```diff
@@ -0,0 +1,18 @@
...
+from pingdriver import PingDriver
+from pingutils import PingDeviceDescriptor
+
+
+class Ping1DDriver(PingDriver):
+    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+        super().__init__(ping, port)
+
+    def start(self) -> None:
+        bridges = which("bridges")
```
patrickelectric, line 14: cool

core/services/ping/pingprober.py
```diff
@@ -0,0 +1,102 @@
...
+        device was not detected
+        """
+
+        try:
+            ping = PingDevice()
+            ping.connect_serial(port.device, 115200)
+
+        except Exception as exception:
+            if exception.args[0] and "Errno 16" in exception.args[0]:
+                logging.info(f"device {port.hwid} is busy. Re-trying...")
```
patrickelectric, line 66: Some log messages start capitalized and others not, be consistent.

patrickelectric (changes requested):

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,83 @@
...
+            logging.info(f"baudrate {baud} {baud==last_valid_baud}... going up...")
+        return last_valid_baud
+
+    def start(self) -> None:
+        baud = self.detect_highest_baud()
+        self.set_sensor_baudrate(self.ping.port.device, baud)
+        self.set_low_latency()
+        bridges = which("bridges")
+        command_line = f"{bridges} -u 0.0.0.0:{self.port} -p {self.ping.port.device}:{baud}"
+        logging.info(f"launching {command_line}")
```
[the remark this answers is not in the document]
Williangalvani: while there is some duplicated code right now, the ping1D implementarion will change soon to something that actually talks ping and sends messages to the autopilot.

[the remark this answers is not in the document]
Williangalvani: I'm not so sure about creating a class to abstract bridges alone. There are 4 lines of code for it. And while this code repeats in Ping1D, the implemetation there will stop using bridges soon, as we'll need a proper man-in-the-middle to extract data and send it to the autopilot.

core/services/ping/ping360_driver.py
```diff
@@ -0,0 +1,83 @@
...
+from subprocess import Popen
+
+import serial
+from brping import PingDevice
+from brping.definitions import COMMON_DEVICE_INFORMATION
+
+from pingdriver import PingDriver
+from pingutils import PingDeviceDescriptor
+
+BAUDRATES = [
```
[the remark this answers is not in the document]
Williangalvani: how relevant is ABR for ping1d? does it make anything faster/better?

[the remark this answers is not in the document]
patrickelectric: Is not necessary, but the latest and future firmwares supports, I'm saying this to unify both serial integrations in a single class to avoid duplicate/unnecessary in both classes and share the same know behavior when using via serial.

core/services/ping/bridges.py
```diff
@@ -0,0 +1,26 @@
...
+        logging.info(f"Launching {command_line}")
+        # pylint: disable=consider-using-with
+        self.process = Popen(shlex.split(command_line))
+
+    def stop(self) -> None:
+        if self.process:
+            self.process.kill()
+
+    def __del__(self) -> None:
+        self.stop()
```
patrickelectric, line 26: :ok_hand:

patrickelectric (changes requested):
LGTM

core/services/ping/pingdriver.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        failure_threshold = 0.1  # allow up to 10% failure rate
+        attempts = 10  # try up to 10 times per baudrate
+        max_failures = attempts * failure_threshold
+
+        last_valid_baud = Baudrates.b115200
+        for baud in Baudrates:
+            # Open device with 2M baud to setup Ping360
+            logging.debug(f"Trying baud {baud}...")
+            set_sensor_baudrate(self.ping.port, baud)
```
jaxxzer, line 30: ping device got it taken care of:
https://github.com/bluerobotics/ping-python/blob/deployment/brping/device.py#L38

core/services/ping/pingutils.py
```diff
@@ -0,0 +1,33 @@
+from dataclasses import dataclass
+from enum import IntEnum
+
+from serial.tools.list_ports_linux import SysFS
+
+
+class PingType(IntEnum):
```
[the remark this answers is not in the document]
Williangalvani: I couldn't find it.

patrickelectric (approved):
LGTM :shipit:

patrickelectric (commented):
And the others points in your bullet ? 
Should be done in this PR ?

Williangalvani:
> And the others points in your bullet ?
> Should be done in this PR ?

I'll extract then onto an issue so we can merge this and get them done in subsequent prs

PR #266, opened 2021-05-21
Core: remove unnecessary wget installations

Williangalvani:


patrickelectric (commented):
Good PR, but for our sake, we should add
```bash
# Exit immediately if a command exits with a non-zero status
set -e
```
after the shebang, just to avoid any future problem if wget is missing.

We also should change wget installation order:
https://github.com/bluerobotics/companion-docker/blob/master/core/Dockerfile#L18

patrickelectric (approved):

PR #268, opened 2021-05-21
Use companion-base image for core

patrickelectric:
Build time improvement for everyone!
image

Replace #227


Williangalvani (approved):
LGTM, but I think you accidentally missed something in the commit message.

PR #271, opened 2021-05-24
core: ardupilot-manager: Add proper logging

rafaellehmkuhl:
- Move "prints" and "warns" to logging when appropriate 
- Make logging output non-verbose by default
- Add option to use verbose output with -v.

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -106,6 +106,7 @@ class ArduPilotManager(metaclass=Singleton):
         boards.sort(key=lambda tup: tup[0].value)
 
         flight_controller_type, place = boards[0]
+        logger.info(f"Board in use: {flight_controller_type.name}.")
 
         if FlightControllerType.Navigator == flight_controller_type:
             self.start_navigator()
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -4,9 +4,9 @@ import stat
 import subprocess
 import time
 from copy import deepcopy
-from pprint import pprint
 from typing import Any, List, Optional, Set, Tuple
-from warnings import warn
+
+from loguru import logger
 
 from firmware_download.FirmwareDownload import FirmwareDownload, Vehicle
 from flight_controller_detector.Detector import Detector as BoardDetector
@@ -24,8 +24,8 @@ class ArduPilotManager(metaclass=Singleton):
 
         # Load settings and do the initial configuration
         if self.settings.load():
-            print(f"Loaded settings from {self.settings.settings_file}:")
-            pprint(self.settings.content)
+            logger.info(f"Loaded settings from {self.settings.settings_file}.")
+            logger.debug(self.settings.content)
         else:
             self.settings.create_settings()
 
@@ -38,7 +38,7 @@ class ArduPilotManager(metaclass=Singleton):
         ArduPilotManager.check_running_as_root()
 
         while not self.start_board(BoardDetector.detect()):
-            print("Flight controller board not detected, will try again.")
+            logger.warning("Flight controller board not detected, will try again.")
             time.sleep(2)
 
     @staticmethod
@@ -91,7 +91,7 @@ class ArduPilotManager(metaclass=Singleton):
                 {Endpoint("MAVLink2Rest", self.settings.app_name, "udpout", "127.0.0.1", 14000, protected=True)}
             )
         except Exception as error:
-            warn(f"Could not create default GCS endpoint: {error}")
+            logger.error(f"Could not create default GCS endpoint: {error}")
         self.mavlink_manager.set_master_endpoint(device)
         self.mavlink_manager.start()
 
@@ -100,7 +100,7 @@ class ArduPilotManager(metaclass=Singleton):
             return False
 
         if len(boards) > 1:
-            print(f"More than a single board detected: {boards}")
+            logger.warning(f"More than a single board detected: {boards}")
 
         # Sort by priority
         boards.sort(key=lambda tup: tup[0].value)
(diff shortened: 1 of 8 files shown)


core/services/ardupilot_manager/settings.py
```diff
@@ -88,7 +89,30 @@ def save(self, content: Dict[str, Any]) -> None:
...
+            "()": "uvicorn.logger.DefaultFormatter",
+            "fmt": "%(levelprefix)s %(asctime)s %(message)s",
+            "datefmt": "%Y-%m-%d %H:%M:%S",
+        },
+    },
+    "handlers": {
+        "default": {
+            "formatter": "default",
+            "class": "logger.StreamHandler",
+            "stream": "ext://sys.stderr",
```
patrickelectric, line 112: Is there a reason to push it to stderr and not stdout ?

core/services/ardupilot_manager/settings.py
```diff
@@ -88,7 +89,30 @@ def save(self, content: Dict[str, Any]) -> None:
...
             with open(self.settings_file, "w+") as file:
-                print(f"Updating settings file: {self.settings_file}")
+                logger.info(f"Updating settings file: {self.settings_file}")
                 json.dump(self.root, file, sort_keys=True, indent=4)
         except Exception as error:
-            warn(f"Could not save settings to disk: {error}")
+            logger.warning(f"Could not save settings to disk: {error}")
+
+
+log_config = {
```
patrickelectric, line 98: Where is this being used ?

core/services/ardupilot_manager/main.py
```diff
@@ -68,4 +104,4 @@ def remove_endpoints(response: Response, endpoints: Set[Endpoint] = Body(...)) -
 
 if __name__ == "__main__":
     autopilot.run()
-    uvicorn.run(app, host="0.0.0.0", port=8000)
+    uvicorn.run(app, host="0.0.0.0", port=8000, log_config=None)
```
patrickelectric, line 107: Why log_config is None ? Can you add a comment explaining ?

patrickelectric (changes requested):
The patch looks good, but some things are not clear, maybe a couple of comments may help.

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: First thing: this settings.json change was not supposed to go ahead, so I'm removing that.

Despite that, the loguru default call specifies that, so it's still a valid discussion.

I basically have no strong reason, besides it being the default place that both [python](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial) and [loguru](https://github.com/Delgan/loguru/blob/master/loguru/__init__.py) use as sink.

core/services/ardupilot_manager/settings.py
rafaellehmkuhl: Not being. It was a test and I forgot to remove it.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: Because we are capturing the uvicorn logging output to loguru, so we disable output by uvicorn itself.
Do you mean comment on the code or the commit message?

rafaellehmkuhl:
Regarding the type-errors we were having before, I'm suppressing them as the library itself is still on the process of implementing the type-hints. 

I've filed [an issue](https://github.com/Delgan/loguru/issues/449) to track that there.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: Added a comment on the line before.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -100,7 +100,7 @@ def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
             return False
 
         if len(boards) > 1:
-            print(f"More than a single board detected: {boards}")
+            logger.warning(f"More than a single board detected: {boards}")
```
Williangalvani, line 103: does it say which one is going to be used?

rafaellehmkuhl: No. Can add that.

core/services/ardupilot_manager/main.py
```diff
@@ -15,6 +19,36 @@
 FRONTEND_FOLDER = Path.joinpath(Path(__file__).parent.absolute(), "frontend")
 
 
+parser = argparse.ArgumentParser(description="ArduPilot Manager service for Bluerobotics Companion")
```
patrickelectric, line 22:
```suggestion
parser = argparse.ArgumentParser(description="ArduPilot Manager service for Blue Robotics Companion")
```

core/services/ardupilot_manager/main.py
```diff
@@ -15,6 +19,36 @@
...
+parser.add_argument("-v", "--verbose", help="increase output verbosity", action="store_true")
+
+args = parser.parse_args()
+
+if not args.verbose:
```
patrickelectric, line 27: Just to be sure, will this print errors and warnings by default (verbose as false) ?

patrickelectric (changes requested):

core/services/ardupilot_manager/main.py
rafaellehmkuhl: Yes. Verbose here is in the meaning that more information regarding each log message will be outputed (module and code line).

As an open source project with no sensitive information being exposed, I assume we should log debugs and warnings by default.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/main.py
patrickelectric: just to clear, the only difference between verbose and not verbose is only the log format and not the output of such messages by category ?

rafaellehmkuhl: Yes.

patrickelectric: So, this is not what verbose means generally. 
I would recommend to make the 'pretty' format default and remove the verbose command at all since this does not change the verbosity and only the format.

rafaellehmkuhl: Ok, but the different format adds relevant information for debug.
Do you don't think we should completely remove it or have some suggestion on how to better implement that?

rafaellehmkuhl: Oh, ok. So to remove the current default (which does not have the module name and code line)?

patrickelectric: yes

rafaellehmkuhl: Done.

core/services/ardupilot_manager/main.py
```diff
@@ -15,6 +17,26 @@
 FRONTEND_FOLDER = Path.joinpath(Path(__file__).parent.absolute(), "frontend")
 
 
+class InterceptHandler(logging.Handler):
+    def emit(self, record) -> None:  # type: ignore
+        # Get corresponding Loguru level if it exists
+        try:
```
patrickelectric, line 23: Just to make sure, since we'll use loguru for now on, is this try still necessary ?

patrickelectric (approved):

PR #282, opened 2021-05-26
Core: Add tmux configuration file

### LINKED ISSUE #279: core: tmux: Add configuration file for human friendly controls

patrickelectric:
Fix #279

### DIFF
--- a/core/configuration/tmux.conf
+++ b/core/configuration/tmux.conf
@@ -1,0 +1,8 @@
+# Basic configuration
+set-option -g history-limit 10000
+
+# Enable mouse over splits
+set -g mouse on
+
+# Force utf-8 usage for non ASCII characters
+setw -gq utf8 on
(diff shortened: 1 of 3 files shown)


core/configuration/tmux.conf
```diff
@@ -0,0 +1,8 @@
+# Basic configuration
+set-option -g history-limit 10000
+
+# Enable mouse over splits
+set -g mouse on
+
+# Force utf-8 usage for non ASCII characters
+setw -gq utf8 on
```
Williangalvani, line 8: what does this do in practice?

patrickelectric: Turning utf8 on solves some problems when running tmux under windows terminal and others, so if someone is using it and tmux is the default endpoint, we should at least enable to avoid any related problem.

Issue #201, opened 2021-04-28
Tests missing on ArduPilotManager


rafaellehmkuhl:
ArduPilotManager service has tests justo on the MavlinkProxy part.
Both the ArduPilotManager class itself, and the main file (REST API) don't have tests.

rafaellehmkuhl:
Improvements:
 - Test ArduPilotManager class
 - Test REST API
 - Test if master/output endpoints truly work (try to connect to them)
 - Test latency on mavlink messages (between flight-controller/router/output)

Issue #170, opened 2021-04-19
Navigator ardusub does not come back after mavlink reboot command


jaxxzer:
In the past we would restart the binary after the process exits: https://github.com/bluerobotics/companion-docker/blob/8bfb13a55f3941b6986ea3c0b4d7982dd19bc6bd/core/services/ardupilot_manager/main.py#L32

patrickelectric:
ref: https://github.com/ArduPilot/ardupilot/issues/17572

PR #62, opened 2020-10-01
Create Version Chooser

Williangalvani:
This PR adds a Version Chooser service.

It lists all available Companion images, online and locally.
Pulls new images from dockerhub if necessary.
Updates `startup.json` to use another available image.
KIlls companion-core.

After companion-core dies. Bootstrap takes over and attempts to launch it from the edited startup.json. If the launch fails, it reverts to a default startup.json (contained in bootstrap itself).

Convenience startup.json:
```

{
    "core": {
        "tag": "versionchooser",
        "image": "williangalvani/companion-core",
        "enabled": true,
        "webui": false,
        "network": "host",
        "binds": {
            "/dev/": {
                "bind": "/dev/",
                "mode": "rw"
            },
            "/var/run/wpa_supplicant/wlan0": {
                "bind": "/var/run/wpa_supplicant/wlan0",
                "mode": "rw"
            },
            "/tmp/wpa_playground": {
                "bind": "/tmp/wpa_playground",
                "mode": "rw"
            },
            "/var/run/docker.sock": {
                "bind": "/var/run/docker.sock",
                "mode": "rw"
            }
        },
        "privileged": true
    }
}




```

PR #289, opened 2021-05-26
Add SITL to ArduPilot Manager

### LINKED ISSUE #98: Include SITL in ArduPilotManager
Opens the possibility for the users to test the Companion functionalities without a Pixhawk or Navigator connected.

rafaellehmkuhl:
Fix #98 

Use `--sitl` to activate it.

When [this PR](https://github.com/ArduPilot/ardupilot/pull/17573) gets approved we should reintroduce validation of the SITL binary.

~~Missing tests with Rasp/Navigator.~~
Tested with x86 and ARMv7 platforms.

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -10,7 +10,7 @@ from loguru import logger
 from firmware_download.FirmwareDownload import FirmwareDownload, Platform, Vehicle
 from flight_controller_detector.Detector import Detector as BoardDetector
 from flight_controller_detector.Detector import FlightControllerType
-from mavlink_proxy.Endpoint import Endpoint
+from mavlink_proxy.Endpoint import Endpoint, EndpointType
 from mavlink_proxy.Manager import Manager as MavlinkManager
 from settings import Settings
 from Singleton import Singleton
@@ -53,7 +53,9 @@ class ArduPilotManager(metaclass=Singleton):
             shutil.move(str(temporary_file), firmware)
 
         # ArduPilot process will connect as a client on the UDP server created by the mavlink router
-        master_endpoint = Endpoint("Master", self.settings.app_name, "udpin", "127.0.0.1", 8852, protected=True)
+        master_endpoint = Endpoint(
+            "Master", self.settings.app_name, EndpointType.UDPServer, "127.0.0.1", 8852, protected=True
+        )
 
         # Run ardupilot inside while loop to avoid exiting after reboot command
         ## Can be changed back to a simple command after https://github.com/ArduPilot/ardupilot/issues/17572
@@ -73,7 +75,9 @@ class ArduPilotManager(metaclass=Singleton):
         self.start_mavlink_manager(master_endpoint)
 
     def start_serial(self, device: str) -> None:
-        self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
+        self.start_mavlink_manager(
+            Endpoint("Master", self.settings.app_name, EndpointType.Serial, device, 115200, protected=True)
+        )
 
     def run_with_sitl(self, vehicle: str = "vectored") -> None:
         firmware = os.path.join(self.settings.firmware_path, "sitl")
@@ -83,7 +87,9 @@ class ArduPilotManager(metaclass=Singleton):
             shutil.move(str(temporary_file), firmware)
 
         # ArduPilot SITL binary will bind TCP port 5760 (server) and the mavlink router will connect to it as a client
-        master_endpoint = Endpoint("Master", self.settings.app_name, "tcpin", "127.0.0.1", 5760, protected=True)
+        master_endpoint = Endpoint(
+            "Master", self.settings.app_name, EndpointType.TCPServer, "127.0.0.1", 5760, protected=True
+        )
         # pylint: disable=consider-using-with
         self.subprocess = subprocess.Popen(
             [
@@ -105,10 +111,23 @@ class ArduPilotManager(metaclass=Singleton):
     def start_mavlink_manager(self, device: Endpoint) -> None:
         try:
             self.add_new_endpoints(
-                {Endpoint("GCS Link", self.settings.app_name, "udpout", "192.168.2.1", 14550, protected=True)}
+                {
+                    Endpoint(
+                        "GCS Link", self.settings.app_name, EndpointType.UDPClient, "192.168.2.1", 14550, protected=True
+                    )
+                }
             )
             self.add_new_endpoints(
-                {Endpoint("MAVLink2Rest", self.settings.app_name, "udpout", "127.0.0.1", 14000, protected=True)}
+                {
+                    Endpoint(
+                        "MAVLink2Rest",
+                        self.settings.app_name,
+                        EndpointType.UDPClient,
+                        "127.0.0.1",
+                        14000,
+                        protected=True,
+                    )
+                }
             )
         except Exception as error:
             logger.error(f"Could not create default GCS endpoint: {error}")
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -1,6 +1,5 @@
 import os
 import shutil
-import stat
 import subprocess
 import time
 from copy import deepcopy
@@ -34,7 +33,7 @@ class ArduPilotManager(metaclass=Singleton):
         self.subprocess: Optional[Any] = None
         self.firmware_download = FirmwareDownload()
 
-    def run(self) -> None:
+    def run_with_board(self) -> None:
         ArduPilotManager.check_running_as_root()
 
         while not self.start_board(BoardDetector.detect()):
@@ -52,12 +51,6 @@ class ArduPilotManager(metaclass=Singleton):
             temporary_file = self.firmware_download.download(Vehicle.Sub, Platform.Navigator)
             assert temporary_file, "Failed to download navigator binary."
             shutil.move(str(temporary_file), firmware)
-            # Make the binary executable
-            os.chmod(firmware, stat.S_IXOTH)
-        try:
-            subprocess.check_output([firmware, "--help"])
-        except Exception as error:
-            raise RuntimeError(f"Failed to start navigator: {error}") from error
 
         # ArduPilot process will connect as a client on the UDP server created by the mavlink router
         master_endpoint = Endpoint("Master", self.settings.app_name, "udpin", "127.0.0.1", 8852, protected=True)
@@ -81,6 +74,33 @@ class ArduPilotManager(metaclass=Singleton):
 
     def start_serial(self, device: str) -> None:
         self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
+
+    def run_with_sitl(self, vehicle: str = "vectored") -> None:
+        firmware = os.path.join(self.settings.firmware_path, "sitl")
+        if not os.path.exists(firmware):
+            temporary_file = self.firmware_download.download(Vehicle.Sub, Platform.SITL)
+            assert temporary_file, "Failed to download SITL binary."
+            shutil.move(str(temporary_file), firmware)
+
+        # ArduPilot SITL binary will bind TCP port 5760 (server) and the mavlink router will connect to it as a client
+        master_endpoint = Endpoint("Master", self.settings.app_name, "tcpin", "127.0.0.1", 5760, protected=True)
+        # pylint: disable=consider-using-with
+        self.subprocess = subprocess.Popen(
+            [
+                firmware,
+                "--model",
+                vehicle,
+                "--base-port",
+                str(master_endpoint.argument),
+                "--home",
+                "-27.563,-48.459,0.0,270.0",
+            ],
+            shell=False,
+            encoding="utf-8",
+            errors="ignore",
+        )
+
+        self.start_mavlink_manager(master_endpoint)
 
     def start_mavlink_manager(self, device: Endpoint) -> None:
         try:
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -8,7 +8,7 @@ from typing import Any, List, Optional, Set, Tuple
 
 from loguru import logger
 
-from firmware_download.FirmwareDownload import FirmwareDownload, Vehicle
+from firmware_download.FirmwareDownload import FirmwareDownload, Platform, Vehicle
 from flight_controller_detector.Detector import Detector as BoardDetector
 from flight_controller_detector.Detector import FlightControllerType
 from mavlink_proxy.Endpoint import Endpoint
@@ -49,7 +49,7 @@ class ArduPilotManager(metaclass=Singleton):
     def start_navigator(self) -> None:
         firmware = os.path.join(self.settings.firmware_path, "ardusub")
         if not os.path.isfile(firmware):
-            temporary_file = self.firmware_download.download(Vehicle.Sub, "Navigator")
+            temporary_file = self.firmware_download.download(Vehicle.Sub, Platform.Navigator)
             assert temporary_file, "Failed to download navigator binary."
             shutil.move(str(temporary_file), firmware)
             # Make the binary executable
--- a/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
+++ b/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
@@ -4,9 +4,12 @@ import os
 import pathlib
 import random
 import ssl
+import stat
 import string
+import subprocess
 import tempfile
 from enum import Enum
+from platform import machine
 from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
 from urllib.request import urlopen, urlretrieve
@@ -18,6 +21,12 @@ from packaging.version import Version
 # Disable SSL verification
 if not os.environ.get("PYTHONHTTPSVERIFY", "") and getattr(ssl, "_create_unverified_context", None):
     ssl._create_default_https_context = ssl._create_unverified_context
+
+
+def get_sitl_platform_name() -> str:
+    if machine() == "x86_64":
+        return "SITL_x86_64_linux_gnu"
+    return "SITL_arm_linux_gnueabihf"
 
 
 class Vehicle(str, Enum):
@@ -36,11 +45,24 @@ class Platform(str, Enum):
 
     Pixhawk1 = "Pixhawk1"
     Navigator = "Navigator"
+    SITL = get_sitl_platform_name()
+
+
+class FirmwareFormat(str, Enum):
+    """Valid firmware formats to download.
+    The Enum values are 1:1 representations of the formats available on the ArduPilot manifest."""
+
+    APJ = "apj"
+    ELF = "ELF"
 
 
 class FirmwareDownload:
     _manifest_remote = "https://firmware.ardupilot.org/manifest.json.gz"
-    _supported_firmware_format = "apj"
+    _supported_firmware_formats = {
+        Platform.SITL: FirmwareFormat.ELF,
+        Platform.Pixhawk1: FirmwareFormat.APJ,
+        Platform.Navigator: FirmwareFormat.ELF,
+    }
 
     def __init__(self) -> None:
         self._manifest: Dict[str, Any] = {}
@@ -103,11 +125,18 @@ class FirmwareDownload:
         Returns:
             bool: True if valid, False if otherwise.
         """
-        with open(firmware_path, "r") as firmware_file:
-            data = json.load(firmware_file)
-            keys = data.keys()
-            if "image_size" in keys and "image" in keys:
+        if FirmwareFormat.APJ.value in firmware_path.suffix:
+            with open(firmware_path, "r") as firmware_file:
+                data = json.load(firmware_file)
+                keys = data.keys()
+                if "image_size" in keys and "image" in keys:
+                    return True
+        else:
+            try:
+                subprocess.check_output([firmware_path, "--help"])
                 return True
+            except Exception as error:
+                logger.error(f"Firmware help menu not available. Binary not validated. {error}")
 
         return False
 
@@ -182,12 +211,11 @@ class FirmwareDownload:
         if not self._manifest_is_valid():
             return available_versions
 
-        items = self._find_version_item(
-            vehicletype=vehicle.value, platform=platform.value, format=FirmwareDownload._supported_firmware_format
-        )
+        items = self._find_version_item(vehicletype=vehicle.value, platform=platform.value)
 
         for item in items:
-            available_versions.append(item["mav-firmware-version-type"])
+            if item["format"] == FirmwareDownload._supported_firmware_formats[platform]:
+                available_versions.append(item["mav-firmware-version-type"])
 
         return available_versions
 
@@ -207,7 +235,10 @@ class FirmwareDownload:
             return FirmwareDownload._download_navigator()
 
         versions = self.get_available_versions(vehicle, platform)
+        logger.debug(f"Got following versions for {vehicle} running {platform}: {versions}")
+
         if version and versions and version not in versions:
+            logger.error(f"Specified version not found for this configuration ({vehicle} and {platform}).")
             return None
 
         # Autodetect the latest stable version
@@ -219,22 +250,30 @@ class FirmwareDownload:
                 if not newest_version or Version(newest_version) < Version(semver_version):
                     newest_version = semver_version
             if not newest_version:
+                logger.error(f"No firmware versions found for this configuration ({vehicle} and {platform}).")
                 return None
             version = f"STABLE-{newest_version}"
+
+        firmware_format = FirmwareDownload._supported_firmware_formats[platform]
 
         items = self._find_version_item(
             vehicletype=vehicle.value,
             platform=platform.value,
             mav_firmware_version_type=version,
-            format=FirmwareDownload._supported_firmware_format,
+            format=firmware_format,
         )
 
         if len(items) != 1:
-            logger.error(f"Invalid number of candidates to download: {len(items)}")
+            logger.error(f"Invalid number of candidates to download ({len(items)}): {items}")
             return None
 
         item = items[0]
+        logger.debug(f"Downloading following firmware: {item}")
         path = FirmwareDownload._download(item["url"])
+        if firmware_format == FirmwareFormat.ELF:
+            # Make the binary executable
+            os.chmod(str(path), stat.S_IXOTH)
+
         if not path or not FirmwareDownload._validate_firmware(path):
             logger.error("Unable to validate firmware file.")
             return None
--- a/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
+++ b/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
@@ -6,7 +6,7 @@ import random
 import ssl
 import string
 import tempfile
-from enum import IntEnum
+from enum import Enum
 from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
 from urllib.request import urlopen, urlretrieve
@@ -20,13 +20,13 @@ if not os.environ.get("PYTHONHTTPSVERIFY", "") and getattr(ssl, "_create_unverif
     ssl._create_default_https_context = ssl._create_unverified_context
 
 
-class Vehicle(IntEnum):
+class Vehicle(str, Enum):
     """Valid vehicle types to download"""
 
-    Sub = 1
-    Rover = 2
-    Plane = 3
-    Copter = 4
+    Sub = "Sub"
+    Rover = "Rover"
+    Plane = "Plane"
+    Copter = "Copter"
 
 
 class FirmwareDownload:
@@ -174,7 +174,7 @@ class FirmwareDownload:
             return available_versions
 
         items = self._find_version_item(
-            vehicletype=vehicle.name, platform=platform, format=FirmwareDownload._supported_firmware_format
+            vehicletype=vehicle.value, platform=platform, format=FirmwareDownload._supported_firmware_format
         )
 
         for item in items:
@@ -214,7 +214,7 @@ class FirmwareDownload:
             version = f"STABLE-{newest_version}"
 
         items = self._find_version_item(
-            vehicletype=vehicle.name,
+            vehicletype=vehicle.value,
             platform=platform,
             mav_firmware_version_type=version,
             format=FirmwareDownload._supported_firmware_format,
--- a/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
+++ b/core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
@@ -241,20 +241,26 @@ class FirmwareDownload:
             logger.error(f"Specified version not found for this configuration ({vehicle} and {platform}).")
             return None
 
-        # Autodetect the latest stable version
-        if not version:
-            stable_versions = [version for version in versions if "STABLE" in version]
-            newest_version: Optional[str] = None
-            for stable_version in stable_versions:
-                semver_version = stable_version.split("-")[1]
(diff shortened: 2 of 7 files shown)


core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
```diff
@@ -37,6 +37,8 @@ def convert_endpoint(endpoint: Endpoint) -> str:
                 master_string = f"{master.place}"
             else:
                 master_string = f"{serial_endpoint_as_input(master)}"
+        if master.connection_type == EndpointType.TCPServer:
+            master_string = f"tcp:{master.place}:{master.argument}"
```
patrickelectric, line 41: Is there a test for it ?

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -60,7 +60,7 @@ def convert_endpoint(endpoint: Endpoint) -> str:
         log = f"--log {self.logdir().resolve()}"
 
         if master.connection_type == EndpointType.TCPServer:
-            return f"{self.binary()} --tcp-port {master.argument} {log} {endpoints}"
+            return f"{self.binary()} --tcp-endpoint {master.place}:{master.argument} --tcp-port 0 {log} {endpoints}"
```
patrickelectric, line 63: Add a comment explaining about `tcp-port 0` and why is necessary

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +79,50 @@ def start_navigator(self) -> None:
     def start_serial(self, device: str) -> None:
         self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
 
+    @staticmethod
+    def get_current_arch() -> str:
+        """Maps paltform.machine() outputs to docker architectures"""
+        arch_map = {"x86_64": "amd64", "aarch64": "arm", "armv7l": "arm"}  # TODO: differentiate arm64v8/ arm32v7
+        machine = platform.machine()
```
patrickelectric, line 86:
```suggestion
        local_arch = platform.machine()
```

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +79,50 @@ def start_navigator(self) -> None:
     def start_serial(self, device: str) -> None:
         self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
 
+    @staticmethod
+    def get_current_arch() -> str:
+        """Maps paltform.machine() outputs to docker architectures"""
+        arch_map = {"x86_64": "amd64", "aarch64": "arm", "armv7l": "arm"}  # TODO: differentiate arm64v8/ arm32v7
```
patrickelectric, line 85: if it does not contains x86 is arm, you don't need to have all necessary cases for it, ardupilot only supports x86 and arm.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -34,11 +36,25 @@ class Platform(str, Enum):
 
     Pixhawk1 = "Pixhawk1"
     Navigator = "Navigator"
+    SITL_ARM = "SITL_arm_linux_gnueabihf"
+    SITL_X86 = "SITL_x86_64_linux_gnu"
```
patrickelectric, line 40: This is unnecessary, you can have it as:
`SITL = detect_sitl_string()`

This will avoid all logic related to both SITL versions.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +79,50 @@ def start_navigator(self) -> None:
...
+
+    def start_sitl(self) -> None:
+        current_arch = self.get_current_arch()
+        logger.debug(f"Architecture detected: {current_arch}")
+        firmware = os.path.join(self.settings.firmware_path, "sitl_ardusub")
+        if not os.path.exists(firmware):
+            if current_arch == "amd64":
+                temporary_file = self.firmware_download.download(Vehicle.Sub, Platform.SITL_X86)
+            elif current_arch == "arm":
+                temporary_file = self.firmware_download.download(Vehicle.Sub, Platform.SITL_ARM)
```
patrickelectric, line 100: `plataform = Platform.SITL_X86 if current_arch != "arm" else platform.SITL_ARM`

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -34,11 +36,25 @@ class Platform(str, Enum):
...
+
+
+class FirmwareFormat(str, Enum):
+    """Valid firmware formats to download"""
+
+    APJ = "apj"
+    ELF = "ELF"
```
patrickelectric, line 47: Is there a reason for the capitalization here ? Can you add a comment explaining ?

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -34,12 +34,15 @@ def __init__(self) -> None:
...
 
-    def run(self) -> None:
+    def run(self, sitl: bool = False) -> None:
         ArduPilotManager.check_running_as_root()
 
-        while not self.start_board(BoardDetector.detect()):
-            logger.warning("Flight controller board not detected, will try again.")
-            time.sleep(2)
+        if sitl:
+            self.start_sitl()
```
patrickelectric, line 41: Is better to have two different runs than a run with a bool specific for sitl, this makes the logic a bit more confuse here.

patrickelectric (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Just to make sure, you're saying to remove aarch64 (ARM64), correct?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: But they are different binaries.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Agree

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
patrickelectric: Yes, since we are not going to run X86 in ARM and ARM in x86, it does not make sense to have both, we can have the string for comparison done in runtime and remove the extra enum and all the logic behind both.

core/services/ardupilot_manager/mavlink_proxy/MAVProxy.py
rafaellehmkuhl: There is, but the tests currently only verify if the binary runs, but not it's output.

There are several limitations to the ardupilot-manager tests today. Issue #201 is tracking that and I will open a PR focused on extending and fixing them.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Ok. Got it now.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: I'm trying to keep the Enums 1:1 with the manifest. They are like this there.

patrickelectric: So, can you add this information in the code documentation, otherwise this is not clear.

rafaellehmkuhl: Done.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Removed aarch64.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -60,7 +60,8 @@ def convert_endpoint(endpoint: Endpoint) -> str:
         log = f"--log {self.logdir().resolve()}"
 
         if master.connection_type == EndpointType.TCPServer:
-            return f"{self.binary()} --tcp-port {master.argument} {log} {endpoints}"
+            # The "--tcp-port 0" argument is used to prevent the router from opening a default tcp port on 5760
```
patrickelectric, line 63:
```suggestion
            # The "--tcp-port 0" argument is used to prevent the router to bind by default tcp port on 5760
```

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +76,46 @@ def start_navigator(self) -> None:
     def start_serial(self, device: str) -> None:
         self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
 
+    @staticmethod
+    def get_current_arch() -> str:
+        """Maps paltform.machine() outputs to docker architectures"""
+        arch_map = {"x86_64": "amd64", "armv7l": "arm"}  # TODO: differentiate arm64v8/ arm32v7
+        local_arch = platform.machine()
+        arch = arch_map.get(local_arch, None)
```
patrickelectric, line 84: Sorry if I was not clear, but what I'm saying is that:
if matches X86 is X86_64, otherwise it should be arm, the dict creates an unnecessary complex logic. 
We don't need to check each type of arm, any arm will run armv6 and since we only supports armv7, everything else that is not armv6 should run armv7.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +76,46 @@ def start_navigator(self) -> None:
...
+        if not arch:
+            raise Exception(f"Unknown architecture ('{local_arch}') is not supported.")
+        return arch
+
+    def run_with_sitl(self) -> None:
```
patrickelectric, line 89: run_with_sitl should receive an argument of vehicle type.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
patrickelectric: The problem behind the actual approach is that we need to have the control over the architecture inside our code, and not inside the enum for free, making the logic unnecessary inside run_with_sitl. can you provide the diff or patch for such change ?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -101,11 +118,18 @@ def _validate_firmware(firmware_path: pathlib.Path) -> bool:
         Returns:
             bool: True if valid, False if otherwise.
         """
-        with open(firmware_path, "r") as firmware_file:
-            data = json.load(firmware_file)
-            keys = data.keys()
-            if "image_size" in keys and "image" in keys:
+        if firmware_path.suffix == ".apj":
```
patrickelectric, line 121: Why not `FirmwareFormat.APJ` ?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -217,24 +243,34 @@ def download(self, vehicle: Vehicle, platform: Platform, version: str = "") -> O
...
+        logger.debug(f"Downloading following firmware: {item}")
         path = FirmwareDownload._download(item["url"])
-        if not path or not FirmwareDownload._validate_firmware(path):
-            logger.error("Unable to validate firmware file.")
-            return None
+        if firmware_format == FirmwareFormat.ELF:
+            # Make the binary executable
+            os.chmod(str(path), stat.S_IXOTH)
+
+        # TODO: Reincorporate firmware validation for SITL once Ardupilot fixes non-zero return on --help
```
patrickelectric, line 270: Add ardupilot issue number here

patrickelectric (changes requested):

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Exactly, and this is how it's done. The SITL binaries are platform specific, like the firmware binaries are. Not having them on the Enums seems like creating one more logic for a problem that is exactly what the others are.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Don't agree. In your scenario, if it detects an ARM64 it will classify it as ARM and try to download the armv6/7 binary, which will not necessarily run on it.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: In fact this whole method came from PR #62, which was approved, and I copied it here in the intent to have common logic that will enter the common library on core in some time.
I would prefer that we try to keep the same method here and elaborate on the changes in it once we  have them on the final place, if that's possible.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Pushed last commit as patch. Let me know what you think.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -38,8 +39,7 @@ class Platform(str, Enum):
 
     Pixhawk1 = "Pixhawk1"
     Navigator = "Navigator"
-    SITL_ARM = "SITL_arm_linux_gnueabihf"
-    SITL_X86 = "SITL_x86_64_linux_gnu"
+    SITL = "SITL"
```
patrickelectric, line 42: That is wrong, you should use a function that generates the string, check my first comment

patrickelectric (changes requested):

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Ohhhh ok, now I get it. This seems more reasonable. Will change.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: The archtecture from last update seems better, that's true.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: What do you think about this last update?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -20,18 +23,54 @@
     ssl._create_default_https_context = ssl._create_unverified_context
 
 
-class Vehicle(IntEnum):
+def get_current_arch() -> str:
+    """Maps paltform.machine() outputs to docker architectures"""
+    arch_map = {"x86_64": "amd64", "armv7l": "arm"}  # TODO: differentiate arm64v8/ arm32v7
```
patrickelectric, line 28: from https://github.com/bluerobotics/companion-docker/pull/289#discussion_r640801726:

They are different things, what are you using as reference is a docker PR that sadly we need to have track of the architecture to download the correct one based on docker API.

>  if it detects an ARM64 it will classify it as ARM and try to download the armv6/7 binary, which will not necessarily run on it.

That's not true, aarch64 can run both armv6 and armv7.

ArduPilot only deploys armv6 binaries, so no armv7 or aarch64, different from docker.
Also, armv6 runs in both armv7 and aarch64, so we don't need to worry.
In docker we do not support armv6 and we probably not will since some binaries are only provided in armv7.

patrickelectric (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +75,33 @@ def start_navigator(self) -> None:
     def start_serial(self, device: str) -> None:
         self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
 
+    def run_with_sitl(self, vehicle: str = "vectored") -> None:
+        firmware = os.path.join(self.settings.firmware_path, "sitl_ardusub")
```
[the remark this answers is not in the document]
rafaellehmkuhl: Are you talking about the firmware name?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: I usually prefer the safe route, not assuming something will run if we don't know it will, but if you are confident it's not a problem, I'm ok.

core/services/ardupilot_manager/ArduPilotManager.py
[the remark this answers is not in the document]
rafaellehmkuhl: If so, I don't think it makes much sense to fix the binary name if the vehicle is a argument for the method.

What do you think about `f"sitl_{vehicle}"`?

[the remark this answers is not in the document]
rafaellehmkuhl: Forget what I said. The sitl binary is independent of the vehicle. Let's just keep "sitl" there.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -20,6 +23,12 @@
     ssl._create_default_https_context = ssl._create_unverified_context
 
 
+def get_sitl_platform() -> str:
```
Williangalvani, line 26: maybe "get_sitl_binary_name" or something?

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +75,33 @@ def start_navigator(self) -> None:
     def start_serial(self, device: str) -> None:
         self.start_mavlink_manager(Endpoint("Master", self.settings.app_name, "serial", device, 115200, protected=True))
 
+    def run_with_sitl(self, vehicle: str = "vectored") -> None:
+        firmware = os.path.join(self.settings.firmware_path, "sitl")
+        if not os.path.exists(firmware):
+            temporary_file = self.firmware_download.download(Vehicle.Sub, Platform.SITL)
```
Williangalvani, line 81: we shouldn't have to download it everytime. This is less of an issue for SITL, we should have firmware available for flashing when there's no internet connectivity, as we do in the current companion.

This is scope creep, though, so we can just open an issue.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -82,6 +75,33 @@ def start_navigator(self) -> None:
...
+            assert temporary_file, "Failed to download SITL binary."
+            shutil.move(str(temporary_file), firmware)
+
+        # ArduPilot SITL binary will bind TCP port 5760 (server) and the mavlink router will connect to it as a client
+        master_endpoint = Endpoint("Master", self.settings.app_name, "tcpin", "127.0.0.1", 5760, protected=True)
```
Williangalvani, line 86: can we use `Endpoint.TCPServer` ?
Is that going to work right on both mavproxy and mavlinkrouter? I expected TCPClient here

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -219,24 +250,34 @@ def download(self, vehicle: Vehicle, platform: Platform, version: str = "") -> O
...
+        logger.debug(f"Downloading following firmware: {item}")
         path = FirmwareDownload._download(item["url"])
-        if not path or not FirmwareDownload._validate_firmware(path):
-            logger.error("Unable to validate firmware file.")
-            return None
+        if firmware_format == FirmwareFormat.ELF:
+            # Make the binary executable
+            os.chmod(str(path), stat.S_IXOTH)
+
+        # TODO: Reincorporate firmware validation for SITL once Ardupilot fixes non-zero return on --help (PR #17573)
```
Williangalvani, line 277: That got merged to master. we could use ardusub master(dev) for sitl. I have mixed feeling about it though.
What do you think @patrickelectric ?

Williangalvani (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Didn't get it.
Currently we download the firmware only if the binary does not exist already (`if not os.path.exists(firmware)`).

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Yes. I was just replicating how it is currently, to change them all together, but I will append a commit changing that on this PR.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: But is the binary already deployed? We use the manifest to download them, so it need to be deployed to work.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Yup

rafaellehmkuhl: Maybe "get_sitl_platform_name"?

The platform is a object property, so I like the idea of maintaining it's name on the function.

Williangalvani: sunds good.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
Williangalvani: last build was today
https://firmware.ardupilot.org/Sub/latest/SITL_x86_64_linux_gnu/

rafaellehmkuhl: Perfect!
Let me try it and I update the PR.

patrickelectric: @Williangalvani can we cherry pick it to the ardusub version of navigator ?
If its master we are ok.

core/services/ardupilot_manager/ArduPilotManager.py
Williangalvani: I missed that somehow. my bad. the path is in the settings folder, which means it is in the host filesystem, correct? nice

core/services/ardupilot_manager/ArduPilotManager.py
patrickelectric: It may be necessary to use UDP here, if we send a reboot command the TCP client/server logic may fail since it connects only once and it does not reconnect alone.
@Williangalvani you are the network guy, can you confirm this ?
It may be safer to use UDP for our master.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -36,11 +45,24 @@ class Platform(str, Enum):
 
     Pixhawk1 = "Pixhawk1"
     Navigator = "Navigator"
+    SITL = get_sitl_platform_name()
```
patrickelectric, line 48: Yes, that's the way.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Ok, so the latest version is fixed, but as it's not classified as stable yet, we are ignoring it.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Ok, but is this related to sitl or are you suggesting to change it for all other masters?

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -219,24 +250,34 @@ def download(self, vehicle: Vehicle, platform: Platform, version: str = "") -> O
...
         path = FirmwareDownload._download(item["url"])
-        if not path or not FirmwareDownload._validate_firmware(path):
-            logger.error("Unable to validate firmware file.")
-            return None
+        if firmware_format == FirmwareFormat.ELF:
+            # Make the binary executable
+            os.chmod(str(path), stat.S_IXOTH)
+
+        # TODO: Reincorporate firmware validation for SITL once Ardupilot fixes non-zero return on --help (PR #17573)
+        if platform != Platform.SITL:
```
patrickelectric, line 278: Why are you not validating SITL ?

patrickelectric (changes requested):

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: ?

patrickelectric (approved):
Looks good, waiting @Williangalvani to accept.
I believe that we should be validating SITL/Navigator since we are using both from master.
But I'll let @Williangalvani decide.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
patrickelectric: My bad, I just accepted and added my comment there.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
Williangalvani: Yeah. Let's use master for now. both this patch and proper navigator support are on master only.

core/services/ardupilot_manager/ArduPilotManager.py
Williangalvani: Yes, depending on the implementaton on the router side, it may just give up after the connection is lost and never reconnect.
We should move to UDP if possible or at least make sure that the routers reconnect on ardusub restart.
This may affect SITL and Navigator.

patrickelectric: Just to point it, I found this problem while working around #170

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Done.

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -241,21 +241,26 @@ def download(self, vehicle: Vehicle, platform: Platform, version: str = "") -> O
...
-            if not newest_version:
-                logger.error(f"No firmware versions found for this configuration ({vehicle} and {platform}).")
-                return None
-            version = f"STABLE-{newest_version}"
-
         firmware_format = FirmwareDownload._supported_firmware_formats[platform]
 
+        # Autodetect the latest supported version.
+        # For .apj firmwares (e.g. Pixhawk), we use the latest STABLE version while for the others (e.g. SITL and
+        # Navigator) we use latest DEV.
```
patrickelectric, line 248: You are just saying what the code does but not why (the important part).

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -274,10 +279,8 @@ def download(self, vehicle: Vehicle, platform: Platform, version: str = "") -> O
             # Make the binary executable
             os.chmod(str(path), stat.S_IXOTH)
 
-        # TODO: Reincorporate firmware validation for SITL once Ardupilot fixes non-zero return on --help (PR #17573)
-        if platform != Platform.SITL:
-            if path and not FirmwareDownload._validate_firmware(path):
-                logger.error("Unable to validate firmware file.")
-                return None
+        if path and not FirmwareDownload._validate_firmware(path):
```
patrickelectric, line 282: Can you apply this in the correct commit ?

patrickelectric (changes requested):

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: No prob.
Yes, that's the default behavior. It's usually under `/root/.config/ardupilot-manager/`

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
```diff
@@ -241,21 +241,27 @@ def download(self, vehicle: Vehicle, platform: Platform, version: str = "") -> O
             logger.error(f"Specified version not found for this configuration ({vehicle} and {platform}).")
             return None
 
-        # Autodetect the latest stable version
```
patrickelectric, line 204: Commit message is wrong:
`- SITL validation is back in this commit, since the latest DEV fixes the non-zero return on --help call.`
This does not exist here.

patrickelectric (changes requested):

core/services/ardupilot_manager/firmware_download/FirmwareDownload.py
rafaellehmkuhl: Fixed

patrickelectric (approved):

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Apparently there's no way to output SITL to UDP, according to [the ardupilot docs](https://ardupilot.org/dev/docs/using-sitl-for-ardupilot-testing.html#sitl-with-mavproxy-udp) and the `--help`:

image

patrickelectric: try uartA as udpin

rafaellehmkuhl:
```Starting sketch 'ArduSub'
Starting SITL input
Using Irlock at port : 9005
PANIC: Invalid device path: udpin
Running: sh dumpstack.sh 89024 >dumpstack_sitl.89024.out 2>&1
Failed```

Williangalvani: while that should technically work, I'm afraid that some things could be output only to the main port.

Williangalvani (approved):

PR #291, opened 2021-05-27
CI: use the same version of mavlink-router used in deployment

### LINKED ISSUE #187: Using different versions of MavlinkRouter on deploy and tests
The Github test action is builing MavlinkRouter from scratch and using a different commit than the one used on Bootstrap.

We should use the same.

Williangalvani:
fix #187

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,8 @@ jobs:
     -
       name: Install dependencies
       run: |
+        # This should be the same hash as in core/tools/mavlink_router/bootstrap.py#L10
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     -
       name: Install dependencies
       run: |
+        MAVLINK_ROUTER_COMMIT="94c4e3c6a9ff7c517b20b417e232caf52f12a6c6"
```
patrickelectric, line 24: Add a comment explaining where this magic hash is necessary and where we should look to find if is not in sync anymore.

patrickelectric (approved):

PR #296, opened 2021-05-28
added user to docker group

### LINKED ISSUE #292: Docker commands demanding root privilegies
Newly installed image:

```bash
pi@companion:~ $ docker ps
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/json: dial unix /var/run/docker.sock: connect: permission denied
pi@companion:~ $ sudo docker ps
CONTAINER ID   IMAGE                                     COMMAND                  CREATED          STATUS          PORTS     NAMES
13d341bd40f3   bluerobotics/companion-core:master        "/bin/sh -c '/usr/bi…"   14 minutes ago   Up 14 minutes             companion-core
c47870608bec   bluerobotics/companion-bootstrap:master   "/bin/sh -c /main.py"    17 minutes ago   Up 14 minutes             companion-bootstrap
pi@companion:~ $ 
```

I think we should [enable docker for non-root users](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user).

slayer321:
fix #292

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -50,7 +50,7 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
 )
 systemctl enable docker
 
-# creating a Unix group called docker and adding users to it
+# Add user to 'docker' group allowing docker commands without root privilegies
 echo "Adding default user to 'docker' group."
 usermod -aG docker $USER
 
--- a/install/install.sh
+++ b/install/install.sh
@@ -50,6 +50,11 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
 )
 systemctl enable docker
 
+#creating a Unix group called docker and adding users to it.
+echo "Created the docker group"
+usermod -aG docker $USER
+
+
 # Stop and remove all docker if NO_CLEAN is not defined
 test $NO_CLEAN || (
     # Check if there is any docker installed
--- a/install/install.sh
+++ b/install/install.sh
@@ -50,10 +50,9 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
 )
 systemctl enable docker
 
-#creating a Unix group called docker and adding users to it.
-echo "Created the docker group"
+# creating a Unix group called docker and adding users to it
+echo "Adding default user to 'docker' group."
 usermod -aG docker $USER
-
 
 # Stop and remove all docker if NO_CLEAN is not defined
 test $NO_CLEAN || (


install/install.sh
```diff
@@ -50,6 +50,11 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
 )
 systemctl enable docker
 
+#creating a Unix group called docker and adding users to it.
+echo "Created the docker group"
```
rafaellehmkuhl, line 54: "Creating the docker group." seems to replicate better the other messages writing style.

Also, this command is in fact adding the default user to the pre-existing docker group, so something like "Adding default user to 'docker' group." seems better.

install/install.sh
```diff
@@ -50,6 +50,11 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
...
+usermod -aG docker $USER
+
+
```
rafaellehmkuhl, line 57: Better to remove this extra line.

install/install.sh
```diff
@@ -50,6 +50,11 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
 )
 systemctl enable docker
 
+#creating a Unix group called docker and adding users to it.
```
rafaellehmkuhl, line 53: Please atent to the space after the `#` and also the message, as the following comment explains.

rafaellehmkuhl (changes requested):

install/install.sh
```diff
@@ -50,11 +50,10 @@ docker --version || curl -fsSL https://get.docker.com | env -u VERSION sh || (
 )
 systemctl enable docker
 
-#creating a Unix group called docker and adding users to it.
-echo "Created the docker group"
+# creating a Unix group called docker and adding users to it
```
rafaellehmkuhl, line 53: This comment also needs to change.
Something like "Add user to 'docker' group allowing docker commands without root privilegies"

rafaellehmkuhl (changes requested):

rafaellehmkuhl:
@patrickelectric as I understand the idea is to change from this to using the `pi` user on dockers, right?

ES-Alexander:
I just came across this PR while looking for something else.

> as I understand the idea is to change from this to using the `pi` user on dockers, right?

@rafaellehmkuhl that could cause issues given we should generally be moving away from `pi` user dependence, per #1073

slayer321:
Hey @rafaellehmkuhl , just wanted to know that is this PR still relevant or we can close it ?

joaoantoniocardoso:
Thanks for the contribution @slayer321, it ended up being superseded by #2979.

PR #301, opened 2021-06-03
Add DHCP Server to companion

### LINKED ISSUE #287: No DHCP Server on Cable Guy

rafaellehmkuhl:
Fix #287 

- Add dnsmasq as DHCP Server provider
- Use 192.168.2.2 as gateway
- Offer IPs in the 192.168.2.100-200 range
- Changes default  GCS Link to 192.168.2.2
- Add proper logging to cable-guy

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
+++ b/core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
@@ -1,0 +1,80 @@
+import pathlib
+import shutil
+import subprocess
+from typing import Any, Optional, Union
+
+from loguru import logger
+
+
+class Dnsmasq:
+    def __init__(self, config_path: pathlib.Path) -> None:
+        self._subprocess: Optional[Any] = None
+
+        binary_path = shutil.which(self.binary_name())
+        if binary_path is None:
+            logger.error("Dnsmasq binary not found on system's PATH.")
+            raise ValueError
+
+        self._binary = pathlib.Path(binary_path)
+        assert self.is_binary_working()
+
+        self._config_path = config_path
+        assert self.is_valid_config_file()
+
+    @staticmethod
+    def binary_name() -> str:
+        return "dnsmasq"
+
+    def binary(self) -> pathlib.Path:
+        return self._binary
+
+    def is_binary_working(self) -> bool:
+        if self.binary() is None:
+            return False
+
+        try:
+            subprocess.check_output([self.binary()])
+            return True
+        except subprocess.CalledProcessError as error:
+            logger.error(f"Invalid binary: {error}")
+            return False
+
+    def config_path(self) -> pathlib.Path:
+        return self._config_path
+
+    def is_valid_config_file(self) -> bool:
+        try:
+            subprocess.check_output([*self.command_list(), "--test"])
+            return True
+        except subprocess.CalledProcessError as error:
+            logger.error(f"Invalid configuration file: {error}")
+            return False
+
+    def command_list(self) -> list[Union[str, pathlib.Path]]:
+        return [self.binary(), "--no-daemon", f"--conf-file={self.config_path()}"]
+
+    def start(self) -> None:
+        try:
+            # pylint: disable=consider-using-with
+            self._subprocess = subprocess.Popen(self.command_list(), shell=False, encoding="utf-8", errors="ignore")
+            logger.info("DHCP Server started.")
+        except Exception as error:
+            logger.error(f"Unable to start DHCP Server: {error}")
+
+    def stop(self) -> None:
+        if self.is_running():
+            assert self._subprocess is not None
+            self._subprocess.kill()
+            logger.info("DHCP Server stopped.")
+        else:
+            logger.info("Tried to stop DHCP Server, but it was already not running.")
+
+    def restart(self) -> None:
+        self.stop()
+        self.start()
+
+    def is_running(self) -> bool:
+        return self._subprocess is not None and self._subprocess.poll() is None
+
+    def __del__(self) -> None:
+        self.stop()
--- a/core/services/cable_guy/api/manager.py
+++ b/core/services/cable_guy/api/manager.py
@@ -5,6 +5,7 @@ from socket import AddressFamily
 from typing import Any, List, Optional, Tuple
 
 import psutil
+from loguru import logger
 from pydantic import BaseModel
 from pyroute2 import IW, NDB, IPRoute
 from pyroute2.netlink.rtnl.ifaddrmsg import ifaddrmsg
@@ -49,25 +50,25 @@ class EthernetManager:
 
         # Load settings and do the initial configuration
         if not self.settings.load():
-            print(f"Failed to load previous settings. Using default configuration: {default_config}")
+            logger.error(f"Failed to load previous settings. Using default configuration: {default_config}")
             self.set_configuration(default_config)
             return
 
-        print("Previous settings loaded:")
+        logger.info("Previous settings loaded:")
         for item in self.settings.root["content"]:
-            print(f"Configuration with: {item}")
+            logger.info(f"Configuration with: {item}")
             if not self.set_configuration(EthernetInterface(**item)):
-                print("Failed.")
+                logger.error("Failed.")
 
     def save(self) -> None:
         """Save actual configuration"""
         try:
             self.get_interfaces()
         except Exception as exception:
-            print(f"Failed to fetch actual configuration, going to use the previous info: {exception}")
+            logger.error(f"Failed to fetch actual configuration, going to use the previous info: {exception}")
 
         if not self.result:
-            print("Configuration is empty, aborting.")
+            logger.error("Configuration is empty, aborting.")
             return
 
         result = [interface.dict(exclude={"info"}) for interface in self.result]
@@ -83,6 +84,7 @@ class EthernetManager:
             bool: Configuration was accepted
         """
         interfaces = self.get_interfaces()
+        logger.debug(f"Found following ethernet interfaces: {interfaces}.")
         valid_names = [interface.name for interface in interfaces]
 
         name = interface.name
@@ -90,15 +92,19 @@ class EthernetManager:
         mode = interface.configuration.mode
 
         if name not in valid_names:
+            logger.error(f"Invalid interface name ('{name}'). Valid names are: {valid_names}")
             return False
 
         if mode == InterfaceMode.Client:
             self.set_dynamic_ip(name)
+            logger.info(f"Interface '{name}' configured with dynamic IP.")
             return True
         if mode == InterfaceMode.Unmanaged:
             self.set_static_ip(name, ip)
+            logger.info(f"Interface '{name}' configured with static IP.")
             return True
 
+        logger.error(f"Could not configure interface '{name}'.")
         return False
 
     def _get_wifi_interfaces(self) -> List[str]:
@@ -126,9 +132,11 @@ class EthernetManager:
             bool: True if valid, False if not
         """
         blacklist = ["lo", "ham.*", "docker.*"]
-        blacklist += self._get_wifi_interfaces()
+        wifi_interfaces = self._get_wifi_interfaces()
+        blacklist += wifi_interfaces
 
         if not interface_name:
+            logger.error("Interface name cannot be blank or null.")
             return False
 
         for pattern in blacklist:
@@ -204,6 +212,7 @@ class EthernetManager:
         """
         interface_index = self._get_interface_index(interface_name)
         self.ipr.flush_addr(index=interface_index)
+        logger.info(f"Flushing IP addresses from interface {interface_name}.")
 
     def enable_interface(self, interface_name: str, enable: bool = True) -> None:
         """Enable interface
@@ -215,6 +224,7 @@ class EthernetManager:
         interface_index = self._get_interface_index(interface_name)
         interface_state = "up" if enable else "down"
         self.ipr.link("set", index=interface_index, state=interface_state)
+        logger.info(f"Setting interface {interface_name} to '{interface_state}' state.")
 
     async def _trigger_dhcp_service(self, interface_name: str) -> None:
         """Internal async trigger for dhcp service
@@ -243,6 +253,7 @@ class EthernetManager:
         """
         interface_index = self._get_interface_index(interface_name)
         self.ipr.addr("add", index=interface_index, address=ip, prefixlen=24)
+        logger.info(f"Setting interface {interface_name} to IP '{ip}'.")
 
     def set_dynamic_ip(self, interface_name: str) -> None:
         """Set interface to use dynamic ip address
@@ -254,6 +265,7 @@ class EthernetManager:
         self.flush_interface(interface_name)
         # Trigger DHCP service to add a new dynamic ip address
         self.trigger_dhcp_service(interface_name)
+        logger.info(f"Getting dynamic IP to interface {interface_name}.")
 
     def set_static_ip(self, interface_name: str, ip: str) -> None:
         """Set interface to use static ip address
--- a/core/services/cable_guy/api/manager.py
+++ b/core/services/cable_guy/api/manager.py
@@ -1,10 +1,12 @@
 import asyncio
+import pathlib
 import re
 from enum import Enum
 from socket import AddressFamily
 from typing import Any, List, Optional, Tuple
 
 import psutil
+from commonwealth.utils.DHCPServerManager import Dnsmasq
 from loguru import logger
 from pydantic import BaseModel
 from pyroute2 import IW, NDB, IPRoute
@@ -45,8 +47,12 @@ class EthernetManager:
 
     result: List[EthernetInterface] = []
 
-    def __init__(self, default_config: EthernetInterface) -> None:
+    def __init__(self, default_config: EthernetInterface, dhcp_gateway: str) -> None:
         self.settings = settings.Settings()
+
+        self._config_path = pathlib.Path(__file__).parent.absolute().joinpath("settings", "dnsmasq.conf")
+        self._server = Dnsmasq(self._config_path)
+        self._dhcp_server_gateway = dhcp_gateway
 
         # Load settings and do the initial configuration
         if not self.settings.load():
@@ -95,9 +101,18 @@ class EthernetManager:
             logger.error(f"Invalid interface name ('{name}'). Valid names are: {valid_names}")
             return False
 
+        if mode != InterfaceMode.Server and self._server.is_running():
+            self._server.stop()
+
         if mode == InterfaceMode.Client:
             self.set_dynamic_ip(name)
             logger.info(f"Interface '{name}' configured with dynamic IP.")
+            return True
+        if mode == InterfaceMode.Server:
+            self.set_static_ip(name, self._dhcp_server_gateway)
+            if not self._server.is_running():
+                self._server.start()
+            logger.info(f"Interface '{name}' configured as DHCP server with static IP.")
             return True
         if mode == InterfaceMode.Unmanaged:
             self.set_static_ip(name, ip)
@@ -297,9 +312,13 @@ class EthernetManager:
                 ip = address.address if valid_ip else "undefined"
 
                 is_static_ip = self.is_static_ip(ip)
+                is_gateway_ip = ip == self._dhcp_server_gateway
 
                 # Populate our output item
-                mode = InterfaceMode.Unmanaged if is_static_ip and valid_ip else InterfaceMode.Client
+                if self._server.is_running() and is_gateway_ip:
+                    mode = InterfaceMode.Server
+                else:
+                    mode = InterfaceMode.Unmanaged if is_static_ip and valid_ip else InterfaceMode.Client
                 info = self.get_interface_info(interface)
                 data = EthernetInterface(
                     name=interface, configuration=InterfaceConfiguration(ip=ip, mode=mode), info=info
--- a/core/services/cable_guy/main.py
+++ b/core/services/cable_guy/main.py
@@ -37,8 +37,9 @@ if args.default_config == "bluerov2":
     default_config = EthernetInterface(
         name="eth0", configuration=InterfaceConfiguration(ip="192.168.2.2", mode=InterfaceMode.Unmanaged)
     )
+    dhcp_gateway = "192.168.2.2"
 
-manager = EthernetManager(default_config)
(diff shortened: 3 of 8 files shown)


patrickelectric (changes requested):

install/network/dnsmasq.sh
```diff
@@ -0,0 +1,37 @@
...
+REMOTE="${REMOTE:-https://raw.githubusercontent.com/bluerobotics/companion-docker}"
+REMOTE="$REMOTE/$VERSION"
+CONFIG_NETWORK_PATH="$REMOTE/install/network"
+
+# Exit if something goes wrong
+set -e
+
+# Dnsmasq provides DHCP and DNS servers for small networks
+echo "Installing dnsmasq."
+apt install -y --no-install-recommends dnsmasq
```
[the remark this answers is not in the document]
patrickelectric: Can it run in the dockerfile ?

install/network/dnsmasq.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Dnsmasq needs to be stopped before finishing the configuration because by default
+# it shadows the system's default DNS service (e.g. systemd-resolved)
+echo "Stopping dnsmasq."
+systemctl stop dnsmasq
+
+echo "Configuring dnsmasq."
+DNSMASQ_CONF_PATH="/etc/"
+[ ! -d "${DNSMASQ_CONF_PATH}" ] && (
```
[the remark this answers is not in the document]
rafaellehmkuhl: Oh, yeah. This really passed by. I had a script before, but them I decided to replicate the one we are using with Avahi for maintaining purposes and totally didn't realized that. I will check for the config file.

install/network/dnsmasq.sh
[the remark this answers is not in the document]
rafaellehmkuhl: For the first part, you mean installing from somewhere (maybe even the debian repository) but avoiding the apt tool?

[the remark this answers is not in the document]
patrickelectric: we can't use debian repository, with that you'll need to install to install others tools to install .deb files under non-debian  distros.

I believe that the best solution, if possible, is to install it under docker.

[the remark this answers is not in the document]
rafaellehmkuhl: I really don't think it makes much sense to run a host os service under a docker container...

Besides the logical view, that would also add a good amount of work, for a problem that its not a priori, and that we currently have already (g++, python3-dev, python3-pip and python3-setuptools are all installed with apt).

I believe we should merge this and think about a final solution for all packages together, not for this one specifically.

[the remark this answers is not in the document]
patrickelectric: > apt

Inside docker, this script runs in the host computer, and you can't guarantee that apt will be available.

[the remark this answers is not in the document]
rafaellehmkuhl: Yes, but as with them, we can't guarantee that apt will be available either.

What does the fact that they run inside a container changes?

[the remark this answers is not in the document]
patrickelectric: > we can't guarantee that apt will be available either.

Yes we can, we use debian based docker images. 

The host computer can run debian, arch or any other distro that the user desires. We have control over the docker, so we can guarantee that ap exists.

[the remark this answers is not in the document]
Williangalvani: > Besides the logical view, that would also add a good amount of work, for a problem that its not a priori, and that we currently have already (g++, python3-dev, python3-pip and python3-setuptools are all installed with apt).

Those run inside the debian container. We know the container is debian, but we can't say the host os is, too.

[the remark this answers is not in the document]
rafaellehmkuhl: Oh, you're problem is with the os itself, not the apt tool. Got it.

rafaellehmkuhl:
Did some changes on the PR today so it becomes more consistent with the current architecture and allow a future hotspot PR to  be enabled without much hassle.

I consider it ready and tested. Image available under rafaellehmkuhl/companion-core:dhcp.

core/tools/dnsmasq/bootstrap.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+# Exit if something goes wrong
+set -e
+
+# Dnsmasq provides DHCP and DNS servers for small networks
+echo "Installing dnsmasq."
+apt install -y --no-install-recommends dnsmasq
+
+# Stops dnsmasq service. It will be run independently.
```
Williangalvani, line 10: this looks weird. is the install step starting the service? even then, this runs in build time, I expected something analog to `service disable", not stop.

can you make explain better in comments?

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -113,7 +113,7 @@ def start_mavlink_manager(self, device: Endpoint) -> None:
             self.add_new_endpoints(
                 {
                     Endpoint(
-                        "GCS Link", self.settings.app_name, EndpointType.UDPClient, "192.168.2.1", 14550, protected=True
+                        "GCS Link", self.settings.app_name, EndpointType.UDPClient, "192.168.2.2", 14550, protected=True
```
Williangalvani, line 116: this will break all current setups out there, right? we should talk about about it. Ideally we should be able to make both work. If we can't, this is the priority.

Williangalvani (changes requested):
this is looking pretty good. I got a few comments though.

Can you upload an image for testing?

rafaellehmkuhl:
> this is looking pretty good. I got a few comments though.
> 
> Can you upload an image for testing?

Image online on rafaellehmkuhl/companion-core:dhcp

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: Yes, altough current setups are all for testing :smile: 

I'm inclined to say it's a decision better to took now them later.

bootstrap/bootstrap/bootstrap.py
```diff
@@ -143,6 +143,7 @@ def start_core(self) -> bool:
         image = core["image"]
         core_version = core["tag"]
         binds = core["binds"]
+        ports = core["ports"]
```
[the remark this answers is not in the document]
rafaellehmkuhl: Fair. Will check.

Can we maintain this but remove the port bind?

core/tools/dnsmasq/bootstrap.sh
rafaellehmkuhl: I'm not much familiar with non-systemd managers, as this one, and it doesn't have 'disable'.

During tests I remember to find out that maybe the stop command also disabled the service, but I can be wrong. Will test without it.

core/services/cable_guy/main.py
```diff
@@ -9,11 +11,41 @@
...
+
+args = parser.parse_args()
+
+if not args.verbose:
+    logger.remove()
+    logger.add(sys.stderr, format="<green>{time:HH:mm:ss.SSS}</green> | <level>{message}</level>")
+
+
+class InterceptHandler(logging.Handler):
+    def emit(self, record) -> None:  # type: ignore
```
patrickelectric, line 31: `from logging import LogRecord`

patrickelectric (changes requested):

core/services/cable_guy/main.py
patrickelectric: Also, check: loguru_mypy

rafaellehmkuhl: This is basically the same discussion we had in the SITL PR. When introduced, LogRecord raises a lot of other typing problems, that I'm discussing with the Loguru author.

As its replicated code with ArduPilotManager, can we keep it like it is and solve both together on the common library?

core/services/cable_guy/main.py
```diff
@@ -9,11 +11,41 @@
...
 
 from api import manager
 from api.manager import EthernetInterface
 
+parser = argparse.ArgumentParser(description="Cable Guy service for Blue Robotics Companion")
+parser.add_argument("-v", "--verbose", help="increase output verbosity", action="store_true")
+
+args = parser.parse_args()
+
+if not args.verbose:
```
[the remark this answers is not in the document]
rafaellehmkuhl: The opposite, but like on SITL I will remove this option.

core/services/cable_guy/main.py
patrickelectric: I got it working here without problems only doing what I said

rafaellehmkuhl:
```
core/services/cable_guy/main.py:27: error: Incompatible types in assignment (expression has type "int", variable has type "str")
Found 1 error in 1 file (checked 1 source file)
```

You're sure you're not receiving this as soon as LogRecord is introduced?

patrickelectric: I just used: level: Union[int, str]

rafaellehmkuhl: Tracked on #313

rafaellehmkuhl:
Fix #313

core/tools/dnsmasq/bootstrap.sh
rafaellehmkuhl: Testes again and its not needed indeed. Removed.

bootstrap/bootstrap/bootstrap.py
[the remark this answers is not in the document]
rafaellehmkuhl: Maintained function but removed bind. Tested and working.

core/services/ardupilot_manager/ArduPilotManager.py
jaxxzer: let's keep the dhcp server running at 192.168.2.2, like companion 0

patrickelectric (changes requested):

bootstrap/test_bootstrap.py
```diff
@@ -30,6 +30,7 @@
                 "mode": "rw"
             }
         },
+        "ports": {},
```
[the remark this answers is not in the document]
rafaellehmkuhl: The consensus was to maintain it, but I can remove, no problem.

core/tools/dnsmasq/bootstrap.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "Installing dnsmasq."
+apt install -y --no-install-recommends dnsmasq
+
+echo "Configuring dnsmasq."
+CONFIG_ORIG_PATH="/home/pi/tools/dnsmasq/dnsmasq.conf"
+CONFIG_DEST_PATH="/etc/dnsmasq.conf"
+[ ! -f "${CONFIG_DEST_PATH}" ] && (
+    echo "Dnsmasq configuration file does not exist: ${CONFIG_DEST_PATH}"
+    exit 1
+)
```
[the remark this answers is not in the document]
rafaellehmkuhl: It's checks if the installation process indeed created the original file, preventing changes in this behavior to silent affect out application.

If you find it unnecessary, there's no problem in removing it.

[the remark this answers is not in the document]
patrickelectric: Can you just add a comment about it ? The logic behind the code is not clear without this explanation.

bootstrap/test_bootstrap.py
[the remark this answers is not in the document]
Williangalvani: we talked about it and I agree this should be introduced.
It IS out of scope for this pr, but it should get in eventually.
I have no strong feelings about keeping it here or not.

core/services/cable_guy/api/manager.py
```diff
@@ -98,6 +100,10 @@ def set_configuration(self, interface: EthernetInterface) -> bool:
             self.set_dynamic_ip(name)
             logger.info(f"Interface '{name}' configured with dynamic IP.")
             return True
+        if mode == InterfaceMode.Server:
+            self.set_static_ip(name, self.dhcp_server_gateway)
```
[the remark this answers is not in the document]
rafaellehmkuhl: The DHCP Server is always on, yes.

core/services/cable_guy/api/manager.py
```diff
@@ -296,9 +302,13 @@ def get_interfaces(self) -> List[EthernetInterface]:
                 ip = address.address if valid_ip else "undefined"
 
                 is_static_ip = self.is_static_ip(ip)
+                is_gateway_ip = ip == self.dhcp_server_gateway
 
                 # Populate our output item
-                mode = InterfaceMode.Unmanaged if is_static_ip and valid_ip else InterfaceMode.Client
+                if is_gateway_ip:
```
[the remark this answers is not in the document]
rafaellehmkuhl: Yes, that's the correct behavior.

core/services/ardupilot_manager/main.py
```diff
@@ -24,17 +26,19 @@
...
+    def emit(self, record: LogRecord) -> None:
         # Get corresponding Loguru level if it exists
+        level: Union[int, str]
         try:
             level = logger.level(record.levelname).name
         except ValueError:
             level = record.levelno
 
         # Find caller from where originated the logged message
+        frame: Optional[FrameType]
```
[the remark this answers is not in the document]
patrickelectric: Yes, where the code that you are fixing was provided.
As general rule, fixing commit that was not merged yet should be done in the same commit and not in a new one. 
Otherwise you are creating unnecessary commits.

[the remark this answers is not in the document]
rafaellehmkuhl: Your comment is on the ardupilot-manager service, not on cable-guy.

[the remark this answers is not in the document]
rafaellehmkuhl: They are two different code bases.

core/tools/dnsmasq/dnsmasq.conf
```diff
@@ -0,0 +1,38 @@
...
+interface=eth0
+
+# IPv4 lease range
+dhcp-range=192.168.2.100,192.168.2.200,255.255.255.0,24h
+
+# IPv4 gateway (this device)
+dhcp-option=option:router,192.168.2.2
+
+# IPv6 lease range
+dhcp-range=::1681:ff,::1681:ffff,constructor:eth0,ra-names,slaac,24h
```
patrickelectric, line 11: Mixed feelings about enabling ipv6, in the end it only results in problems, I think that would be more sane to just disable it.

rafaellehmkuhl: Can you test without it?

I'm not a DHCP expert, so I don't know if this is a regular behavior, but for my machine, not having a IPv6 provided when the machine asked for it would prevent also the IPv4 to be provided.

patrickelectric: I just removed and appears to be working.

rafaellehmkuhl: Will check if there's anything on my machine regarding this configuration and test with another one.

core/tools/dnsmasq/dnsmasq.conf
```diff
@@ -0,0 +1,38 @@
...
+
+# Adk client to poll for option changes every 6 hours
```
Williangalvani, line 13: ask?

core/tools/dnsmasq/dnsmasq.conf
rafaellehmkuhl: On tests, although I can see the message on failure to provide an IPv6, this time the IPv4 got leased, so I assume it was a coincidence what I have seem previously. 
Removing.

core/tools/dnsmasq/dnsmasq.conf
rafaellehmkuhl: Fixed.

Williangalvani:
I don't get it. IS the DHCP server always on? The user should be able to disable it.
I expect a few devices will behave differently in each case.

rafaellehmkuhl:
> I don't get it. IS the DHCP server always on? The user should be able to disable it.
> I expect a few devices will behave differently in each case.

Right now it's always on, yes.
Do you some specific problem with this behavior?

I was thinking about it and didn't see an use case on a first moment.

My concern on a first moment, that made me not enable/disable dnsmasq was that I would not want cable-guy for disabling the wifi DHCP server when we implemented it, but sometime after I realized we can have two DHCP server (dnsmasq instances) running separated, one for ethernet and another for wifi.

core/services/cable_guy/api/manager.py
```diff
@@ -125,9 +131,11 @@ def is_valid_interface_name(self, interface_name: str) -> bool:
             bool: True if valid, False if not
         """
         blacklist = ["lo", "ham.*", "docker.*"]
-        blacklist += self._get_wifi_interfaces()
+        wifi_interfaces = self._get_wifi_interfaces()
```
patrickelectric, line 140: Try to avoid such changes in a non related commit.

core/tools/dnsmasq/bootstrap.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "Installing dnsmasq."
+apt install -y --no-install-recommends dnsmasq
+
+echo "Configuring dnsmasq."
+CONFIG_ORIG_PATH="/home/pi/tools/dnsmasq/dnsmasq.conf"
+CONFIG_DEST_PATH="/etc/dnsmasq.conf"
+
+# Checks if the installation process indeed created the original file,
+# preventing changes in this behavior to silent affect our application.
+[ ! -f "${CONFIG_DEST_PATH}" ] && (
```
patrickelectric, line 16: Since the important thing here is the binary itself, we may need just to check the binary installation or if the binary itself exists.

core/tools/dnsmasq/bootstrap.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Dnsmasq configuration file does not exist: ${CONFIG_DEST_PATH}"
+    exit 1
+)
+
+cp $CONFIG_ORIG_PATH $CONFIG_DEST_PATH
```
patrickelectric, line 21: This is not necessary since we can provide the path for the configuration file with `--conf-file` as an argument.

core/tools/dnsmasq/bootstrap.sh
```diff
@@ -0,0 +1,25 @@
...
+
+echo "Dnsmasq configuration:"
+cat $CONFIG_DEST_PATH
```
patrickelectric, line 24: Why is this necessary ? Will just generate unnecessary noise.

patrickelectric (changes requested):
I'm not sure if I'm ok with it, having the DHCP server always on could result in problems under the network or in unreliable behavior when using unmanaged mode.

rafaellehmkuhl:
> I'm not sure if I'm ok with it, having the DHCP server always on could result in problems under the network or in unreliable behavior when using unmanaged mode.

Yes. I've discussed that with @Williangalvani and we decided to change the behavior. Forgot to comment here, my bad.

I'm just waiting for the approval of #330 right now, since the manager for `dnsmasq` will be there.

patrickelectric:
>I'm just waiting for the approval of #330 right now, since the manager for dnsmasq will be there.

what ? Why dnsmasq should be in commonwealth ?

rafaellehmkuhl:
> > I'm just waiting for the approval of #330 right now, since the manager for dnsmasq will be there.
> 
> what ? Why dnsmasq should be in commonwealth ?

Dnsmasq will also be used on a separate instance for the Wifi DHCP server.

rafaellehmkuhl:
> > > I'm just waiting for the approval of #330 right now, since the manager for dnsmasq will be there.
> > 
> > 
> > what ? Why dnsmasq should be in commonwealth ?
> 
> Dnsmasq will also be used on a separate instance for the Wifi DHCP server.

To better explain it:

We should have a wifi hotspot provided by the wifi service soon. This hotspot will also need a DHCP server. So we can keep both services completely separated, with one not interfeiring on the other, the idea is for each of them to control their own instance of dnsmasq (DHCP server provider), with its own configuration.

rafaellehmkuhl:
Finished and tested.

The image is available for tests under `rafaellehmkuhl/companion-core:dhcp`.

core/services/cable_guy/api/manager.py
```diff
@@ -48,6 +50,10 @@ class EthernetManager:
     def __init__(self, default_config: EthernetInterface) -> None:
         self.settings = settings.Settings()
 
+        self._config_path = pathlib.Path(__file__).parent.absolute().joinpath("settings", "dnsmasq.conf")
+        self._server = Dnsmasq(self._config_path)
+        self._dhcp_server_gateway = "192.168.2.2"
```
patrickelectric, line 55: I think that we should allow any IP address but have this one as default for bluerov and frontend.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -0,0 +1,75 @@
...
+    @staticmethod
+    def binary_name() -> str:
+        return "dnsmasq"
+
+    def binary(self) -> pathlib.Path:
+        return self._binary
+
+    def is_binary_working(self) -> bool:
+        if self.binary() is not None:
+            for line in subprocess.check_output([self.binary(), "--version"]).decode("utf-8").split("\n"):
```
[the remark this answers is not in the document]
rafaellehmkuhl: Agree. Will just test for the error-code.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @staticmethod
+    def binary_name() -> str:
+        return "dnsmasq"
+
+    def binary(self) -> pathlib.Path:
+        return self._binary
+
+    def is_binary_working(self) -> bool:
+        if self.binary() is not None:
```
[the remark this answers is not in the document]
rafaellehmkuhl: Yes.

core/services/cable_guy/api/manager.py
rafaellehmkuhl: I think that would be cool, but this will add some good work and test.

Can we open an enhancement issue to track that and postpone it? I would like to merge the DHCP server as it is, as it adds value already, and focus on the firmware management and frontend for now.

rafaellehmkuhl: Opened #336 to track custom configurations.

patrickelectric (approved):

Williangalvani (approved):

PR #311, opened 2021-06-07
core: ardupilot_manager: map serial ports to ardusub's ports

Williangalvani:
Tested with an ublox GPS.
Ardusub didn't like the gps on ports other than SERIAL1, so I made all of them serial1 individually to test.
This should allow us to use serial... stuff.

the mapping comes from correlating our the navigtor mapping:
image

With these docs:
https://ardupilot.org/dev/docs/sitl-serial-mapping.html

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -61,11 +61,26 @@ class ArduPilotManager(metaclass=Singleton):
         ## Can be changed back to a simple command after https://github.com/ArduPilot/ardupilot/issues/17572
         ## gets fixed.
         # pylint: disable=consider-using-with
+        #
+        # The mapping of serial ports works as in the following table:
+        #
+        # |    ArduSub   |       Navigator         |
+        # | -C = Serial1 | Serial1 => /dev/ttyS0   |
+        # | -B = Serial3 | Serial3 => /dev/ttyAMA1 |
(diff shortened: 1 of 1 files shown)


rafaellehmkuhl (changes requested):
Will, could you add a comment before the subprocess call to explain those extra params?

patrickelectric (approved):
LGTM, a comment explaining the serial number on the board with the maps would be great.

rafaellehmkuhl (approved):

jaxxzer (approved):

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -61,11 +61,26 @@ def start_navigator(self) -> None:
         ## Can be changed back to a simple command after https://github.com/ArduPilot/ardupilot/issues/17572
         ## gets fixed.
         # pylint: disable=consider-using-with
+        #
+        # The mapping of serial ports works as in the following table:
+        #
+        # |    ArduSub   |       Navigator         |
+        # | -C = Serial1 | Serial1 => /dev/ttyS0   |
```
jaxxzer, line 68: @willian, -C = SERIAL1_*?, use the parameter indices

Williangalvani: Yes, that is how I did it

Issue #75, opened 2020-10-29
Move to pyproject.toml over setup.py


patrickelectric:
pyproject is the default way to manage software development with python3
https://www.python.org/dev/peps/pep-0518/

patrickelectric:
pyproject does not have support for nested projects and also needs to specify a configuration to install projects in the system: `settings.virtualenvs.create false`
Ref: https://github.com/python-poetry/poetry/issues/2270

PR #320, opened 2021-06-09
pyproject: Add pytest integration

patrickelectric:
Helps #75

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ addopts = "--cov-report html --cov-report term"
 [tool.coverage.run]
 branch = true
 omit = [
+    "*/setup.py",
     "*/__init__.py",
     ]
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -18,6 +18,19 @@ mypy = "^0.900"
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
+[tool.pytest]
+addopts = "--cov-report html --cov-report term"
+
+[tool.coverage.run]
+branch = true
+omit = [
+    "*/setup.py",
```
Williangalvani, line 27: this is new, why is it?

Williangalvani (approved):

pyproject.toml
patrickelectric: It appears that setup.py is also used by coverage to calculate the coverage range \shrug

PR #323, opened 2021-06-09
bootstrap: Allow port binding through startup.json

rafaellehmkuhl:


patrickelectric:
Can you provide an explanation on the commit body why is this necessary since we bind as network host ?

rafaellehmkuhl:
> Can you provide an explanation on the commit body why is this necessary since we bind as network host ?

In fact it is not, but a comment on the other PR would have saved some useful time, I guess...

PR #330, opened 2021-06-10
Introducing `commonwealth`

rafaellehmkuhl:
The Commonwealth of Services, generally known simply as the Commonwealth, is a technological association of Companion services to share common knowledge.

### DIFF
--- a/core/services/install-services.sh
+++ b/core/services/install-services.sh
@@ -16,6 +16,9 @@ RUNTIME_PACKAGES=(
 apt install -y --no-install-recommends ${BUILD_PACKAGES[*]}
 apt install -y --no-install-recommends ${RUNTIME_PACKAGES[*]}
 
+# Commonwealth library:
+cd /home/pi/services/commonwealth/ && python3 setup.py install
+
 # Wifi service:
 ## Bind path for wpa
 mkdir -p /var/run/wpa_supplicant/
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -9,6 +9,10 @@ RUN /home/pi/tools/linux2rest/bootstrap.sh
 RUN /home/pi/tools/bridges/bootstrap.sh
 RUN /home/pi/tools/ttyd/bootstrap.sh
 
+# Install custom libraries
+COPY libs /home/pi/libs
+RUN /home/pi/libs/install-libs.sh
+
 # Set tmux configuration file
 COPY configuration/tmux.conf /etc/tmux.conf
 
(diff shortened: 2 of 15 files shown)


core/services/commonwealth/setup.py
```diff
@@ -0,0 +1,18 @@
...
+    name="commonwealth",
+    version="0.1.0",
+    description="Companion library to share common code.",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    license="MIT",
+    packages=setuptools.find_packages(),
+    python_requires=">=3.9",
+    install_requires=[
+        "starlette == 0.13.6",
```
patrickelectric, line 16: missing loguru

core/services/install-services.sh
```diff
@@ -36,6 +36,9 @@ cd /home/pi/services/versionchooser/ && python3 setup.py install
 # Ping service:
 cd /home/pi/services/ping/ && python3 setup.py install
 
+# Commonwealth library:
+cd /home/pi/services/commonwealth/ && python3 setup.py install
```
patrickelectric, line 40: why this is running after all services if it appears as base dependence for all of them ?

core/services/commonwealth/README.md
```diff
@@ -0,0 +1,3 @@
+# Commonwealth of Services
```
patrickelectric, line 1: What is the reason to have an empty .gitignore ?
We can add it when necessary.

patrickelectric (changes requested):
nice cleanup

core/services/commonwealth/README.md
rafaellehmkuhl: No specific reason besides o.c.d :smile:

core/services/install-services.sh
rafaellehmkuhl: Moving up.

core/services/commonwealth/setup.py
rafaellehmkuhl: I had put it there but apparently I removed on the last rebase. Fixing.

Williangalvani:
it looks like a library to me. I would place it in core/libs/commonwealth.

patrickelectric (changes requested):
wifi and ardupilot_manager are missing in setup.py the dependency for `commonwealth == 0.1`

rafaellehmkuhl:
> it looks like a library to me. I would place it in core/libs/commonwealth.

@patrickelectric can we settle with /core/libs/commonwealth so?

patrickelectric (approved):

PR #331, opened 2021-06-11
core: cable_guy: Use default config when no settings file is found

### LINKED ISSUE #286: cable-guy: add --default=bluerov2
This should, if there is no settings file for the cable guy yet,  set the ip address to unmanaged and 192.168.2.2.

This will ensure that the current companion works well with our current rov without extra configuration needed.

rafaellehmkuhl:
Fix #286 
Default config is static IP `192.168.2.2` on interface `eth0`.

### DIFF
--- a/core/services/cable_guy/api/manager.py
+++ b/core/services/cable_guy/api/manager.py
@@ -44,12 +44,13 @@ class EthernetManager:
 
     result: List[EthernetInterface] = []
 
-    def __init__(self) -> None:
+    def __init__(self, default_config: EthernetInterface) -> None:
         self.settings = settings.Settings()
 
         # Load settings and do the initial configuration
         if not self.settings.load():
-            print("Failed to load previous settings.")
+            print(f"Failed to load previous settings. Using default configuration: {default_config}")
+            self.set_configuration(default_config)
             return
 
         print("Previous settings loaded:")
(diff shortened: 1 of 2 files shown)


Williangalvani (approved):

core/services/cable_guy/api/manager.py
```diff
@@ -34,6 +34,11 @@ class EthernetInterface(BaseModel):
     info: Optional[InterfaceInfo]
 
 
+default_config = EthernetInterface(
+    name="eth0", configuration=InterfaceConfiguration(ip="192.168.2.2", mode=InterfaceMode.Unmanaged)
```
patrickelectric, line 38: Since this will break when running locally, could this at least be configured as an argument ?

patrickelectric (changes requested):

core/services/cable_guy/api/manager.py
rafaellehmkuhl: You mean when running directly on your machine, for example? 
It doesn't break because it uses "enpxsx" instead of "ethx". Ubuntu also moved back in 2016.

But even in this case, using it with an argument seems more reasonable. Adding.

patrickelectric (approved):

rafaellehmkuhl:
Will test again after the rebase/conflict-solve just to make sure.

rafaellehmkuhl:
Tested and it's working. Ready to merge.

patrickelectric (approved):

PR #341, opened 2021-06-17
Adding Nginx

Williangalvani:
Image available at williangalvani/companion-core:nginx

services will be accessible both at the regular place and at /\<service-name\>/.
the trailing / is required for now (but that is fixeable via nginx).

some paths of static files had to be changed to relative paths so they work in a different url path.

it still needs cleaning, reorganization, and testing.  Let me know what you think.

(I will cleanup the .conf file, please don't worry about whitespaces or whatever there)

### DIFF
--- a/core/tools/nginx/nginx.conf
+++ b/core/tools/nginx/nginx.conf
@@ -1,0 +1,93 @@
+user www-data;
+worker_processes auto;
+pid /run/nginx.pid;
+
+events {
+        worker_connections 768;
+        # multi_accept on;
+}
+
+http {
+    client_max_body_size 10M;
+    ##
+    # Basic Settings
+    ##
+
+    sendfile on;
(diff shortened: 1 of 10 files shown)


rafaellehmkuhl (approved):

core/tools/nginx/nginx.conf
```diff
@@ -0,0 +1,96 @@
...
+    ##
+    # Basic Settings
+    ##
+
+    sendfile on;
+    tcp_nopush on;
+    tcp_nodelay on;
+    keepalive_timeout 65;
+    types_hash_max_size 2048;
+    # server_tokens off;
```
patrickelectric, line 21: We could just remove this lost comments here and we are ready to go.

patrickelectric (approved):

PR #342, opened 2021-06-18
Add main frontend

### LINKED ISSUE #207: Core: Add general web application
May 27: 
Since I'll be working half time/period, I'll explain what I did and what was the problems that I found while doing a introspection about the technologies and libraries for our webpage idea. 
- Typescript is a must thing and it improves the software architecture and development.
- We are aiming to have an API in our frontend that we'll be used to abstract mostly of the services and provide an easy way for us to manage companion and also the user experience without the pain to maintain it as the original companion. 
- Integrate services and also 3rd party web pages as a plugin-like system for companion, the object behind this:
    - The same powerful and easy abstraction in the frontend can be used for the plugin system (services and 3rd party web pages)
    - Provides easy and generic integration 
    - Allows the same look and feel (since css and components will be used from the main webpage)
    - Better integration compared to others alternatives besides vue components.
    - But there are problems:
        - I was unable to make it work with typescript, but it was a great experience using javascript: [vue-sfc-loader](https://github.com/FranckFreiburger/vue3-sfc-loader)

About UI libraries, right now we are using Metro UI, that have a great ecosystem and component library but is there is no vue support, resulting in a hard time for developers to maintain and create a great UI for our clients.
An alternative for it was searched and Quasar was the best alternative for now, it supports both Vue 3 (the next release) and full typescript support. The only problem with Quasar is that `quasar dev` takes long times to rebuild the page in development mode and has a poor integration with developmental pages that are already open when a restart is necessary in the command, since it fails to catch some typescript components. 

References:
https://vuejs.org/v2/guide/installation.html#Explanation-of-Different-Builds
https://forum.vuejs.org/t/what-is-the-compiler-included-build/13239

rafaellehmkuhl:
Main frontend for companion-docker

Setup is:
- Vue 2
- SASS (with dart-sass)
- Typescript
- PWA support
- Router (with history mode)
- Vuex
- Vuetify

Adds test job for installing and building the frontend project.

This fixes #207, although there are import considerations there we should open new issue for in the future (regarding dynamic plugins).

Image available for test under `rafaellehmkuhl/companion-core:frontend`


core/services/bff/tsconfig.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "strict": false,
```
patrickelectric, line 5: We should use strict mode as a good practice.

core/services/wifi/main.py
```diff
@@ -9,6 +9,7 @@
 import uvicorn
 from commonwealth.utils.apis import PrettyJSONResponse
 from fastapi import FastAPI, HTTPException, status
+from fastapi.middleware.cors import CORSMiddleware
```
patrickelectric, line 12: this is not necessary, use nginx.

core/services/bff/src/App.vue
```diff
@@ -0,0 +1,91 @@
...
+      >
+        <v-list-item-group
+          active-class="deep-blue--text text--accent-4"
+        >
+          <v-card elevation="0" class="d-flex justify-center">
+            <v-img
+              alt="Blue Robotics Logo"
+              class="shrink mr-2"
+              contain
+              src="https://bluerobotics.com/wp-content/uploads/2019/03/BlueRobotics-Logo-1024x1024.png"
```
patrickelectric, line 41: We can't rely on external dependencies.

patrickelectric (changes requested):

core/services/bff/.gitignore
```diff
@@ -0,0 +1,23 @@
+.DS_Store
```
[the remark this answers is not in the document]
rafaellehmkuhl: Didn't get it. Why would we have a dockerignore file on the service?

core/services/bff/README.md
```diff
@@ -0,0 +1,3 @@
+# BFF
```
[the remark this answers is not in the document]
rafaellehmkuhl: Oh, we don't have one yet. That's just a placeholder.

Any suggestion? Was thinking about "Companion Ground Station".

core/services/bff/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "bff",
```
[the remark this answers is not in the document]
rafaellehmkuhl: Answered before.

core/services/bff/.editorconfig
```diff
@@ -0,0 +1,7 @@
+[*.{js,jsx,ts,tsx,vue}]
```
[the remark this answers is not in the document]
rafaellehmkuhl: Yes. This was commited by mistake.

core/services/bff/.gitignore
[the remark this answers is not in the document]
patrickelectric: docker will track and hash everything that is not in a dockerignore, any modification in any of this files that are in gitignore will invalidate the cache and also be copied and create an unnecessary amount of space.

core/services/bff/src/registerServiceWorker.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-console */
+
+import { register } from 'register-service-worker';
```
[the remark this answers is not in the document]
rafaellehmkuhl: Will take a look for an alternative way.
If you have any suggestion on how to do it, let me know.

core/services/install-services.sh
```diff
@@ -36,6 +36,9 @@ cd /home/pi/services/versionchooser/ && python3 setup.py install
 # Ping service:
 cd /home/pi/services/ping/ && python3 setup.py install
 
+# BFF service:
+cd /home/pi/services/bff/ && yarn install && yarn build
```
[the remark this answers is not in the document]
rafaellehmkuhl: Agree. That's there just for the tests to work without external changes.

core/services/bff/tsconfig.json
rafaellehmkuhl: That's why there's a big [TEMP] on this commit :sweat_smile: 

I will remove it when finishing the Wifi service componentization.

core/services/wifi/main.py
rafaellehmkuhl: As the [TEMP] states, that's there just for the tests, as nginx was not yet merged. Will remove after rebasing.

core/services/bff/src/App.vue
rafaellehmkuhl: Forgot to change this one. Good catch.

core/services/bff/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow:
```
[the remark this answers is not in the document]
rafaellehmkuhl: Agree

core/services/bff/src/components/BigMap.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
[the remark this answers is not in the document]
rafaellehmkuhl: Removing the .png.

core/services/bff/.gitignore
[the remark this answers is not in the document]
rafaellehmkuhl: Fair.
We currently don't have a dockerignore for each service. Should we keep using the one that's already in core or make a specific one?

core/services/bff/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "eslint-plugin-import": "^2.20.2",
+    "eslint-plugin-vue": "^6.2.2",
+    "sass": "~1.32.0",
+    "sass-loader": "^10.0.0",
+    "typescript": "~4.1.5",
+    "vue-cli-plugin-vuetify": "~2.4.1",
+    "vue-template-compiler": "^2.6.11",
+    "vuetify-loader": "^1.7.0"
+  }
+}
```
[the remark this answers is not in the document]
rafaellehmkuhl: The mobile icons and favicon are somewhat different from the original .svg. They have custom background and margins.

I've removed most of them and just kept the favicon, Android and iOS ones.

core/services/bff/.gitignore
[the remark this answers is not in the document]
rafaellehmkuhl: Since those were all dev specific and our last discussion headed in the way of not including dev specific ignores, I'm removing the .gitignore.

core/frontend/README.md
```diff
@@ -0,0 +1,3 @@
+# CompanionOS
```
patrickelectric, line 1: What

patrickelectric (changes requested):

core/frontend/README.md
rafaellehmkuhl: ?

Williangalvani, line 1: Companion OS is what we were considering calling the project as a whole, not the frontend. I think the frontend could be just "frontend". And do we even need a readme here?

rafaellehmkuhl: Got it wrong on the meeting.
Let's stay with frontend.

core/frontend/src/main.ts
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue';
+import App from './App.vue';
+import router from './router';
+import store from './store';
+import vuetify from './plugins/vuetify';
+
+Vue.config.productionTip = false;
```
Williangalvani, line 7: why is this? I don't see any harm in having a tip in the console when running dev mode.

rafaellehmkuhl: Agree.

core/frontend/tsconfig.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "strict": false,
```
patrickelectric, line 5: This should be true, as I said in my previous comment, we should use the strict mode to avoid unnecessary pain and problems in the future as developers.

core/frontend/src/main.ts
```diff
@@ -0,0 +1,12 @@
...
+import router from './router';
+import store from './store';
+import vuetify from './plugins/vuetify';
+
+new Vue({
+  router,
+  store,
+  vuetify,
+  render: (h) => h(App),
+}).$mount('#app');
```
patrickelectric, line 12: You are missing the service worker registration for PWA:
```
if("serviceWorker" in navigator) {
    navigator.serviceWorker
        .register("service-worker.js")
        .then(function() { console.log("Service Worker Registered.") })
}
```

`service-worker.js` should be generated by the PWA system that you are using here (check your package.json and vue.config.js)

patrickelectric (changes requested):
The are some things that are conceptually wrong.
Files that are a size transformation should be created/generated by the build system, avoid unnecessary tracking and management of unnecessary files. 
The patch is also missing a gitignore and dockerignore files, the first one is to ignore and help us as developers to avoid dealing with unnecessary noise under git diff, the second is important to avoid problems with docker cache system, check my comment for further explanation.
There is also PWA packages and some configuration files, but it's missing the configuration for service-worker.

I believe that is a good patch, but to get it merged we need to have at least a bare minimum correct structure and dependencies to avoid any pain for us in further development, once we hit the sweet spot of a good basic configuration for the typescript frontend we should be good to merge.

core/frontend/src/main.ts
rafaellehmkuhl: It's not needed. Tested here and it's working without it. It's generated by vue.

rafaellehmkuhl:
> The are some things that are conceptually wrong.
> Files that are a size transformation should be created/generated by the build system, avoid unnecessary tracking and management of unnecessary files.

Please, check my comment on your first review. Those files are not just size transformation. There's background and margins added in the process. That's the only solution, unless we want to learn how to use and add a cli-image-editor-like tool on our build because of the tracking of 3 images that are all together. 

> The patch is also missing a gitignore and dockerignore files, the first one is to ignore and help us as developers to avoid dealing with unnecessary noise under git diff, the second is important to avoid problems with docker cache system, check my comment for further explanation.

You've previously suggested that we should not have gitignore for node_modules folder. Have you changed your mind on that?
About the dockerignore, I will add one.

> There is also PWA packages and some configuration files, but it's missing the configuration for service-worker.

It's not, like I've answered on the first review. Vue automatically adds a default service worker during build time. We only need to specify one if we want custom behavior.

> It's also failing in the CI.

Yes, this is failing since we removed the dedicated eslintrc file, like you suggested (which I found is reasonable), but it's needed on the frontend folder during build time. Do you have any suggestion on that?

patrickelectric:
> suggested that we should not have gitignore for node_modules 

Please link the commend, I said that dockerignore is missing and not that gitignore should not exist.

> dedicated eslintrc file, like you suggested 

As said, the file should not exist if you merge it with the project configuration. Check MGroundStation example

rafaellehmkuhl:
> Please link the commend, I said that dockerignore is missing and not that gitignore should not exist.

That was outside github, but ok

> As said, the file should not exist if you merge it with the project configuration. Check MGroundStation example

You're not understanding. That's not the case with MGS. That's a build problem from the fact that the dockerfile is on a sub-folder with relation to the eslintrc file, so there's no way to copy.

I've found a possible solution disabling the eslint on build, though.

core/frontend/.browserslistrc
```diff
@@ -0,0 +1,3 @@
+> 1%
```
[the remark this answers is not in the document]
rafaellehmkuhl: Done.

`tsconfig.json` is the only one which [does not currently support](https://github.com/microsoft/TypeScript/issues/32830) to be placed there.

core/Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM node:current-buster-slim AS frontendBuilder
+
+# Build frontend
+COPY frontend /home/pi/frontend
+RUN cd /home/pi/frontend && yarn install && yarn build
```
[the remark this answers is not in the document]
rafaellehmkuhl: >This is wrong, as you can see in the line 33 of the current file, the multistage will take care of the frontend build process.

Could you elaborate on that? Line 33 is merely copying the dist folder *after* its build (on line 5).

> The copy of the frontend is also not correct, since this project is missing the dockerignore, otherwise we are going break the docker cache system calculation that is based in file fingerprint.

You've already stated the missing dockerignore...

rafaellehmkuhl:
Building problems are fixed. Strangly enough @Williangalvani found out that `yarn build`, which is an alias to `vue-cli-service build` works, but `vue-cli-service build` itself does not work. Will assume it's javascript magic/madness.

There are lint errors now, because of the different eslint rules I was using. Fixing those.

core/services/bff/.gitignore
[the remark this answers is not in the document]
patrickelectric: > We currently don't have a dockerignore for each service. Should we keep using the one that's already in core or make a specific one?

We have a global dockerignore for all services

rafaellehmkuhl:
Tested and working.
Local development and test also works fine.

Important thing to notice is that this PR moves the frontend to the root path ("/"), so the helper service is only available through port 81. I've tried creating the nginx redirect on `/helper`, but the HTML seems to have something preventing the fetchs to work correctly. I will work with that on parallel. If I archieve success on it I rebase this PR before merging.

core/frontend/.gitignore
```diff
@@ -0,0 +1,3 @@
+dist/
+node_modules/
+.eslintrc.js
```
Williangalvani, line 3: is this required? why?

Williangalvani (commented):
Looking good so far.
27289a5 commit message is either missing "dockerfile" or has an extra colon

rafaellehmkuhl:
The problem with the helper service is because it's fetching the backend through the `latest` path on the frontend, instead of the `vX.Y` path, like the other services.

image

For some strange reason, when this is done, the redirect URL is removing the `/helper` on the beggining of the URL, as can be seen in the image above.

As all other services are using `v1.0` URLs for fetching, I moved this fetch on helper to `v1.0` also and opened issue # to track this problem (inability to use `/latest`paths on the frontends).

It was working before by a coincidence. It was the one that should not work, but the fact that it was on the root path (`/`) masked the problem.

.eslintrc.js
```diff
@@ -4,7 +4,11 @@ module.exports = {
     browser: true,
     es2021: true,
   },
-  extends: ['eslint:recommended', 'plugin:vue/vue3-strongly-recommended'],
+  extends: [
+    'eslint:recommended',
+    'plugin:vue/vue3-strongly-recommended',
```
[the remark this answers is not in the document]
rafaellehmkuhl: No, we should be using `plugin:vue/strongly-recommended`. I didn't change that just because it's not the scope of this PR. If you disagree, let me know and I add it.

It is working, indeed. `plugin:vue/vue3-strongly-recommended` seems to cover the two versions, although [the documentation indicates it's aimed only for vue3](https://eslint.vuejs.org/rules/#priority-b-strongly-recommended-improving-readability-for-vue-js-3-x).

core/frontend/.gitignore
rafaellehmkuhl: You mean the three of them or some specific one?

`dist/` for sure is not if one is only going to use `yarn serve` for development (thus not building it). `node_modules/` and `.eslintrc.js` I'm not really sure since the discovery on how eslnt find's it's packages.

I will take a look to see if any of those are not required.

Williangalvani: There's currently no .eslintrc.js file in core/frontend, is there?
Unless something is creating it, I don't think it is necessary to ignore it.

rafaellehmkuhl: Ok, just tested. 
The `.eslintrc.js` file does not need to be copied to that folder. `eslint-loader` is able to find the root one.
The `dist/` folder is there to allow the developer to try building the frontend locally (instead of serving it dinamically). Do you think it should be removed?
The `node_modules/` folder is indeed required, as one needs to `yarn install` on the frontend folder before serving it.

Current workflow is:

1. go to the root folder and install the packages need for testing, with the same command available on `.github/workflows/tests.yml` (as previously needed already)
2. go to the frontend folder and `yarn install`
3. `yarn serve` to begin testing/developing

rafaellehmkuhl: > There's currently no .eslintrc.js file in core/frontend, is there?
> Unless something is creating it, I don't think it is necessary to ignore it.

There's not. I just though before that for developing we would need to copy the eslint file to the frontend folder, but as the comment before states, it's not needed.

.github/workflows/tests.yml
```diff
@@ -24,7 +24,10 @@ jobs:
         # This should be the same hash as in core/tools/mavlink_router/bootstrap.py#L10
         MAVLINK_ROUTER_COMMIT="94c4e3c6a9ff7c517b20b417e232caf52f12a6c6"
         sudo apt install shellcheck
-        npm install eslint eslint-plugin-vue eslint-plugin-html --save-dev
+        npm install \
+          eslint eslint-plugin-vue eslint-plugin-html \
+          @vue/eslint-config-typescript @typescript-eslint/parser @typescript-eslint/eslint-plugin typescript \
+          --save-dev
```
patrickelectric, line 30: @Williangalvani I believe that we should avoid adding frontend tests in this part of the test CI, maybe creating a second job in this file? 
We could also avoid the usage of npm in the second step and use the github docker action step with the yarn docker directly.

.eslintrc.js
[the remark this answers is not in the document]
rafaellehmkuhl: Yeah. Maybe the style-guide is the same for both versions and the vue3 one just adds the new typescript styles.

.github/workflows/tests.yml
rafaellehmkuhl: I agree on that, but as its the current approach, I just followed it for this PR, and we can open a following one to attack that. What do you think?

core/Dockerfile
```diff
@@ -23,5 +29,8 @@ COPY services /home/pi/services
 COPY start-companion-core /usr/bin/start-companion-core
 RUN /home/pi/services/install-services.sh
 
+# Copy frontend built on frontendBuild to this stage
```
patrickelectric, line 32: why `frontendBuild` is in camel case ?

rafaellehmkuhl: No specif reason. Any suggestion?

rafaellehmkuhl: @Williangalvani do you know if there's any limitation on the stage's name or a name convention?

patrickelectric: 'Copy multistage frontend output to docker'

rafaellehmkuhl: Ok, but any suggestion on the name of the stage?

Williangalvani: that is probably on me. The multi-stage examples only have 1-word names, and they are all lower-case. `frontendbuild` gets hard to read, so I made it camelCase.

patrickelectric: Ha, I just saw that this is the name of the multistage

.github/workflows/tests.yml
Williangalvani: let's talk about it. there's a lot to consider here...

rafaellehmkuhl: From meeting: decision is to stay as it is and do a PR focusing on refactoring our tests, so they become more atomic.

Williangalvani (approved):

.github/workflows/tests.yml
```diff
@@ -54,4 +54,14 @@ jobs:
...
\ No newline at end of file
+        path: htmlcov
+  frontend-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: borales/actions-yarn@v2.3.0
+        with:
+          cmd: --cwd ./core/frontend install
+      - uses: borales/actions-yarn@v2.3.0
```
patrickelectric, line 65: missing linter step, as we discussed before, we should run linter to make sure that is really running and we do not break the implicit behaviour of linter on build.

rafaellehmkuhl: Done.

patrickelectric (approved):
LGTM :shipit:

PR #346, opened 2021-06-21
Add new mavlink-router with tlog

patrickelectric:


core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -57,7 +57,7 @@ def convert_endpoint(endpoint: Endpoint) -> str:
         ]:
             raise ValueError(f"Master endpoint of type {master.connection_type} not supported on MavlinkRouter.")
 
-        log = f"--log {self.logdir().resolve()}"
+        log = f"--log {self.logdir().resolve()} --tlog {self.logdir().resolve()}"
```
rafaellehmkuhl, line 60: this should be just `log = f"--tlog {self.logdir().resolve()}"` right?

rafaellehmkuhl (changes requested):

rafaellehmkuhl:
From the CI, it appears the 'tlog' tag does not exist on your repo:

`error: pathspec 'tlog' did not match any file(s) known to git`

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
patrickelectric: I was aiming to have both logs, but Ok.

Williangalvani:
this should be kept in sync with https://github.com/bluerobotics/companion-docker/blob/master/.github/workflows/tests.yml#L25
(feel free to add a comment saying so, while you are at it)

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
rafaellehmkuhl: I'm not against it. We just need to keep this in sync with `mavproxy`.

.github/workflows/tests.yml
```diff
@@ -21,28 +21,15 @@ jobs:
...
-        cd ${MAVLINK_ROUTER_PATH}
-        git fetch origin ${MAVLINK_ROUTER_COMMIT}
-        git checkout ${MAVLINK_ROUTER_COMMIT}
-        git submodule update --init --recursive --quiet
-        ./autogen.sh
-        ./configure --disable-systemd CFLAGS='-g -O2' --sysconfdir=/etc --localstatedir=/var --libdir=/usr/lib64 --prefix=/usr
-        sudo make install
-        rm -rf {MAVLINK_ROUTER_PATH}
+        # Install mavlink-router
+        sudo ./core/tools/mavlink_router/bootstrap.py
```
rafaellehmkuhl, line 32: I'm remember we've tried to use the same bootstrap python script to install mavlinkrouter on the tests but it didn't work (needed some change on the script).

patrickelectric: Done.

core/tools/mavlink_router/bootstrap.py
```diff
@@ -35,6 +35,10 @@ def run_apt_install(package: str) -> None:
 
 
 def run_apt_uninstall(package: str) -> None:
+    if "NOUNINSTALL" in os.environ:
+        print("Skipping uninstall step.")
+        return
+
```
rafaellehmkuhl, line 41: I think it would make more sense for this skip statement to be on the call to the function, on line 79. It's more explicit there. Here we call uninstall and it does not uninstall, which is somewhat weird. 

What do you think?

rafaellehmkuhl (approved):
Seems good to go. I would change that statement, but its up to you.

Williangalvani (approved):
lgtm, but I would do @rafaellehmkuhl 's suggested change before merging.

core/tools/mavlink_router/bootstrap.py
patrickelectric: This is a global rule for the uninstall step, if the function is called more than once the same if would not to exist.

rafaellehmkuhl: Didn't get it. Why would someone need to call uninstall twice?

patrickelectric: To encapsulate point specific uninstalls for better organization, like: build time dependencies and runtime dependencies, as we have for GST in companion-core.

rafaellehmkuhl: Yes. I'm not suggesting to remove it, just to move it to the call of the function, instead of the function itself, like that:

```python
if "NOUNINSTALL" not in os.environ:
    run_apt_uninstall("autoconf g++ git libtool make pkg-config python3-future")
else:
    print("Skipping uninstall step.")
```

patrickelectric: Sorry if I was not clear, but as I said in my previous comments, multiple calls for the uninstall function may happen in the future and the idea behind NOUNINSTALL environment variable is to be a global behavior change and not scope specific, that's why it's inside the function and not outside as a scope specific behavior control.
If that's not clear yet, imagine the following scenario, that as I said previous, is based under the docker-base with GST installation.
```
BUILD_LIBS = [A, B, C]
RUNTIME_LIBS = [D, E, F]

// code
install(BUILD_LIBS)
//code for build
uninstall(BUILD_LIBS)

// more code
install(RUNTIME_LIBS)
//code for runtime
uninstall(RUNTIME_LIBS)
```

With the current patch this code does not change since we are controlling the *global* behavior and not scope specific behavior. Otherwise we'll need a if check for each uninstall.

rafaellehmkuhl: Totally get it. Just saying it sounds weird to me calling a function named "uninstall_something" and this something is not uninstalled.

But hey, as I said, its up to you. That's totally a minor.

PR #348, opened 2021-06-22
Add apt update before calls of apt install

patrickelectric:
This fix problems related to not updated mirrors


Williangalvani:
I worry a little about the image size. an "apt update" on a vannila python 3.9-buster-slim image creates a layer with 18mb.
We can worry about reducing image size later, though. This pr improves reliability.

Williangalvani (approved):

PR #347, opened 2021-06-22
core: mavlink-camera-manager: Update version with relative URL calls

patrickelectric:
Fix problem relative to URL redirect with HTTP server

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

PR #349, opened 2021-06-22
core: mavlink2rest: Update version with relative URL calls

patrickelectric:


Williangalvani:
I found an issue:
image

I believe it can be fixed by replacing `URL('delete_stream')` for `URL('delete_stream', window.location)` at https://github.com/patrickelectric/mavlink-camera-manager/blob/master/src/html/index.html#L115

patrickelectric:
@Williangalvani I think that your comment is about #347 right ?

Williangalvani (approved):
tested with some dvl stuff and it worked well on both paths.

PR #350, opened 2021-06-22
core: frontend: libs: Add MAVLink2Rest library

patrickelectric:
To test: patrickelectric/companion-core:mavlink2rest_ts
(you'll see MAV_TYPE_SUBMARINE at 10Hz)

Check console output, you'll see the attitude values. 

The code is the following:
```js
mavlink2rest.startListening('HEARTBEAT').setCallback((message) => {
  console.log(message.mavtype.type)
}).setFrequency(10)
```

### DIFF
--- a/core/frontend/src/libs/MAVLink2Rest.ts
+++ b/core/frontend/src/libs/MAVLink2Rest.ts
@@ -1,0 +1,186 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+// The library is an interface for MAVLink objects, messages can by of any type
+
+interface Dictionary<T> {
+  [key: string]: T;
+}
+
+class Listener {
+  callback: (msg: any) => void = () => { console.log('Listener not assigned a callback') };
+  parent: Endpoint;
+  frequency = 1
+  interval = -1;
+
+  constructor (parent: Endpoint) {
+    this.parent = parent
+    this.setFrequency(1)
+  }
+
+  /**
+   * Define callback to be used when a new message is available
+   * @param  {(msg:any)=>void} callback
+   * @returns Listener
+   */
+  setCallback (callback: (msg: any) => void): Listener {
+    this.callback = callback
+    return this
+  }
+
+  /**
+   * Set desired frequency for the callback
+   * @param  {number} frequency
+   * @returns Listener
+   */
+  setFrequency (frequency: number): Listener {
+    clearInterval(this.interval)
+    this.interval = window.setInterval(() => {
+      if (this.parent.latestData !== null) {
+        this.callback(this.parent.latestData)
+      }
+    }, 1000 / frequency)
+    return this
+  }
+
+  discard () {
+    clearInterval(this.interval)
+    this.parent.removeListener(this)
+  }
+}
+
+class Endpoint {
+  socket: WebSocket;
+  listeners: Array<Listener> = [];
+  latestData: any = null;
+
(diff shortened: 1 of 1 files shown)


core/frontend/src/libs/MAVLink2Rest.ts
```diff
@@ -0,0 +1,186 @@
...
+  baseUrl: string;
+  // Dictionary mapping endpoints to websocket
+  endpoints: Dictionary<Endpoint> = {}
+  baseUrlCandidates: Array<string>
+
+  private static instance: Mavlink2RestManager;
+
+  private constructor () {
+    this.baseUrl = `${this.getWebsocketPrefix()}://${window.location.host}/mavlink2rest/ws/mavlink`
+    this.baseUrlCandidates = [this.baseUrl, 'ws://localhost:8088/ws/mavlink']
```
Williangalvani, line 109: what is up with the second candidate?
IIRC in the original code it was for local development, but I'm not sure if that still applies.

core/frontend/src/libs/MAVLink2Rest.ts
```diff
@@ -0,0 +1,186 @@
...
+    if (!Mavlink2RestManager.instance) {
+      Mavlink2RestManager.instance = new Mavlink2RestManager()
+    }
+    return Mavlink2RestManager.instance
+  }
+
+  /**
+   * Check for valid base URL from REST API
+   */
+  probeBaseUrlCandidates () {
```
Williangalvani, line 127: is this working to find the correct url? is it required still?

Williangalvani (commented):
I got a couple of questions.

I like the code overall. it feels... familiar :eyes:

core/frontend/src/libs/MAVLink2Rest.ts
patrickelectric: With that we could run the frontend and the service locally without running entire companion core

core/frontend/src/libs/MAVLink2Rest.ts
patrickelectric: Check my previous comment

Williangalvani (approved):

rafaellehmkuhl:
Code looks good.
Was wondering though if wouldn't it be better for it to be available as an external dependency, like the marvlink2rest itself?

patrickelectric:
> Was wondering though if wouldn't it be better for it to be available as an external dependency, like the marvlink2rest itself?

How so ?

rafaellehmkuhl:
> > Was wondering though if wouldn't it be better for it to be available as an external dependency, like the marvlink2rest itself?
> 
> How so ?

Publishing on NPM.

patrickelectric:
> Publishing on NPM.

For the same reason that our python services are not in pypa, we are not going to use it outside this project and it's an unnecessary bureaucracy.

rafaellehmkuhl:
I was not suggesting for bureaucracy reasons, but as a feature to your package, since mavlink2rest itself is already a library and probably it's users would benefit from having a javascript library available also.

If you prefer not to, it's ok.

patrickelectric:
> I was not suggesting for bureaucracy reasons, but as a feature to your package, since mavlink2rest itself is already a library and probably it's users would benefit from having a javascript library available also.

Oh, good idea!
Sadly I'm not going to have time to do such thing, but maybe in a future where I could have typescript object definition for each supported mavlink message.

PR #356, opened 2021-06-23
Add Service Scanner to frontend

Williangalvani:
image

questions:
make it not as wide? I don't like my current solution.
is src/components a good place for the component?
is it worth declaring an interface to remove the "anys" in there?

image coming soon to a docker registry near you. (williangalvani/companion-core:helper)

### DIFF
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "vue": {
     "transpileDependencies": [
-      "vuetify"
+      "vuetify",
+      "vuex-module-decorators"
     ],
     "pwa": {
       "name": "Companion",
@@ -45,7 +46,8 @@
     "vue": "^2.6.11",
     "vue-router": "^3.2.0",
     "vuetify": "^2.4.0",
-    "vuex": "^3.4.0"
+    "vuex": "^3.4.0",
+    "vuex-module-decorators": "^1.0.1"
   },
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^4.18.0",
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@mdi/font": "^5.9.55",
+    "axios": "^0.21.1",
     "core-js": "^3.6.5",
     "register-service-worker": "^1.7.1",
     "roboto-fontface": "*",
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -62,14 +62,20 @@
     <v-main>
       <router-view />
     </v-main>
+    <services-scanner />
   </v-app>
 </template>
 
 <script lang="ts">
 import Vue from 'vue'
+import ServicesScanner from './components/scanner/servicesScanner.vue'
 
 export default Vue.extend({
   name: 'App',
+
+  components: {
+    'services-scanner': ServicesScanner,
+  },
 
   data: () => ({
     drawer: false,
--- a/core/frontend/src/components/scanner/servicesScanner.vue
+++ b/core/frontend/src/components/scanner/servicesScanner.vue
@@ -1,0 +1,58 @@
+<template>
+  <span />
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import axios from 'axios'
+import { getModule } from 'vuex-module-decorators'
+import ServicesScannerStore from '@/store/servicesScanner'
+import {Service} from '@/types/SERVICE'
+
+const servicesHelper: ServicesScannerStore = getModule(ServicesScannerStore)
+
+/**
+ * Actual scanner for running services.
+ * This periodically fetches /helper/latest/web_services
+ * and updates the ServiceHelperStore with the available services
+ * @displayName Services Scanner
+ */
+export default Vue.extend({
+  name: 'ServicesFetcher',
+  data () {
+    return {
+      // IntervalId is a number (id) returned by setInterval()
+      intervalId: 0,
+    }
+  },
+  mounted () {
+    // Fetch network data
+    this.requestData()
+
+    // Fetch periodic API request
+    this.startPeriodicRequest()
+  },
+  beforeDestroy () {
+    clearInterval(this.intervalId)
+  },
+  methods: {
+    startPeriodicRequest () {
+      this.intervalId = setInterval(() => {
+        this.requestData()
+      }, 5000)
+    },
+    requestData () {
+      axios.get('/helper/latest/web_services').then((response) => {
+        // Sort services by port number
+        const services = response.data.sort(
+          (first: Service, second: Service) => first.port - second.port,
+        )
+        servicesHelper.updateFoundServices(services)
+      }).catch((error) => {
+        console.log('Error scanning for services:', error)
+        servicesHelper.updateFoundServices([])
+      })
+    },
+  },
+})
+</script>
--- a/core/frontend/src/store/index.ts
+++ b/core/frontend/src/store/index.ts
@@ -3,13 +3,4 @@ import Vuex from 'vuex'
 
 Vue.use(Vuex)
 
-export default new Vuex.Store({
-  state: {
-  },
-  mutations: {
-  },
-  actions: {
-  },
-  modules: {
-  },
-})
+export default new Vuex.Store({})
--- a/core/frontend/src/components/scanner/availableServicesTable.vue
+++ b/core/frontend/src/components/scanner/availableServicesTable.vue
@@ -1,0 +1,81 @@
+<template>
+  <v-simple-table class="helper-table">
+    <thead>
+      <tr>
+        <th>Port</th>
+        <th>Service Name</th>
+        <th>Webpage</th>
+        <th>API Documentation</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr
+        v-for="service in availableServices"
+        :key="service.port"
+      >
+        <td>{{ service.port }}</td>
+        <td>{{ service.title }}</td>
+        <td v-html="createWebpageHyperlink(service.port)" />
+        <td
+          v-if="service.documentation_url"
+          v-html="createWebpageHyperlink(service.port, service.documentation_url)"
+        />
+        <td v-else>
+          No API documentation
+        </td>
+      </tr>
+    </tbody>
+  </v-simple-table>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { getModule } from 'vuex-module-decorators'
+import ServicesScannerStore from '@/store/servicesScanner'
+
+const servicesHelperStore: ServicesScannerStore = getModule(ServicesScannerStore)
+
+/**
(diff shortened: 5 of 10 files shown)


rafaellehmkuhl:
Looks good. I would just try to centralize it.

API docs are not working?

Williangalvani:
> Looks good. I would just try to centralize it.
> 
> API docs are not working?

this was on my laptop. I started the docker with no binds or anything, so a bunch of services aren't working

rafaellehmkuhl:
Cool

rafaellehmkuhl (approved):

core/frontend/src/store/serviceHelper.ts
```diff
@@ -0,0 +1,27 @@
...
+  port: number;
+}
+
+@Module({
+  dynamic: true,
+  store,
+  name: 'serviceHelper',
+})
+
+@Module
```
rafaellehmkuhl, line 18: Those two lines should not exist, I think.

core/frontend/src/store/serviceHelper.ts
```diff
@@ -0,0 +1,27 @@
+import { Module, VuexModule, Mutation } from 'vuex-module-decorators'
+import store from '@/store'
+
+
+export interface Service {
+  valid: boolean;
+  title: string;
+  documentationUrl: string;
+  port: number;
+}
```
rafaellehmkuhl, line 10: I think it would be good for us to keep all the types/interfaces on `@/types/SERVICE.d.ts`.

core/frontend/src/store/serviceHelper.ts
```diff
@@ -0,0 +1,27 @@
...
+  name: 'serviceHelper',
+})
+
+@Module
+export default class ServiceHelperStore extends VuexModule {
+  services: Array<Service> = []
+
+
+  @Mutation
+  updateFoundServices (services: Array<Service>): void {
```
rafaellehmkuhl, line 24: `Service[]` is a cool syntax. Learned recently.

core/frontend/src/store/index.ts
```diff
@@ -1,5 +1,6 @@
 import Vue from 'vue'
 import Vuex from 'vuex'
+import ServiceHelper from './serviceHelper'
```
rafaellehmkuhl: By putting the last two commits on the beggining, this one can be removed.

core/frontend/src/App.vue
```diff
@@ -62,15 +62,21 @@
...
 import Vue from 'vue'
+import ServicesFetcher from './components/servicesFetcher.vue'
 
 export default Vue.extend({
   name: 'App',
 
+  components: {
+    'services-fetcher': ServicesFetcher,
+  },
+
```
rafaellehmkuhl, line 79: I have a small problem with the 'ServicesFetcher' name. It seems like it's responsible for fetching service data, not info about the service. Something like 'ServiceInfoFetcher' wouldn't be better?

core/frontend/package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "vue": {
     "transpileDependencies": [
-      "vuetify"
+      "vuetify",
```
rafaellehmkuhl, line 7: This commit belongs to the beggining of the PR.

core/frontend/src/store/index.ts
```diff
@@ -1,17 +1,6 @@
 import Vue from 'vue'
 import Vuex from 'vuex'
```
rafaellehmkuhl: This commit substitutes the first one adding the Helper store to the modules.

rafaellehmkuhl (changes requested):

core/frontend/package.json
Williangalvani: done

core/frontend/src/store/serviceHelper.ts
Williangalvani: sounds good. so each one gets a file?

rafaellehmkuhl: Yup.
This way we keep the same separation for store, components and types.

core/frontend/src/components/helper.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <v-simple-table class="helper-table">
+    <thead>
```
rafaellehmkuhl, line 3: Forgot to add, but it would be better if this and eventually other helper components stay under `@/components/helper/`. This way we keep the service-specific components separated.

rafaellehmkuhl (changes requested):

core/frontend/src/App.vue
Williangalvani: Yeah... let's see if I come up with a better name overnight...

core/frontend/package.json
```diff
@@ -4,7 +4,8 @@
...
+      "vuex-module-decorators"
```
patrickelectric, line 8: I dream for a day where package.json accepts trialing commas

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,59 @@
...
+  mounted () {
+    // Fetch network data
+    this.requestData()
+
+    // Fetch periodic API request
+    this.startPeriodicRequest()
+  },
+  methods: {
+    startPeriodicRequest () {
+      setInterval(() => {
```
patrickelectric, line 32: Just to make sure, if we mount, by mistake, this module more than one time, will this create two periodic requests ?
I'm saying that since the return variable for the `setInterval` is not being used and the RAII will not run (?).
Maybe we should add a variable for it and delete and delete in the destructor, but I'm not sure how typescript will deal with the actual scenario.

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,59 @@
...
+        this.requestData()
+      }, 5000)
+    },
+    requestData () {
+      axios.get('http://localhost/helper/latest/web_services').then((response) => {
```
patrickelectric, line 37: localhost ?
Shouldn't this be current host ?

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,59 @@
...
+        // Sort services by port number
+        const services = response.data.sort(
+          (first: Service, second: Service) => first.port - second.port,
```
patrickelectric, line 40: :ok_hand:

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,59 @@
...
+        )
+        servicesHelper.updateFoundServices(services)
+      }).catch((error) => {
+        console.log('Error scanning for services:')
```
patrickelectric, line 44: Maybe:
```suggestion
        console.log(`Error scanning for services: ${error}`)
```

patrickelectric (commented):
Just some minor questions, but the patch looks great

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,59 @@
...
+  },
+})
+</script>
+
+<style scoped>
+.helper-table {
+    max-width: 70%;
+    margin: auto;
+}
+</style>
```
rafaellehmkuhl, line 59: This is not being used here, right?

rafaellehmkuhl (changes requested):

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: good catch!

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: the explanation is in the next line. doing this prints the error as a string, logging int alone give us better information.

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: yes, it is, this just compresses the table horizontally a bit. It probably should be a media query, but I can't really do those.

core/frontend/src/components/scanner/servicesScanner.vue
patrickelectric: About that, I'm not sure if both lines will run without interruption of another `console.log` when a interruption/event/async happens. But I'm ok if you think that the current approach should work.

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: weird. I had some local changes I hadn't pushed yet. It is there now. line 2

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: I used that for running locally on port 8080 but talking to the helper in a docker, fixed.

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,66 @@
...
+ * Actual scanner for running services.
+ * This periodically fetches /helper/latest/web_services
+ * and updates the ServiceHelperStore with the available services
+ * @displayName Services Scanner
+ */
+export default Vue.extend({
+  name: 'ServicesFetcher',
+  data () {
+    return {
+      interval: 0,
```
patrickelectric, line 24: For someone reading the code, it may think that interval is the interval in time, and not the interval timer ID. 
Maybe: `interval_id`, `timer_id` or `request_timer_id` may help, also a small comment such as "We define the request timer ID by default as 0"

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: I really prefer the extra details we get from logging the error on its own.

patrickelectric: Does `console.log('Error scanning for services: ', error)` works for you ?

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: done.

core/frontend/src/components/scanner/servicesScanner.vue
Williangalvani: yeah, that works. updated.

core/frontend/src/components/scanner/servicesScanner.vue
```diff
@@ -0,0 +1,65 @@
...
+      })
+    },
+  },
+})
+</script>
+
+<style scoped>
+.helper-table {
+    max-width: 70%;
+    margin: auto;
```
rafaellehmkuhl, line 63: To remove

rafaellehmkuhl (changes requested):

rafaellehmkuhl (approved):

patrickelectric (approved):

Issue #354, opened 2021-06-22
All settings and logs are being stored at /root/.config instead of in the user home (in the host os)


Williangalvani:

patrickelectric:
Just to make it clear, we are talking about the raspberry bind and not inside the docker.

Williangalvani:
We decided to stay on /root/.config.

PR #362, opened 2021-06-24
Move tlogs to `.config` folder making them persistent

### LINKED ISSUE #353: core: ardupilot-manager: Logs are going to /tmp folder and being lost
`/tmp` folder is not binded on core, so they are getting lost on power down, and even if we bind it, `/tmp` is cleared on boot.

Where should we keep the logs? `/root/config/ardupilot-manager/logs` or something like that?

rafaellehmkuhl:
This PR fixes #353 by allowing mavlink-manager to set the logs directory and changing it to the `logs` folder, created under the app folder.

To allow for better execution and catching behaviors correctly, I separated the methods to create the folders and to create the settings file. This way we always make sure the app folders are there, independently of having or not a settings file (e.g. there could be the case where the logs folder is not there but the settings file is, thus the logs folder not being created).

This fixes the problem of lost logs for now and makes the path for the future decision of where to better put the files, raised by @Williangalvani on #354.

### DIFF
--- a/core/services/ardupilot_manager/settings.py
+++ b/core/services/ardupilot_manager/settings.py
@@ -12,7 +12,8 @@ class Settings:
     settings_path = Path(appdirs.user_config_dir(app_name))
     settings_file = Path.joinpath(settings_path, "settings.json")
     firmware_path = Path.joinpath(settings_path, "firmware")
-    app_folders = [settings_path, firmware_path]
+    log_path = Path.joinpath(settings_path, "logs")
+    app_folders = [settings_path, firmware_path, log_path]
 
     def __init__(self) -> None:
         self.root: Dict[str, Union[int, Dict[str, Any]]] = {"version": 0, "content": {}}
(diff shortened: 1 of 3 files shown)


core/services/ardupilot_manager/settings.py
```diff
@@ -12,15 +12,25 @@ class Settings:
...
+        self.create_app_folders()
+
+    @staticmethod
+    def create_app_folders() -> None:
+        """Create the necessary folders for proper app function."""
+        for folder in Settings.app_folders:
+            try:
+                Path.mkdir(folder, exist_ok=True)
+            except FileExistsError:
+                logger.warning(f"Found file {folder} where a folder should be. Removing file and creating folder.")
```
Williangalvani, line 29: I don't think it is creating the folder after this line. is it?

rafaellehmkuhl: Oh, yes. Good catch. It does not retry.

rafaellehmkuhl: Solved.

Williangalvani (approved):

patrickelectric (approved):
Good clean patch

PR #369, opened 2021-06-24
use wheel for aiohttp in armv7

Williangalvani:
This should speed up arm builds.
This also uses PIP to install the services as vannila setup-tools is unable to handle [PEP 508](https://www.python.org/dev/peps/pep-0508/)

### DIFF
--- a/core/services/install-services.sh
+++ b/core/services/install-services.sh
@@ -10,6 +10,13 @@ BUILD_PACKAGES=(
 apt update
 apt install -y --no-install-recommends ${BUILD_PACKAGES[*]}
 
+# Pre-Build dependencies:
+# For convenience, we build ourselves the .wheel packages for dependencies
+# which have no armv7 wheel in pypi. This saves a lot of build time in docker
+if [[ "$(uname -m)" == "armv7l"* ]]; then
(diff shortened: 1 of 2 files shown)


core/services/install-services.sh
```diff
@@ -25,7 +25,7 @@ cd /home/pi/services/ardupilot_manager/ && python3 setup.py install
 cd /home/pi/services/helper/ && python3 setup.py install
 
 # Version Chooser service:
-cd /home/pi/services/versionchooser/ && python3 setup.py install
+cd /home/pi/services/versionchooser/ && pip install .
```
patrickelectric, line 28: Should we move everything to pip install for consistency ?
Also, there is no comment explaining the reason why pip is used here and not in the following lines

Williangalvani: I think we can change them all to pip. Any thoughts @rafaellehmkuhl ?

rafaellehmkuhl: Definitely agree. Was going to do this on the commonwealth PR but forgot.

rafaellehmkuhl (approved):
Would just move them all to pip and it's ready to go.

Williangalvani:
It looks like pip install . is not downloading the static files to the static files folder.

patrickelectric (approved):
image

PR #370, opened 2021-06-25
core: No more mavlink-router builds, download binary directly

### LINKED ISSUE #352: Install mavlink-router as binary and not from scratch
This helps to improve the build time

patrickelectric:
Fix #352


core/tools/mavlink_router/bootstrap.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+LOCAL_BINARY_PATH="/usr/bin/mavlink-routerd"
+VERSION="companion-core-devlopment-0.1"
```
Williangalvani, line 7:
```suggestion
VERSION="companion-core-development-0.1"
```

patrickelectric: Ops, I'll need to create the tag and build for that.

.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       name: Install dependencies
       run: |
         # Install mavlink-router
```
rafaellehmkuhl, line 24: This line is not from the PR itself, but it would be good to remove or modify it, because the CI is interpreting all commands below it (apt, npm, etc) as if they were under it.

rafaellehmkuhl (commented):
Seems good. The test apparently is failling to execute the binary, though.

.github/workflows/tests.yml
patrickelectric: Done.

rafaellehmkuhl (approved):

PR #377, opened 2021-06-29
Corrections around start-companion-core

patrickelectric:


rafaellehmkuhl (approved):

Williangalvani (approved):

core/Dockerfile
```diff
@@ -46,5 +46,10 @@ COPY --from=downloadBinaries \
 # Copy frontend built on frontendBuild to this stage
 COPY --from=frontendBuilder /home/pi/frontend/dist /home/pi/frontend
 
+# Asserts
+## Make sure that .config does not exist, this needs to be populated at runtime,
+## any file that exists under this folder will not be accessible after the docker bind.
+RUN [ ! -d "/root/.config" ]
+
```
rafaellehmkuhl, line 53: To better understand, what will be the final behavior with this change?
If we have something under `/root/.config`, what will happen? And if it is not there?

patrickelectric: As the comment says, is an Assert. If the folder exists the build will be canceled.

rafaellehmkuhl: Just the bind will be canceled, or all the building?

patrickelectric: Build, is a docker step.

rafaellehmkuhl: So, after the folder is created the first time, it would not work anymore? I think I'm missing something.

rafaellehmkuhl (approved):

core/Dockerfile
patrickelectric: As I said, it's an assert. The build will fail if the folder exist. We do not want the folder to exist, because if it does exist we are doing something wrong since the path will have a bind in the host computer and everything inside will not be available for it.

rafaellehmkuhl: Oh, it checks if the folder exists *inside* the docker image. Of couse. Got it upside down.

rafaellehmkuhl (approved):
I would just modify the comment about the .config folder a little bit, to be more clear. As it is right now, it appears that the build will fail if the folder exists on host.

patrickelectric:
> I would just modify the comment about the .config folder a little bit, to be more clear. As it is right now, it appears that the build will fail if the folder exists on host.

What are you saying is not possible with docker..
The code is pretty clear as the comment. If the folder exists we fail. Dockerfiles run inside docker builds, so we are checking the line inside docker.

rafaellehmkuhl:
> > I would just modify the comment about the .config folder a little bit, to be more clear. As it is right now, it appears that the build will fail if the folder exists on host.
> 
> What are you saying is not possible with docker..
> The code is pretty clear as the comment. If the folder exists we fail. Dockerfiles run inside docker builds, so we are checking the line inside docker.

I understand it's not possible. I'm just talking about the semantic of the comment. It induces to misinterpretation.

patrickelectric:
> I'm just talking about the semantic of the comment. It induces to misinterpretation.

I'm open to suggestions

rafaellehmkuhl:
> > I'm just talking about the semantic of the comment. It induces to misinterpretation.
> 
> I'm open to suggestions

Something like this seems good:

```
## Make sure that .config folder is not being created inside the image. As 
## this needs to be populated at runtime, any file that exists under this folder
## on the host OS would not be accessible after the docker bind if the folder
## was also created here.
```

patrickelectric:
> > > I'm just talking about the semantic of the comment. It induces to misinterpretation.
> > 
> > 
> > I'm open to suggestions
> 
> Something like this seems good:
> 
> ```
> ## Make sure that .config folder is not being created inside the image. As 
> ## this needs to be populated at runtime, any file that exists under this folder
> ## on the host OS would not be accessible after the docker bind if the folder
> ## was also created here.
> ```

This comment is wrong, the host OS Will be accessible and what is inside the image is not. That's why you got confused, if you read the comment is the other way around. 
I would recommend to read: https://docs.docker.com/storage/bind-mounts/#:~:text=Bind%20mounts%20have%20been%20around,path%20on%20the%20host%20machine.

rafaellehmkuhl:
> > > > I'm just talking about the semantic of the comment. It induces to misinterpretation.
> > > 
> > > 
> > > I'm open to suggestions
> > 
> > 
> > Something like this seems good:
> > ```
> > ## Make sure that .config folder is not being created inside the image. As 
> > ## this needs to be populated at runtime, any file that exists under this folder
> > ## on the host OS would not be accessible after the docker bind if the folder
> > ## was also created here.
> > ```
> 
> This comment is wrong, the host OS Will be accessible and what is inside the image is not. That's why you got confused, if you read the original comment is the other way around.
> I would recommend to read: https://docs.docker.com/storage/bind-mounts/#:~:text=Bind%20mounts%20have%20been%20around,path%20on%20the%20host%20machine.

Got it. 
Anyway, works the other way around:

```
## Make sure that .config folder is not being created inside the image. As 
## this needs to be populated at runtime, any file that exists under this folder
## on the container would not be accessible after the docker bind if the folder
## already exists in the OS.
```

patrickelectric:
> ## Make sure that .config folder is not being created inside the image. As 
> ## this needs to be populated at runtime, any file that exists under this folder
> ## on the container would not be accessible after the docker bind if the folder
> ## already exists in the OS.

This is also incorrect, if exists or not does not matter. Docker will generate the folder on the host machine.

rafaellehmkuhl:
> > ## Make sure that .config folder is not being created inside the image. As
> > ## this needs to be populated at runtime, any file that exists under this folder
> > ## on the container would not be accessible after the docker bind if the folder
> > ## already exists in the OS.
> 
> This is also incorrect, if exists or not does not matter. Docker will generate the folder on the host machine.

So it will subscribe the folder in the host machine?

patrickelectric:
> So it will subscribe the folder in the host machine?

No, **if** the folder exists, it exists. **If not** docker will create it.

rafaellehmkuhl:
So what case exactly are we trying to prevent here?

Williangalvani:
> So what case exactly are we trying to prevent here?

When running, the .config folder in the docker is not accessible. It gets shadowed by the Host's .config folder.
If we have anything in the .config folder in the image (in build step), it will not be accessible (when running), which means we put it there by mistake.

rafaellehmkuhl:
> > So what case exactly are we trying to prevent here?
> 
> When running, the .config folder in the docker is not accessible. It gets shadowed by the Host's .config folder.
> If we have anything in the .config folder in the image (in build step), it will not be accessible (when running), which means we put it there by mistake.
> 
> We are trying to prevent developer errors of placing a file in the wrong place, where it can't be used in runtime.

I think your comment can be put there as it is. It's very clear of what is happening and why the command is there.

PR #380, opened 2021-07-05
Add SITL to REST API

rafaellehmkuhl:
Allows for toggling between SITL and boards from the REST API.

Depends on #378


rafaellehmkuhl:
Closed in favor of #490

PR #383, opened 2021-07-13
Refactor wifi service

### LINKED ISSUE #211: Connecting to a wifi can be unsuccessful but yet return successful
Sometimes there's a network that can be seen when scanning, but has a weak signal. 
When trying to connect to this network, sometimes the wifi service will return a successful result, but when user checks wifi status, it's on the same network as it was before.
It's not clear to me if the service was indeed able to connect to the new network but wifi_supplicant decided to revert to the previous one because of the weak signal, or if it didn't connect at all.

This is the log from the wifi service:

```
> ADD_NETWORK
< 4
> SET_NETWORK 4 ssid "Network1234"
< OK
> SET_NETWORK 4 psk "abcd1234"
< OK
> ENABLE_NETWORK 4
< OK
> SAVE_CONFIG
< OK
> RECONFIGURE
< OK
> STATUS
< wpa_state=SCANNING
ip_address=192.168.1.103
p2p_device_address=de:a6:32:0f:a8:71
address=dc:a6:32:0f:a8:71
uuid=5410d263-b03f-5bc6-a7a8-586a400f62ee
```

### LINKED ISSUE #238: Allow connection to hidden wifi network
Maybe a button "Connect to hidden network".

### LINKED ISSUE #294: core: wifi: New wifi connections to existing networks generates new entries on wpa_supplicant.conf
Currently the wifi service does not check if the network of a new connection already exists on the wpa_supplicant.conf file, so creating a new entry.

Expected behavior should be:

1. Check if the entry exists
2. If it exists, check if the password is the same
3. If the password is the same, don't create a new entry
4. If the password is not the same, override the password on the existing entry

rafaellehmkuhl:
Definitely a big refactor, but a much needed one.
Among the changes:
- Introduces timeouts for WPA operations (they could be slow or busy sometimes, making requests fail)
- Everything is async now (YES!) (scan operations, which can take from 5 to 15 seconds do not block other requests anymore)
- Scan logic is refactored so in-progress scan results are used on new fetchs (e.g: 10 requests will not necessarily create 10 scan operations. If there's one scan happening when you make a request, its result will be used)
- Network-registering is now separated from network-connecting, so new connections to known/saved networks don't create infinite new entries
- We can now connect to unprotected networks
- We can now forget saved networks
- Connections are now forced by default (currently if there were two known networks nearby and you were connected in one of them, you couldn't connect on the other)
- Allow connecting to hidden networks
- Improved readability on some stuff

Fix #294 
Fix #211 
Fix #238

### DIFF
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -138,7 +138,10 @@ class WifiManager:
             network_number = await self.wpa.send_command_add_network()
 
             await self.wpa.send_command_set_network(network_number, "ssid", f'"{credentials.ssid}"')
-            await self.wpa.send_command_set_network(network_number, "psk", f'"{credentials.password}"')
+            if not credentials.password:
+                await self.wpa.send_command_set_network(network_number, "key_mgmt", "NONE")
+            else:
+                await self.wpa.send_command_set_network(network_number, "psk", f'"{credentials.password}"')
             await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
             return network_number
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -135,14 +135,13 @@ class WifiManager:
             credentials {[WifiCredentials]} -- object containing ssid and password of the network
         """
         try:
-            data = await self.wpa.send_command_add_network()
+            network_number = await self.wpa.send_command_add_network()
 
-            network_number = data.decode("utf-8")
             await self.wpa.send_command_set_network(network_number, "ssid", f'"{credentials.ssid}"')
             await self.wpa.send_command_set_network(network_number, "psk", f'"{credentials.password}"')
             await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
-            return int(network_number)
+            return network_number
         except Exception as error:
             raise ConnectionError("Failed to set new network.") from error
 
@@ -153,7 +152,7 @@ class WifiManager:
             network_id {int} -- Network ID provided by WPA Supplicant
         """
         try:
-            await self.wpa.send_command_enable_network(str(network_id))
+            await self.wpa.send_command_enable_network(network_id)
             await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
             await self.wpa.send_command_reconnect()
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -165,7 +165,12 @@ class WifiManager:
             network_id {int} -- Network ID provided by WPA Supplicant
         """
         try:
-            await self.wpa.send_command_enable_network(network_id)
+            # Using select_network forces connection on the desired network by disabling all the others. To ensure we
+            # can still automatically connect to the other saved networks when far from this one, we re-enable them.
+            await self.wpa.send_command_select_network(network_id)
+            saved_networks = await self.get_saved_wifi_network()
+            for network in saved_networks:
+                await self.wpa.send_command_enable_network(network.networkid)
             await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
             await self.wpa.send_command_reconnect()
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -84,7 +84,7 @@ class WifiManager:
     async def get_wifi_available(self) -> List[ScannedWifiNetwork]:
         """Get a dict from the wifi signals available"""
         try:
-            await self.wpa.send_command_scan()
+            await self.wpa.send_command_scan(timeout=15)
             data = await self.wpa.send_command_scan_results()
             networks_list = WifiManager.__dict_from_table(data)
             return [ScannedWifiNetwork(**network) for network in networks_list]
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -1,4 +1,7 @@
-from typing import Any, Dict, List
+import asyncio
+from typing import Any, Dict, List, Optional
+
+from loguru import logger
 
 from exceptions import FetchError, ParseError
 from typedefs import SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials
@@ -15,6 +18,8 @@ class WifiManager:
             path {[tuple/str]} -- Can be a tuple to connect (ip/port) or unix socket file
         """
         self.wpa.run(path)
+        self._scan_task: Optional[asyncio.Task[bytes]] = None
+        self._updated_scan_results: Optional[List[ScannedWifiNetwork]] = None
 
     @staticmethod
     def __decode_escaped(data: bytes) -> str:
@@ -81,13 +86,38 @@ class WifiManager:
 
     async def get_wifi_available(self) -> List[ScannedWifiNetwork]:
         """Get a dict from the wifi signals available"""
-        try:
-            await self.wpa.send_command_scan(timeout=15)
-            data = await self.wpa.send_command_scan_results()
-            networks_list = WifiManager.__dict_from_table(data)
-            return [ScannedWifiNetwork(**network) for network in networks_list]
-        except Exception as error:
-            raise FetchError("Failed to fetch wifi list.") from error
+
+        async def perform_new_scan() -> None:
+            try:
+                # We store the scan task here so that new scan requests that happen in the interval
+                # where this one is running can check when it has finished.
+                # Otherwise, it could happen that a new scan is initiated before the ones waiting know
+                # the previous one has finished, making them stay on the loop unnecessarily.
+                self._scan_task = asyncio.create_task(self.wpa.send_command_scan(timeout=15))
+                await self._scan_task
+                data = await self.wpa.send_command_scan_results()
+                networks_list = WifiManager.__dict_from_table(data)
+                self._updated_scan_results = [ScannedWifiNetwork(**network) for network in networks_list]
+            except Exception as error:
+                if self._scan_task is not None:
+                    self._scan_task.cancel()
+                self._updated_scan_results = None
+                raise FetchError("Failed to fetch wifi list.") from error
+
+        # Performs a new scan only if it's the first one or the last one is already done
+        # In case there's one running already, wait for it to finish and use its result
+        if self._scan_task is None or self._scan_task.done():
+            await perform_new_scan()
+        else:
+            awaited_scan_task_name = self._scan_task.get_name()
+            while self._scan_task.get_name() == awaited_scan_task_name and not self._scan_task.done():
+                logger.info(f"Waiting for {awaited_scan_task_name} results.")
+                await asyncio.sleep(0.5)
+
+        if self._updated_scan_results is None:
+            raise FetchError("Failed to fetch wifi list.")
+
+        return self._updated_scan_results
 
     async def get_saved_wifi_network(self) -> List[SavedWifiNetwork]:
         """Get a list of saved wifi networks"""
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -2,19 +2,8 @@ from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
 
+from exceptions import BusyError, FetchError, ParseError
 from wpa_supplicant import WPASupplicant
-
-
-class ParseError(ValueError):
-    """Raise for errors regarding fail parsing string data to proper structs."""
-
-
-class FetchError(ValueError):
-    """Raise for errors regarding fail on fetching data."""
-
-
-class BusyError(ValueError):
-    """Raise for errors regarding excessive number of requests on a given time."""
 
 
 class ScannedWifiNetwork(BaseModel):
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -128,7 +128,7 @@ class WifiManager:
         except Exception as error:
             raise FetchError("Failed to fetch saved networks list.") from error
 
-    async def set_wifi_password(self, credentials: WifiCredentials) -> Any:
+    async def add_network(self, credentials: WifiCredentials) -> int:
         """Set network ssid and password
 
         Arguments:
@@ -140,13 +140,26 @@ class WifiManager:
             network_number = data.decode("utf-8")
             await self.wpa.send_command_set_network(network_number, "ssid", '"{}"'.format(credentials.ssid))
             await self.wpa.send_command_set_network(network_number, "psk", '"{}"'.format(credentials.password))
-            await self.wpa.send_command_enable_network(network_number)
-            await self.wpa.send_command_reconnect()
 
             await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
+            return int(network_number)
         except Exception as error:
             raise ConnectionError("Failed to set new network.") from error
+
+    async def connect_to_network(self, network_id: int) -> None:
+        """Connect to wifi network
+
+        Arguments:
+            network_id {int} -- Network ID provided by WPA Supplicant
+        """
+        try:
+            await self.wpa.send_command_enable_network(str(network_id))
+            await self.wpa.send_command_save_config()
+            await self.wpa.send_command_reconfigure()
+            await self.wpa.send_command_reconnect()
+        except Exception as error:
+            raise ConnectionError("Failed to connect to network.") from error
 
     async def status(self) -> Dict[str, Any]:
         """Check wpa_supplicant status"""
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -1,6 +1,6 @@
 from typing import Any, Dict, List
 
-from exceptions import BusyError, FetchError, ParseError
+from exceptions import FetchError, ParseError
 from typedefs import SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials
 from wpa_supplicant import WPASupplicant
 
@@ -84,14 +84,10 @@ class WifiManager:
     def get_wifi_available(self) -> List[ScannedWifiNetwork]:
         """Get a dict from the wifi signals available"""
         try:
-            scan_data = self.wpa.send_command_scan()
-            if b"BUSY" in scan_data:
-                raise BusyError
+            self.wpa.send_command_scan()
             data = self.wpa.send_command_scan_results()
             networks_list = WifiManager.__dict_from_table(data)
             return [ScannedWifiNetwork(**network) for network in networks_list]
-        except BusyError as error:
-            raise BusyError("WPA service is busy. Try to space network scans by at least 4 seconds.") from error
         except Exception as error:
             raise FetchError("Failed to fetch wifi list.") from error
 
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -108,9 +108,6 @@ class WifiManager:
         """
         try:
             data = await self.wpa.send_command_add_network()
-            data = data.strip()
-            if data == b"Fail":
-                raise RuntimeError("Failed to add new network.")
 
             network_number = data.decode("utf-8")
             await self.wpa.send_command_set_network(network_number, "ssid", '"{}"'.format(credentials.ssid))
@@ -118,11 +115,7 @@ class WifiManager:
             await self.wpa.send_command_enable_network(network_number)
             await self.wpa.send_command_reconnect()
 
-            answer = await self.wpa.send_command_save_config()
-            answer = answer.strip()
-            if answer == b"FAIL":
-                raise RuntimeError(str(answer))
-
+            await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
         except Exception as error:
             raise ConnectionError("Failed to set new network.") from error
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -81,74 +81,74 @@ class WifiManager:
 
         return output
 
-    def get_wifi_available(self) -> List[ScannedWifiNetwork]:
+    async def get_wifi_available(self) -> List[ScannedWifiNetwork]:
         """Get a dict from the wifi signals available"""
         try:
-            self.wpa.send_command_scan()
-            data = self.wpa.send_command_scan_results()
+            await self.wpa.send_command_scan()
+            data = await self.wpa.send_command_scan_results()
             networks_list = WifiManager.__dict_from_table(data)
             return [ScannedWifiNetwork(**network) for network in networks_list]
         except Exception as error:
             raise FetchError("Failed to fetch wifi list.") from error
 
-    def get_saved_wifi_network(self) -> List[SavedWifiNetwork]:
+    async def get_saved_wifi_network(self) -> List[SavedWifiNetwork]:
         """Get a list of saved wifi networks"""
         try:
-            data = self.wpa.send_command_list_networks()
+            data = await self.wpa.send_command_list_networks()
             networks_list = WifiManager.__dict_from_table(data)
             return [SavedWifiNetwork(**network) for network in networks_list]
         except Exception as error:
             raise FetchError("Failed to fetch saved networks list.") from error
 
-    def set_wifi_password(self, credentials: WifiCredentials) -> Any:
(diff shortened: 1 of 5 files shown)


core/services/wifi/wpa_supplicant.py
```diff
@@ -47,223 +49,228 @@ def run(self, target: Union[Tuple[str, int], str] = target) -> None:
...
 
         try:
-            self.sock.send(command.encode("utf-8"))
-        except Exception as error:
-            raise SockCommError("Could not send data to WPA Supplicant socket.") from error
-
-        try:
-            data, _ = self.sock.recvfrom(self.BUFFER_SIZE)
+            timeout_start = time.time()
+            while True:
```
patrickelectric: This could be refactored to a while else

core/services/wifi/wpa_supplicant.py
```diff
@@ -47,223 +49,228 @@ def run(self, target: Union[Tuple[str, int], str] = target) -> None:
         self.sock.settimeout(10)
         self.sock.connect(self.target)
 
-    def send_command(self, command: str) -> bytes:
+    def send_command(self, command: str, timeout: float) -> bytes:
```
patrickelectric: timeout type is not specified (years, ms, s, ns ?)

core/services/wifi/WifiManager.py
```diff
@@ -79,74 +79,74 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
...
             data = data.strip()
             if data == b"Fail":
                 raise RuntimeError("Failed to add new network.")
 
             network_number = data.decode("utf-8")
-            self.wpa.send_command_set_network(network_number, "ssid", '"{}"'.format(credentials.ssid))
-            self.wpa.send_command_set_network(network_number, "psk", '"{}"'.format(credentials.password))
-            self.wpa.send_command_enable_network(network_number)
-            self.wpa.send_command_reconnect()
+            await self.wpa.send_command_set_network(str(network_number), "ssid", '"{}"'.format(credentials.ssid))
```
patrickelectric: Why network_number should needs a cast with await ?

core/services/wifi/WifiManager.py
```diff
@@ -141,18 +141,24 @@ async def set_wifi_password(self, credentials: WifiCredentials) -> Any:
...
             await self.wpa.send_command_set_network(str(network_number), "psk", '"{}"'.format(credentials.password))
-            await self.wpa.send_command_enable_network(network_number)
-            await self.wpa.send_command_reconnect()
-
-            answer = await self.wpa.send_command_save_config()
-            answer = answer.strip()
-            if answer == b"FAIL":
-                raise RuntimeError(str(answer))
-
+            await self.wpa.send_command_save_config()
```
patrickelectric, line 149: why are you ignoring save config return ? What happens if the operation fail ?

core/services/wifi/WifiManager.py
```diff
@@ -141,18 +141,24 @@ async def set_wifi_password(self, credentials: WifiCredentials) -> Any:
...
             await self.wpa.send_command_reconfigure()
+            return network_number
         except Exception as error:
             raise ConnectionError("Failed to set new network.") from error
 
+    async def connect_to_network(self, network_id: int) -> None:
+        """Connect to wifi network
+
+        Arguments:
+            ssid {str} -- Network SSID
```
patrickelectric, line 172: the documentation is wrong

core/services/wifi/main.py
```diff
@@ -62,7 +62,14 @@ async def saved() -> Any:
 @version(1, 0)
 async def connect(credentials: WifiCredentials) -> Any:
     try:
-        await wifi_manager.set_wifi_password(credentials)
+        saved_networks = await wifi_manager.get_saved_wifi_network()
+        for network in saved_networks:
```
patrickelectric, line 75: The code could be simpler with the inverse logic:
```
network = find(lambda network: network.ssid == credentials.ssid, networks)
if not network:
    network_id = await wifi_manager.add_network(credentials)
else:
    network_id = network.id
```

core/services/wifi/WifiManager.py
```diff
@@ -135,12 +135,12 @@ async def add_network(self, credentials: WifiCredentials) -> int:
         try:
             data = await self.wpa.send_command_add_network()
             data = data.strip()
-            if data == b"Fail":
+            if not data.isdigit():
                 raise RuntimeError("Failed to add new network.")
 
-            network_number = data.decode("utf-8")
-            await self.wpa.send_command_set_network(str(network_number), "ssid", f'"{credentials.ssid}"')
```
patrickelectric: Is this in the wrong commit ? Should be a fixup, check my previous comment

core/services/wifi/WifiManager.py
```diff
@@ -160,13 +160,17 @@ async def remove_network(self, network_id: int) -> None:
         except Exception as error:
             raise ConnectionError("Failed to remove existing network.") from error
 
-    async def connect_to_network(self, network_id: int) -> None:
+    async def connect_to_network(self, network_id: int, force_connection: bool = True) -> None:
```
patrickelectric, line 168: I believe that a connect_to_work should always force the connection, so the force_connection variable is unnecessary

core/services/wifi/wpa_supplicant.py
```diff
@@ -74,6 +74,9 @@ async def send_command(self, command: str, timeout: float) -> bytes:
         if self.verbose and data:
             logger.info(f"<\n{data.decode('utf-8').strip()}")
 
+        if data == b"FAIL":
+            raise WPAOperationFail("WPA operation {command} failed.")
```
patrickelectric, line 78: missing fstring

patrickelectric (changes requested):
Looks good, just some minor corrections

core/services/wifi/wpa_supplicant.py
rafaellehmkuhl: Humn, didn't get exaclty how.

Also, it seems like Github is still with the "Outdated" problem. Is this suggestion still true for the most recent version of the file?

core/services/wifi/wpa_supplicant.py
rafaellehmkuhl: Good one. Will improve the docstrings for this function.

core/services/wifi/WifiManager.py
rafaellehmkuhl: It doesn't need, it was a rebase mistake. Will fix. (still, this is a middle change. the final version refactors this part)

core/services/wifi/WifiManager.py
rafaellehmkuhl: I removed the fail check for the individual operations and added it as default on send_command so every operation is checked. Anyway, I should have removed the check on the save_config together with this other commit. Will change.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Good catch

core/services/wifi/main.py
rafaellehmkuhl: Definitely better. Changing.

core/services/wifi/WifiManager.py
rafaellehmkuhl: This part is not, but the previous one is. Will change there.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Agree.

core/services/wifi/wpa_supplicant.py
rafaellehmkuhl: Good catch. Changing.

core/services/wifi/WifiManager.py
```diff
@@ -113,78 +84,125 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
...
+                    self._scan_task.cancel()
+                self._updated_scan_results = None
+                raise FetchError("Failed to fetch wifi list.") from error
+
+        # Performs a new scan only if it's the first one or the last one is already done
+        # In case there's one running already, wait for it to finish and use its result
+        if self._scan_task is None:
+            await perform_new_scan()
+        elif self._scan_task.done():
+            await perform_new_scan()
```
Williangalvani, line 108: This should work:
```suggestion
        if self._scan_task is None or self._scan_task.done():
            await perform_new_scan()
```
The second clause should only run if the first one is false:
```
>>> batata = None
>>> if batata is None or batata.biscoito():
...     print("wat")
... 
wat

```

core/services/wifi/WifiManager.py
```diff
@@ -113,78 +84,125 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
...
+        else:
+            awaited_scan_task_name = self._scan_task.get_name()
+            while self._scan_task.get_name() == awaited_scan_task_name and not self._scan_task.done():
```
Williangalvani, line 111: I would add a comment explaining that this is to check the eventual case where a new task is started before the while loop resumes.

core/services/wifi/WifiManager.py
```diff
@@ -113,78 +84,125 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
...
             return [SavedWifiNetwork(**network) for network in networks_list]
         except Exception as error:
             raise FetchError("Failed to fetch saved networks list.") from error
 
-    def set_wifi_password(self, credentials: WifiCredentials) -> Any:
+    async def add_network(self, credentials: WifiCredentials, hidden: bool = False) -> int:
         """Set network ssid and password
 
         Arguments:
             credentials {[WifiCredentials]} -- object containing ssid and password of the network
```
Williangalvani, line 133: missing the hidden argument in the docstring

Williangalvani (changes requested):
looking good

core/services/wifi/wpa_supplicant.py
patrickelectric: I did the review as soon I received the review request notification

rafaellehmkuhl: I assume this was just because a following commit change this code again.
Anyway, can you elaborate on the suggestion? I didn't understand.

patrickelectric:
```
while time.time() - timeout_start > timeout:
    self.sock.send(command.encode("utf-8"))
    data, _ = self.sock.recvfrom(self.BUFFER_SIZE)
    if b"FAIL-BUSY" not in data:
        break

    print(f"Busy during {command} operation. Trying again...")
    time.sleep(0.5)
else:
    raise BusyError(f"{command} operation took more than specified timeout ({timeout}. Cancelling.")
```

patrickelectric: It also appears that the while logic of your code is wrong, it should be `< timeout`

rafaellehmkuhl: Got the idea!

The logic on mine is right, and yours is also. Is just that I *raise* when the timeout is reached and you're *running while* the timeout is *not* reached.

Both work, but yours is cleaner. Changing.

patrickelectric: From your original code:
`while time.time() - timeout_start > timeout:`
When you start will be:
`while timeout_start - timeout_start > timeout:`
And them:
`while 0 > timeout:`

That is false.

rafaellehmkuhl: No, my original code is:

```python
while True:
    if time.time() - timeout_start > timeout:
        raise BusyError
```

It's there yet, you can check.

patrickelectric: I see now, thanks!

rafaellehmkuhl: No prob. Thanks for the `while else` suggestion. I'm not used to it, but it truly makes the code cleaner.

core/services/wifi/wpa_supplicant.py
rafaellehmkuhl: Done

core/services/wifi/WifiManager.py
rafaellehmkuhl: Fixed.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Fixed.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Fixed.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Fixed.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Changed.

core/services/wifi/wpa_supplicant.py
rafaellehmkuhl: Fixed.

core/services/wifi/WifiManager.py
rafaellehmkuhl: This one passed by.
Changed.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Added.

core/services/wifi/WifiManager.py
rafaellehmkuhl: Fixed.

core/services/wifi/main.py
rafaellehmkuhl: Changed.

core/services/wifi/main.py
```diff
@@ -84,6 +84,22 @@ async def connect(credentials: WifiCredentials) -> Any:
...
 
+@app.post("/remove", summary="Remove saved wifi network.")
+@version(1, 0)
+async def remove(ssid: str) -> Any:
+    try:
+        saved_networks = await wifi_manager.get_saved_wifi_network()
+        for network in saved_networks:
+            if network.ssid == ssid:
+                await wifi_manager.remove_network(network.networkid)
+                return
```
patrickelectric:
```suggestion
    try:
        saved_networks = await wifi_manager.get_saved_wifi_network()
        network = next(filter(lambda network: network.ssid = ssid, saved_networks))
        await wifi_manager.remove_network(network.networkid)
        return
```

core/services/wifi/WifiManager.py
```diff
@@ -166,6 +166,9 @@ async def connect_to_network(self, network_id: int) -> None:
             network_id {int} -- Network ID provided by WPA Supplicant
         """
         try:
+            saved_networks = await self.get_saved_wifi_network()
+            for network in saved_networks:
+                await self.wpa.send_command_disable_network(network.networkid)
```
patrickelectric, line 178: I'm not sure if if this is the correct approach, disabling every single network when connecting to something else will result in a non connection when restarting the ROV.

Imagine the following scenario: 
Connect the ROV in House1 Wifi
Go to House2 and connect the ROV (House1 wifi not available)
Go back to House1 and there'll be no ethernet connection, since the network is disabled by default. 
We should force connection and not disabling everything else.

core/services/wifi/WifiManager.py
```diff
@@ -139,7 +139,10 @@ async def add_network(self, credentials: WifiCredentials) -> int:
 
             network_number = int(data)
             await self.wpa.send_command_set_network(network_number, "ssid", f'"{credentials.ssid}"')
-            await self.wpa.send_command_set_network(network_number, "psk", f'"{credentials.password}"')
+            if not credentials.password:
```
patrickelectric, line 144: What happens if the user just want to connect in a network that he has previous configured and don't want to set the password (since we have it stored) ?

core/services/wifi/WifiManager.py
```diff
@@ -128,11 +128,13 @@ async def get_saved_wifi_network(self) -> List[SavedWifiNetwork]:
...
             raise FetchError("Failed to fetch saved networks list.") from error
 
-    async def add_network(self, credentials: WifiCredentials) -> int:
+    async def add_network(self, credentials: WifiCredentials, hidden: bool = False) -> int:
         """Set network ssid and password
 
         Arguments:
             credentials {[WifiCredentials]} -- object containing ssid and password of the network
+            hidden {bool} -- Should be set as true when adding hidden networks (that have SSID as
+            blank or null). False by default.
```
patrickelectric, line 137: Can't we just do it if WifiCredentials.ssid is empty ?

patrickelectric (changes requested):
The patch looks good, just some minor corrections.

core/services/wifi/wpa_supplicant.py
```diff
@@ -47,223 +49,233 @@ def run(self, target: Union[Tuple[str, int], str] = target) -> None:
...
-
-        try:
-            data, _ = self.sock.recvfrom(self.BUFFER_SIZE)
+            timeout_start = time.time()
+            while time.time() - timeout_start < timeout:
+                self.sock.send(command.encode("utf-8"))
+                data, _ = self.sock.recvfrom(self.BUFFER_SIZE)
+                if b"FAIL-BUSY" in data:
+                    print(f"Busy during {command} operation. Trying again...")
+                    time.sleep(0.5)
```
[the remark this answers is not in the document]
rafaellehmkuhl: We can make it 0.1s, there's no problem.
With the except of SCAN, most commands rarelly return a busy state.

core/services/wifi/WifiManager.py
```diff
@@ -108,21 +108,14 @@ async def set_wifi_password(self, credentials: WifiCredentials) -> Any:
         """
         try:
             data = await self.wpa.send_command_add_network()
-            data = data.strip()
-            if data == b"Fail":
```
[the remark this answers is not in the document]
rafaellehmkuhl: I think you are mixing them. This one was already there, but I've removed since we now check for FAIL by default in every command.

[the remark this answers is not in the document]
patrickelectric: But capitalized FAIL is different from Fail, that is also different from FAIL-BUSY, capitalization matter on the string comparison. 
Where is the current patch checking all possible fails strings ?

core/services/wifi/WifiManager.py
```diff
@@ -81,13 +86,38 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
...
+                self._updated_scan_results = None
+                raise FetchError("Failed to fetch wifi list.") from error
+
+        # Performs a new scan only if it's the first one or the last one is already done
+        # In case there's one running already, wait for it to finish and use its result
+        if self._scan_task is None or self._scan_task.done():
+            await perform_new_scan()
+        else:
+            awaited_scan_task_name = self._scan_task.get_name()
+            while self._scan_task.get_name() == awaited_scan_task_name and not self._scan_task.done():
```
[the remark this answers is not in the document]
rafaellehmkuhl: The comment before tries to explain the problem and why do we use this.
Think about the following scenario:
1. API receives scan call.
2. No scan operation is in place, so a new one is started
3. API receives another scan call.
4. There's already a scan operation in place, so it sleeps, waiting for it to be finished
5. First scan is finished and gets out of the loop
6. API receives another scan call.
7. No scan operation is in place, so a new one is started
8. Second scan call gets out of the sleep and checks if the scan task is done, but it is not (the third call restarted it), so it sleeps again

Got how the second scan got trapped by the third one? With the name approach, we guarantee this won't happen, since it checks if the scan that is in place is the one it was waiting the first time.

core/services/wifi/WifiManager.py
```diff
@@ -137,12 +137,12 @@ async def add_network(self, credentials: WifiCredentials) -> int:
         try:
             data = await self.wpa.send_command_add_network()
 
-            network_number = data.decode("utf-8")
+            network_number = int(data)
```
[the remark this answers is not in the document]
rafaellehmkuhl: Yes. ADD_NETWORK always return an int representing the index of the network on the conf file or a FAIL. Since we already cover the FAIL case, it will always return bytes directly convertible to ints.

core/services/wifi/wpa_supplicant.py
```diff
@@ -66,15 +68,15 @@ async def send_command(self, command: str, timeout: float) -> bytes:
...
             if b"FAIL-BUSY" in data:
-                print(f"Busy during {command} operation. Trying again...")
+                logger.info(f"Busy during {command} operation. Trying again...")
                 await asyncio.sleep(0.5)
                 continue
             break
         else:
             raise BusyError(f"{command} operation took more than specified timeout ({timeout}). Cancelling.")
 
         if self.verbose and data:
```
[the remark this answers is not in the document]
rafaellehmkuhl: It is, no?

core/services/wifi/WifiManager.py
rafaellehmkuhl: Got it and agree with you that there is this case. I don't know about a method for forcing the connection without disabling the others, though. Do you know how to do it?

core/services/wifi/WifiManager.py
rafaellehmkuhl: This case is handled by `connect_to_network` method. This method here is just for adding new ones.

core/services/wifi/WifiManager.py
rafaellehmkuhl: You're not mixing hidden with unprotected?

core/services/wifi/WifiManager.py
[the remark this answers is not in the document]
rafaellehmkuhl: I'm pretty sure wpa_cli uses capitalized FAIL for all cases. Did you ever experienced a non-capitalized one?
Anyway, if you're afraid this can happen, we can just capitalize the received answer on the `send_command` check and compare just one possibility, covering all.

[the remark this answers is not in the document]
rafaellehmkuhl: Just checked [here](https://github.com/cozybit/hostap-sae/blob/master/wpa_supplicant/wpa_cli.c) and it's always capitalized, indeed.
This check on the `add_network` was in fact never working.

core/services/wifi/WifiManager.py
rafaellehmkuhl: From what I could understand from WPA supplicant and WPA cli, "SELECT_NETWORK" does this (disables all networks and connect to the desired one), but it enables them again after, so it forces the connection but keep the enabled ones...enabled.

I will try to see if it indeed works like this. If not, we can add a start logic stating that if all networks are disabled and one of them is available nearby, we enable it.

rafaellehmkuhl: No, "SELECT_NETWORK" indeed disables all the others but the desired, but doesn't enable them back. I've tried doing this (enabling them back after connection) and apparently this approach works (it doesn't connect again to the later enabled, even if it's stronger.

Will add this logic to the method.

core/services/wifi/WifiManager.py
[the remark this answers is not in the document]
patrickelectric: So send_command_add_network should return an int, not a bytearray

core/services/wifi/wpa_supplicant.py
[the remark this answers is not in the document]
patrickelectric: My point is: `self.verbose` is still necessary ? This should be handle by the log backend, no ?

core/services/wifi/WifiManager.py
patrickelectric: WifiCredentials has ssid information, if ssid is empty so the network is in hidden mode, no ?

core/services/wifi/wpa_supplicant.py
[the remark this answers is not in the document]
rafaellehmkuhl: I was keeping the previous behavior there, but we can change, for sure.

core/services/wifi/WifiManager.py
rafaellehmkuhl: This is the add_network method. To add the network, we need it's SSID. The SSID is just not available on the scan, but we still need it, so in this case, it's provided by the user on the frontend (when the frontend detects a hidden network, it shows an input box for the SSID, togheter with the one for the password, and sends, together with the WifiCredentials, a hidden flag.

patrickelectric: > This is the add_network method. To add the network, we need it's SSID.

So the comment in the code is wrong: `that have SSID as blank or null`
It should be: `that does not broadcast the SSID`

rafaellehmkuhl: Better. Changing.

patrickelectric (approved):

Williangalvani (approved):

PR #391, opened 2021-07-22
Add Wifi manager to main frontend

rafaellehmkuhl:
The wifi manager allows among other things:
- showing available networks (hidden or not) and it's properties;
- connecting to new networks (hidden or not);
- connect to existing networks (without pass);
- forgetting knows networks;
- forcing password change;
- disconnecting from current network;
- showing wifi adapter properties (including local IP, security type, signal level and others).

It is mobile-friendly, asynchronous and tries to follow interface patterns already familiar to most users.

Docker image available at `rafaellehmkuhl/companion-core:wifi-frontend`

Preview:
frontend

### DIFF
--- a/core/frontend/src/components/common/PasswordInput.vue
+++ b/core/frontend/src/components/common/PasswordInput.vue
@@ -1,0 +1,54 @@
+<template>
+  <v-text-field
+    :value="value"
+    :append-icon="password_show_icon"
+    :type="input_field_type"
+    label="Password"
+    hide-details="auto"
+    class="pa-2"
+    @input="emitInput"
+    @keyup.enter="emitSubmit"
+    @click:append="togglePasswordShow"
+  />
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+
+export default Vue.extend({
+  name: 'PasswordInput',
+  props: {
+    value: {
+      type: String,
+      default: '',
+    },
+  },
+  data() {
+    return {
+      show_password: false,
+    }
+  },
+  computed: {
+    password_show_icon(): string {
+      return this.show_password ? 'mdi-eye' : 'mdi-eye-off'
+    },
+    input_field_type(): string {
+      return this.show_password ? 'text' : 'password'
+    },
+  },
+  methods: {
+    togglePasswordShow() {
+      this.show_password = !this.show_password
+    },
+    emitInput(value: string) {
+      this.$emit('input', value)
+    },
+    emitSubmit() {
+      this.$emit('submit')
+    },
+  },
+})
+</script>
+
+<style>
+</style>
--- a/core/frontend/src/components/common/SpinningLogo.vue
+++ b/core/frontend/src/components/common/SpinningLogo.vue
@@ -1,0 +1,34 @@
+<template>
+  <v-img
+    class="lds-dual-ring"
+    src="../../assets/logo.svg"
+  />
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+
+export default Vue.extend({
+  name: 'SpinningLogo',
+})
+</script>
+
+<style>
+  .lds-dual-ring {
+    display: block;
+    width: 30%;
+    height: 30%;
+    margin: 5% auto;
+    animation-name: rotation360;
+    animation-duration: 2s;
+    animation-iteration-count: infinite;
+  }
+  @keyframes rotation360 {
+    0% {
+        transform: rotate(0deg);
+    }
+    100% {
+        transform: rotate(360deg);
+    }
+  }
+</style>
--- a/core/frontend/src/components/wifi/ConnectionDialog.vue
+++ b/core/frontend/src/components/wifi/ConnectionDialog.vue
@@ -1,0 +1,222 @@
+<template>
+  <v-dialog
+    width="300"
+    :value="show"
+    @input="showDialog"
+  >
+    <v-card>
+      <v-card-title>
+        {{ real_ssid }}
+
+        <v-spacer />
+
+        <v-btn
+          icon
+          @click="toggleInfoShow()"
+        >
+          <v-icon>{{ show_more_info_icon }}</v-icon>
+        </v-btn>
+      </v-card-title>
+
+      <v-expand-transition>
+        <div v-show="show_more_info">
+          <v-card-text>
+            <v-simple-table
+              dense
+            >
+              <template #default>
+                <tbody>
+                  <tr
+                    v-for="(value, name) in network"
+                    :key="name"
+                  >
+                    <td>{{ name }}</td>
+                    <td>{{ value }}</td>
+                  </tr>
+                </tbody>
+              </template>
+            </v-simple-table>
+          </v-card-text>
+        </div>
+      </v-expand-transition>
+
+      <v-card-actions class="d-flex flex-column">
+        <v-container>
+          <v-row>
+            <v-text-field
+              v-if="is_hidden"
+              v-model="inputed_ssid"
+              label="SSID"
+              hide-details="auto"
+              class="pa-2"
+            />
+          </v-row>
+
+          <v-row>
+            <v-card
+              v-if="!show_password_input_box"
+              elevation="0"
+              @click="toggleForcePassword()"
+            >
+              <v-card-text>
+                Force new password
+              </v-card-text>
+            </v-card>
+            <password-input
+              v-if="show_password_input_box"
+              v-model="password"
+              @submit="connectToWifiNetwork"
+            />
+          </v-row>
+
+          <v-row
+            class="d-flex justify-space-around py-4"
+          >
+            <v-btn
+              v-if="network.saved"
+              dark
+              color="red lighten-1"
+              depressed
+              @click="removeSavedWifiNetwork"
+            >
+              Forget
+            </v-btn>
+
+            <v-btn
+              dark
+              color="blue darken-1"
+              depressed
+              @click="connectToWifiNetwork"
+            >
+              Connect
+            </v-btn>
+          </v-row>
+        </v-container>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import axios from 'axios'
+import Vue, { PropType } from 'vue'
+import { getModule } from 'vuex-module-decorators'
+
+import NotificationsStore from '@/store/notifications'
+import WifiStore from '@/store/wifi'
+import { wifi_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { Network, NetworkCredentials } from '@/types/wifi.d'
+
+import PasswordInput from '../common/PasswordInput.vue'
+
+const notification_store: NotificationsStore = getModule(NotificationsStore)
+const wifi_store: WifiStore = getModule(WifiStore)
+
+export default Vue.extend({
+  name: 'ConnectionDialog',
+  components: {
+    PasswordInput,
+  },
+  model: {
+    prop: 'show',
+    event: 'change',
+  },
+  props: {
+    network: {
+      type: Object as PropType<Network>,
+      required: true,
+    },
+    show: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data() {
+    return {
+      password: '',
+      force_password: false,
+      inputed_ssid: 'HIDDEN SSID',
+      show_more_info: false,
+    }
+  },
+  computed: {
+    show_more_info_icon(): string {
+      return this.show_more_info ? 'mdi-chevron-up' : 'mdi-information'
+    },
+    real_ssid(): string {
+      return this.is_hidden ? this.inputed_ssid : this.network.ssid
+    },
+    is_hidden(): boolean {
+      return this.network.ssid === null || this.network.ssid === ''
+    },
+    show_password_input_box(): boolean {
+      if (this.force_password) {
+        return true
+      }
+      if (this.network.saved || !this.network.locked) {
+        return false
+      }
+      return true
+    },
+  },
+  methods: {
+    toggleInfoShow(): void {
+      this.show_more_info = !this.show_more_info
+    },
+    toggleForcePassword(): void {
+      this.force_password = !this.force_password
+    },
+    async connectToWifiNetwork(): Promise<void> {
+      const credentials: NetworkCredentials = { ssid: this.real_ssid, password: this.password }
+      await axios({
+        method: 'post',
+        url: `${wifi_store.API_URL}/connect`,
+        timeout: 2000,
+        data: credentials,
+        params: { hidden: this.is_hidden },
+      })
+        .then(() => {
+          wifi_store.setNetworkStatus(null)
+        })
+        .catch((error) => {
+          notification_store.pushNotification(new LiveNotification(
+            NotificationLevel.Error,
+            wifi_service,
+            'WIFI_CONNECT_FAIL',
+            `Could not connect to wifi network: ${error.message}.`,
+          ))
+        })
+        .finally(() => {
+          this.password = ''
+          this.showDialog(false)
+        })
+    },
+    async removeSavedWifiNetwork(): Promise<void> {
+      await axios({
+        method: 'post',
+        url: `${wifi_store.API_URL}/remove`,
+        timeout: 10000,
+        params: { ssid: this.network.ssid },
+      })
+        .catch((error) => {
+          notification_store.pushNotification(new LiveNotification(
+            NotificationLevel.Error,
+            wifi_service,
+            'WIFI_FORGET_FAIL',
+            `Could not remove saved wifi network: ${error.message}.`,
+          ))
+        })
+        .finally(() => {
+          this.showDialog(false)
+        })
+    },
+    showDialog(state: boolean) {
+      this.$emit('change', state)
+    },
+  },
+})
+</script>
+
+<style>
+</style>
--- a/core/frontend/src/components/wifi/DisconnectionDialog.vue
+++ b/core/frontend/src/components/wifi/DisconnectionDialog.vue
@@ -1,0 +1,141 @@
+<template>
+  <v-dialog
+    width="500"
+    :value="show"
+    @input="showDialog"
+  >
+    <v-card>
+      <v-card-title class="text-h5">
+        {{ network.ssid }}
+
+        <v-spacer />
+
+        <v-btn
+          icon
+          @click="toggleInfoShow()"
+        >
+          <v-icon>{{ show_more_info_icon }}</v-icon>
+        </v-btn>
+      </v-card-title>
+
+      <v-expand-transition>
+        <div v-show="show_more_info">
+          <v-card-text>
+            <v-simple-table
+              dense
+            >
+              <template #default>
+                <tbody>
+                  <tr
+                    v-for="(value, name) in connection_info"
+                    :key="name"
+                  >
+                    <td>{{ name }}</td>
+                    <td>{{ value }}</td>
+                  </tr>
+                </tbody>
+              </template>
+            </v-simple-table>
+          </v-card-text>
+        </div>
+      </v-expand-transition>
+
+      <v-card-actions class="d-flex flex-column py-6">
+        <v-btn
+          elevation="1"
+          dark
+          color="red darken-1"
+          @click="disconnectFromWifiNetwork"
+        >
+          Disconnect
+        </v-btn>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import axios from 'axios'
+import Vue, { PropType } from 'vue'
+import { getModule } from 'vuex-module-decorators'
+
+import NotificationStore from '@/store/notifications'
+import WifiStore from '@/store/wifi'
+import { wifi_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { Network, WifiStatus } from '@/types/wifi.d'
+
+const notification_store: NotificationStore = getModule(NotificationStore)
+const wifi_store: WifiStore = getModule(WifiStore)
+
+export default Vue.extend({
+  name: 'DisconnectionDialog',
+  model: {
+    prop: 'show',
+    event: 'change',
+  },
+  props: {
+    network: {
+      type: Object as PropType<Network>,
+      required: true,
+    },
+    status: {
+      type: Object as PropType<WifiStatus | null>,
+      required: true,
+    },
+    show: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data() {
+    return {
+      show_more_info: false,
+    }
+  },
+  computed: {
+    show_more_info_icon(): string {
+      return this.show_more_info ? 'mdi-chevron-up' : 'mdi-information'
+    },
+    connection_info(): Network | (Network & WifiStatus) {
+      if (this.status === null) {
+        return this.network
+      }
+      return { ...this.network, ...this.status }
+    },
+  },
+  methods: {
+    toggleInfoShow(): void {
+      this.show_more_info = !this.show_more_info
+    },
+    async disconnectFromWifiNetwork(): Promise<void> {
+      await axios({
+        method: 'get',
+        url: `${wifi_store.API_URL}/disconnect`,
+        timeout: 10000,
+      })
+        .then(() => {
+          wifi_store.setNetworkStatus(null)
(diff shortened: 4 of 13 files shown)


core/frontend/src/components/common/PasswordInput.vue
```diff
@@ -0,0 +1,43 @@
...
+  name: 'PasswordInput',
+  props: {
+    value: {
+      type: String,
+      default: '',
+    },
+  },
+  data () {
+    return {
+      show_password: '',
```
Williangalvani, line 28: why not a a boolean?

core/frontend/src/types/wifi.d.ts
```diff
@@ -0,0 +1,43 @@
...
+    wifi_generation: int
+    pairwise_cipher: string
+    group_cipher: string
+    key_mgmt: string
+    wpa_state: string
+    ip_address: string
+    p2p_device_address: string
+    address: string
+    uuid: string
+}
```
rafaellehmkuhl, line 23: number instead of int

core/frontend/src/types/wifi.d.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface SavedNetwork {
+    networkid: number
+    ssid: string
+    bssid: string
+    flags: string
+}
+
+export class NetworkCredentials {
```
rafaellehmkuhl, line 40: interface, not class

rafaellehmkuhl (commented):
Just found some typing errors. Tracking.

core/frontend/src/components/wifi/WifiUpdater.vue
```diff
@@ -0,0 +1,129 @@
...
+            index: ErrorIndex.WifiSavedFail,
+            message: `Could not fetch saved networks: ${error.message}.`,
+          })
+          wifi_store.setSavedNetworks([])
+        })
+    },
+    sleep (ms: number): Promise<void> {
+      return new Promise((resolve) => setTimeout(resolve, ms))
+    },
+    async updateNetworkStatus (): Promise<void> {
```
Williangalvani, line 112: are these loops that run every 3s? There must be a better way to do this. 

wouldn't 
```
while (true) {
     await this.doStuff()
     await this.sleep(3000)
}
``` 
yield the same result, but in a more obvious way?

core/frontend/src/components/wifi/WifiUpdater.vue
```diff
@@ -0,0 +1,129 @@
...
+      await this.fetchNetworkStatus()
+      await this.sleep(3000)
+      this.updateNetworkStatus()
+    },
+    async updateSavedNetworks (): Promise<void> {
```
Williangalvani, line 117: I don't think we need to constantly fetch the saved and available networks.

Williangalvani (changes requested):
This is looking good! I just did a broad review, I'll test it today.

core/frontend/src/types/wifi.d.ts
patrickelectric: Wait, `int` is not a valid type, how the tests are passing ??

core/frontend/src/store/common.ts
```diff
@@ -0,0 +1,44 @@
...
+  name: 'common_store',
+})
+
+@Module
+export default class CommonStore extends VuexModule {
+  errors: Error[] = []
+  ignored_errors_indexes: ErrorIndex[] = []
+
+  @Mutation
+  openError (error: Error): void {
```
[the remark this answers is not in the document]
rafaellehmkuhl: "add" seems to fit better, indeed, as "open" sounds like it will for sure open, which is not true if the user is ignoring this error.
Will change.

core/frontend/src/components/common/PasswordInput.vue
rafaellehmkuhl: No idea why this is a string. It was a boolean for sure. More strange is the fact that this works. Fixing.

core/frontend/src/components/wifi/WifiUpdater.vue
rafaellehmkuhl: Better, indeed. Changing.

core/frontend/src/components/wifi/WifiUpdater.vue
rafaellehmkuhl: Didn't get the idea, exactly. You mean to just fetch at boot?

core/frontend/src/components/common/ErrorDialog.vue
```diff
@@ -0,0 +1,74 @@
...
+
+const common_store: CommonStore = getModule(CommonStore)
+
+export default Vue.extend({
+  name: 'Error',
+  computed: {
+    error (): Error | null {
+      return common_store.current_error
+    },
+    error_message (): string {
```
patrickelectric, line 43: `error` already provides the abstract for message, we should rely on the components that already exist. 
An abstraction just to access a member of a function that already return sounds as not a good interface practice.

core/frontend/src/components/common/ErrorDialog.vue
```diff
@@ -0,0 +1,74 @@
...
+      return this.error === null ? '' : this.error.message
+    },
+    show_dialog: {
+      get (): boolean {
+        return this.error === null ? false : true
```
patrickelectric, line 48:
```suggestion
        return this.error !== null
```

core/frontend/src/components/common/ErrorDialog.vue
```diff
@@ -0,0 +1,74 @@
...
+      },
+      set (value: boolean): void {
```
patrickelectric, line 50: `true` is not supported ? If so this should not receive arguments.

core/frontend/src/types/wifi.d.ts
rafaellehmkuhl: It's strange, indeed. Any idea?

core/frontend/src/components/common/PasswordInput.vue
patrickelectric: How typescript failed to identify the wrong use of string here ??

core/frontend/src/components/common/SpinningLogo.vue
```diff
@@ -0,0 +1,34 @@
...
+
+export default Vue.extend({
+  name: 'SpinningLogo',
+})
+</script>
+
+<style>
+  .lds-dual-ring {
+    display: block;
+    width: 80px;
```
patrickelectric, line 19: avoid `px`, use `em` to support high definition displays.

core/frontend/src/components/wifi/WifiUpdater.vue
Williangalvani: I still haven't run it here, but I got the impression that this is always running.
we should have it scanning only when the wifi widget is open.

core/frontend/src/components/wifi/ConnectionDialog.vue
```diff
@@ -0,0 +1,178 @@
...
+      return this.is_hidden ? this.inputed_ssid : this.network.ssid
+    },
+    is_hidden (): boolean {
+      return this.network.ssid === null || this.network.ssid === ''
+    },
+    show_password_input_box (): boolean {
+      if (this.force_password) {
+        return true
+      }
+      else if (this.network.saved || !this.network.locked) {
```
patrickelectric, line 124: remove else after return

core/frontend/src/components/wifi/ConnectionDialog.vue
```diff
@@ -0,0 +1,178 @@
...
+      this.password = ''
+      this.emitChange(false)
+
+      const connect_response = await axios({
+        method: 'post',
+        url: `${wifi_store.API_URL}/connect`,
+        timeout: 500,
+        data: {ssid: this.real_ssid, password},
+        params: { hidden: this.is_hidden },
+      })
```
patrickelectric, line 142: why not use `then/catch`  for better readability?

core/frontend/src/types/wifi.d.ts
Williangalvani: we should add a test to test the tests :eyes:

core/frontend/src/components/wifi/ConnectionDialog.vue
```diff
@@ -0,0 +1,178 @@
...
+        .then(() => {
+          common_store.closeError(ErrorIndex.WifiForgetFail)
+        })
+        .catch((error) => {
+          common_store.openError({
+            index: ErrorIndex.WifiForgetFail,
+            message: `Could not remove saved wifi network: ${error.message}.`})
+        })
+    },
+    emitChange (state: boolean) {
```
patrickelectric, line 170: What changed ? The function should have a better name

core/frontend/src/components/wifi/ConnectionDialog.vue
```diff
@@ -0,0 +1,178 @@
...
+        params: { hidden: this.is_hidden },
+      })
+
+      if (connect_response.status != 200) {
+        common_store.openError({index: ErrorIndex.WifiConnectFail, message: 'Could not connect to wifi network.'})
+        return
+      }
+      common_store.closeError(ErrorIndex.WifiConnectFail)
+
+      wifi_store.setNetworkStatus(null)
```
patrickelectric, line 150: What is a null state ? This should be more clear.

core/frontend/src/components/wifi/NetworkCard.vue
```diff
@@ -0,0 +1,84 @@
...
+      // | -70 dBm         | Okay      | Minimum signal strength for reliable packet delivery.                                                                                      |
+      // | -80 dBm         | Not Good  | Minimum signal strength for basic connectivity. Packet delivery may be unreliable.                                                         |
+      // | -90 dBm         | Unusable  | Approaching or drowning in the noise floor. Any functionality is highly unlikely.                                                           |
+      // Reference: metageek.com/training/resources/wifi-signal-strength-basics.html
+      /* eslint-enable */
+
+      if (this.network.signal >= -67) return 'mdi-wifi-strength-4'
+      if (this.network.signal >= -70) return 'mdi-wifi-strength-3'
+      if (this.network.signal >= -80) return 'mdi-wifi-strength-2'
+      return 'mdi-wifi-strength-1'
```
patrickelectric, line 74: since this is "unusable", we should use `mdi-wifi-strength-outline`, the others we could do -1 and also include >= 30 with strength-4

core/frontend/src/components/wifi/WifiManager.vue
```diff
@@ -0,0 +1,122 @@
...
+  data () {
+    return {
+      selected_network: null as Network | null,
+      show_connection_dialog: false,
+      show_disconnection_dialog: false,
+    }
+  },
+  computed: {
+    are_networks_available (): boolean {
+      return wifi_store.available_networks.length === 0 ? false : true
```
patrickelectric, line 76: As general tip, if you are checking for true/false, you don't need to return true or false, this is a bad practice.

core/frontend/src/components/wifi/WifiUpdater.vue
```diff
@@ -0,0 +1,129 @@
...
+          )
+          const saved_network = wifi_store.saved_networks.find(
+            (network) => network.ssid === response.data.ssid,
+          )
+
+          wifi_store.setCurrentNetwork({
+            ssid: response.data.ssid,
+            signal: scanned_network ? scanned_network.signal : 0,
+            locked: response.data.key_mgmt.includes('WPA'),
+            saved: saved_network ? true : false,
```
patrickelectric, line 52: As said.

core/frontend/src/components/wifi/WifiUpdater.vue
```diff
@@ -0,0 +1,129 @@
...
+              locked: network.flags.includes('WPA'),
+              saved: saved_networks_ssids.includes(network.ssid),
+            }
+          })
+          wifi_store.setAvailableNetworks(available_networks)
+        })
+        .catch((error) => {
+          common_store.openError({
+            index: ErrorIndex.WifiScanFail,
+            message: `Could not scan wifi networks: ${error.message}`,
```
patrickelectric, line 86: `Could not scan for Wi-Fi networks`

core/frontend/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
+export interface Error {
+    index: ErrorIndex
+    message: string
+}
+
+export enum ErrorIndex {
+    WifiForgetFail = 'wifi_forget_fail',
```
patrickelectric, line 7: Why the strings are necessary for this enum ?

core/frontend/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
+export interface Error {
```
patrickelectric, line 1: This should be in an Error file and not in a common.ts

core/frontend/src/store/common.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (this.ignored_errors_indexes.includes(error.index)) {
+      console.log(`Errors of type ${error.index} are being ignored. Not opening error dialog.`)
```
patrickelectric, line 19: console.warning ?

core/frontend/src/store/common.ts
```diff
@@ -0,0 +1,44 @@
...
+      return
+    }
+    if (this.errors.includes(error)) {
+      console.log(`Error of type ${error.index} is already open.`)
```
patrickelectric, line 23: console.error ?

patrickelectric (changes requested):

core/frontend/src/types/wifi.d.ts
patrickelectric: This a must thing to fix and do it asap, if the typescript checks are not working, something is really wrong with the project.

core/frontend/src/components/wifi/WifiUpdater.vue
rafaellehmkuhl: It's an option, for sure.
Was like that before, in fact, but I changed.

core/frontend/src/components/common/ErrorDialog.vue
rafaellehmkuhl: Will check on using it.

core/frontend/src/components/common/ErrorDialog.vue
rafaellehmkuhl: Didn't get it. get/set for are an existing interface that I'm implementing.

patrickelectric:
I'm receiving error messages every time that I open the widget 
image
image

patrickelectric:
Wifi appears as saved but asks for password again
image

patrickelectric:
The wifi popup appears with a unnecessary scrollbar and unnecessary nerd information, we could add it as an "i" information icon or dropdown.

image

rafaellehmkuhl:
> I'm receiving error messages every time that I open the widget
> image
> image
> image

Could you send a print of the network tab on the inspector?
Are you receiving this constantly? From the start or after some time?

rafaellehmkuhl:
> Wifi appears as saved but asks for password again
> image

You're sure you didn't click the "force new password" button?
Could you post here the JSON that comes from both 'scan' and 'saved'?

rafaellehmkuhl:
> The wifi popup appears with a unnecessary scrollbar and unnecessary nerd information, we could add it as an "i" information icon or dropdown.
> 
> image

Sounds like a good idea.

patrickelectric:
> > I'm receiving error messages every time that I open the widget
> > image
> > image
> > image
> 
> Could you send a print of the network tab on the inspector?
> Are you receiving this constantly? From the start or after some time?

When I click in connect:
image
When I click to connect in another network:
image

patrickelectric:
> > Wifi appears as saved but asks for password again
> > image
> 
> You're sure you didn't click the "force new password" button?
> Could you post here the JSON that comes from both 'scan' and 'saved'?

saved:
image


scan:
image

rafaellehmkuhl:
> > > I'm receiving error messages every time that I open the widget
> > > image
> > > image
> > > image
> > 
> > 
> > Could you send a print of the network tab on the inspector?
> > Are you receiving this constantly? From the start or after some time?
> 
> When I click in connect:
> image
> When I click to connect in another network:
> image

Cool! That really helps!
It could be that when connecting the WPA Supplicant gets busier and delivers the other results with more delay than I was expecting.
I will try to reproduce it here to see if that's just a matter of changing the timeouts or if we could change the logic to improve that.

rafaellehmkuhl:
> > > Wifi appears as saved but asks for password again
> > > image
> > 
> > 
> > You're sure you didn't click the "force new password" button?
> > Could you post here the JSON that comes from both 'scan' and 'saved'?
> 
> saved:
> image
> 
> scan:
> image

This one I'm not understanding why it happens. 
Does it stays like this or it's a momentary problem?
Do the other saved networks have the same behavior? I tried with some of them here and couldn't reproduce.

core/frontend/src/types/wifi.d.ts
rafaellehmkuhl: Apparently what's happening is that the 'int' was casted to 'unknown', thus accepting literally anything.
[There's a rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) for setting that.I will take a better look to see if it's the responsible for this behavior.

core/frontend/src/components/common/PasswordInput.vue
rafaellehmkuhl: What happened what that, when using `show_password ?`, the empty string was being cast to a boolean, returning `false`, and on the first call to `show_password = !show_password`, it was turned into a *de facto* boolean. All the logic works fine and the operations are valid.

The problem I see is that the Typescript linter was unable to see that a variable was having its type altered on runtime. This happened because this toggle operation happened inside the `<template>`. When I move the operation to a method, inside the `<script>` tag, the linter is able to catch the type-change mistake and forces me to declare it initially as a boolean.

This enforces that we should avoid any operation inside the template, and use `computed` and `methods` whenever possible.

core/frontend/src/components/common/SpinningLogo.vue
rafaellehmkuhl: Using `%` instead, as it's the recommended way to scale relative to parent elements. `em` are font-size dependent, so it wouldn't be the best choice here.

core/frontend/src/components/wifi/ConnectionDialog.vue
rafaellehmkuhl: Forgot to change this one.

core/frontend/src/components/common/SpinningLogo.vue
patrickelectric: `em` are not font-size dependent, they are relative to resolution, as fonts.

> Pixel is a static measurement, while percent and EM are relative measurements. The size of an EM or percent depends on its parent. If the text size of body is 16 pixels, then 150% or 1.5 EM will be 24 pixels (1.5 * 16).

rafaellehmkuhl: https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units

Relative length units section.

core/frontend/src/types/common.ts
rafaellehmkuhl: They are not. Removing.

core/frontend/src/components/wifi/ConnectionDialog.vue
rafaellehmkuhl: It's a `null` only when there's no status information available, as in the case of API request errors.

core/frontend/src/components/wifi/NetworkCard.vue
rafaellehmkuhl: Liked the idea for the outline, but got confused with the other one. Are you suggesting this?

```javascript
if (this.network.signal >= -30) return 'mdi-wifi-strength-4'
if (this.network.signal >= -67) return 'mdi-wifi-strength-3'
if (this.network.signal >= -70) return 'mdi-wifi-strength-2'
if (this.network.signal >= -80) return 'mdi-wifi-strength-1'
return 'mdi-wifi-strength-outline'
```

core/frontend/src/components/wifi/WifiManager.vue
rafaellehmkuhl: True. I usually write the whole phrase as a sanity-check and fix it after, but sometimes I forgot to fix rs

core/frontend/src/components/wifi/WifiUpdater.vue
rafaellehmkuhl: Will, in fact this approach will lock javascript in the while block. Do you know some better alternative considering the async/await approach?

rafaellehmkuhl:
@patrickelectric can you test if you're still experiencing issues regarding the timeouts with the new image?

core/frontend/src/components/wifi/WifiUpdater.vue
```diff
@@ -0,0 +1,120 @@
...
+    },
+    async startNetworkStatusUpdateLoop (): Promise<void> {
+      await this.fetchNetworkStatus()
+      await this.sleep(5000)
+      this.startNetworkStatusUpdateLoop()
+    },
+    async startNetworksScanLoop (): Promise<void> {
+      await this.fetchAvailableNetworks()
+      await this.sleep(5000)
+      this.startNetworksScanLoop()
```
patrickelectric, line 116: We could abstract this in a better way.  maybe a periodicRequest function ?

patrickelectric (changes requested):
there is still some problems related to the behavior, we did talk via chat.

core/frontend/src/components/wifi/WifiUpdater.vue
rafaellehmkuhl: Good idea!

core/frontend/src/components/wifi/DisconnectionDialog.vue
```diff
@@ -0,0 +1,127 @@
...
+    show: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  computed: {
+    show_more_info_icon (): string {
+      return this.show_more_info ? 'mdi-chevron-up' : 'mdi-chevron-down'
+    },
+    connection_info (): Network | Network & WifiStatus {
```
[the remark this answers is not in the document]
rafaellehmkuhl: Although I understand the suggestion, the problem is that some properties like the SSID are duplicated on the two. That's why I merged then in this way. 

The type is very strict check as it is. There's only the possibility of returning a Network object alone or a merged Network+WifiStatus object. If I add or remove any property, Typescript blames me.

For clarity of syntax I've added brackets around the `Network & WifiStatus`.

core/frontend/src/components/wifi/ConnectionDialog.vue
```diff
@@ -0,0 +1,199 @@
...
+      const password = this.password
+      this.password = ''
+      this.showDialog(false)
+
+      await axios({
+        method: 'post',
+        url: `${wifi_store.API_URL}/connect`,
+        timeout: 2000,
+        data: {ssid: this.real_ssid, password},
+        params: { hidden: this.is_hidden },
```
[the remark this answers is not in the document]
patrickelectric: Just to be sure, this was fixed manually or via eslint ?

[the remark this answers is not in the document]
rafaellehmkuhl: Let me check.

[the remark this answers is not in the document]
patrickelectric: Let me rephrase that, was eslint able to detect the style problem ?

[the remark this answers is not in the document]
rafaellehmkuhl: If you're talking about the spaces around the curly braces, this is being done automatically. Just tried to make it bad and eslint doesn't allow.

rafaellehmkuhl:
I'm building a new image for you guys to test. Should be ready in 10 minutes.

core/frontend/src/components/wifi/ConnectionDialog.vue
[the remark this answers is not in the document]
rafaellehmkuhl: > Let me rephrase that, was eslint able to detect the style problem ?

Yes.

rafaellehmkuhl:
> I'm building a new image for you guys to test. Should be ready in 10 minutes.

Already available at `rafaellehmkuhl/companion-core:wifi-frontend`.

patrickelectric (approved):

PR #396, opened 2021-08-04
Add notification system

rafaellehmkuhl:
Allows services to push notifications and add a notification manager to the top-left corner.

Image available at: `rafaellehmkuhl/companion-core:notifications`

notifications

### DIFF
--- a/core/frontend/src/components/notifications/ConfigMenu.vue
+++ b/core/frontend/src/components/notifications/ConfigMenu.vue
@@ -1,0 +1,38 @@
+<template>
+  <v-card
+    elevation="1"
+    width="250"
+  >
+    <v-list>
+      <v-list-item>
+        <v-switch
+          :value="showOld"
+          color="blue lighten-1"
+          @change="emitShowOldChange"
+        />
+        <div>Show old messages</div>
+      </v-list-item>
+    </v-list>
+  </v-card>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+
+export default Vue.extend({
+  name: 'NotificationsConfigMenu',
+  props: {
+    showOld: {
+      type: Boolean,
+    },
+  },
+  methods: {
+    emitShowOldChange(value: boolean): void {
+      this.$emit('update:show-old', value)
+    },
+  },
+})
+</script>
+
+<style>
+</style>
--- a/core/frontend/src/components/notifications/NotificationCard.vue
+++ b/core/frontend/src/components/notifications/NotificationCard.vue
@@ -1,0 +1,97 @@
+<template>
+  <v-lazy
+    :options="{threshold: .5}"
+    transition="slide-x-transition"
+  >
+    <v-card elevation="0">
+      <v-list-item>
+        <v-list-item-icon>
+          <v-icon :color="level_color">
+            {{ level_icon }}
+          </v-icon>
+        </v-list-item-icon>
+
+        <v-list-item-content>
+          <v-list-item-title>{{ notification.service.name }}</v-list-item-title>
+          <v-list-item-subtitle>{{ notification.type }} - {{ notification.message }}</v-list-item-subtitle>
+          <v-list-item-subtitle>{{ time_since }} ({{ count }} times)</v-list-item-subtitle>
+        </v-list-item-content>
+
+        <v-list-item-action>
+          <v-btn icon>
+            <v-icon
+              color="grey lighten-1"
+              size="medium"
+              @click="dismiss"
+            >
+              mdi-close
+            </v-icon>
+          </v-btn>
+        </v-list-item-action>
+      </v-list-item>
+      <v-divider />
+    </v-card>
+  </v-lazy>
+</template>
+
+<script lang="ts">
+import { formatDistance } from 'date-fns'
+import Vue, { PropType } from 'vue'
+
+import { Notification, NotificationLevel } from '@/types/notifications'
+
+export default Vue.extend({
+  name: 'NotificationCard',
+  props: {
+    count: {
+      type: Number,
+      required: true,
+    },
+    notification: {
+      type: Object as PropType<Notification>,
+      required: true,
+    },
+    timestamp: {
+      type: Date,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      updated_time: new Date(),
+    }
+  },
+  computed: {
+    time_since(): string {
+      const date_now = new Date(this.timestamp)
+      const date_notification = new Date(this.notification.time_created)
+      return `${formatDistance(date_now, date_notification)} ago`
+    },
+    level_icon(): string {
+      switch (this.notification.level) {
+        case NotificationLevel.Success: return 'mdi-check-bold'
+        case NotificationLevel.Error: return 'mdi-close-circle'
+        case NotificationLevel.Info: return 'mdi-information'
+        case NotificationLevel.Warning: return 'mdi-alert'
+        case NotificationLevel.Critical: return 'mdi-skull'
+        default: return 'mdi-help-circle'
+      }
+    },
+    level_color(): string {
+      switch (this.notification.level) {
+        case NotificationLevel.Success: return 'green accent-4'
+        case NotificationLevel.Error: return 'red darken-2'
+        case NotificationLevel.Info: return 'indigo darken-1'
+        case NotificationLevel.Warning: return 'yellow lighten-1'
+        case NotificationLevel.Critical: return 'black'
+        default: return 'mdi-help-circle'
+      }
+    },
+  },
+  methods: {
+    dismiss(): void {
+      this.$emit('dismiss', this.notification)
+    },
+  },
+})
+</script>
--- a/core/frontend/src/components/notifications/NotificationManager.vue
+++ b/core/frontend/src/components/notifications/NotificationManager.vue
@@ -1,0 +1,145 @@
+<template>
+  <v-card width="500">
+    <v-app-bar
+      elevate-on-scroll
+      scroll-target="#notifications-list"
+    >
+      <v-toolbar-title>Notifications</v-toolbar-title>
+      <v-spacer />
+      <v-btn
+        icon
+        @click="dismissAllNotifications()"
+      >
+        <v-icon>mdi-broom</v-icon>
+      </v-btn>
+      <v-menu
+        v-model="show_config_menu"
+        :close-on-content-click="false"
+        :nudge-x="400"
+        offset-y
+      >
+        <template #activator="{on, attrs}">
+          <v-btn
+            icon
+            v-bind="attrs"
+            v-on="on"
+          >
+            <v-icon>mdi-cog</v-icon>
+          </v-btn>
+        </template>
+        <notifications-config-menu
+          :show-old.sync="show_old_messages"
+        />
+      </v-menu>
+    </v-app-bar>
+
+    <v-sheet
+      id="notifications-list"
+      class="overflow-y-auto"
+    >
+      <v-list
+        three-line
+        dense
+        class="overflow-y-auto"
+        tile
+        max-height="50vh"
+      >
+        <v-list-item-group
+          v-if="notifications_to_show.length != 0"
+        >
+          <notification-card
+            v-for="({notification, count}, index) in notifications_to_show"
+            :key="index"
+            :notification="notification"
+            :count="count"
+            :timestamp="timestamp"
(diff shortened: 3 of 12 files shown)


core/frontend/src/components/notifications/ConfigMenu.vue
```diff
@@ -0,0 +1,38 @@
...
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+
+export default Vue.extend({
+  name: 'NotificationsConfigMenu',
+  props: {
+    showOld: {
+      type: Boolean,
```
patrickelectric, line 26: How eslint is not detecting this ?? boolean should be not capitalized.

core/frontend/src/components/notifications/NotificationCard.vue
```diff
@@ -0,0 +1,96 @@
...
+        <v-list-item-icon>
+          <v-icon :color="level_color">
+            {{ level_icon }}
+          </v-icon>
+        </v-list-item-icon>
+
+        <v-list-item-content>
+          <v-list-item-title> {{ notification.service.name }} </v-list-item-title>
+          <v-list-item-subtitle> {{ notification.type }} - {{ notification.message }} </v-list-item-subtitle>
+          <v-list-item-subtitle>{{ time_since }} ({{ this.count }} times)</v-list-item-subtitle>
```
patrickelectric, line 17: The space between < and {{ is not consistent, can't we have eslint to detect that ?

core/frontend/src/components/notifications/NotificationManager.vue
```diff
@@ -0,0 +1,134 @@
...
+    notifications (): CumulatedNotification[] {
+      return notification_store.active_cumulated_notifications
+    },
+    recent_notifications (): CumulatedNotification[] {
+      return this.notifications.filter((cumulated) => {
+        return cumulated.notification.secondsTill(new Date()) < this.seconds_recent
+      })
+    },
+    notifications_to_show (): CumulatedNotification[] {
+      const notifications_to_show = this.show_old_messages? this.notifications : this.recent_notifications
```
patrickelectric, line 110: inconsistent style of spaces between `?`, where is eslint ?

core/frontend/src/components/notifications/NotificationManager.vue
```diff
@@ -0,0 +1,134 @@
...
+      return notifications_to_show.slice().reverse()
```
patrickelectric, line 111: why .slice.reverse ? Please add a comment

core/frontend/src/components/notifications/NotificationManager.vue
```diff
@@ -0,0 +1,134 @@
...
+    },
+  },
+  methods: {
+    showOldMessages (value: boolean) {
+      this.show_old_messages = value
+    },
+    dismissAllNotifications (): void {
```
patrickelectric, line 118: Why there are functions that return avoid and functions that does not ? is eslint not working ?

core/frontend/src/types/services.ts
```diff
@@ -0,0 +1,8 @@
+import { Service } from './common'
```
patrickelectric, line 1: maybe frontend_services as filename ?

core/frontend/src/types/services.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export const service_scanner_service: Service = {
+  name: 'Service Scanner',
+  description: 'Actual scanner for running services.',
+  company: 'Blue Robotics',
+  version: '0.1.0',
```
patrickelectric, line 7: How this number will be updated and how we are going to track it ?

core/frontend/src/types/notifications.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  /** Returns the time in seconds between time_created and given Date object */
+  secondsTill (date: Date): number {
+    const notification_time = new Date(this.time_created)
+    const time_diff = date.getTime() - notification_time.getTime()
+    return Math.floor(time_diff/1e3)
+  }
+
+  /** Compare if all properties (except time_created) between this and other match */
+  similarTo (other: Notification): boolean {
```
patrickelectric, line 53: `isEqual` maybe ? This make clear that will return a boolean

core/frontend/src/types/notifications.ts
```diff
@@ -0,0 +1,77 @@
...
+    const { time_created: t_this, ...similar_this } = this // eslint-disable-line @typescript-eslint/no-unused-vars
+    const { time_created: t_other, ...similar_other } = other // eslint-disable-line @typescript-eslint/no-unused-vars
+    return JSON.stringify(similar_this) === JSON.stringify(similar_other)
```
patrickelectric, line 56: use hash

core/frontend/src/types/notifications.ts
```diff
@@ -0,0 +1,77 @@
...
+  /** Returns a stringfied version of the object to be used as hash */
+  hash (): string {
+    return JSON.stringify(this)
+  }
+
+  /** Returns the time in seconds between time_created and given Date object */
+  secondsTill (date: Date): number {
+    const notification_time = new Date(this.time_created)
+    const time_diff = date.getTime() - notification_time.getTime()
+    return Math.floor(time_diff/1e3)
```
patrickelectric, line 49: What is the problem with [getSeconds](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Date/getSeconds) ?

core/frontend/src/store/notifications.ts
```diff
@@ -0,0 +1,53 @@
...
+
+@Module
+export default class NotificationStore extends VuexModule {
+  notifications: Notification[] = []
+  cumulated_notifications: CumulatedNotification[] = []
+  dismissed_notifications_hashes: string[] = []
+
+  @Mutation
+  pushNotification (notification: Notification): void {
+    this.notifications.push(notification)
```
patrickelectric, line 20: Can we use console here to have also the notification/message log in the browser console and not only in the notification web ui ?

patrickelectric (changes requested):

core/frontend/src/components/notifications/ConfigMenu.vue
rafaellehmkuhl: It's not that. Read [this](https://vuejs.org/v2/guide/components-props.html#Prop-Types).

core/frontend/src/types/services.ts
rafaellehmkuhl: This comes from the backend Helper service.

core/frontend/src/types/notifications.ts
rafaellehmkuhl: Will change to `isSimilarTo`. Equal would not be a good fit here, since they are not exactly equal (their timestamps are different).

core/frontend/src/types/notifications.ts
rafaellehmkuhl: Simply didn't know it existed. Thanks.

core/frontend/src/store/notifications.ts
rafaellehmkuhl: Will add.

core/frontend/src/components/notifications/NotificationCard.vue
rafaellehmkuhl: I couldn't found a rule to enforce for that :/

core/frontend/src/components/notifications/NotificationManager.vue
rafaellehmkuhl: Fixed with the last eslint update.

core/frontend/src/types/notifications.ts
rafaellehmkuhl: As they are not Notification instances anymore, this is not possible.

core/frontend/src/components/notifications/NotificationManager.vue
```diff
@@ -0,0 +1,135 @@
...
+  methods: {
+    showOldMessages(value: boolean): void {
+      this.show_old_messages = value
+    },
+    dismissAllNotifications(): void {
+      notification_store.notifications.forEach((notification) => {
+        this.dismissNotification(notification)
+      })
+    },
+    dismissNotification(notification: Notification): void {
```
patrickelectric, line 124: I'm not a huge fun to have explicit return type for void, @Williangalvani ?

rafaellehmkuhl: And it was in fact [rejected](https://github.com/typescript-eslint/typescript-eslint/issues/50) (with really bad arguments cof cof).

patrickelectric: ouch..

core/frontend/src/store/notifications.ts
```diff
@@ -0,0 +1,74 @@
...
+  name: 'notification_store',
+})
+
+@Module
+export default class NotificationStore extends VuexModule {
+  notifications: Notification[] = []
+
+  cumulated_notifications: CumulatedNotification[] = []
+
+  dismissed_notifications_hashes: string[] = []
```
Williangalvani, line 17: while I don't think it will matter in this application, this should probably be a set instead. it is more efficient in checking if an element is present or not.

core/frontend/src/components/notifications/NotificationCard.vue
```diff
@@ -0,0 +1,100 @@
...
+  },
+  data() {
+    return {
+      updated_time: new Date(),
+    }
+  },
+  computed: {
+    time_since(): string {
+      const seconds_passed = this.notification.secondsTill(new Date(this.timestamp))
+      if (seconds_passed < 10) {
```
Williangalvani, line 65: check https://github.com/bluerobotics/companion-docker/blob/master/core/services/versionchooser/setup.py#L18

core/frontend/src/components/notifications/NotificationManager.vue
```diff
@@ -0,0 +1,135 @@
...
+      )
+    },
+    notifications_to_show(): CumulatedNotification[] {
+      const notifications_to_show = this.show_old_messages ? this.notifications : this.recent_notifications
+      // Returning the notifications reversed to show the last ones first
+      return notifications_to_show.slice().reverse()
+    },
+  },
+  mounted() {
+    setInterval(this.updateTimestamp, 1000)
```
Williangalvani, line 113: there's no `clearInterval`anywhere.

Williangalvani (changes requested):
It's looking good, but I still haven't tested it.

Can you ping me once there's an image up with the versionchooser fixed?

core/frontend/src/store/notifications.ts
rafaellehmkuhl: Didn't know javascript had sets. Will change.

core/frontend/src/components/notifications/NotificationManager.vue
patrickelectric: @Williangalvani, since this component will only exist in our top bar and will only exist one, does it still have a problem ?
That does not invalidate the point that we should add a RAII here to avoid problem.

Williangalvani: I thought of that. it isn't an issue right now, it is just a matter of future-proofing things.

core/frontend/src/store/notifications.ts
rafaellehmkuhl: For some reason I don't understand yet Vuex is not detecting changes to Set instances, thus not updating the DOM automatically.

I will keep using arrays for now and try to do some research on that.

core/frontend/src/store/notifications.ts
```diff
@@ -0,0 +1,72 @@
...
+        console.info(notification.message)
+        break
+      case NotificationLevel.Warning:
+        console.warn(notification.message)
+        break
+      case NotificationLevel.Critical:
+        console.exception(notification.message)
+        break
+      default:
+        break
```
patrickelectric, line 41: Can we add a message here when a invalid or not exhaustive switch case happen ?
E.g: A new notification level was added but we have not updated this part of the code

patrickelectric (approved):
Just added a minor request, the PR should also be rebased to remove the fixups

core/frontend/src/store/notifications.ts
rafaellehmkuhl: Good one! Adding.

Williangalvani (approved):

PR #412, opened 2021-08-13
Add ethernet manager to main frontend

rafaellehmkuhl:
Docker image available at rafaellehmkuhl/companion-core:ethernet-frontend


core/frontend/src/utils/helper_functions.ts
```diff
@@ -0,0 +1,18 @@
...
+/* Call a given function periodically for a given interval. */
+/**
+ * @param func - Function to be called.
+ * @param interval - Time in milliseconds to wait after the previous request is done
+* */
+export async function callPeriodically(func: () => Promise<void>, interval: number): Promise<void> {
+  await func()
+  await sleep(interval)
+  callPeriodically(func, interval)
+}
```
patrickelectric, line 18: That's a bit dangerous, how can we turn off the `callPeriodically` if necessary ?

patrickelectric (changes requested):

core/frontend/src/utils/helper_functions.ts
rafaellehmkuhl: Let me give a double-check, but as it's a set-interval I think it will die for itself when the component is mounted.

Is this your question or your concerned about something else?

patrickelectric: I'm concerned about:
```
let potato = {
    a: async () => console.log("Potato")
}

callPeriodically(potato.a, 1000)

# somehow potato is dead
```

What happens if the object that callPeriodically is using dies or the variables used inside potato.a dies ?

patrickelectric:
We should filter virtual interfaces and not display them in the interface, virtual interfaces start with `veth` prefix.

rafaellehmkuhl:
> 
> 
> We should filter virtual interfaces and not display them in the frontend, virtual interfaces start with `veth` prefix.

Definitely. Should we filter them on the backend or here? I'm not sure if they have any value on the ethernet backend, so maybe we should filter them out there.

rafaellehmkuhl:
Image available at `rafaellehmkuhl/companion-core:ethernet-frontend`

core/services/cable_guy/api/manager.py
```diff
@@ -302,6 +302,10 @@ def get_interfaces(self) -> List[EthernetInterface]:
         """
         result = []
         for interface, addresses in psutil.net_if_addrs().items():
+            # We don't care about virtual ethernet interfaces
+            if str(interface).startswith("veth"):
+                break
```
patrickelectric, line 307: why break ?
This should be a continue.

patrickelectric (changes requested):

core/frontend/src/utils/helper_functions.ts
patrickelectric: ping @Williangalvani

core/services/cable_guy/api/manager.py
rafaellehmkuhl: Oh god. Glad you saw this haha

core/frontend/src/utils/helper_functions.ts
rafaellehmkuhl: Opened #425 to discuss that as it was already merged with #391

patrickelectric (changes requested):
The frontend is saying that is running as static, but it's running as dynamic.
Is also only showing a single IP address and not both.

rafaellehmkuhl:
> The frontend is saying that is running as static, but it's running as dynamic.
> Is also only showing a single IP address and not both.
> 
> 

Unfortunately we don't have support on the backend for more than one IP on a given interface yet.
I've opened #428 to track that.
As it's unrelated to this PR, I would suggest we approve it. What do you think?

core/services/cable_guy/api/manager.py
```diff
@@ -302,6 +302,10 @@ def get_interfaces(self) -> List[EthernetInterface]:
         """
         result = []
         for interface, addresses in psutil.net_if_addrs().items():
+            # We don't care about virtual ethernet interfaces
+            if str(interface).startswith("veth"):
```
patrickelectric, line 306: The comment is ok, but it does not explain the reason of 'why', and that may be not clear to someone reading the code
```suggestion
            # Ignore virtual ethernet interfaces
            ## Virtual interfaces are created by programs such as docker
            ## and they are an abstract of real interfaces, the ones that we want to configure
```

patrickelectric (approved):
Just a small comment, but overall it looks great.
There is #428 and I believe that we should deal with that as soon this PR get merged. 
I have check the code and basic functionalities. I recommend someone else also to check before merging it.

core/services/cable_guy/api/manager.py
rafaellehmkuhl: Better. Changing.

Issue #418, opened 2021-08-18
propose: Change in Companion's default communication behavior


rafaellehmkuhl:
We are currently using a hardcoded UDP client endpoint (`192.168.2.1`) for the top machine on companion-docker. This is the same behavior we have on companion0. There are some problems with this approach. Two are the main ones:
- We need the user to configure it's top machine with a static IP
- With `mavlink-router`, if there's no UDP server online with this hardcoded IP on the local network, it will fail to route all mavlink communication

My proposal for a new behavior is the following:
- Remove the hardcoded IP from companion
- Make companion start as a DHCP server by default if it hasn't received a DHCP handshake from an external route. This allows someone to just connect companion with an ethernet cable to it's top computer, which will receive a dynamic IP and enter companion's network
- Make companion start a Wifi hotspot with a DHCP server whenever there's no known network around for more than 10 seconds. This allows the same behavior as the ethernet DHCP server, but with wifi.

rafaellehmkuhl:
I really, really dislike needing to set up the connection manually in QGC.

_Originally posted by @Williangalvani in https://github.com/bluerobotics/companion-docker/issues/419#issuecomment-902089507_

Another possibility is to do the following:
1. Ask the user to add an UDP client endpoint on companion pointing to the top machine's IP
2.a. Add some logic on companion to check if the IP exists on the network, and if not, disable this endpoint
2.b. Work on `mavlink-router` to prevent that behavior

patrickelectric:
We should also think about the automatic connection / discovery of the companion computer via the GCS. 
As reference: https://discuss.px4.io/t/qgc-companion-computer-integration-idea/22831

Williangalvani:
We should take some time to think about this, considering multiple clients, boats, and remote clients as well...

So.. some loose requirements, it should:
 - [ ] be as plug and play as possible
 - [ ] support additional clients connecting with no further setup
 - [ ] support  multiple links/redundancy (not sure how this works) for boats.
 - [ ] support remote links for boat, preferably including both telemetry and video.
 
 
I would make the pi a dhcp by default, so there's no need to setup the fxti.

Then we should make it possible to find all qgc instances in the network, and just send the streams to these instances. Ideally, mDNS would work. other options are broadcasting like the Ping360, or even rudely scanning for open ports.

For remote access, I vote we add a server so we don't need to deal with NAT issues. 
I think redundancy is somewhat embeded in QGC, but it needs testing.

[...]

PR #419, opened 2021-08-18
core: ardupilot-manager: Make default GCS endpoint an UDP server

rafaellehmkuhl:
First step to solve #418 

Image available at `rafaellehmkuhl/companion-core:gcs-endpoint`

### DIFF
--- a/core/frontend/src/views/MainView.vue
+++ b/core/frontend/src/views/MainView.vue
@@ -1,5 +1,21 @@
 <template>
   <v-container>
+    <v-bottom-sheet
+      :value="true"
+      inset
+    >
+      <v-card>
+        <v-card-title>Mavlink connection change</v-card-title>
+        <v-card-text>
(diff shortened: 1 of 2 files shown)


patrickelectric (approved):
I believe that this is the correct approach.

Williangalvani:
I really, really dislike needing to set up the connection manually in QGC.

patrickelectric:
> I really, really dislike needing to set up the connection manually in QGC.

Can we make it temporary until we sort the companion GCS integration ?
The idea is to make QGC find companion in the network.

rafaellehmkuhl:
I've added to this discussion on #418 @Williangalvani

Williangalvani (approved):
sounds good for now. we should add a note in the frontend to notify users they need to add a new connection in QGC

patrickelectric:
Maybe a popup ?

patrickelectric (approved):

rafaellehmkuhl:
New alert:

image

PR #439, opened 2021-08-27
Firmware on the frontend

rafaellehmkuhl:
Image available at `rafaellehmkuhl/companion-core:firmware-frontend`.

### DIFF
--- a/core/frontend/src/components/autopilot/FirmwareManager.vue
+++ b/core/frontend/src/components/autopilot/FirmwareManager.vue
@@ -1,0 +1,294 @@
+<template>
+  <v-card
+    elevation="0"
+    class="mb-12 pb-12 text-center"
+  >
+    <v-item-group
+      v-model="upload_type"
+      mandatory
+      class="ma-6"
+    >
+      <v-item
+        v-slot="{ active, toggle }"
+        :value="UploadType.Cloud"
+      >
+        <v-btn
+          :color="active ? 'primary' : ''"
+          class="mr-2"
+          @click="toggle"
+        >
+          <v-icon class="mr-2">
+            mdi-cloud
+          </v-icon>
+          <div>Choose firmware from repository</div>
+        </v-btn>
+      </v-item>
+      <v-item
+        v-slot="{ active, toggle }"
+        :value="UploadType.File"
+      >
+        <v-btn
+          :color="active ? 'primary' : ''"
+          @click="toggle"
+        >
+          <v-icon class="mr-2">
+            mdi-paperclip
+          </v-icon>
+          <div>Upload custom firmware file</div>
+        </v-btn>
+      </v-item>
+    </v-item-group>
+
+    <v-form
+      v-if="upload_type === UploadType.Cloud"
+      ref="form"
+      lazy-validation
+    >
+      <v-select
+        v-model="chosen_vehicle"
+        :items="vehicle_types"
+        label="Vehicle"
+        required
+        @change="updateAvailableFirmwares"
+      />
+      <v-select
+        v-model="chosen_firmware_url"
+        :disabled="disable_firmware_selection"
+        :items="showable_firmwares"
+        :label="firmware_selector_label"
+        :loading="loading_firmware_options"
+        required
+        @change="setCloudFirmwareChosen"
+      />
+    </v-form>
+
+    <v-file-input
+      v-if="upload_type === UploadType.File"
+      show-size
+      label="Firmware file"
+      @change="setFileFirmware"
+    />
+
+    <v-btn
+      :disabled="!allow_installing"
+      class="mt-6"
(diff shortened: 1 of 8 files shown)


core/frontend/src/components/autopilot/FirmwareManager.vue
```diff
@@ -0,0 +1,269 @@
...
+enum UploadType {
+  Cloud,
+  File
+}
+
+export default Vue.extend({
+  name: 'FirmwareManager',
+  data() {
+    return {
+      UploadType,
```
patrickelectric, line 138: I'm not sure if I follow, what is this UploadType here ?

core/frontend/src/components/autopilot/FirmwareManager.vue
```diff
@@ -0,0 +1,269 @@
...
+    },
+    setCloudFirmwareAvailable(): void {
+      this.cloud_firmware_available = true
+    },
+    setFileFirmwareAvailable(file: Blob): void {
+      this.firmware_file = file
+      this.file_firmware_available = true
+    },
+    async installFirmware(): Promise<void> {
+      this.installing = true
```
patrickelectric, line 207: This looks like an enum disguised as a series of booleans

core/frontend/src/components/autopilot/FirmwareManager.vue
```diff
@@ -0,0 +1,269 @@
...
+            NotificationLevel.Error,
+            autopilot_service,
+            'FILE_FIRMWARE_UPLOAD_FAIL',
+            `Could not upload firmware: ${error}.`,
+          ))
+          this.successfull_install = false
+          try {
+            this.install_result_message = error.response.data.message
+          } catch {
+            this.install_result_message = ''
```
patrickelectric, line 256: This should contains something like "Invalid backend error message."

core/frontend/src/types/autopilot.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export enum Vehicle {
+    Sub = 'Sub',
+    Rover = 'Rover',
+    Plane = 'Plane',
+    Copter = 'Copter',
+}
+
+export enum SITLFrame {
```
patrickelectric, line 13: Is it possible somehow to get these enums from the backend ? To avoid maintaining two enums in sync.

patrickelectric (changes requested):

core/frontend/src/components/autopilot/FirmwareManager.vue
rafaellehmkuhl: This is injecting UploadType enum into the Vue instance, making it usable inside template.

core/frontend/src/components/autopilot/FirmwareManager.vue
rafaellehmkuhl: Definitely. Will make an Enum for install status.

core/frontend/src/components/autopilot/FirmwareManager.vue
rafaellehmkuhl: Good one.

core/frontend/src/types/autopilot.ts
rafaellehmkuhl: Maybe...
I like the idea, but have to think on how to do it properly, since languages are different and so are restrictions (last time I've checked the enum support on json is not yet rounded for both languages).

I've opened issue #451 so we can study that for the other types as well.

I suspect we will have problems with type-check and auto-complete, though, since they would be dynamically loaded.

Williangalvani, line 13: Is this whole enum unused?

Williangalvani (approved):
Nice!

I had an issue loading the available firmwares for SUB  (timeout) but github lost my previous review draft. That was likely to blame on the backend, and I couldn't replicate anyway.

I onyl have one request: Can we make it so we can drag a file onto the file field?

core/frontend/src/types/autopilot.ts
rafaellehmkuhl: Yup. I was adding this to allow SITL configuration, but I can add this later on the following PR.

rafaellehmkuhl:
> Nice!
> 
> I had an issue loading the available firmwares for SUB (timeout) but github lost my previous review draft. That was likely to blame on the backend, and I couldn't replicate anyway.
> 
> I onyl have one request: Can we make it so we can drag a file onto the file field?

Sure! I will take a look at adding this. If it's no big deal I will add to this PR already.

About the issue, if you get to replicate it please let me know.

rafaellehmkuhl:
Image updated.

patrickelectric (approved):
Code looks good, but I'm unable to test it at the moment.

PR #449, opened 2021-08-30
core: file-browser: Add baseurl argument on start

rafaellehmkuhl:
`baseurl` argument is needed to make file-browser work properly with reverse-proxys (such as nginx).


Williangalvani (approved):

patrickelectric:
Can you explain what is the problem ?

rafaellehmkuhl:
> Can you explain what is the problem ?

Without `baseurl` set Filebrowser will try to fetch it's statics from `/` (root), so not working with the `/file-browser` address.

PR #450, opened 2021-08-31
Add Logviewer

Williangalvani:
image is up at williangalvani/companion-core:logviewer

it is served by nginx at /logviewer, and isn't picked up by the helper as it isn't an independent server.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -21,6 +21,7 @@ RUN /home/pi/tools/ardupilot_tools/bootstrap.sh
 RUN /home/pi/tools/dnsmasq/bootstrap.sh
 RUN /home/pi/tools/filebrowser/bootstrap.sh
 RUN /home/pi/tools/linux2rest/bootstrap.sh
+RUN /home/pi/tools/logviewer/bootstrap.sh
 RUN /home/pi/tools/nginx/bootstrap.sh
 
 # Install custom libraries
--- a/core/tools/nginx/nginx.conf
+++ b/core/tools/nginx/nginx.conf
@@ -113,6 +113,10 @@ http {
             proxy_pass http://127.0.0.1:81;
         }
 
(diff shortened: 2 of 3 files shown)


core/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN /home/pi/tools/dnsmasq/bootstrap.sh
 RUN /home/pi/tools/filebrowser/bootstrap.sh
 RUN /home/pi/tools/linux2rest/bootstrap.sh
 RUN /home/pi/tools/nginx/bootstrap.sh
+RUN /home/pi/tools/logviewer/bootstrap.sh
```
patrickelectric, line 25: The list is in alphabetical order, can you fix logviewer position ?

core/tools/nginx/nginx.conf
```diff
@@ -113,6 +113,10 @@ http {
             proxy_pass http://127.0.0.1:81;
         }
 
+        location /logviewer {
+            root /home/pi/tools;
```
patrickelectric, line 117: Should it be /home/pi/tools/logviewer ?

patrickelectric (changes requested):

core/Dockerfile
Williangalvani: done

core/tools/nginx/nginx.conf
Williangalvani: no, the current way the full path ends up at /home/pi/tools/logviewer (root + location)

patrickelectric: Thanks! That was what I was imagining

patrickelectric:
Just a last thing, the commits message should have a `core:` prefix.
image

patrickelectric (approved):

PR #454, opened 2021-08-31
Add log browser

rafaellehmkuhl:
Log browser allow to download, delete and play autopilot logs (both `.tlog` and `.bin`).

Log browser uses `filebrowser` API to deal with the files and `UAV Logviewer` to play them.

image

Docker image available at `rafaellehmkuhl/companion-core/add-log-browser`

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -131,6 +131,11 @@ export default Vue.extend({
             icon: 'mdi-chip',
             route: '/firmware',
           },
+          {
+            title: 'Log Browser',
+            icon: 'mdi-file-multiple',
+            route: '/logs',
+          },
         ],
       },
     ],
--- a/core/frontend/src/components/logs/LogManager.vue
+++ b/core/frontend/src/components/logs/LogManager.vue
@@ -1,0 +1,157 @@
+<template>
+  <v-container>
+    <v-card
+      v-if="!fetching_logs"
+      flat
+    >
+      <v-card-title>
+        Autopilot logs
+        <v-spacer />
+        <v-btn
+          icon
+          @click="fetchAvailableLogs"
+        >
+          <v-icon>mdi-update</v-icon>
+        </v-btn>
+        <v-btn
+          icon
+          color="green darken-1"
+          :disabled="disable_batch_operations"
+          @click="downloadSelectedLogs"
+        >
+          <v-icon>mdi-download-multiple</v-icon>
+        </v-btn>
+        <v-btn
+          icon
+          color="red darken-1"
+          :disabled="disable_batch_operations"
+          @click="removeLogs"
+        >
+          <v-icon>mdi-trash-can</v-icon>
+        </v-btn>
+      </v-card-title>
+      <v-card-text>
+        <v-data-table
+          v-model="selected_logs"
+          :headers="headers"
+          :items="parsed_logs"
+          item-key="name"
+          show-select
+          :sort-by.sync="sortBy"
+          :sort-desc.sync="sortDesc"
+        >
+          <template #item.actions="{ item }">
+            <v-icon @click="replay_log(item)">
+              mdi-play
+            </v-icon>
+            <v-icon @click="downloadLogs([item])">
+              mdi-download
+            </v-icon>
+          </template>
+        </v-data-table>
+      </v-card-text>
+    </v-card>
+    <v-container v-else>
+      <spinning-logo />
+    </v-container>
+  </v-container>
+</template>
+
+<script lang="ts">
+import { format } from 'date-fns'
+import Vue from 'vue'
+
+import filebrowser from '@/libs/filebrowser'
+import { FilebrowserFile } from '@/types/filebrowser'
+
+import SpinningLogo from '../common/SpinningLogo.vue'
+
+export default Vue.extend({
+  name: 'LogManager',
+  components: {
+    SpinningLogo,
+  },
+  data() {
+    return {
+      sortBy: 'modified',
+      sortDesc: true,
+      auth_key: '',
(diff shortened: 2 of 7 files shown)


core/frontend/src/App.vue
```diff
@@ -59,6 +59,12 @@
             </v-list-item-icon>
             <v-list-item-title>Main</v-list-item-title>
           </v-list-item>
+          <v-list-item to="/logs">
```
patrickelectric, line 62: We should break this here with a line to divide in a "Vehicle" category and use a different icon other than home.
Maybe:
image

core/frontend/src/components/logs/LogManager.vue
```diff
@@ -0,0 +1,231 @@
...
+    are_logs_available(): boolean {
+      return this.available_logs.length !== 0
+    },
+  },
+  async mounted() {
+    await this.authenticateOnFileBrowser()
+    await this.fetchAvailableLogs()
+  },
+  methods: {
+    async authenticateOnFileBrowser(): Promise<void> {
```
patrickelectric, line 133: We should break all this file handle logic API in a different class to be common for all our components.

patrickelectric (changes requested):
Besides that, this PR looks awesome

Williangalvani (commented):
Nice!

I agree with @patrickelectric suggestion about the log icon.

Plus, should we filter for filetypes?
image

The size looks off too, I'd try to make it human-readable.

core/frontend/src/App.vue
rafaellehmkuhl: Didn't found this on Material Design Icons.

Think I will use the `file-multiple`:

image

rafaellehmkuhl:
> Nice!
> 
> I agree with @patrickelectric suggestion about the log icon.
> 
> Plus, should we filter for filetypes?
> image
> 
> The size looks off too, I'd try to make it human-readable.

Cool.
Will filter for `.tlog` and `.bin` files.

patrickelectric (changes requested):
- Disable download and delete buttons when no file is selected
    - This will download or remove / !
    - Add also a check to make sure that we are never going to use it with empty strings! 
- We should deal when one of the folder does not exist, like firmware/logs will not exist if navigator was never connected, this will make everything fail and no file will be available.
- The garbage icon should be red for DANGER and the play icon could be green
- The size of the files should be human friendly and not 0.0000085, maybe 85Bytes, 123Kb and etc, and not everything in MB.

Williangalvani:
clicking the refresh button loses the sorting. I'm trying to get the latest log.
I would also consider just making it sorted by newest by default... thoughts?

rafaellehmkuhl:
> clicking the refresh button loses the sorting. I'm trying to get the latest log.
> I would also consider just making it sorted by newest by default... thoughts?

Agree

rafaellehmkuhl:
> clicking the refresh button loses the sorting. I'm trying to get the latest log.
> I would also consider just making it sorted by newest by default... thoughts?

Fixed sorting reset.

core/frontend/src/components/logs/LogManager.vue
rafaellehmkuhl: Done!

rafaellehmkuhl:
Image updated.

core/frontend/src/libs/filebrowser.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /* Download files (single or multiple). */
+  /**
+   * @param files - FilebrowserFile objects array
+  * */
+  downloadFiles(files: FilebrowserFile[]): void {
+    let url = ''
+    if (files.length === 1) {
+      url = this.singleFileRelativeURL(files[0])
```
Williangalvani, line 143: at a first glance, this looks unnecessary. isn't `multipleFilesRelativeUrl` able to handle a list with only one element?

Williangalvani: oh, it zips all the files together when it's multiple files? then I agree it is best to leave it as is.

core/frontend/src/App.vue
```diff
@@ -90,6 +116,30 @@ export default Vue.extend({
...
+          },
+        ],
+      },
+    ],
   }),
 })
 </script>
+
+<style>
+.active_menu {
```
Williangalvani, line 142: is this used? I couldn't find it.

Williangalvani (approved):
Looking great.

I have a couple of observations, but for me this is already good to go in. I'd be fine with tracking these in an issue or just opening a subsequent pr:

- Add "title" field to buttons (so their function shows up when hovering the mouse)
- The "page size" is lost when updating the table
- We should consider a confirmation dialog (or similar) when deleting files
- Human-readable sizes

core/frontend/src/libs/filebrowser.ts
rafaellehmkuhl: > oh, it zips all the files together when it's multiple files? then I agree it is best to leave it as is.

Yup

core/frontend/src/App.vue
rafaellehmkuhl: No. Just updated without it.

rafaellehmkuhl:
> Looking great.
> 
> I have a couple of observations, but for me this is already good to go in. I'd be fine with tracking these in an issue or just opening a subsequent pr:
> 
> * Add "title" field to buttons (so their function shows up when hovering the mouse)
> * The "page size" is lost when updating the table
> * We should consider a confirmation dialog (or similar) when deleting files
> * Human-readable sizes

Agree with all, but as you suggested I prefer to merge this as it is and open an issue with suggested improvements.

PR #455, opened 2021-09-01
.eslint: remove ForOFStatement restriction

Williangalvani:
We believe ForOf statements are more readable, and readability comes first. Let the developer decide which one to use at its own discretion.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -27,6 +27,25 @@ module.exports = {
     'no-alert': 'off',
     'no-console': 'off',
     'no-extra-parens': ['error', 'all'],
+    // modified https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L339
+    // In our opinion, readability comes first and ForOF statements are more readable,
+    // so we remove the ForOfStatement block.
+    'no-restricted-syntax': [
+      'error',
+      {
+        selector: 'ForInStatement',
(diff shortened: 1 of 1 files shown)


patrickelectric:
My point is that forEach function can be replaced with the map function.

patrickelectric (approved):

rafaellehmkuhl (approved):

rafaellehmkuhl:
> My point is that forEach function can be replaced with the map function.

We should not use map functions to apply side-effects, though. Eslint will block, also.

patrickelectric:
> > My point is that forEach function can be replaced with the map function.
> 
> We should not use map functions to apply side-effects, though. Eslint will block, also.

which side-effects ? map functions are much clear than raw for loops

rafaellehmkuhl:
> > > My point is that forEach function can be replaced with the map function.
> > 
> > 
> > We should not use map functions to apply side-effects, though. Eslint will block, also.
> 
> which side-effects ? map function are much clear than raw for loops

`map` is functional. It should just take an original array and map it into another, without side-effects.

When you want to cause side-effects you should use a forEach or for loop.

patrickelectric:
> `map` is functional. It should just take an original array and map it into another, without side-effects.
> 
> When you want to cause side-effects you should use a forEach or for loop.

Ok, I understand what you want to said now, thanks!

PR #460, opened 2021-09-02
core: frontend: add proxy for APIs in dev mode

Williangalvani:
This allows running the frontend in the local machine while still talking to the actual API running in the companion.

### DIFF
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -3,6 +3,46 @@
   "version": "0.1.0",
   "private": true,
   "vue": {
+    "devServer": {
+      "proxy": {
+        "^/ardupilot-manager": {
+          "target": "http://companion.local/"
+        },
+        "^/cable-guy": {
+          "target": "http://companion.local/"
+        },
+        "^/docker": {
+          "target": "http://companion.local/"
+        },
+        "^/file-browser": {
+          "target": "http://companion.local/"
+        },
+        "^/helper": {
+          "target": "http://companion.local/"
+        },
+        "^/logviewer": {
+          "target": "http://companion.local/"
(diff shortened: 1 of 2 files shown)


core/frontend/package.json
```diff
@@ -3,6 +3,46 @@
...
+        "^/wifi-manager": {
+          "target": "http://companion.local/"
+        },
+        "^/cable-guy": {
+          "target": "http://companion.local/"
+        },
+        "^/helper": {
+          "target": "http://companion.local/"
+        },
+        "^/logviewer": {
```
patrickelectric, line 41: Can we make this list alphabetic ?

Williangalvani: This is following the list in nginx.conf. we should probably make both

rafaellehmkuhl: Agree.

core/frontend/package.json
```diff
@@ -3,6 +3,46 @@
   "version": "0.1.0",
   "private": true,
   "vue": {
+    "devServer": {
```
rafaellehmkuhl, line 6: How should I use it?

Williangalvani: yarn run serve will use this

Williangalvani: then you should be able to open localhost:8080 and everything should work automagically.
you can delete some of this lines or point it to your machine if you want to run some apis locally.

rafaellehmkuhl: Cool!

rafaellehmkuhl (approved):

patrickelectric (approved):

PR #466, opened 2021-09-07
Create a Health indicator using (for now) mavlink data.

Williangalvani:
image up at williangalvani/companion-core:mavlink

image

This creates a mavlink store and uses this mavlink store to fetch the data from the backend.

There's a bunch of things to consider here:

 - [ ] do we actually want a mavlink store? should we just use the mavlink lib api instead? 
 - [ ] should we have the messagename->messageid map in here?
 - others as comments in the code

### DIFF
--- a/core/frontend/src/components/health/HealthTrayMenu.vue
+++ b/core/frontend/src/components/health/HealthTrayMenu.vue
@@ -1,0 +1,96 @@
+<template>
+  <v-menu
+    :close-on-content-click="false"
+    nudge-left="150"
+    nudge-bottom="25"
+  >
+    <template
+      #activator="{ on, attrs }"
+    >
+      <v-card
+        elevation="0"
+        color="transparent"
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon
+          class="px-1"
+          color="white"
+        >
+          mdi-submarine
+        </v-icon>
+      </v-card>
+    </template>
+
+    <v-card
+      elevation="1"
+      width="250"
+    >
+      <v-container>
+        <div>
+          <table style="width: 100%">
+            <tr>
+              <td>Core Temperature:</td>
+              <td>-- ºC</td>
+            </tr>
+            <tr>
+              <td>Voltage:</td>
+              <td>{{ batt_voltage.toFixed(2) }} V</td>
+            </tr>
+            <tr>
+              <td>Current:</td>
+              <td> {{ batt_current.toFixed(2) }} A</td>
+            </tr>
+          </table>
+        </div>
+      </v-container>
+    </v-card>
+  </v-menu>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { getModule } from 'vuex-module-decorators'
+
+import MavlinkStore from '@/store/mavlink'
+
+const mavlink_store: MavlinkStore = getModule(MavlinkStore)
+
+export default Vue.extend({
+  name: 'HealthTrayMenu',
+  components: {
+  },
+
+  computed: {
+    batt_voltage(): number {
+      if (mavlink_store.available_messages.SYS_STATUS) {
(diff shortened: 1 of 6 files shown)


core/frontend/src/libs/MAVLink2Rest/index.ts
```diff
@@ -1,11 +1,17 @@
...
+
+import { Dictionary } from '@/types/common'
+
 import Endpoint from './Endpoint'
 import Listener from './Listener'
 
-interface Dictionary<T> {
-  [key: string]: T;
+// Maps message names to message IDs
+// TODO: should we get this generated somehow?
```
Williangalvani, line 12: I would rather not have mavlink as a submodule in this project, if possible, could mavlink2rest provide this info, @patrickelectric ? I would even manually curate this list rather than include pymavlink...

core/frontend/src/components/health/HealthTrayMenu.vue
```diff
@@ -0,0 +1,96 @@
...
+
+const mavlink_store: MavlinkStore = getModule(MavlinkStore)
+
+export default Vue.extend({
+  name: 'HealthTrayMenu',
+  components: {
+  },
+
+  computed: {
+    batt_voltage(): number {
```
Williangalvani, line 65: @rafaellehmkuhl  could/should these be turned into getters or something in the store?

core/frontend/src/components/health/HealthTrayMenu.vue
```diff
@@ -0,0 +1,96 @@
...
+      if (mavlink_store.available_messages.SYS_STATUS) {
+        const data = mavlink_store.available_messages.SYS_STATUS.messageData.voltage_battery
+        if (typeof data === 'number') {
+          return data / 1000
+        }
+      }
+      return -1
```
Williangalvani, line 72: should we just make this return (value.toFixed(2) | "N/A") ?

rafaellehmkuhl: I think this should be component-wise. In this case, I totally agree.

core/frontend/src/components/health/HealthTrayMenu.vue
rafaellehmkuhl: As from our conversation, the suggestion is to have two getters on the store:
- getMessageData(message_type_name: string)
- getMessageStatus(message_type_name: string)

core/frontend/src/libs/MAVLink2Rest/index.ts
```diff
@@ -76,6 +84,48 @@ class Mavlink2RestManager {
...
+        param7: 0.0,
+        command: {
+          type: 'MAV_CMD_SET_MESSAGE_INTERVAL',
+        },
+        target_system: 0,
+        target_component: 0,
+        confirmation: 0,
+      },
+    }
+    axios.post(`${this.baseUrl}/mavlink`.replace('/ws/mavlink', '').replace('ws', 'http'), payload)
```
Williangalvani, line 121: get rid of this hack

core/frontend/src/components/health/HealthTrayMenu.vue
```diff
@@ -0,0 +1,96 @@
...
+              <td>Core Temperature:</td>
+              <td>-- ºC</td>
+            </tr>
+            <tr>
+              <td>Voltage:</td>
+              <td>{{ batt_voltage.toFixed(2) }} V</td>
+            </tr>
+            <tr>
+              <td>Current:</td>
+              <td> {{ batt_current.toFixed(2) }} A</td>
```
rafaellehmkuhl, line 42: It would be better to move this `.toFixed(2)`  to the computed itself so it can be typechecked.

Williangalvani:
superseded by #508

Issue #445, opened 2021-08-30
frontend: Add Autopilot manager


rafaellehmkuhl:
- [x] Allow creation of new endpoints
- [x] Allow deleting existing endpoints
- [x] Show list of running endpoints
- [ ] Allow system restart
- [ ] Show current running platform

Issue #389, opened 2021-07-21
core: ardupilot-manager: Fails to create udpserver endpoint


patrickelectric:
How to replicate:
- Create udp server  to 0.0.0.0 14550
- It'll not work
- Create an udp client with another ip address, udp server now works

rafaellehmkuhl:
Was able to reproduce. Will check.

rafaellehmkuhl:
I was able to reproduce it in another way:

1. Create UDP Server endpoint at 0.0.0.0 14332 ---> It stops other original UDP Server (0.0.0.0 14550)
2. Create UDP Server endpoint at 0.0.0.0 143333 ---> Eveybody works again

patrickelectric:
Reference #481

rafaellehmkuhl:
Just tested with QGC and it's working properly, without need to create extra endpoints after the first one.

PR #471, opened 2021-09-09
Endpoints frontend

rafaellehmkuhl:
Add page for listing, creating and deleting autopilot endpoints.
Creation forms include IP/Path and Port/Baudrate validation.

Image soon on `rafaellehmkuhl/companion-core:endpoints-frontend`.

Partially solves #445

### DIFF
--- a/core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
+++ b/core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
@@ -1,0 +1,54 @@
+<template>
+  <span />
+</template>
+
+<script lang="ts">
+import axios from 'axios'
+import Vue from 'vue'
+import { getModule } from 'vuex-module-decorators'
+
+import AutopilotManagerStore from '@/store/autopilot_manager'
+import NotificationStore from '@/store/notifications'
+import { autopilot_manager_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { callPeriodically } from '@/utils/helper_functions'
+
+const notification_store: NotificationStore = getModule(NotificationStore)
+const autopilot_manager_store: AutopilotManagerStore = getModule(AutopilotManagerStore)
+
+/**
+ * Responsible for updating autopilot-manager-related data.
+ * This component periodically fetches external APIs to gather information
+ * related to autopilot-manager functions, like mavlink endpoints.
+ * @displayName Autopilot Updater
+ */
+
+export default Vue.extend({
+  name: 'AutopilotManagerUpdater',
+  async mounted() {
+    await callPeriodically(this.fetchAvailableEndpoints, 5000)
+  },
+  methods: {
+    async fetchAvailableEndpoints(): Promise<void> {
+      await axios({
+        method: 'get',
+        url: `${autopilot_manager_store.API_URL}/endpoints`,
+        timeout: 10000,
+      })
+        .then((response) => {
+          const available_endpoints = response.data
+          autopilot_manager_store.setAvailableEndpoints(available_endpoints)
+        })
+        .catch((error) => {
+          notification_store.pushNotification(new LiveNotification(
+            NotificationLevel.Error,
+            autopilot_manager_service,
+            'AUTOPILOT_ENDPOINT_FETCH_FAIL',
+            `Could not fetch available MAVLink endpoints: ${error.message}`,
+          ))
+          autopilot_manager_store.setAvailableEndpoints([])
+        })
+    },
+  },
+})
+</script>
--- a/core/frontend/src/components/autopilot/EndpointCreationDialog.vue
+++ b/core/frontend/src/components/autopilot/EndpointCreationDialog.vue
@@ -1,0 +1,179 @@
+<template>
+  <v-dialog
+    width="500"
+    :value="show"
+    @input="showDialog"
+  >
+    <v-card>
+      <v-card-title>
+        New endpoint
+      </v-card-title>
+
+      <v-card-text class="d-flex flex-column">
+        <v-form
+          ref="form"
+          lazy-validation
+        >
+          <v-text-field
+            v-model="endpoint.name"
+            :counter="50"
+            label="Name"
+            :rules="[validate_required_field]"
+          />
+
+          <v-text-field
+            v-model="endpoint.owner"
+            :counter="50"
+            label="Owner"
+            :rules="[validate_required_field]"
+          />
+
+          <v-select
+            v-model="endpoint.connection_type"
+            :items="endpoint_types"
+            label="Type"
+            :rules="[validate_required_field]"
+          />
+
+          <v-text-field
+            v-model="endpoint.place"
+            :rules="[validate_required_field, is_ip_address_path]"
+            label="IP/Device"
+          />
+
+          <v-text-field
+            v-model="endpoint.argument"
+            :counter="50"
+            label="Port/Baudrate"
+            :rules="[validate_required_field, is_socket_port_baudrate]"
+          />
+
+          <v-checkbox
+            v-model="endpoint.persistent"
+            label="Save endpoint between system sessions?"
+          />
+
+          <v-checkbox
+            v-model="endpoint.protected"
+            label="Protect endpoint from being deleted?"
+            disabled
+          />
+
+          <v-btn
+            color="success"
+            class="mr-4"
+            @click="createEndpoint"
+          >
+            Create
+          </v-btn>
+        </v-form>
+      </v-card-text>
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import axios from 'axios'
+import Vue from 'vue'
+import { getModule } from 'vuex-module-decorators'
+
+import AutopilotManagerStore from '@/store/autopilot_manager'
+import NotificationStore from '@/store/notifications'
+import { EndpointType, userFriendlyEndpointType } from '@/types/autopilot'
+import { autopilot_manager_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { VForm } from '@/types/vuetify'
+import {
+  isBaudrate, isFilepath, isIntegerString, isIpAddress, isNotEmpty, isSocketPort,
+} from '@/utils/pattern_validators'
+
+const notification_store: NotificationStore = getModule(NotificationStore)
+const autopilot_manager_store: AutopilotManagerStore = getModule(AutopilotManagerStore)
+
+export default Vue.extend({
+  name: 'ConnectionDialog',
+  model: {
+    prop: 'show',
+    event: 'change',
+  },
+  props: {
+    show: {
+      type: Boolean,
+      default: false,
+    },
+  },
+
+  data() {
+    return {
+      endpoint: {
+        name: '',
+        owner: '',
+        connection_type: EndpointType.udpin,
+        place: '',
+        argument: '',
+        protected: false,
+        persistent: false,
+      },
+    }
+  },
+  computed: {
+    endpoint_types(): {value: EndpointType, text: string}[] {
+      return Object.entries(EndpointType).map(
+        (type) => ({ value: type[1], text: userFriendlyEndpointType(type[1]) }),
+      )
+    },
+    form(): VForm {
+      return this.$refs.form as VForm
+    },
+  },
+  methods: {
+    validate_required_field(input: string): (true | string) {
+      return isNotEmpty(input) ? true : 'Required field.'
+    },
+    is_ip_address_path(input: string): (true | string) {
+      return isIpAddress(input) || isFilepath(input) ? true : 'Invalid IP/Device-path.'
+    },
+    is_socket_port_baudrate(input: string): (true | string) {
+      if (!isIntegerString(input)) {
+        return 'Please use an integer value.'
+      }
+      const int_input = parseInt(input, 10)
+      return isSocketPort(int_input) || isBaudrate(int_input) ? true : 'Invalid Port/Baudrate.'
+    },
+    async createEndpoint(): Promise<boolean> {
+      // Validate form before proceeding with API request
+      if (!this.form.validate()) {
+        return false
+      }
+
+      autopilot_manager_store.setUpdatingEndpoints(true)
+      this.showDialog(false)
+
+      await axios({
+        method: 'post',
+        url: `${autopilot_manager_store.API_URL}/endpoints`,
+        timeout: 10000,
+        data: [this.endpoint],
+      })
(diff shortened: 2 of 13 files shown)


core/frontend/src/components/autopilot/AutopilotUpdater.vue
```diff
@@ -0,0 +1,47 @@
...
+        .then((response) => {
+          const available_endpoints = response.data
+          autopilot_store.setAvailableEndpoints(available_endpoints)
+        })
+        .catch((error) => {
+          notification_store.pushNotification(new LiveNotification(
+            NotificationLevel.Error,
+            autopilot_service,
+            'AUTOPILOT_ENDPOINT_FETCH_FAIL',
+            `Could not fetch for available endpoints: ${error.message}`,
```
Williangalvani, line 40:
```suggestion
            `Could not fetch available MAVLink endpoints: ${error.message}`,
```

Williangalvani (commented):
The code looks fine, I'll test it soon

core/frontend/src/components/autopilot/AutopilotUpdater.vue
```diff
@@ -0,0 +1,47 @@
...
+import NotificationStore from '@/store/notifications'
+import { autopilot_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { callPeriodically } from '@/utils/helper_functions'
+
+const notification_store: NotificationStore = getModule(NotificationStore)
+const autopilot_store: AutopilotStore = getModule(AutopilotStore)
+
+export default Vue.extend({
+  name: 'AutopilotUpdater',
```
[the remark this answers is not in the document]
rafaellehmkuhl: The idea is for it to take more responsibilities, yes.
Will add a docstring.

core/frontend/src/utils/pattern_validators.ts
```diff
@@ -0,0 +1,33 @@
...
+  const ip_pattern = /^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
+  return ip_pattern.test(ip)
+}
+
+export function validate_filepath(filepath: string): boolean {
+  const filepath_pattern = /^(.+)\/([^/]+)$/
+  return filepath_pattern.test(filepath)
+}
+
+export function validate_non_empty(input: string): boolean {
```
patrickelectric, line 31: I believe that is better to make it generic to work for any scenario.
```ts
function is_empty<T>(sizable: Array<T> | string | undefined | null) {
    return !sizable || sizable.length == 0
}
```

rafaellehmkuhl: Cool! Will update.

Williangalvani (commented):
a few things that bothered me:

 - it allowed me to create an endpoint with no name, which just didn't show up.
 - Ideally we would disable the delete button for non-deleteble endpoints.
 - I can't get it to send data to the topside computer with the UDP Client optioon. I tried both `mavproxy.py --master=udpin:0.0.0.0:14550` and creating an udp server in QGC:
 
image

I'm having some other issues as well, to these may not be related to the pr.

rafaellehmkuhl:
> * it allowed me to create an endpoint with no name, which just didn't show up.
I will check it, because this is a required field, so this should not be possible.

> * Ideally we would disable the delete button for non-deleteble endpoints.
Yup. Will add that.

> * I can't get it to send data to the topside computer with the UDP Client optioon. I tried both `mavproxy.py --master=udpin:0.0.0.0:14550` and creating an udp server in QGC:
This seems like a problem with the autopilot backend. Probably related to #389

core/frontend/src/components/autopilot/AutopilotUpdater.vue
```diff
@@ -0,0 +1,54 @@
+<template>
```
patrickelectric, line 1: **AutoPilotUpdater** sounds like a firmware update thing, or closer to it, can we change it to **AutopilotManagerData** or something closer to it.

core/frontend/src/components/autopilot/AutopilotUpdater.vue
```diff
@@ -0,0 +1,54 @@
...
+import NotificationStore from '@/store/notifications'
+import { autopilot_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { callPeriodically } from '@/utils/helper_functions'
+
+const notification_store: NotificationStore = getModule(NotificationStore)
+const autopilot_store: AutopilotStore = getModule(AutopilotStore)
+
+/**
+ * Responsible for updating autopilot-related data.
```
patrickelectric, line 20: autopilot manager data ?

core/frontend/src/components/autopilot/AutopilotUpdater.vue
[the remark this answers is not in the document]
patrickelectric: I agree with @Williangalvani at this point, but this appears to deal only with the autopilot manager and not the autopilot itself.

core/frontend/src/components/autopilot/EndpointCreationDialog.vue
```diff
@@ -0,0 +1,176 @@
...
+        (type) => ({ value: type[1], text: userFriendlyEndpointType(type[1]) }),
+      )
+    },
+    form(): VForm {
+      return this.$refs.form as VForm
+    },
+  },
+  methods: {
+    required_rule(input: string): (true | string) {
+      return validate_non_empty(input) ? true : 'Required field.'
```
patrickelectric, line 131: This function is not clear, with the return type.
You can have a "required rule true" and you can have a 'Required field' ?
The idea is that when it's `true` you have a valid input and when you have a string you have an invalid input ?
If that's the case, maybe we could rename it to `validate_required_field` ?

patrickelectric: If that's not the case, we could have what is the actual behavior of this `_rules` as a general comment in methods or anywhere, since this appears to be not obvious looking at the code.

core/frontend/src/utils/pattern_validators.ts
```diff
@@ -0,0 +1,40 @@
...
+    return false
+  }
+  if (public_range) {
+    return port >= 1024
+  }
+  return port >= 0
+}
+
+export function validate_integer_string(input: string): boolean {
+  return input.match(/^[0-9]+$/) != null
```
patrickelectric, line 18: What is the problem with `Number.isInteger()` ? Having a function only for that is not overkill ?
If you can't use this function with any type under typescript, It may be better to have `!isNaN(Number(input))` to be more explicit than using regex magic.

core/frontend/src/utils/pattern_validators.ts
```diff
@@ -0,0 +1,40 @@
+export function validate_port(port: number, public_range = true): boolean {
```
patrickelectric, line 1: `validate_socket_port`, we are not validating serial, usb, ethernet ports.

core/frontend/src/utils/pattern_validators.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (typeof port !== 'number') {
```
patrickelectric, line 2: How is the possible with typescript ?

core/frontend/src/types/autopilot.ts
```diff
@@ -9,3 +9,31 @@ export enum Vehicle {
...
+  switch (type) {
+    case EndpointType.udpin: return 'UDP Server'
+    case EndpointType.udpout: return 'UDP Client'
+    case EndpointType.tcpin: return 'TCP Server'
+    case EndpointType.tcpout: return 'TCP Client'
+    case EndpointType.serial: return 'Serial'
+    default: return 'Undefined type'
+  }
+}
+export interface AutopilotEndpoint {
```
patrickelectric, line 31: isn't there a rule for empty lines between definitions ?

patrickelectric (changes requested):

core/frontend/src/components/autopilot/EndpointCreationDialog.vue
rafaellehmkuhl: This is the Vuetify way to handle validation rules. You have a function that returns `true` (validated) or a `string` (not validated). The `string` should contain the reason why it's not validated.

patrickelectric: Thanks for the explanation, but the function name does not completely matches what you are saying, I would recommend change it from `required_` to `validate_`, since this is what the Vuetify field does.

core/frontend/src/utils/pattern_validators.ts
rafaellehmkuhl: `Number.isInteger()` will return `true` for both `111.0` and `111sasdad`. @Williangalvani noticed the problem and the solution I've found was to check for digit-only strings.

Also, typescript performs build-time tests only. It cannot validate runtime (e.g.: user-input) data.

core/frontend/src/utils/pattern_validators.ts
rafaellehmkuhl: True.

core/frontend/src/utils/pattern_validators.ts
rafaellehmkuhl: Good catch. This was from when I was prototyping. Will remove.

core/frontend/src/types/autopilot.ts
rafaellehmkuhl: Humn, apparently not.

core/frontend/src/components/autopilot/AutopilotUpdater.vue
rafaellehmkuhl: True. Maybe *AutoPilotDataUpdater*?

core/frontend/src/components/autopilot/EndpointCreationDialog.vue
rafaellehmkuhl: And I liked the idea of renaming it to `validate_required_field`. Sounds more clear.

core/frontend/src/components/autopilot/AutopilotUpdater.vue
patrickelectric: This still implies on the data of the autopilot itself and not from the manager.. but that's not a problem that could hold the PR.
Maybe only **AutopilotManager** since it'll deal with everything related to it.

core/frontend/src/utils/pattern_validators.ts
patrickelectric: > Also, typescript performs build-time tests only. It cannot validate runtime (e.g.: user-input) data.

I was talking about `Number.isInteger` to receive an string, if that's a valid typescript input since has build-time analysis. 

And what about the second case ?

> If you can't use this function with any type under typescript, It may be better to have `!isNaN(Number(input))` to be more explicit than using regex magic.

core/frontend/src/components/autopilot/AutopilotUpdater.vue
rafaellehmkuhl: Got it.
**AutoPilotManagerUpdater** so?

patrickelectric: :shipit:

core/frontend/src/utils/pattern_validators.ts
rafaellehmkuhl: Just tested `!isNaN(Number(input))` and unfortunately it stills return `true` for the "1.0" case.

core/frontend/src/components/autopilot/AutopilotUpdater.vue
Williangalvani: Agree. AutopilotStore sounds like it will have all Autopilot(MAVLink) data

core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
```diff
@@ -0,0 +1,54 @@
...
+import NotificationStore from '@/store/notifications'
+import { autopilot_manager_service } from '@/types/frontend_services'
+import { LiveNotification, NotificationLevel } from '@/types/notifications'
+import { callPeriodically } from '@/utils/helper_functions'
+
+const notification_store: NotificationStore = getModule(NotificationStore)
+const autopilot_manager_store: AutopilotManagerStore = getModule(AutopilotManagerStore)
+
+/**
+ * Responsible for updating autopilot-related data.
```
patrickelectric, line 20: autopilot manager data

core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
```diff
@@ -0,0 +1,54 @@
...
+ * This component periodically fetches external APIs to gather information
+ * related to autopilot functions, like mavlink endpoints.
```
patrickelectric, line 22: autopilot manager

core/frontend/src/utils/pattern_validators.ts
```diff
@@ -0,0 +1,37 @@
+export function validate_socket_port(port: number, public_range = true): boolean {
```
patrickelectric, line 1: Wait, are we using CamelCase or snake_case for files on this project ? I can see both types in this patch.

patrickelectric (changes requested):

core/frontend/src/utils/pattern_validators.ts
rafaellehmkuhl: Just added a note about style on the Wiki.

patrickelectric (approved):

core/frontend/src/components/autopilot/EndpointCreationDialog.vue
```diff
@@ -0,0 +1,179 @@
...
+    },
+    form(): VForm {
+      return this.$refs.form as VForm
+    },
+  },
+  methods: {
+    validate_required_field(input: string): (true | string) {
+      return isNotEmpty(input) ? true : 'Required field.'
+    },
+    isIpAddress_path(input: string): (true | string) {
```
Williangalvani, line 133: wait what?

Williangalvani (changes requested):

core/frontend/src/components/autopilot/EndpointCreationDialog.vue
rafaellehmkuhl: ?

rafaellehmkuhl: Oh, the case fushadfasuidfhuisda

Williangalvani (approved):

PR #491, opened 2021-09-21
ardupilot-manager: Add default firmwares and allow restoring

### LINKED ISSUE #444: frontend: Add Firmware manager
- [x] Allow installation of firmwares
  - [x] From file
  - [x] From Ardupilot's database
- [x] Allow restoring default firmware (factory firmware)

rafaellehmkuhl:
Fix #444

### DIFF
--- a/core/services/ardupilot_manager/setup.py
+++ b/core/services/ardupilot_manager/setup.py
@@ -9,12 +9,17 @@ from warnings import warn
 
 import setuptools
 
+from settings import Settings
+
 ssl._create_default_https_context = ssl._create_unverified_context
 
 
 current_folder = pathlib.Path(__file__).parent.absolute()
 frontend_folder = pathlib.Path.joinpath(current_folder, "frontend")
 static_folder = pathlib.Path.joinpath(frontend_folder, "static")
+
+autopilot_settings = Settings()
+defaults_folder = autopilot_settings.defaults_folder
 
 StaticFile = namedtuple("StaticFile", "parent filename url")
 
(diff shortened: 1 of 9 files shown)


install/install.sh
```diff
@@ -69,6 +69,15 @@ echo "Going to install companion-docker version ${VERSION}."
...
 curl -fsSL $ROOT/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
 
+echo "Download default Ardusub firmwares for Navigator and Pixhawk1."
+curl -fsSL https://firmware.ardupilot.org/Sub/latest/navigator/ardusub \
+    -o $HOME/.config/companion/ardupilot-manager/firmware/ardupilot_navigator_default
+
+curl -fsSL https://firmware.ardupilot.org/Sub/latest/Pixhawk1/ardusub.apj \
+    -o $HOME/.config/companion/ardupilot-manager/firmware/ardupilot_pixhawk1_default
+
+curl -fsSL $ROOT/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
```
patrickelectric, line 79: duplicated ?

rafaellehmkuhl: Yup haha

Fixed

install/install.sh
```diff
@@ -69,6 +69,13 @@ echo "Going to install companion-docker version ${VERSION}."
 echo "Downloading and installing udev rules."
 curl -fsSL $ROOT/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
 
+echo "Download default Ardusub firmwares for Navigator and Pixhawk1."
```
patrickelectric, line 72: I don't think that this is the best place to install it, since is totally not relevant for the installation step and is only specific for the ardupilot manager, can't we have it in the setup.py as we do for the js files ?

rafaellehmkuhl: Makes sense. Will transfer.

core/services/ardupilot_manager/setup.py
```diff
@@ -9,13 +9,18 @@
```
patrickelectric, line 9: The commit title should start with `core: ardupilot-amanger`

patrickelectric (commented):
The code looks good, the only thing necessary is the docker to test

patrickelectric (approved):
Great Job!

PR #492, opened 2021-09-23
core: frontend: Use /version-chooser as view route

rafaellehmkuhl:
This is a "gambiarra", a "chuncho" to make the version-chooser service call the correct URL on API requests.

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -160,7 +160,7 @@ export default Vue.extend({
           {
             title: 'Version-chooser',
             icon: 'mdi-cellphone-arrow-down',
-            route: '/versionchooser',
+            route: '/version-chooser',
           },
         ],
       },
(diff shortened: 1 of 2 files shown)


patrickelectric (approved):

patrickelectric:
> This is a "gambiarra", a "chuncho" to make the version-chooser service call the correct URL on API requests.

Do we need to add a comment or something to remember that ? What is the correct way to fix this ?

rafaellehmkuhl:
> > This is a "gambiarra", a "chuncho" to make the version-chooser service call the correct URL on API requests.
> 
> Do we need to add a comment or something to remember that ? What is the correct way to fix this ?

I dont think so. There will be a dedicated frontend interface for version-chooser in no time, so this will become deprecated.

Williangalvani (approved):

PR #506, opened 2021-09-28
core: frontend: Update panel icons based on the wifi signal strength and ethernet connection status

patrickelectric:
image

Docker: patrickelectric/companion-core:icons


rafaellehmkuhl (approved):
Flawless

PR #508, opened 2021-09-29
Health indicator: Revolution 

patrickelectric:
image

Docker: patrickelectric/companion-core:health

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -96,6 +96,7 @@
     <autopilot-manager-updater />
     <ethernet-updater />
     <wifi-updater />
+    <mavlink-updater />
   </v-app>
 </template>
 
@@ -106,6 +107,7 @@ import AutopilotManagerUpdater from './components/autopilot/AutopilotManagerUpda
 import EthernetTrayMenu from './components/ethernet/EthernetTrayMenu.vue'
 import EthernetUpdater from './components/ethernet/EthernetUpdater.vue'
 import HealthTrayMenu from './components/health/HealthTrayMenu.vue'
+import MavlinkUpdater from './components/mavlink/MavlinkUpdater.vue'
 import NotificationTrayButton from './components/notifications/TrayButton.vue'
 import ServicesScanner from './components/scanner/servicesScanner.vue'
 import WifiTrayMenu from './components/wifi/WifiTrayMenu.vue'
@@ -123,6 +125,7 @@ export default Vue.extend({
     'ethernet-updater': EthernetUpdater,
     'autopilot-manager-updater': AutopilotManagerUpdater,
     'health-tray-menu': HealthTrayMenu,
+    'mavlink-updater': MavlinkUpdater,
   },
 
   data: () => ({
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -13,7 +13,7 @@
         />
 
         <v-spacer />
-
+        <health-tray-menu />
         <wifi-tray-menu />
         <ethernet-tray-menu />
         <notification-tray-button />
@@ -105,6 +105,7 @@ import Vue from 'vue'
 import AutopilotManagerUpdater from './components/autopilot/AutopilotManagerUpdater.vue'
 import EthernetTrayMenu from './components/ethernet/EthernetTrayMenu.vue'
 import EthernetUpdater from './components/ethernet/EthernetUpdater.vue'
+import HealthTrayMenu from './components/health/HealthTrayMenu.vue'
 import NotificationTrayButton from './components/notifications/TrayButton.vue'
 import ServicesScanner from './components/scanner/servicesScanner.vue'
 import WifiTrayMenu from './components/wifi/WifiTrayMenu.vue'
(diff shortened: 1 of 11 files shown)


core/frontend/src/App.vue
```diff
@@ -123,6 +125,7 @@ export default Vue.extend({
     'ethernet-updater': EthernetUpdater,
     'autopilot-manager-updater': AutopilotManagerUpdater,
     'health-tray-menu': HealthTrayMenu,
+    'mavlink-updater': MavlinkUpdaer,
```
Williangalvani, line 128: typo "Updaer"

core/frontend/src/components/mavlink/MavlinkUpdater.vue
```diff
@@ -0,0 +1,20 @@
...
+import MavlinkStore from '@/store/mavlink'
+
+const mavlink_store: MavlinkStore = getModule(MavlinkStore)
+
+export default Vue.extend({
+  name: 'MavlinkUpdater',
+  async mounted() {
+    mavlink_store.setMessageRefreshRate({ messageName: 'HEARTBEAT', refreshRate: 1 })
+    mavlink_store.setMessageRefreshRate({ messageName: 'SYS_STATUS', refreshRate: 3 })
+  },
```
rafaellehmkuhl, line 18: Id `setMessageRefreshRate` async? If yes its missing an await here. If not there's no need for `mounted` to be async, I think.

core/frontend/src/store/mavlink.ts
```diff
@@ -0,0 +1,64 @@
...
+  message_listeners: Dictionary<Listener> = {}
+
+  @Action({ commit: 'updateMessage' })
+  setMessageRefreshRate(rate: messsageRefreshRate): void {
+    const { messageName, refreshRate } = rate
+    if (refreshRate < 0) {
+      console.warn(`Invalid request rate requested for message ${messageName}@${refreshRate}Hz`)
+    }
+
+    mavlink2rest.requestMessageRate(messageName, refreshRate)
```
rafaellehmkuhl, line 35: Is this async? If so it would be good to await for it.

core/frontend/src/store/mavlink.ts
```diff
@@ -0,0 +1,64 @@
...
+        messageData: receivedMessage,
+        requestedMessageRate: refreshRate,
+      })
+    }).setFrequency(refreshRate)
+  }
+
+  @Mutation
+  updateMessage(message: MavlinkMessage): void {
+    if (message) {
+      Vue.set(this.available_messages, message.messageName, message)
```
rafaellehmkuhl, line 61: I don't know this `Vue.set` function, but I assume this is similar to `Object.assign`, right? If so, the later would be a more common choice, easier to read.

rafaellehmkuhl (changes requested):

core/frontend/src/store/mavlink.ts
patrickelectric: I believe that doing with Vue.set will register and update the change of the value to update any other variable that needs to be updated in an vue event.

patrickelectric:
```
Vue does not allow dynamically adding new root-level reactive properties to an already created instance. However, it’s possible to add reactive properties to a nested object using the Vue.set(object, propertyName, value) method:
```

patrickelectric: I was unable to find something related to it in Vue3

core/frontend/src/store/mavlink.ts
patrickelectric: This is not async, it'll only request a message rate to the backend.

rafaellehmkuhl: It's definitely not mandatory, but it would be good for `requestMessageRate` to be async, since there's a post post request there, right?

patrickelectric: I'm not sure if I follow, what is the post post request ?

rafaellehmkuhl: https://github.com/bluerobotics/companion-docker/blob/6db2eee1dfea7b9ecdc712752bc52f7aa404e3f3/core/frontend/src/libs/MAVLink2Rest/index.ts#L123

core/frontend/src/store/mavlink.ts
Williangalvani: I have a feeling that Vue.set() will be a lot more efficient here. This function has the potential to be called at a pretty high rate and to hold quite a bit of messages.
It may also trigger updates in non-changed data if we update the whole dict.

For the readability, a comment would make it clear.

rafaellehmkuhl: Maybe. Really don't know about performance comparisons. It sounds kind of abusing the method though (what is the relationship between the available_messages vuex state and Vue?).

Although, I don't think we need to block this PR because of this.

patrickelectric: I'll add a todo about it.

core/frontend/src/store/mavlink.ts
patrickelectric: This is handled inside the function, there is no need to deal with it for every single call of `requestMessageRate`

rafaellehmkuhl: But it's blocking hehe

rafaellehmkuhl (approved):

core/frontend/src/store/mavlink.ts
rafaellehmkuhl: Apparently not

Issue #509, opened 2021-09-29
core: bridget: Improve functionality


rafaellehmkuhl:
- [ ] List available serial ports for user to choose from
- [ ] Do not list serial ports associated with autopilot
- [ ] Do not list serial ports associated with ping
- [ ] Do not list serial ports that are blocked

patrickelectric:
Autopilot should lock the ports. The same thing for ports associated with ping. Udev information can be accessed by linux2rest.

Williangalvani:
We should also allow the user to take ports away from ardupilot (navigator), so they are free to be used in companion for user code.

patrickelectric:
Ref: #1076

PR #511, opened 2021-09-29
Add bridges

### LINKED ISSUE #489: core: Add bridges between udp/tcp and serial ports

rafaellehmkuhl:
- Allows fetching, creating and deleting bridge links
- Has validation for all user-inputs

Missing: things specified in #509 

Fix #489


core/frontend/src/components/bridges/BridgeCard.vue
```diff
@@ -0,0 +1,82 @@
...
+  props: {
+    bridge: {
+      type: Object as PropType<Bridge>,
+      required: true,
+    },
+  },
+  methods: {
+    async removeBridge(): Promise<void> {
+      bridget_store.setUpdatingBridges(true)
+      await axios({
```
patrickelectric, line 65: I believe that you don't need to await if you don't need to wait it to finish after this call.

patrickelectric (changes requested):
Baudrate should be a combobox
The same thing should be applied to the serial port showing a list of valid ports
We should also get a default port for the user to use like: 15000
The user should also be able to change between udp/tcp server and client. 
image

core/frontend/src/components/bridges/BridgeCard.vue
rafaellehmkuhl: > Baudrate should be a combobox The same thing should be applied to the serial port showing a list of valid ports We should also get a default port for the user to use like: 15000 The user should also be able to change between udp/tcp server and client. image

1. About baudrate, agree. Will change.
2. About serial port, I can do it, but it will show all available, inclusing autopilot, ok?
3. About default port, ok.
4. About UDP/TCP, isn't bridges UDP only? It says so on both the documentation and help menu.

patrickelectric: > 4\. About UDP/TCP, isn't bridges UDP only? It says so on both the documentation and help menu.

Yep, it's ok to be only udp for now.

patrickelectric: The original questions was also not answered.

core/frontend/src/components/bridges/BridgeCreationDialog.vue
```diff
@@ -0,0 +1,180 @@
...
+      const string_input = String(input)
+      return isNotEmpty(string_input) ? true : 'Required field.'
+    },
+    is_ip_address(input: string): (true | string) {
+      return isIpAddress(input) ? true : 'Invalid IP.'
+    },
+    is_path(input: string): (true | string) {
+      return isFilepath(input) ? true : 'Invalid path.'
+    },
+    is_socket_port(input: string): (true | string) {
```
patrickelectric, line 132: We should also check if it's inside the valid public port range, no ?

core/frontend/src/components/bridges/BridgeCreationDialog.vue
```diff
@@ -0,0 +1,180 @@
...
+      return isSocketPort(int_input) ? true : 'Invalid port.'
+    },
+    is_baudrate(input: number): (true | string) {
+      return isBaudrate(input) ? true : 'Invalid baudrate.'
+    },
+    async createBridge(): Promise<boolean> {
+      // Validate form before proceeding with API request
+      if (!this.form.validate()) {
+        console.log(this.form)
+        console.log(this.bridge)
```
patrickelectric, line 146: Is this lost debug code ?

core/frontend/src/components/bridges/BridgeCreationDialog.vue
rafaellehmkuhl: Maybe. Not completely sure.

core/services/bridget/bridget.py
```diff
@@ -0,0 +1,72 @@
...
+                pass
+            return True
+        except Exception:
+            return False
+
+    @staticmethod
+    def available_serial_ports() -> List[str]:
+        available_ports = []
+        for port in comports():
+            if not Bridget.is_port_available(port.device):
```
patrickelectric, line 44: We should also filter ports that are already in use by bridges

image

patrickelectric (changes requested):

core/services/bridget/bridget.py
rafaellehmkuhl: Fair

patrickelectric (approved):

Williangalvani:
not a blocker, but it looks like this has no "index" which makes helper fail to find the title

rafaellehmkuhl:
Owww let me check that tomorrow. Will add.

Williangalvani (approved):
lgtm.
Just fix CI first =]

Issue #394, opened 2021-07-28
Connecting to a network still doesn't force connection


rafaellehmkuhl:
#383 was supposed to fix #211, but it didn't.

The strategy was to use the `select` command to force the connection and after that use `enable` on all other saved networks.
This re-enable was necessary so the user could be auto-reconnected to other saved networks if away from the away from the last selected one (as the `select` command disables all networks apart from the selected one).

The problem is that this re-enable sometimes makes WPA Supplicant reconnect to one of those re-enabled networks, ignoring the selected one. How WPA decides that is still a question.

A approach that will definitely work is just `select` ing the desired network (not re-enabling the others) and having another asynchronous loop for auto-reconnect.
This loop would check if the system is not connected to any wifi network for the last 10 seconds (for example) and in that case, re-enable all saved networks.
This way we guarantee that, if the system got away from the last connected network or this network was dropped, it would auto-reconnect to any nearby saved network.

PR #534, opened 2021-10-11
core: wifi-manager: Add connection watchdog

rafaellehmkuhl:
Auto-enable and try reconnection to known networks if disconnected for more than specified time.

By removing the reenable of all saved networks on new connections, those changes also ensure connections are guaranteed (preventing impossibility of connecting to another network if there's a known one with bether signal nearby).

Solves #394

### DIFF
--- a/core/services/wifi/WifiManager.py
+++ b/core/services/wifi/WifiManager.py
@@ -1,10 +1,16 @@
 import asyncio
+import time
 from typing import Any, Dict, List, Optional
 
 from loguru import logger
 
 from exceptions import FetchError, ParseError
-from typedefs import SavedWifiNetwork, ScannedWifiNetwork, WifiCredentials
+from typedefs import (
+    ConnectionStatus,
+    SavedWifiNetwork,
+    ScannedWifiNetwork,
+    WifiCredentials,
+)
 from wpa_supplicant import WPASupplicant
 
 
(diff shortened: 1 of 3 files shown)


Williangalvani:
The code looks clean.

Is there a docker image?
how can we test this?
what is the behavior before/after this patch?

core/services/wifi/WifiManager.py
```diff
@@ -209,3 +205,51 @@ async def disconnect(self) -> None:
...
+                logger.debug(f"Seconds since disconnection: {seconds_disconnected}.")
+            if not networks_reenabled and seconds_disconnected >= seconds_before_reconnecting:
+                await self.enable_saved_networks()
+                await self.wpa.send_command_reconnect()
+                networks_reenabled = True
+                logger.debug("Enabled all networks.")
+            if not was_connected and is_connected:
+                seconds_disconnected = 0
+                networks_reenabled = False
+                was_connected = True
```
patrickelectric, line 253: this two booleans appears to be working as enum. It may worth to think a bit more and think if that is the case to improve the logic and readability of the code.

core/services/wifi/WifiManager.py
```diff
@@ -209,3 +205,51 @@ async def disconnect(self) -> None:
...
+                logger.debug("Regained connection.")
+            await asyncio.sleep(2.5)
```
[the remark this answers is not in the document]
rafaellehmkuhl: Cool. Never thought about that.

core/services/wifi/WifiManager.py
rafaellehmkuhl: A enum would definitely help readability here. Will do.

patrickelectric (approved):

PR #559, opened 2021-10-13
Make version-chooser properly async

Williangalvani:
williangalvani/companion-core:async-chooser

### DIFF
--- a/core/services/versionchooser/utils/chooser.py
+++ b/core/services/versionchooser/utils/chooser.py
@@ -2,10 +2,11 @@ import json
 import logging
 import pathlib
 from dataclasses import asdict
+from datetime import datetime
 from typing import Dict, List, Optional, Tuple
 
+import aiodocker
 import appdirs
-import docker
 from aiohttp import web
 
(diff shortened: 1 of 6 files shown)


core/services/versionchooser/utils/chooser.py
```diff
@@ -128,12 +125,12 @@ async def set_version(self, image: str, tag: str) -> web.StreamResponse:
...
-                bootstrap = self.client.containers.get("companion-bootstrap")
-                bootstrap.start()
+                bootstrap = await self.client.containers.get("companion-bootstrap")  # type: ignore
+                await bootstrap.start()
 
                 logging.info("Stopping core...")
-                core = self.client.containers.get("companion-core")
-                core.kill()
+                core = await self.client.containers.get("companion-core")  # type: ignore
+                await core.kill()
```
rafaellehmkuhl, line 133: Just to understand, what is the type problem here?

Williangalvani: IIRC it is just that there are no stubs for aiodocker, so mypy was complaining about no type information

PR #567, opened 2021-10-14
core: add motd message

Williangalvani:


core/configuration/motd
```diff
@@ -0,0 +1,16 @@
+
+Welcome to Companion 1.0!
+
+What you currently see is a tmux session built into Companion docker.
+This means most of the data in here is not-persistent, except for things in /root/.config
```
patrickelectric, line 5: It's better to just say that nothing is persistent and do not tell about /root/.config, otherwise people will start using it to do sutff

core/configuration/motd
```diff
@@ -0,0 +1,16 @@
+
+Welcome to Companion 1.0!
+
+What you currently see is a tmux session built into Companion docker.
```
ES-Alexander, line 4:
```suggestion
What you currently see is a tmux session built into Companion's docker container.
```

PR #595, opened 2021-10-18
core: frontend: Fix available wifi networks

rafaellehmkuhl:
- Show when no networks are available
- Fix filtering out current network from connectable ones
- Use null when no proper information is available


Williangalvani (approved):
lgtm. @patrickelectric is right about splitting it into multiple commits, but I'm willing to let you off the hook with a warning this time =]

PR #598, opened 2021-10-19
core: Remove linting from old frontends

rafaellehmkuhl:
Keep linter just on the main frontend, as the old ones will be removed with time.


patrickelectric (changes requested):
We should decrease the scope of the linter to be only for frontend and not remove the linter entirely

PR #635, opened 2021-11-03
Ardupilot-manager refactor #1

rafaellehmkuhl:
I will send the refactor as concise PRs, separating critical changes from easy ones, so it gets better to review.

Main changes for this first were:

- Proper process kill on keyboard-interrupt
- Fix message on Endpoints page when no endpoints are available (disconnected, for example)

Image available under 'rafaellehmkuhl/companion-core'


core/services/ardupilot_manager/typedefs.py
```diff
@@ -57,10 +57,10 @@ class SITLFrame(str, Enum):
     UNDEFINED = " undefined"
 
 
-def get_sitl_platform_name() -> str:
+def get_sitl_platform_name(machine_arch: str) -> str:
     """Get SITL platform name based on machine architecture."""
 
-    if machine() == "x86_64":
+    if machine_arch == "x86_64":
```
patrickelectric, line 63: since this receives a string as input, we would be more safe if we check if `machine_arch` contains `arm`, since it's a more generic check for arm and arm64, it'll also make more simpler and less error prone to detect x86_64 arch if the string is different.

patrickelectric (changes requested):

core/services/ardupilot_manager/typedefs.py
patrickelectric: This is just a minor request, everything else looks ready to approve.

rafaellehmkuhl: Good idea

patrickelectric (approved):
Great job! 

Just take in mind that we should rework also the others services to reflect the changes on the architecture and organization of this project.
It would be nice to write something about the typedefs, errors, how to detect cotrl-c and etc. 
Maybe an issue can help us to track such necessary global changes.

rafaellehmkuhl:
> Great job!
> 
> Just take in mind that we should rework also the others services to reflect the changes on the architecture and organization of this project. It would be nice to write something about the typedefs, errors, how to detect ctrl-c and etc. Maybe an issue can help us to track such necessary global changes.

Definitely.
Will add an issue for that.

PR #646, opened 2021-11-09
Ardupilot-manager refactor #2

### LINKED ISSUE #556: ardupilot-manager: Raise when failing to start mavlink router
Right now, if a mavlink router subprocess dies on creation, it dies silently.

Something like this on the AbstractRouter should be enough:

```python
        import time
        self._subprocess = subprocess.Popen(
            shlex.split(command), encoding="utf-8", stdout=subprocess.PIPE, stderr=subprocess.PIPE
        )
        time.sleep(1.0)
        if self._subprocess.poll() is not None:
            error = self._subprocess.communicate()[1]
            raise RuntimeError(f"Failed to start router: {self._subprocess.returncode} - {error}.")
```

Unfortunately this makes the tests not pass because the serial port being used on the tests (`/dev/autopilot`) is not available there, so the routers indeed die. The solution would be to also create a fake serial port at `/dev/autopilot` so the tests can go on.

rafaellehmkuhl:
Folllwing #635 this is a put-order-in-the-house PR.

Main changes are:
- Moving the router watchdog method to the Mavlink Manager, since it's the correct responsible for making the router work
- Raise on Mavlink Router start fail, making sure we know when it happens and correctly deal with it
- Create the FlightController type, which is used in favor of the generic tuples to retain information about available flight-controllers. This deprecates FlightControllerType in favor of Platform, since they share basically the same info, and any additional one (e.g. board name) is available on FlightController itself

Image available already.

Fix #556

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -200,14 +200,6 @@ class ArduPilotManager(metaclass=Singleton):
         master_endpoint = Endpoint(
             "Master", self.settings.app_name, EndpointType.TCPServer, "127.0.0.1", 5760, protected=True
         )
-        # The mapping of serial ports works as in the following table:
-        #
-        # |    ArduSub   |       Navigator         |
-        # | -C = Serial1 | Serial1 => /dev/ttyS0   |
-        # | -B = Serial3 | Serial3 => /dev/ttyAMA1 |
-        # | -E = Serial4 | Serial4 => /dev/ttyAMA2 |
-        # | -F = Serial5 | Serial5 => /dev/ttyAMA3 |
-        #
         # pylint: disable=consider-using-with
         self.ardupilot_subprocess = subprocess.Popen(
             [
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -79,27 +79,8 @@ class ArduPilotManager(metaclass=Singleton):
                 self.should_be_running = True
             await asyncio.sleep(5.0)
 
-    async def auto_restart_router(self) -> None:
-        """Auto-restart Mavlink router process if it dies."""
-        while True:
-            try:
-                subprocess_stopped = (
-                    self.mavlink_manager.router_process() is not None
-                    and self.mavlink_manager.router_process().poll() is not None
-                )
-                if self.should_be_running and subprocess_stopped:
-                    logger.debug("Trying to restart Mavlink router...")
-                    self.mavlink_manager.restart()
-                    await asyncio.sleep(1)
-                    if self.mavlink_manager.router_process().poll() is not None:
-                        error = self.mavlink_manager.router_process().communicate()[1]
-                        raise RuntimeError(error)
-                    logger.debug("Mavlink router successfully restarted.")
-            except AssertionError:
-                logger.debug("Mavlink router did not start yet.")
-            except Exception as error:
-                logger.debug(f"Could not restart Mavlink router: {error}. Will try again soon.")
-            await asyncio.sleep(5.0)
+    async def start_mavlink_manager_watchdog(self) -> None:
+        await self.mavlink_manager.auto_restart_router()
 
     def run_with_board(self) -> None:
         if not self.start_board(BoardDetector.detect()):
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -3,7 +3,7 @@ import os
 import pathlib
 import subprocess
 from copy import deepcopy
-from typing import Any, List, Optional, Set, Tuple
+from typing import Any, List, Optional, Set
 
 import psutil
 from commonwealth.mavlink_comm.VehicleManager import VehicleManager
@@ -25,8 +25,9 @@ from settings import Settings
 from typedefs import (
     EndpointType,
     Firmware,
-    FlightControllerType,
+    FlightController,
     Platform,
+    PlatformType,
     SITLFrame,
     Vehicle,
 )
@@ -64,7 +65,7 @@ class ArduPilotManager(metaclass=Singleton):
                 self.ardupilot_subprocess is not None and self.ardupilot_subprocess.poll() is not None
             ) or len(self.running_ardupilot_processes()) == 0
             needs_restart = self.should_be_running and (
-                (self.current_platform in [Platform.SITL, Platform.Navigator] and process_not_running)
+                (self.current_platform.type in [PlatformType.SITL, PlatformType.Linux] and process_not_running)
                 or self.current_platform == Platform.Undefined
             )
             if needs_restart:
@@ -113,10 +114,10 @@ class ArduPilotManager(metaclass=Singleton):
     def current_firmware_path(self) -> pathlib.Path:
         return self.firmware_manager.firmware_path(self.current_platform)
 
-    def start_navigator(self, navigator_type: FlightControllerType) -> None:
-        self.current_platform = Platform.Navigator
+    def start_navigator(self, board: FlightController) -> None:
+        self.current_platform = board.platform
         if not self.firmware_manager.is_firmware_installed(self.current_platform):
-            if navigator_type == FlightControllerType.NavigatorR3:
+            if board.platform == Platform.NavigatorR3:
                 self.firmware_manager.install_firmware_from_params(Vehicle.Sub, self.current_platform)
             else:
                 self.install_firmware_from_file(
@@ -161,10 +162,12 @@ class ArduPilotManager(metaclass=Singleton):
 
         self.start_mavlink_manager(master_endpoint)
 
-    def start_serial(self, device: str) -> None:
-        self.current_platform = Platform.Pixhawk1
+    def start_serial(self, board: FlightController) -> None:
+        if not board.path:
+            raise ValueError(f"Could not find device path for board {board.name}.")
+        self.current_platform = board.platform
         self.start_mavlink_manager(
-            Endpoint("Master", self.settings.app_name, EndpointType.Serial, device, 115200, protected=True)
+            Endpoint("Master", self.settings.app_name, EndpointType.Serial, board.path, 115200, protected=True)
         )
 
     def run_with_sitl(self, frame: SITLFrame = SITLFrame.VECTORED) -> None:
@@ -238,7 +241,7 @@ class ArduPilotManager(metaclass=Singleton):
         self.mavlink_manager.set_master_endpoint(device)
         self.mavlink_manager.start()
 
-    def start_board(self, boards: List[Tuple[FlightControllerType, str]]) -> bool:
+    def start_board(self, boards: List[FlightController]) -> bool:
         if not boards:
             return False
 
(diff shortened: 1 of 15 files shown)


core/services/ardupilot_manager/mavlink_proxy/test_all.py
```diff
@@ -82,34 +83,35 @@ def test_endpoint_validators() -> None:
...
-        EndpointType.Serial,
-        EndpointType.UDPServer,
-        EndpointType.UDPClient,
-        EndpointType.TCPServer,
-        EndpointType.TCPClient,
-    ]
-    run_common_routing_tests(
-        mavproxy, allowed_output_types, allowed_master_types, valid_output_endpoints, valid_master_endpoints
-    )
+# TODO: MavProxy is failling to start for several endpoint combinations. Fix before reactivating its tests.
```
patrickelectric, line 86: You can just use a decorator over commenting all the code. 
`@pytest.mark.skip(reason="MavProxy tests are failling for several endpoint combinations. Since it's not being used and it's not a priority to support it, they are being temporarly disabled.")`

patrickelectric (changes requested):
The code looks great, just some minor corrections and comments.

patrickelectric:
@rafaellehmkuhl can we close this PR ? what is the status ?

rafaellehmkuhl:
> @rafaellehmkuhl can we close this PR ? what is the status ?

We should not close it. It's WIP, as some more tests are needed. It should be retaken in the next weeks.

core/services/ardupilot_manager/mavlink_proxy/Manager.py
```diff
@@ -81,5 +89,21 @@ def router_name(self) -> str:
...
+        while True:
+            await asyncio.sleep(5.0)
+
+            needs_restart = self.should_be_running and not self.is_running()
+
+            if not needs_restart:
+                continue
+
+            try:
+                logger.debug("Trying to restart Mavlink router...")
```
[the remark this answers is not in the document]
rafaellehmkuhl: Nope. Moving.

core/services/ardupilot_manager/mavlink_proxy/test_all.py
rafaellehmkuhl: Didn't know this one. Cool!

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
```diff
@@ -79,6 +81,12 @@ def start(self, vehicle_endpoint: Optional[Endpoint] = None, _verbose: bool = Fa
         # pylint: disable=consider-using-with
         self._subprocess = subprocess.Popen(shlex.split(command), shell=False, encoding="utf-8", errors="ignore")
 
+        # Since the process takes some time to successfully start or fail, we need to wait before checking it's state
+        time.sleep(1)
+        if not self.is_running():
+            exit_code = self._subprocess.returncode
```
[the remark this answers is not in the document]
rafaellehmkuhl: That's a good idea.

With current implementation, this would require somewhat cumbersome to do, since the best way (I think) would be to do this asynchronously, using [asyncio subprocess](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.asyncio.subprocess.Process.wait) and also adding the restart logic there.

In one of the next commits, which will enter the following PR, I already make the manager-start async, making this much easier to accomplish, so I suggest we stay like that for now and introduce this better-logging on the next PR.

What do you think?

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -64,7 +64,10 @@ async def auto_restart_ardupilot(self) -> None:
                 self.ardupilot_subprocess is not None and self.ardupilot_subprocess.poll() is not None
             ) or len(self.running_ardupilot_processes()) == 0
             needs_restart = self.should_be_running and (
-                (self.current_platform in [Platform.SITL, Platform.Navigator] and process_not_running)
+                (
+                    self.current_platform in [Platform.SITL, Platform.Navigator, Platform.NavigatorR3]
```
patrickelectric, line 68: I'm not sure if I follow, why Navigator R3 is here if it follows the same string as Navigator ?

patrickelectric, line 68: Navigator stands for R4, R5 or both ? It may be better to have a name for the version and not a generic name.

patrickelectric (changes requested):
Some minor questions and comments.

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: That part was written when we still had only R3 and R4, and I forgot to consider R5. Will do that specifically, as it is on R3.

rafaellehmkuhl: Because the only specific enum is Platform, and NavigatorR4 is a different platform from NavigatorR3 (they happen to have the same string, but comparing strings gets unnecessarily unsafe).

We can keep just a Platform.NavigatorR5 there, but if we happen to change the strings for Navigator platforms (R3, R4, etc) that won't be tracked, and we may miss something.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -200,14 +200,6 @@ def run_with_sitl(self, frame: SITLFrame = SITLFrame.VECTORED) -> None:
...
             "Master", self.settings.app_name, EndpointType.TCPServer, "127.0.0.1", 5760, protected=True
         )
-        # The mapping of serial ports works as in the following table:
-        #
-        # |    ArduSub   |       Navigator         |
-        # | -C = Serial1 | Serial1 => /dev/ttyS0   |
-        # | -B = Serial3 | Serial3 => /dev/ttyAMA1 |
-        # | -E = Serial4 | Serial4 => /dev/ttyAMA2 |
-        # | -F = Serial5 | Serial5 => /dev/ttyAMA3 |
-        #
```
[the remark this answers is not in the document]
rafaellehmkuhl: It is valid, but we don't use it anymore for SITL, so there's no need to keep it there (the one for Navigator is still there as we use it).

core/services/ardupilot_manager/ArduPilotManager.py
rafaellehmkuhl: I'm thinking adding an extra field like `type`, which could be Serial or Linux would help to get this less verbose (specially when we get to support more boards).

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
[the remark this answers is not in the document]
patrickelectric: ping @rafaellehmkuhl

core/services/ardupilot_manager/flight_controller_detector/Detector.py
```diff
@@ -18,89 +19,114 @@ def _is_root() -> bool:
...
+                AK09915_address = 0x0C
+                bus.read_byte_data(AK09915_address, 0)
+
+                BME280_address = 0x76
+                bus.read_byte_data(BME280_address, 0)
+
+                bus = SMBus(4)
+                PCA9685_address = 0x40
+                bus.read_byte_data(PCA9685_address, 0)
+                return True
```
patrickelectric, line 45: It's missing the check for mmc5983 via SPI1, the code is the same as Navigator4

core/services/ardupilot_manager/flight_controller_detector/Detector.py
```diff
@@ -18,89 +19,114 @@ def _is_root() -> bool:
         return os.geteuid() == 0
 
     @staticmethod
-    def detect_navigator_r3() -> Tuple[bool, str]:
-        """Check if navigator R3 is connected using the sensors on the I²C BUS
+    def get_navigator_if_connected() -> Optional[FlightController]:
```
patrickelectric, line 22: This should follow the same function name pattern as `detect_serial_flight_controllers`, like: `detect_navitator_board`

core/services/ardupilot_manager/typedefs.py
```diff
@@ -100,12 +103,31 @@ class FirmwareFormat(str, Enum):
     ELF = "ELF"
 
 
-class FlightControllerType(IntEnum):
-    """Supported flight-controller types."""
+class PlatformType(str, Enum):
```
patrickelectric, line 106: You could convert the string automatically 1x1 with auto enum:
```python
class LowerStringEnum(Enum):
    def __str__(self):
        return self.name.lower()

class PlatformType(LowerStringEnum):
    Serial = auto()
    Linux = auto()
    SITL = auto()
    Unknown = auto()
```

core/services/ardupilot_manager/typedefs.py
```diff
@@ -100,12 +103,31 @@ class FirmwareFormat(str, Enum):
...
+
+def get_platform_type(platform: Platform) -> PlatformType:
+    platform_types = {
+        Platform.Pixhawk1: PlatformType.Serial,
+        Platform.NavigatorR3: PlatformType.Linux,
+        Platform.NavigatorR4: PlatformType.Linux,
+        Platform.NavigatorR5: PlatformType.Linux,
+        Platform.SITL: PlatformType.SITL,
+    }
+    return platform_types.get(platform, PlatformType.Unknown)
```
patrickelectric, line 121: You can move the function to be inside the enum class, this will make it more clear to use. 
```python
class Platform(str, Enum):
    """Valid Ardupilot platform types.
    The Enum values are 1:1 representations of the platforms available on the ArduPilot manifest."""

    Pixhawk1 = "Pixhawk1"
    NavigatorR3 = "navigator"
    NavigatorR4 = "navigator"
    NavigatorR5 = "navigator"
    #SITL = get_sitl_platform_name(machine())
    Undefined = "Undefined"
    
    def type(self) -> PlatformType:
        platform_types = {
            Platform.Pixhawk1: PlatformType.Serial,
            Platform.NavigatorR3: PlatformType.Linux,
            Platform.NavigatorR4: PlatformType.Linux,
            Platform.NavigatorR5: PlatformType.Linux,
        }
        return platform_types.get(self.value, PlatformType.Unknown)
    
print(Platform.Pixhawk1.type())
```

patrickelectric (changes requested):
Some minor corrections

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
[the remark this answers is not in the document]
rafaellehmkuhl: Forgot to commit it. Done.

core/services/ardupilot_manager/flight_controller_detector/Detector.py
rafaellehmkuhl: Yes. I was just placeholding the code there. Still need to check on this.

rafaellehmkuhl: Deprecated R4

core/services/ardupilot_manager/typedefs.py
rafaellehmkuhl: Good idea

patrickelectric (approved):
Great patch!

PR #659, opened 2021-11-16
core: frontend: Add loading icon with iframe

### LINKED ISSUE #521: core: frontend: Add loading icon on external apps views
The interfaces that are loaded as IFrames (terminal, filebrowser and version-chooser) usually take some seconds to appear. It would be good to have some loading icon so users know the loading is happening (so they don't click vigorously on the button many many times).

DAYS_TO_DEVELOP:1

patrickelectric:
Fix #521

### DIFF
--- a/core/frontend/src/components/common/SpinningLogo.vue
+++ b/core/frontend/src/components/common/SpinningLogo.vue
@@ -1,8 +1,8 @@
 <template>
   <v-img
+    :aspect-ratio="1/1"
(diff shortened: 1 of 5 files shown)


Williangalvani:
note that this does work on some setups

patrickelectric:
Icon aspect ratio issue fixed by 31f89cd

rafaellehmkuhl (approved):

Issue #437, opened 2021-08-27
build: yarn install & build is taking too much time


rafaellehmkuhl:
Yarn is taking a long time (10 minutes on Rasp4) to run install and build commands (`3/27 : RUN cd /home/pi/frontend && yarn install --network-timeout=300000 && yarn build --skip-plugins @vue/cli-plugin-eslint`).

It would be good to take a look at that and see what's going on. On my machine (i7 3gen) it takes at least than times less for both operations.

DAYS_TO_DEVELOP:3

Williangalvani:
Maybe we can separate the Build and Test steps in CI

rafaellehmkuhl:
The timeout on package fetch, which stales, is the result of a bug on yarn that [was fixed only for v2 onwards](https://github.com/yarnpkg/yarn/issues/8242).

To migrate to yarn v2 we should follow [this guide](https://yarnpkg.com/getting-started/migration).

patrickelectric:
Fixed by bun

PR #661, opened 2021-11-17
core: dockerfile: Use cache for node_modules

patrickelectric:
This makes the install step to be cached and when used it takes only
10s to run, before this change it could take up almost 2min.

Helps with #437


Williangalvani (approved):
Nice.

PR #667, opened 2021-11-18
core: tools: bootstraps: Add missing 'set -e' to raise if any command fails

patrickelectric:
I got a situation where ttyd fails to download but the ttyd file is there and empty

### DIFF
--- a/core/tools/ardupilot_tools/bootstrap.sh
+++ b/core/tools/ardupilot_tools/bootstrap.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# Immediately exit on errors
+set -e
+
 ## Download and install necessary tools to user binary folder with the correct permissions
 
 ### Ardupilot's uploader is used to upload firmwares to serial boards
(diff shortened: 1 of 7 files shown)


core/tools/ardupilot_tools/bootstrap.sh
```diff
@@ -9,8 +9,13 @@ set -e
 COMMIT_HASH=4ea8c32c61781fa36dff9748fe3a18cdb5743abb
 LOCAL_PATH_UPLOADER="/usr/bin/ardupilot_fw_uploader.py"
 REMOTE_URL_UPLOADER="https://raw.githubusercontent.com/ArduPilot/ardupilot/${COMMIT_HASH}/Tools/scripts/uploader.py"
-wget "$REMOTE_URL_UPLOADER" -O "$LOCAL_PATH_UPLOADER"
-chmod +x "$LOCAL_PATH_UPLOADER"
+
+# Sudo command is not available on docker and the script is also used in different environmensts
```
Williangalvani, line 13: environments*

core/tools/ardupilot_tools/bootstrap.sh
```diff
@@ -9,8 +9,13 @@ set -e
...
+# The SUDO lias allow the usage of sudo when such command exists and also ignore if it doesn't
```
Williangalvani, line 14: alias*

core/tools/ardupilot_tools/bootstrap.sh
patrickelectric: Fixed.

core/tools/ardupilot_tools/bootstrap.sh
patrickelectric: Fixed.

PR #669, opened 2021-11-18
core: libs: filebrowser: Fix token request

### LINKED ISSUE #665: core: frontend: logs: It's loading forever!
We need to check what is the issue, got the report from Blake

patrickelectric:
The token takes time to be received, we need to await for it
- Fix #665 
- docker: docker.io/patrickelectric/companion-core:fix_filebrowser


rafaellehmkuhl (approved):
Good catch on the async problem!

PR #670, opened 2021-11-18
core: frontend: components: logs: Stop loading if folders does not exist

patrickelectric:
- Needs #669

### DIFF
--- a/core/frontend/src/components/logs/LogManager.vue
+++ b/core/frontend/src/components/logs/LogManager.vue
@@ -1,7 +1,7 @@
 <template>
   <v-container>
     <v-card
-      v-if="!fetching_logs"
+      v-if="logs_fetched"
       flat
     >
       <v-card-title>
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/logs/LogManager.vue
```diff
@@ -115,19 +115,24 @@ export default Vue.extend({
...
 
       const log_folders = ['/root/.config/ardupilot-manager/firmware/logs/', '/root/.config/ardupilot-manager/logs/']
 
-      for (const folder_path of log_folders) {
-        await filebrowser.fetchFolder(folder_path)
-          .then((folder) => {
-            Array.prototype.push.apply(new_logs, folder.items)
-          })
+      // We fetch all paths in parallel and wait for everything to finish
+      // If it fails the folder does not exist, we display a 'No data available' message
```
rafaellehmkuhl, line 122: Where is the 'No data available' message logic?

rafaellehmkuhl (changes requested):

core/frontend/src/components/logs/LogManager.vue
patrickelectric: > Where is the 'No data available' message logic?

It's already on the list component I believe, on vuetify side

rafaellehmkuhl: Perfect. It's the `v-data-table` component behavior.

rafaellehmkuhl (approved):

PR #674, opened 2021-11-18
Improve wifi-manager backend

### LINKED ISSUE #651: frontend: wifi: force new password is not working
I did try the option 3 times and it failed 
The experience if someone puts a wrong password on the wifi is really bad.

rafaellehmkuhl:
- Allow password forcing
- Allow wifi service to be killed on terminal
- Don't reconnect to networks which users disconnected from (on the same session)
- Return HTTP error when failing to connect to network
- Fix watchdog infinite loop
- Disable watchdog while connecting/disconnecting
- Don't store bad network credentials if connection failed
- Improve general wifi service logging

Fix #651 

Image available (contains the frontend change commits)


core/services/wifi/main.py
```diff
@@ -131,18 +131,20 @@ async def disconnect() -> Any:
...
             socket_name = os.listdir(wpa_socket_folder)[0]
         WLAN_SOCKET = os.path.join(wpa_socket_folder, socket_name)
         wifi_manager.connect(WLAN_SOCKET)
-    except Exception as socket_connection_error:
-        print(socket_connection_error)
-        print("Could not connect with provided socket. Connecting via internet socket.")
+    except Exception as error:
+        logger.warning(f"Could not connect with wifi socket. {error}")
         try:
+            logger.info("Connecting via internet wifi socket.")
```
patrickelectric, line 195: This message can be outside of the try block

core/services/wifi/WifiManager.py
```diff
@@ -172,7 +172,7 @@ async def remove_network(self, network_id: int) -> None:
         except Exception as error:
             raise ConnectionError("Failed to remove existing network.") from error
 
-    async def connect_to_network(self, network_id: int) -> None:
+    async def connect_to_network(self, network_id: int, timeout: float = 10.0) -> None:
```
patrickelectric, line 176: timeout is in the wrong commit

core/services/wifi/main.py
```diff
@@ -39,30 +40,42 @@
 @version(1, 0)
 async def network_status() -> Any:
     try:
-        return await wifi_manager.status()
+        status = await wifi_manager.status()
+        logger.info(f"Status:\n {pprint.pformat(status)}")
+        return status
     except FetchError as error:
+        logger.info("Status fetch fail.")
```
patrickelectric, line 47: should this be a warning or error ?

core/services/wifi/WifiManager.py
```diff
@@ -183,6 +183,19 @@ async def connect_to_network(self, network_id: int, timeout: float = 10.0) -> No
             await self.wpa.send_command_save_config()
             await self.wpa.send_command_reconfigure()
             await self.wpa.send_command_reconnect()
+            t0 = time.time()
```
patrickelectric, line 188: I would recommend using `started_time`

core/services/wifi/WifiManager.py
```diff
@@ -183,6 +183,19 @@ async def connect_to_network(self, network_id: int, timeout: float = 10.0) -> No
...
+            while True:
+                status = await self.status()
+                is_connected = "wpa_state" in status and status["wpa_state"] == "COMPLETED"
+                if is_connected:
+                    current_network = await self.get_current_network()
+                    if current_network and current_network.networkid == network_id:
```
patrickelectric, line 194: this missing underscore in `networkid` :eyes:, it appears to be a style problem.
we can fix that in another PR, just reminding that we need to update the frontend as well.

rafaellehmkuhl: It comes like that from wpa_supplicant in fact. We just parse it to a `SavedWifiNetwork`, without changes. Check line 130.

patrickelectric (approved):

PR #698, opened 2021-12-09
Minor updates to install scripts

### LINKED ISSUE #697: Update install script for Raspberry pi OS bullseye
- [ ] add armv7l as supported architecture
 - [ ] rpi-update is probably unnecessary
 - [ ] ~`rfkill unblock all` doesn't seem to be working~

Williangalvani:
Fix #697 

Also sets the baudrates for the i2c buses for navigator. 1MHz for the internal buses and 400kHz for external peripherals.

### DIFF
--- a/install/boards/bcm_27xx.sh
+++ b/install/boards/bcm_27xx.sh
@@ -68,7 +68,7 @@ sudo sed -e 's/console=serial[0-9],[0-9]*\ //' -i /boot/cmdline.txt
 # Update raspberry pi firmware
 # this is required to avoid 'i2c transfer timed out' kernel errors
 # on older firmware versions
-if grep -q ID=raspbian < /etc/os-release; then
+if grep -q VERSION_CODENAME=buster < /etc/os-release; then
     RPI_FIRMWARE_VERSION=1340be4
     if sudo JUST_CHECK=1 rpi-update $RPI_FIRMWARE_VERSION | grep "Firmware update required"; then
         echo "- Run rpi update."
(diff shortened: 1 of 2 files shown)


rafaellehmkuhl (approved):

install/install.sh
```diff
@@ -12,6 +12,7 @@ set -e
 SUPPORTED_ARCHITECTURES=(
   "armhf" # Pi, Pi2, Pi3, Pi4
   "armv7" # Pi2, Pi3, Pi4
+  "armv7l" # Pi4 (Raspberry Pi OS Bullseye)
```
patrickelectric, line 15: Pi2 and Pi3 are both also ARMv8, so `armv7l` comment should also reflect it, or add it on the side of the `armv7` line.

install/boards/bcm_27xx.sh
```diff
@@ -34,9 +34,9 @@ for STRING in \
     "dtoverlay=uart4" \
     "dtoverlay=uart5" \
     "dtparam=i2c_vc=on" \
-    "dtoverlay=i2c1" \
-    "dtoverlay=i2c4,pins_6_7" \
-    "dtoverlay=i2c6,pins_22_23" \
+    "dtoverlay=i2c1,baudrate=1000000" \
+    "dtoverlay=i2c4,pins_6_7,baudrate=1000000" \
+    "dtoverlay=i2c6,pins_22_23,baudrate=400000" \
```
patrickelectric, line 39: the clock can be configured by the driver interface, is there a particular reason to add it here ?

install/install.sh
Williangalvani: done

install/boards/bcm_27xx.sh
Williangalvani: Do you mean the driver in ardupilot? it looks like it isn't implemented there. `bus_clock` is not consumed:
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_HAL_Linux/I2CDevice.cpp#L346

patrickelectric (approved):

PR #716, opened 2022-01-05
ardupilot-manager: change default endpoint for the GCS to the 'legacy' 192.168.2.1

Williangalvani:
this will make the life of tester at the office and new users easier


rafaellehmkuhl (approved):

rafaellehmkuhl:
In this case I think we should also remove the splash message about the endpoints changing.

rafaellehmkuhl (approved):

patrickelectric (approved):

PR #724, opened 2022-01-12
core: tools: mavlink_camera_manager: Update to version t3.1.1

patrickelectric:
docker image (updated to 3.1.1): `patrickelectric/companion-core:update_mavlink_camera_manager_3_1_0`

[changelog](https://github.com/bluerobotics/mavlink-camera-manager/releases/tag/t3.1.1)

### DIFF
--- a/core/tools/mavlink_camera_manager/bootstrap.sh
+++ b/core/tools/mavlink_camera_manager/bootstrap.sh
@@ -5,7 +5,7 @@ set -e
 
 LOCAL_BINARY_PATH="/usr/bin/mavlink-camera-manager"
 
-VERSION=t3.0.11
+VERSION=t3.1.1
 
 # By default we install armv7
 REMOTE_BINARY_URL="https://github.com/bluerobotics/mavlink-camera-manager/releases/download/${VERSION}/mavlink-camera-manager-armv7"


rafaellehmkuhl (approved):
It's already working, so approving.
I will send a small patch to fix the UsbBusType to string, to make the types correct (we were not using this info yet, so it didn't break anything).

Williangalvani:
So.... This is creating streams at 15fps as a default...

image

patrickelectric:
@joaoantoniocardoso ^

joaoantoniocardoso:
Yeah, I noticed that too, but I didn't thought we were using the order in that way. Essentially, the technique we used to prevent duplicates is reordering the sizes vectors (from lower to higher value). As it seems, we are getting the first value from that vector as the default, is that right?

If so, we can simply invert the ordering on the mavlink-camera-manager side.

rafaellehmkuhl:
Strange thing is that, for the ball fake source, the order is correct.

> Yeah, I noticed that too, but I didn't thought we were using the order in that way. Essentially, the technique we used to prevent duplicates is reordering the sizes vectors (from lower to higher value). As it seems, we are getting the first value from that vector as the default, is that right?
> 
> If so, we can simply invert the ordering on the mavlink-camera-manager side.

joaoantoniocardoso:
@rafaellehmkuhl  I think `fakes` aren't set by the same means as the local devices.

This should fix it: https://github.com/bluerobotics/mavlink-camera-manager/pull/25

patrickelectric:
The new version t3.1.1 has a fix, I'm going to update the PR asap.

rafaellehmkuhl (approved):

Issue #230, opened 2021-05-10
How should protected Mavlink endpoints behave?


rafaellehmkuhl:
Today, endpoints in companion have a Boolean field called "protected". 
When protected, those endpoints cannot be removed by the end-user, via REST API. One can only remove this endpoint by directly calling the ArduPilotManager's delete method.

As the end-user cannot remove protected endpoints, it cannot create them also (or no one would ever remove it, with currently implementation).

@patrickelectric commented he doesn't think this should be the final behaviour of this API.

This issue was opened so we can discuss what is the final desired behavior, and modify it accordingly.

PR #726, opened 2022-01-12
Use a shared mavlink-router endpoint for internal services

Williangalvani:
so far, that means Mavlink2Rest and mavlink-camera-manager.
I'm not sure of how ArdupilotManager would handle changing an existing endpoint, so I made a new one...

williangalvani/companion-core:shared-link


rafaellehmkuhl (approved):
The PR is fine, and the connection is working correctly with it.

The only problem I see right now is the old Mavlink2Rest endpoint, which will remain there, without the possibility to be removed. This is a consequence of #230.

It doesn't affect the behavior, though.

PR #728, opened 2022-01-14
Common settings module for all services

rafaellehmkuhl:
I'm trying here to create a common "settings" module for all services.

It has two commits: a first working version and a first not-working version.

The first uses migration files (like the ones used on database migrations), while the second puts the migration instructions inside the versioned setting classes.

The second approach is still not working as I'm trying to figure out how to use the same `upgrade_to_latest_version` method that is declared on the base Settings class, without re-declaring it explicitly on each class (V1, V2, ...). You can see the problem by simply running the `settings.py` module.

The advantage of the second approach is that it does not rely on separated migration files and the migration instructions get cleaner (as they end up using helper methods already present on the base class).

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/settings.py
+++ b/core/libs/commonwealth/commonwealth/utils/settings.py
@@ -1,0 +1,185 @@
+from typing import Any, Dict, Optional
+from loguru import logger
+import appdirs
+import os
+import pathlib
+import json
+import importlib.util
+import time
+
+class NoSettingsFile(ValueError):
+    """Settings file does not exist."""
+
+class BadSettingsFile(ValueError):
+    """Settings file is not valid."""
+
+class SettingsFromTheFuture(ValueError):
+    """Settings file version is from a newer version of the service."""
+
+class OldSettings(ValueError):
+    """Settings file version is old."""
+
+class MigrationFail(RuntimeError):
+    """Could not apply migration."""
+
+
+class Settings:
+    def __init__(self, service_name: str, settings_version: int, settings_filepath: Optional[pathlib.Path] = None) -> None:
+        self.service_name = service_name
+        self.settings_version = settings_version
+        self.settings_filepath = settings_filepath
+
+        if self.settings_filepath is None:
+            logger.warning("Settings filepath not provided. Using default.")
+            self.settings_filepath = self.get_default_settings_filepath()
+
+        try:
+            self.validate_settings_file()
+        except NoSettingsFile as error:
+            if settings_filepath:
+                raise error
+            logger.warning(f"Initial settings file not found. Creating a new one on {self.settings_filepath}.")
+            self.create_initial_settings_file()
+        except BadSettingsFile:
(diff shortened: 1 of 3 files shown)


rafaellehmkuhl:
Validating the settings file for a specific schema is also possible, and I'm planning on adding it on the `validate_attributes` method using [Schema](https://github.com/keleshev/schema).

core/libs/commonwealth/commonwealth/utils/settings.py
```diff
@@ -0,0 +1,197 @@
...
+    """Settings file version is old."""
+
+class MigrationFail(RuntimeError):
+    """Could not apply migration."""
+
+class BadAttributes(BadSettingsFile):
+    """Attributes on settings file are not valid for current version."""
+
+
+class Settings:
```
patrickelectric, line 29: That goes closer to what we have discussed, but there are some things that are a bit dangerous. The access of settings variable via a string argument makes it unsafe to access data and more error prone, that's also a problem related to the entire dict logic.
With a serialization and deserialization library this is fixed, we also ensure that the settings variables are known with static analysis tools.
I wrote a simple example that demonstrates such idea here:
https://github.com/patrickelectric/python_settings_magic/blob/master/settings.py

patrickelectric (changes requested):

core/libs/commonwealth/commonwealth/utils/settings.py
rafaellehmkuhl: Yup, that access directly by dicts is less than ideal. Glad you took a look at that.

In your example we are assuming that variables are just added, not removed, from version to version, or they can also be removed by just not migrating them? In this case, static analysis tools are still going to think they are there, since they are inherited from the parent class, right?

rafaellehmkuhl: If so, a solution would be to simple re-declare an "attributes" variable in each version, which would store the entire attributes structure.

patrickelectric: You can just create another class and replace the migrate without super:
```python
class SettingsV3(SettingsV1):
    VERSION = 3
    new_variable_tresh = pykson.IntegerField(default_value=42)

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        self.VERSION = SettingsV3.VERSION

    def migrate(self, data: dict):
        if data["VERSION"] < SettingsV3.VERSION:
            SettingsV2().migrate(data)

        data["VERSION"] = SettingsV3.VERSION
        data.pop("new_variable_dosh")
```

rafaellehmkuhl: Didn't get where it solves the problem...

patrickelectric: Check the repository for a working example.

rafaellehmkuhl:
Closing in favor of the upcoming @patrickelectric PR.

PR #731, opened 2022-01-18
Fix frontend settings

### LINKED ISSUE #730: Frontend settings not working properly
Frontend settings are not working properly.

Once any variable that is stored on the localStorage is set to true, ever reload of the page will always have this variable started as true on the components, despite being false on the localStorage.

rafaellehmkuhl:
As stated on #730, frontend settings were not working properly.

Basically what was happening was that the variables on the localStorage are always stored as a string or a null type. When loaded onto the vuex store, we were not casting those to proper types. With that, strange things were happening, like switches starting on true state even when a false state was comming from to localStorage. Problem was that the state was the string "false", which is true-ish on Javascript.

Added a `castString` helper function to help with that.
Some auto-fixes from the linter were also performed.

Fix #730

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -13,7 +13,7 @@
         />
 
         <v-spacer />
-        <div 
+        <div
           v-if="settings.is_pirate_mode"
         >
           Ahoy matey! You're running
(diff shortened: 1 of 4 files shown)


rafaellehmkuhl:
Should wait for #732 to be merged, fixing CI.

rafaellehmkuhl:
Image available at `rafaellehmkuhl:fix-settings-frontend`

patrickelectric (approved):

patrickelectric:
Great find!

PR #733, opened 2022-01-19
core: Commonwealth : Add settings

### LINKED ISSUE #707: core: Bridget does not save settings

patrickelectric:
docker: patrickelectric/companion-core:commonwealth_settings

Fix #707

### DIFF
--- a/core/libs/commonwealth/commonwealth/settings/manager.py
+++ b/core/libs/commonwealth/commonwealth/settings/manager.py
@@ -1,0 +1,125 @@
+import pathlib
+import re
+from typing import Any, Optional, Type
+
+import appdirs
+from loguru import logger
+
+from commonwealth.settings.settings import BaseSettings, SettingsFromTheFuture
+
+
+class Manager:
+    SETTINGS_NAME_PREFIX = "settings-"
+
+    def __init__(
+        self,
+        project_name: str,
+        settings_type: Type[BaseSettings],
+        config_folder: Optional[pathlib.Path] = None,
+        load: bool = True,
+    ) -> None:
+        assert project_name, "project_name should be not empty"
+        assert issubclass(settings_type, BaseSettings), "settings_type should use BaseSettings as subclass"
+
+        self.project_name = project_name
+        self.config_folder = (
+            config_folder.joinpath(project_name)
+            if config_folder
+            else pathlib.Path(appdirs.user_config_dir(self.project_name))
+        )
+        self.config_folder.mkdir(parents=True, exist_ok=True)
+        self.settings_type = settings_type
+        self._settings = None
+        logger.debug(
+            f"Starting {project_name} settings with {settings_type.__name__}, configuration path: {config_folder}"
+        )
+        if load:
+            self.load()
+
+    @property
+    def settings(self) -> Any:
+        """Getter point for settings
+
+        Returns:
+            [Type[BaseSettings]]: The settings defined in the constructor
+        """
+        if not self._settings:
+            self.load()
+
+        return self._settings
+
+    @settings.setter
+    def settings(self, value: Any) -> None:
+        """Setter point for settings. Save settings for every change
+
+        Args:
+            value ([Type[BaseSettings]]): The settings defined in the constructor
+        """
+        if not self._settings:
+            self.load()
+
+        self._settings = value
+        self.save()
+
+    def settings_file_path(self) -> pathlib.Path:
+        """Return the settings file for the version specified in the constructor settings
+
+        Returns:
+            pathlib.Path: Path for the settings file
+        """
+        return self.config_folder.joinpath(f"{Manager.SETTINGS_NAME_PREFIX}{self.settings_type.VERSION}.json")
+
+    @staticmethod
+    def load_from_file(settings_type: Type[BaseSettings], file_path: pathlib.Path) -> Any:
+        """Load settings from a generic location and settings type
+
+        Args:
+            settings_type (BaseSettings): Settings type that inherits from BaseSettings.
+            file_path (pathlib.Path): Path for a valid settings file
+
+        Returns:
+            Any: The settings based on settings_type
+        """
+        assert issubclass(settings_type, BaseSettings), "settings_type should use BaseSettings as subclass"
+
+        settings_data = settings_type()
+
+        if file_path.exists():
+            settings_data.load(file_path)
+        else:
+            settings_data.save(file_path)
+
+        return settings_data
+
+    def save(self) -> None:
+        """Save settings"""
+        self.settings.save(self.settings_file_path())
+
+    def load(self) -> None:
+        """Load settings"""
+
+        def get_settings_version_from_filename(filename: pathlib.Path) -> int:
+            result = re.search(f"{Manager.SETTINGS_NAME_PREFIX}(\\d+)", filename.name)
+            assert result
+            assert len(result.groups()) == 1
+            return int(result.group(1))
+
+        # Get all possible settings candidates and sort it by version
+        valid_files = [
+            possible_file
+            for possible_file in self.config_folder.iterdir()
+            if possible_file.name.startswith(Manager.SETTINGS_NAME_PREFIX)
+        ]
+        valid_files.sort(key=get_settings_version_from_filename, reverse=True)
+
+        logger.debug(f"Found possible candidates for settings source: {valid_files}")
+        for valid_file in valid_files:
+            logger.debug(f"Checking {valid_file} for settings")
+            try:
+                self._settings = Manager.load_from_file(self.settings_type, valid_file)
+                logger.debug(f"Using {valid_file} as settings source")
+                return
+            except SettingsFromTheFuture as exception:
+                logger.debug("Invalid settings, going to try another file:", exception)
+
+        self._settings = Manager.load_from_file(self.settings_type, self.settings_file_path())
--- a/core/libs/commonwealth/commonwealth/settings/settings.py
+++ b/core/libs/commonwealth/commonwealth/settings/settings.py
@@ -1,0 +1,100 @@
+import abc
+import json
+import pathlib
+from typing import Any, Dict
+
+import pykson  # type: ignore
+from loguru import logger
+from pykson import Pykson
+
+
+class BadSettingsFile(ValueError):
+    """Settings file is not valid."""
+
+
+class SettingsFromTheFuture(ValueError):
+    """Settings file version is from a newer version of the service."""
+
+
+class MigrationFail(RuntimeError):
+    """Could not apply migration."""
+
+
+class BadAttributes(BadSettingsFile):
+    """Attributes on settings file are not valid."""
+
+
+class BaseSettings(pykson.JsonObject):
+    """Base settings class that has version control and struct based serialization/deserialization"""
+
+    VERSION = pykson.IntegerField(default_value=0)
+
+    def __init__(self, *args: str, **kwargs: int) -> None:
+        super().__init__(*args, **kwargs)
+
+    @abc.abstractmethod
+    def migrate(self, data: Dict[str, Any]) -> None:
+        """Function used to migrate from previous settings verion
+
+        Args:
+            data (dict): Data from the previous version settings
+        """
+        raise RuntimeError("Migrating the setings file does not appears to be possible.")
+
+    def load(self, file_path: pathlib.Path) -> None:
+        """Load settings from file
+
+        Args:
+            file_path (pathlib.Path): Path for settings file
+        """
+        if not file_path.exists():
+            raise RuntimeError(f"Settings file does not exist: {file_path}")
+
+        logger.debug(f"Loading settings from file: {file_path}")
+        with open(file_path, encoding="utf-8") as settings_file:
+            result = json.load(settings_file)
+
+            if "VERSION" not in result.keys():
+                raise BadSettingsFile(f"Settings file does not appears to contain a valid settings format: {result}")
+
+            version = result["VERSION"]
+
+            if version <= 0:
+                raise BadAttributes("Settings file contains invalid version number")
+
+            if version > self.VERSION:
+                raise SettingsFromTheFuture(
+                    f"Settings file comes from a future settings version: {version}, "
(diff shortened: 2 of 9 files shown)


core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+import appdirs
+from loguru import logger
+
+from commonwealth.settings import settings
+
+
+class Manager:
+    SETTINGS_NAME_PREFIX = "settings-"
+
+    def __init__(self, project_name: str, settings_type: Any, config_path: Optional[pathlib.Path] = None):
```
rafaellehmkuhl, line 15: I think `settings_type: typing.Type[settings.BaseSettings]` is the way to indicate for static checkers that `settings_type` inherits from `BaseSettings`.

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+    def settings_file_path(self) -> pathlib.Path:
+        """Return the settings file for the version specified in the constructor settings
+
+        Returns:
+            pathlib.Path: Path for the settings file
+        """
+        return self.config_path.joinpath(f"{Manager.SETTINGS_NAME_PREFIX}{self.settings_type.VERSION}.json")
+
+    @staticmethod
+    def load_from_file(settings_type: Any, file_path: pathlib.Path) -> Any:
```
rafaellehmkuhl, line 66: Same here for `typing.Type`

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+            return int(result.group(1))
+
+        # Get all possible settings candidates and sort it by version
+        valid_files = [
+            possible_file
+            for possible_file in os.listdir(self.config_path)
+            if possible_file.startswith(Manager.SETTINGS_NAME_PREFIX)
+        ]
+        valid_files.sort(key=get_settings_version_from_string)
+        valid_files.reverse()
```
rafaellehmkuhl, line 107: `.sort(..., reverse=True)`

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+            result = re.search(f"{Manager.SETTINGS_NAME_PREFIX}(\\d+)", string)
+            assert result
+            assert len(result.groups()) == 1
+            return int(result.group(1))
+
+        # Get all possible settings candidates and sort it by version
+        valid_files = [
+            possible_file
+            for possible_file in os.listdir(self.config_path)
+            if possible_file.startswith(Manager.SETTINGS_NAME_PREFIX)
```
rafaellehmkuhl, line 104: In the case of using `iterdir()`, which would return the full path for every file, this would become `possible_file.name.startswith...`

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+        ]
+        valid_files.sort(key=get_settings_version_from_string)
+        valid_files.reverse()
+
+        logger.debug(f"Found possible candidates for settings source: {valid_files}")
+        for valid_file in valid_files:
+            logger.debug(f"Checking {valid_file} for settings")
+            file_path = self.config_path.joinpath(valid_file)
```
rafaellehmkuhl, line 112: In the case of using `iterdir()`, there's no need to join here, as the valid_file would be a full Path already.

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+        def get_settings_version_from_string(string: str) -> int:
+            result = re.search(f"{Manager.SETTINGS_NAME_PREFIX}(\\d+)", string)
+            assert result
+            assert len(result.groups()) == 1
+            return int(result.group(1))
+
+        # Get all possible settings candidates and sort it by version
+        valid_files = [
+            possible_file
+            for possible_file in os.listdir(self.config_path)
```
rafaellehmkuhl, line 103: `in self.config_path.iterdir()` would be the modern Pathlib way to do that.

core/libs/commonwealth/commonwealth/settings/settings.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @abc.abstractmethod
+    def migrate(self, data: Dict[str, Any]) -> None:
+        """Function used to migrate from previous settings verion
+
+        Args:
+            data (dict): Data from the previous version settings
+        """
+        raise RuntimeError(
+            "Migrate function from base class was called, settings file appears to be not possible to be migrated"
```
rafaellehmkuhl, line 43: ". Migrating the setings file does not appear to be possible."

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
+import os
```
rafaellehmkuhl, line 1: If using [pathlib.Path native methods](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module), `os` in not needed.

core/libs/commonwealth/commonwealth/settings/settings.py
```diff
@@ -0,0 +1,103 @@
...
+        logger.debug(f"Loading settings from file: {file_path}")
+        with open(file_path, encoding="utf-8") as settings_file:
+            result = json.load(settings_file)
+
+            if "VERSION" not in result.keys():
+                raise BadSettingsFile(f"Settings file does not appears to contain a valid settings format: {result}")
+
+            version = result["VERSION"]
+
+            if version == 0:
```
rafaellehmkuhl, line 64: `<= 0`

core/libs/commonwealth/commonwealth/settings/settings.py
```diff
@@ -0,0 +1,103 @@
...
+    def save(self, file_path: pathlib.Path) -> None:
+        """Save settings to file
+
+        Args:
+            file_path (pathlib.Path): Path for the settings file
+        """
+        # Path for settings file does not exist, lets ensure that it does
+        parent_path = file_path.parent.absolute()
+        if not parent_path.exists():
+            parent_path.mkdir(parents=True, exist_ok=True)
```
rafaellehmkuhl, line 93: Wouldn't just a `file_path.mkdir(parents=True, exis_ok=True) work?

core/libs/commonwealth/commonwealth/settings/settings.py
```diff
@@ -0,0 +1,103 @@
...
+
+        with open(file_path, "w", encoding="utf-8") as settings_file:
+            logger.debug(f"Saving settings on: {file_path}")
+            settings_file.write(Pykson().to_json(self))
+
+    def reset(self) -> None:
+        """Reset internal data to default values"""
+        logger.debug("Reset settings")
```
rafaellehmkuhl, line 101: "Resetting settings"

core/libs/commonwealth/commonwealth/settings/settings.py
```diff
@@ -0,0 +1,103 @@
...
+        new = self.__class__()
+        self.__dict__.update(new.__dict__)
```
rafaellehmkuhl, line 103: Fancy 💯

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class SettingsV1(settings.BaseSettings):
+    VERSION = 1
+    version_1_variable = pykson.IntegerField(default_value=42)
+
+    def __init__(self, *args: str, **kwargs: int) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.VERSION = SettingsV1.VERSION
```
rafaellehmkuhl, line 18: I'm almost sure that is unneeded. `self.VERSION` will already point to the `VERSION` class variable of the instantiated class.

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def test_basic_settings_save_load() -> None:
+    temporary_folder = tempfile.mkdtemp()
+    config_path = pathlib.Path(temporary_folder)
+
+    # Test v1 save
+    settings_manager = manager.Manager("ManagerTest", SettingsV1, config_path)
+    settings_manager.settings.version_1_variable = 2022
+    settings_manager.save()
```
rafaellehmkuhl, line 98: An idea here would be implementing a `__setattr__` with the `save()` call inside, which would make calling `.save()` explicitly unnecessary. This could prevent users forgetting to do that, as well as turning class usage simpler. It should be checked if doing that would affect the internal class behavior, though.

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
```diff
@@ -0,0 +1,211 @@
...
+
+    assert settings_manager.settings.version_1_variable == 2022
+
+    # Test v1 load
+    settings_manager = manager.Manager("ManagerTest", SettingsV1, config_path)
+    settings_manager.load()
```
rafaellehmkuhl, line 104: Would it make sense to put the `load()` call inside `__init__()`, making it unnecessary for the user to explicitly call it?

rafaellehmkuhl (changes requested):
There's no major change request on this review, besides the idea about having `save()` and `load()` implicitly called, which I understand is arguable.

I will send this review like it is and start a better look at the tests, which are really extensive 😅 .

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
patrickelectric: From my tests, `self.VERSION` is initialized from the base class `settings.BaseSettings` when the `super().__init__`  does the pykson initialization

core/libs/commonwealth/commonwealth/settings/manager.py
patrickelectric: Perfect, thanks!

core/libs/commonwealth/commonwealth/settings/manager.py
patrickelectric: good catch!

core/libs/commonwealth/commonwealth/settings/manager.py
patrickelectric: done.

core/libs/commonwealth/commonwealth/settings/manager.py
patrickelectric: Yep, with the corrections the join is unnecessary.

core/libs/commonwealth/commonwealth/settings/settings.py
patrickelectric: This will create a folder with the name of the file, and not the file with the parent folder.

rafaellehmkuhl: True. `file_path.parent.mkdir(...)` would be the way to go.

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
patrickelectric: That creates recursive calls from pykson

core/libs/commonwealth/commonwealth/settings/manager.py
```diff
@@ -0,0 +1,120 @@
...
+from loguru import logger
+
+from commonwealth.settings import settings
+
+
+class Manager:
+    SETTINGS_NAME_PREFIX = "settings-"
+
+    def __init__(self, project_name: str, settings_type: Any, config_path: Optional[pathlib.Path] = None):
+        assert project_name, "project_name should be not empty"
```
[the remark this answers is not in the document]
rafaellehmkuhl: Good point!

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
rafaellehmkuhl: Humnnnn, right. Will take a look into that.

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
rafaellehmkuhl: Was afraid that would be the case...

core/libs/commonwealth/commonwealth/settings/tests/test_manager.py
patrickelectric: Good call.

rafaellehmkuhl (changes requested):

core/services/bridget/bridget.py
```diff
@@ -1,12 +1,55 @@
...
+    def migrate(self, data: Dict[str, Any]) -> None:
+        if data["VERSION"] == SettingsV1.VERSION:
+            return
+
+        if data["VERSION"] < SettingsV1.VERSION:
+            super().migrate(data)
+
+        data["VERSION"] = SettingsV1.VERSION
+
+
```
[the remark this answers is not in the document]
patrickelectric: Done.

rafaellehmkuhl (approved):

core/services/bridget/settings.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any, Dict
+
+import pykson  # type: ignore
+from commonwealth.settings import settings
+
+
+class BridgeSettingsSpec(pykson.JsonObject):
```
rafaellehmkuhl, line 7: Just realized, this class should also include a "V1" at the end, as it could change in the future, right?

rafaellehmkuhl (changes requested):

core/services/bridget/settings.py
patrickelectric: Yes, it makes sense.

rafaellehmkuhl (approved):

PR #739, opened 2022-01-27
frontend: Add system information

### LINKED ISSUE #577: core: frontend: Add system information page
Using linux2rest API information

patrickelectric:
Available in: patrickelectric/companion-core:system_information
Peek 27-01-2022 15-22

Fix #577


core/frontend/src/types/system-information/info.ts
```diff
@@ -0,0 +1,15 @@
+/** Base structure that provides information about OS */
```
rafaellehmkuhl, line 1: Are Info and KernelMessage the same?

core/frontend/src/types/system-information/system.ts
```diff
@@ -0,0 +1,193 @@
...
+    available_space_B: number,
+    filesystem_type: string,
+    mount_point: string,
+    name: string,
+    total_space_B: number,
+    type: string
+}
+
+/** Base structure that provides basic OS/computer information */
+export interface Info {
```
rafaellehmkuhl, line 36: Will this conflict with the other Info interface?

core/frontend/src/components/system-information/AboutThisSystem.vue
```diff
@@ -0,0 +1,88 @@
...
+        { os: 'ubuntu', icon: 'mdi-ubuntu' },
+        { os: 'apple', icon: 'mdi-apple' },
+        { os: 'windows', icon: 'mdi-microsoft-windows' },
+      ]
+
+      if (info) {
+        for (const item of map) {
+          if (info.system_name.toLowerCase().includes(item.os)) {
+            return item.icon
+          }
```
rafaellehmkuhl, line 80: `return map.find(item => info.system_name.toLowerCase().includes(item.os)).icon`

core/frontend/src/views/SystemInformationView.vue
```diff
@@ -0,0 +1,89 @@
...
+
+      <v-col
+        v-if="true"
+        md="9"
+      >
+        <processes v-if="page_selected == 'process'" />
+        <system-condition v-if="page_selected == 'system_condition'" />
+        <network v-if="page_selected == 'network'" />
+        <about-this-system v-if="page_selected == 'about'" />
+      </v-col>
```
rafaellehmkuhl, line 51: Do we want to consider browser history (being able to go back and forth) on those submenus? If so, we should implement routes here. If not, it is ok as is.

core/frontend/src/types/system-information/system.ts
patrickelectric: Only if you include it, but you can also include with a different name.
`import { info as SystemInfo } ...`

core/frontend/src/types/system-information/info.ts
patrickelectric: No, `Info` is OS specific, `KernelMessage` is for each kernel message. 
As in the documentation:
```
/** Base structure that provides information about kernel message */
```

core/frontend/src/views/SystemInformationView.vue
patrickelectric: I think that is not necessary.

core/frontend/src/components/system-information/AboutThisSystem.vue
patrickelectric: Good catch!

patrickelectric: Done.

core/frontend/src/components/system-information/NetworkCard.vue
```diff
@@ -0,0 +1,126 @@
...
+    <v-list-item>
+      <v-list-item-icon>
+        <v-icon>
+          mdi-badge-account-horizontal
+        </v-icon>
+      </v-list-item-icon>
+      <v-list-item-subtitle>MAC:</v-list-item-subtitle>
+      <v-list-item-subtitle
+        class="text-right"
+        :style="show_mac ? '' : 'filter:blur(0.5em)'"
```
rafaellehmkuhl, line 50: Suggestion: using `cursor: pointer` css here on hover (`class:hover{}`) to change the mouse cursor, so user understand this is clickable (when blured). Right now, it's a text-selection cursor.

Not necessary, but pretty.

core/frontend/src/types/system-information/info.ts
rafaellehmkuhl: But are they any different? They seem to have the same parameters.

patrickelectric: Oh, this is a development file. I'll remove.

rafaellehmkuhl (approved):
LGTM

Great PR!

PR #741, opened 2022-01-27
core: version-chooser: fix checking for remote tags with 'companion-core'

### LINKED ISSUE #384: Version chooser is not filtering dockers that are not core
Check second item in the list 
image

Williangalvani:
Fix #384

image being built: williangalvani/companion-core:fix_tag_check

### DIFF
--- a/core/services/versionchooser/utils/chooser.py
+++ b/core/services/versionchooser/utils/chooser.py
@@ -219,7 +219,7 @@ class VersionChooser:
         for image in await self.client.images.list():
             if not image["RepoTags"]:
                 continue
-            if not any("companion-core" in tag for tag in image["RepoTags"]):
+            if not any(tag.startswith("companion-core:") for tag in image["RepoTags"]):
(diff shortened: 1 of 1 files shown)


core/services/versionchooser/utils/chooser.py
```diff
@@ -219,7 +219,7 @@ async def get_available_versions(self, repository: str) -> web.Response:
         for image in await self.client.images.list():
             if not image["RepoTags"]:
                 continue
-            if not any("companion-core" in tag for tag in image["RepoTags"]):
+            if not any(tag.startswith("companion-core:") for tag in image["RepoTags"]):
```
patrickelectric, line 222: Can we also make it valid for blueos-core ?

Williangalvani: I think that belongs on it's own pr adding support for blueos-core

patrickelectric (approved):

PR #747, opened 2022-01-28
frontend: Add error message displayer 

### LINKED ISSUE #744: frontend: Improve user feedback for wrong user input or apply
Right now if the user tries to do something and it fails, we create a notification, that is a silent way to say that something went wrong.

patrickelectric:
Fix #744 
docker: patrickelectric/companion-core:add_error_message
How to test: Add a rtsp endpoint with a port that is not 8554


rafaellehmkuhl (approved):
Very very good addition

PR #748, opened 2022-01-28
src: App: Add href for git_info

patrickelectric:
Docker: patrickelectric/companion-core:add_href_version

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -112,7 +112,18 @@
           <power-menu />
           <settings-menu />
         </v-container>
-        <span class="build_info">Version: {{ git_info }}</span>
+        <span
+          class="build_info"
+        >
+          Version:
+          <a
+            target="_blank"
+            rel="noopener noreferrer"
+            :href="git_info_url"
+          >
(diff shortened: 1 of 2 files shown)


core/frontend/src/App.vue
```diff
@@ -269,6 +279,29 @@ export default Vue.extend({
...
+        if (match && match.groups?.tag) {
+          const { tag } = match.groups
+          return `${project_url}/releases/tag/${tag}`
+        }
+      }
+
+      // Show git source files page for commit
+      const hash = git_describe.slice(git_describe.length - 7)
+      return `${project_url}/tree/${hash}`
+    },
```
rafaellehmkuhl, line 304: The code is good. I just few it should be on some other file, not App.vue, so it don't get bloated with functional logic.

rafaellehmkuhl (changes requested):

core/frontend/src/App.vue
patrickelectric: Done.

rafaellehmkuhl (approved):

PR #752, opened 2022-01-31
Update camera manager to t3.3.0

patrickelectric:
- Add thermal option

docker: patrickelectric/bootstrap:camera-manager-3-3-0

image

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStream.vue
+++ b/core/frontend/src/components/video-manager/VideoStream.vue
@@ -104,6 +104,7 @@ export default Vue.extend({
         },
         interval: this.stream.video_and_stream.stream_information.configuration.frame_interval,
         endpoint: this.stream.video_and_stream.stream_information.endpoints[0],
+        thermal: this.stream.video_and_stream.stream_information?.extended_configuration?.thermal ?? false,
       }
     },
     settings_summary(): string {
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -48,6 +48,22 @@
             :rules="[validate_required_field, is_valid_schema]"
           />
 
+          <v-expansion-panels>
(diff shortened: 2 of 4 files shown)


rafaellehmkuhl:
Are we planning to add much more options on that "extra" menu? If not, I would suggest putting the checkbox on the same dialog, without the extra expanding menu, like we have with the endpoints.

image

If we plan to put much more extra options there, I would just put this whole menu at the end (after stream endpoint), as its extra.

patrickelectric:
> Are we planning to add much more options on that "extra" menu? If not, I would suggest putting the checkbox on the same dialog, without the extra expanding menu, like we have with the endpoints.
> 
> image
> 
> If we plan to put much more extra options there, I would just put this whole menu at the end (after stream endpoint), as its extra.

Yes, we are going to add more later with future versions.

core/frontend/src/components/video-manager/VideoStream.vue
```diff
@@ -104,6 +104,7 @@ export default Vue.extend({
         },
         interval: this.stream.video_and_stream.stream_information.configuration.frame_interval,
         endpoint: this.stream.video_and_stream.stream_information.endpoints[0],
+        thermal: this.stream.video_and_stream.stream_information?.extended_configuration?.thermal ?? false,
```
rafaellehmkuhl, line 107: Cool double-question-mark operator. Didn't know.

rafaellehmkuhl (approved):

rafaellehmkuhl:
> > Are we planning to add much more options on that "extra" menu? If not, I would suggest putting the checkbox on the same dialog, without the extra expanding menu, like we have with the endpoints.
> > image
> > If we plan to put much more extra options there, I would just put this whole menu at the end (after stream endpoint), as its extra.
> 
> Yes, we are going to add more later with future versions.

Ok. In this case, I would suggest just putting it at the end.

patrickelectric:
image

Williangalvani (approved):

PR #758, opened 2022-02-01
github: template: Add ui-corrections

patrickelectric:
Github issue template for UI/UX requests

### DIFF
--- a/.github/ISSUE_TEMPLATE/ui-corrections.yml
+++ b/.github/ISSUE_TEMPLATE/ui-corrections.yml
@@ -1,0 +1,36 @@
+name: Improve User Interface or Experience
+description: Suggestion to improve UI/UX.
+title: "core: frontend: "
+labels: ["core", "enhancement", "triage", "ui"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to suggest improvements to our interface! Please fill out the following details so we can respond and act efficiently.
+
+  - type: textarea
+    attributes:
+      label: Current behaviour
+      description: Describe what you see, why that is bad, and any issues it is causing.
+      placeholder: |
+        Example: "The small margins between the buttons make it harder to distinguish between them, and looks unusual"
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Expected or desired behaviour
+      description: Explain what you expect or would like to see instead, and why.
+      placeholder: |
+        Example: "An extra margin or space between the buttons may help"
+    validations:
+      required: true
+
+  - type: checkboxes
+    id: terms
+    attributes:
+      label: Prerequisites
+      description: Please confirm the following (and check the boxes) before submitting.
+      options:
+        - label: I have checked to make sure that a similar request has not already been filed or fixed.
+          required: true


rafaellehmkuhl (approved):
Should we make exclusive for UI/UX or also allow backend problems? Will the users report backend problems?

.github/ISSUE_TEMPLATE/ui-corrections.yml
```diff
@@ -0,0 +1,36 @@
+name: Improve User Interface or Experience
+description: Suggestion to improve UI/UX.
+title: "core: frontend: "
+labels: ["core", "enhancement", "triage", "ui"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to suggest new changes on our interface! Fill out the form below to provide the necessary information for us.
```
Williangalvani, line 9:
```suggestion
        Thanks for taking the time to suggest new changes to our interface! Please fill out the form below to provide the necessary information for us.
```

.github/ISSUE_TEMPLATE/ui-corrections.yml
```diff
@@ -0,0 +1,36 @@
...
+
+  - type: textarea
+    attributes:
+      label: Current behaviour
+      description: Describe what you have and why it does not work or feels correct.
```
Williangalvani, line 14:
```suggestion
      description: Describe what you see, why it is bad, and what you would like to see instead.
```

Williangalvani (approved):

patrickelectric:
I'll wait for @ES-Alexander review before merging

ES-Alexander (changes requested):
I don't really know how the structure of these works/operates, so my comments are only on the language and reasoning used. Happy for this to be merged once the suggested (or similar) clarifications have been implemented :-)

I expect this and similar templates will be quite useful for people submitting issues/feedback. Likely the largest remaining barrier to people submitting them is needing to have a GitHub account to do so - it's perhaps worth making a submission option that doesn't require logging in (maybe via [a form that automatically has its results turned into a GitHub issue](https://softwareengineering.stackexchange.com/a/339334)?). Of course if that's anonymous then we can't contact them about it if we have follow-up questions, or respond to them to let them know it's in progress/has been completed, but better than nothing. If we do make a form like that we should probably include an optional 'email' field in case they're ok with being asked for extra details, and maybe a box they can check if they would like to know when it's done or something.

In an ideal world we'd be able to set up some API stuff that can auto-send a "issue has been closed" email without needing to make their email public in the issue, and in an even more ideal world we could make comments/responses work via email (still without the user needing a GitHub account), but I expect that's likely too painful and problematic to implement.

patrickelectric:
> I don't really know how the structure of these works/operates, so my comments are only on the language and reasoning used. Happy for this to be merged once the suggested (or similar) clarifications have been implemented :-)
> 
> I expect this and similar templates will be quite useful for people submitting issues/feedback. Likely the largest remaining barrier to people submitting them is needing to have a GitHub account to do so - it's perhaps worth making a submission option that doesn't require logging in (maybe via [a form that automatically has its results turned into a GitHub issue](https://softwareengineering.stackexchange.com/a/339334)?). Of course if that's anonymous then we can't contact them about it if we have follow-up questions, or respond to them to let them know it's in progress/has been completed, but better than nothing. If we do make a form like that we should probably include an optional 'email' field in case they're ok with being asked for extra details, and maybe a box they can check if they would like to know when it's done or something.
> 
> In an ideal world we'd be able to set up some API stuff that can auto-send a "issue has been closed" email without needing to make their email public in the issue, and in an even more ideal world we could make comments/responses work via email (still without the user needing a GitHub account), but I expect that's likely too painful and problematic to implement.

We plan to have such thing in the future, sadly, there is no time to create it right now. We have in the frontend 3 options, github, forum or our own slack (for internal usage). But thanks for the google form reference, it's a pretty neat trick!

ES-Alexander (approved):

ES-Alexander:
Seems like I don't have merge permissions, but happy for this to be merged :-)

Issue #736, opened 2022-01-24
feature: Support using the install script without hardware configuration


rafaellehmkuhl:
Something like a `--no-hardware-config` argument that jumps the hardware configuration step would make us support any Linux board without custom changes. Hardware configuration is not needed if ArduPilot is not going to run on Linux (e.g.: serial boards).

PR #760, opened 2022-02-02
Add option to skip board config during install

matt-bathyscope:
This addresses #736 by adding a `--skip-board-config` argument to the installation script. It also adds some rudimentary usage hints with the `--help` argument. The `get_options` function can easily be extended to accommodate future options (overriding git or docker image repos, for example).

This was tested on a Raspberry Pi 4 1GB running a clean install of `bullseye-lite` with the command:
```
sudo su -c 'curl -fsSL https://raw.githubusercontent.com/matt-bathyscope/companion-docker/master/install/install.sh | bash -s -- --skip-board-config'
```


patrickelectric (approved):
Thank you very much @matt-bathyscope! That's a great contribution!

PR #764, opened 2022-02-02
core: frontend: Update gitignore with yarn files

patrickelectric:
It also helps the IDE to figure out where to search or not

### DIFF
--- a/core/frontend/.gitignore
+++ b/core/frontend/.gitignore
@@ -1,2 +1,5 @@
 node_modules/
 dist/
+.yarn/
+yarn.lock
+


core/frontend/.gitignore
```diff
@@ -1,2 +1,5 @@
 node_modules/
-dist/
\ No newline at end of file
+dist/
+.yarn/
+yarn.lock
```
rafaellehmkuhl, line 4: The other two are fine, but yarn.lock [should be commited](https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/). It ensures everyone uses the same set of dependencies.

rafaellehmkuhl (changes requested):

core/frontend/.gitignore
patrickelectric: It's commited and it's being tracked by git already, this just helps the IDE to avoid it as a search source.

rafaellehmkuhl: If someone clones the repository after this (addition to gitignore) and modify the yarn.lock file, will this change be tracked by git? Will this person be able to commit? I don't remember exactly how the git index/cache is transported to new clones.

patrickelectric: Yes, they are going to be tracked and you'll be able to see the changes and commit.

rafaellehmkuhl: Nice

PR #771, opened 2022-02-04
core: frontend: ReportMenu: Update category_id

patrickelectric:
The category now has a new name and category_id


Williangalvani (approved):

ES-Alexander (approved):
Should the comment just be `BlueOS` so we don't need to change it again later? The ID won't change again when the name changes.

Approved either way :-)

patrickelectric:
> Should the comment just be `BlueOS` so we don't need to change it again later? The ID won't change again when the name changes.
> 
> Approved either way :-)

Good call, fixed.

PR #778, opened 2022-02-05
Add CPU temperature on health menu 

### LINKED ISSUE #776: core: frontend: health: fix missing temperature
### Current behaviour

Temperature still says `WIP` in the health section.

Screen Shot 2022-02-05 at 2 46 53 pm

Should display the current CPU temperature, as is available in the `System Information` service since #180.

### Prerequisites

patrickelectric:
image

Fix #776

### DIFF
--- a/core/frontend/src/components/health/HealthTrayMenu.vue
+++ b/core/frontend/src/components/health/HealthTrayMenu.vue
@@ -52,13 +52,16 @@
 import Vue from 'vue'
 
 import mavlink from '@/store/mavlink'
+import system_information from '@/store/system-information'
 import mavlink_store_get from '@/utils/mavlink'
 
 export default Vue.extend({
   name: 'HealthTrayMenu',
   computed: {
     cpu_temperature(): string {
-      return 'WIP'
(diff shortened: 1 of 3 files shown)


ES-Alexander (approved):
This isn't a codebase I'm familiar with, but it seems to do the job, so approved.

Out of curiosity, would it be reasonably easy to make it only update while the user is looking at it, and if so potentially at a higher frequency (e.g. 1-2s)? If that's simple enough to implement then I think I would personally prefer that, but it's definitely not a big deal if it's left as is with the 5s constant data refresh rate :-)

patrickelectric:
> This isn't a codebase I'm familiar with, but it seems to do the job, so approved.
> 
> Out of curiosity, would it be reasonably easy to make it only update while the user is looking at it, and if so potentially at a higher frequency (e.g. 1-2s)? If that's simple enough to implement then I think I would personally prefer that, but it's definitely not a big deal if it's left as is with the 5s constant data refresh rate :-)

That's a good idea, we could do something about that on the future.

PR #780, opened 2022-02-07
core: services: ardupilot_manager: `Detector.py` improvement

ES-Alexander:
If you require access to the class's properties/methods it's better to use `classmethod`, instead of `staticmethod` with global calls to the class name.


rafaellehmkuhl (approved):

patrickelectric (approved):
Great catch!

PR #787, opened 2022-02-10
Store service logs

rafaellehmkuhl:
This patch introduces storage of the service logs.

All logs are stored under the root user log folder (/root/.cache/log), with each service having its own folder.

image

The files use the following name pattern: `logfile_MM_DD_YY_HH_mm_SS.log`

image

### DIFF
--- a/core/tools/ardupilot_tools/bootstrap.sh
+++ b/core/tools/ardupilot_tools/bootstrap.sh
@@ -23,7 +23,7 @@ USER_SITE_PACKAGES="$(python -m site --user-site)"
 mkdir -p $USER_SITE_PACKAGES
 
 ### Ardupilot's decoder is used to parse and validate firmware ELF files
-COMMIT_HASH=bdb56b20b5cf1f8bf3716c88edb51461f18709a9
+COMMIT_HASH=b839ddcc00f4cb89d89aaa8cf0cb03298d2f00b4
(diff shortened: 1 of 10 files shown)


core/tools/ardupilot_tools/bootstrap.sh
```diff
@@ -23,7 +23,7 @@ USER_SITE_PACKAGES="$(python -m site --user-site)"
 mkdir -p $USER_SITE_PACKAGES
 
 ### Ardupilot's decoder is used to parse and validate firmware ELF files
-COMMIT_HASH=bdb56b20b5cf1f8bf3716c88edb51461f18709a9
+COMMIT_HASH=b839ddcc00f4cb89d89aaa8cf0cb03298d2f00b4
```
Williangalvani, line 26: this commit is from the other patch

patrickelectric: When merging the rebase will take care of it

Williangalvani (approved):
looks good. I think we should discuss using the timestamp for the filenames vs a simple counter.
That will likely be wrong in real life when there's no internet. but this is already a great improvement.

rafaellehmkuhl:
> looks good. I think we should discuss using the timestamp for the filenames vs a simple counter. That will likely be wrong in real life when there's no internet. but this is already a great improvement.

Valid point. Opening an issue to open the discussion.

PR #789, opened 2022-02-11
Add tooltips

rafaellehmkuhl:
- Add tooltip-as-directive lib
- Add tooltips on EndpointCard

image

### DIFF
--- a/core/frontend/src/components/autopilot/EndpointCard.vue
+++ b/core/frontend/src/components/autopilot/EndpointCard.vue
@@ -50,19 +50,22 @@
           >
             <v-card class="elevation-0 d-flex flex-column align-center pa-0">
               <v-icon
+                v-tooltip="persistency.tooltip"
                 class="ma-1"
               >
-                {{ persistent_icon }}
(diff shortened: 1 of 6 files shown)


Williangalvani (changes requested):
Nice! but can we have it show the current state of the thing?
so this ones should show "not protected" and "disabled":
change

rafaellehmkuhl:
> Nice! but can we have it show the current state of the thing? so this ones should show "not protected" and "disabled": change

Oh, for sure haha

core/frontend/src/components/autopilot/EndpointCard.vue
```diff
@@ -50,16 +50,19 @@
           >
             <v-card class="elevation-0 d-flex flex-column align-center pa-0">
               <v-icon
+                v-tooltip="endpoint.persistent ? 'Persistent' : 'Not persistent'"
```
patrickelectric, line 53: Is there a way to force the usage of icons + tooltips for consistency ?

patrickelectric (approved):

PR #794, opened 2022-02-13
Create mavlink-inspector

Williangalvani:
inspector

it needs a new icon and probably some comments in the code before it is ready to go in.

Them, later on, it should be upgraded to support plotting stuff.

Image is up at williangalvani/companion-core:mavlinkinspector

### DIFF
--- a/core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
+++ b/core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
@@ -1,0 +1,191 @@
+<template>
+  <v-container fluid>
+    <v-row>
+      <v-col
+        cols="12"
+        sm="2"
+      >
+        <v-sheet
+          rounded="lg"
+          min-height="268"
+        >
+          <v-card
+            class="mx-auto height-limited"
+            max-height="700px"
+          >
+            <v-list shaped>
+              <v-list-item-group
+                v-model="selected_message_types"
+                multiple
+              >
+                <template v-for="(item, i) in message_types">
+                  <v-list-item
+                    :key="i"
+                    :value="item"
+                    active-class="deep-purple--text text--accent-4"
+                  >
+                    <template #default="{ active }">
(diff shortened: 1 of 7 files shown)


Williangalvani:
While this is not 100% ready, I think it is good and useful enough to go in as is, I intend to keep working on it with time, adding things like live plotting

ES-Alexander (approved):
I'm a fan - nice work! :-)
I reckon it would be useful to include the timestamps in the main list display of the messages, but that can be added later (and was that the thing that needed to be fixed in mavlink2rest first or something?)

Agreed that it's useful enough as-is that it's worth making available now, and improving as you/we have time later :-)

rafaellehmkuhl (approved):

core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
```diff
@@ -0,0 +1,191 @@
...
+            max-height="700px"
+          >
+            <v-list shaped>
+              <v-list-item-group
+                v-model="selected_message_types"
+                multiple
+              >
+                <template v-for="(item, i) in message_types">
+                  <v-list-item
+                    :key="`item-${i}`"
```
patrickelectric, line 23: Just a tip, `i` here should be enough

patrickelectric (changes requested):

core/frontend/src/utils/mavlink_prettifier.ts
```diff
@@ -0,0 +1,45 @@
+import { Dictionary } from '@/types/common'
+
+const formatters = {
+  STATUSTEXT(message: any): string {
```
patrickelectric, line 4: I would just recommend to move it in alphabetic order before merging, this will probably grow a lot soon.

Williangalvani: good point, will do

patrickelectric (approved):

PR #796, opened 2022-02-15
core: frontend: BridgetUpdater: only fetch once when mounted, periodically when requested

Williangalvani:
This prevents unnecessary requests when the component is not in view


core/frontend/src/store/bridget.ts
```diff
@@ -43,6 +45,16 @@ class BridgetStore extends VuexModule {
     this.available_serial_ports = available_serial_ports
     this.updating_bridges = false
   }
+
+  @Mutation
+  startFetching(): void {
```
patrickelectric, line 50: If we are aiming to a generic approach, I would recommend to have this kind of call as a register method, where multiple components can "enable"/"disable" the fetch, and after all components get dismounted the register will hit zero and should_fetch will be false.

core/frontend/src/components/bridges/Bridget.vue
```diff
@@ -94,6 +94,12 @@ export default Vue.extend({
       return this.available_bridges.length !== 0
     },
   },
+  mounted() {
+    bridget.startFetching()
+  },
+  beforeDestroy() {
+    bridget.stopFetching()
```
patrickelectric, line 101: Is there a way that we could use RAII ?

core/frontend/src/components/bridges/BridgetUpdater.vue
```diff
@@ -20,12 +20,21 @@ import { callPeriodically } from '@/utils/helper_functions'
 
 export default Vue.extend({
   name: 'BridgetUpdater',
+  data() {
+    return {
+      prefetched_bridges: false,
+      prefeteched_serial_ports: false,
```
patrickelectric, line 26: This two variables looks more as a state control (that should be an enum) than two non relatable booleans

core/frontend/src/components/bridges/BridgetUpdater.vue
```diff
@@ -20,12 +20,21 @@ import { callPeriodically } from '@/utils/helper_functions'
...
+    }
+  },
   mounted() {
     callPeriodically(this.fetchAvailableBridges, 5000)
     callPeriodically(this.fetchAvailableSerialPorts, 5000)
   },
   methods: {
     async fetchAvailableBridges(): Promise<void> {
+      if (this.prefetched_bridges && !bridget.should_fetch) {
```
patrickelectric, line 35: It would be nice that we could encapsulate the should_fetch logic inside a component that works with axios to be used in the others components without copy/paste or duplicated code.

core/frontend/src/store/bridget.ts
```diff
@@ -43,9 +53,91 @@ class BridgetStore extends VuexModule {
...
+    const ref = new WeakRef(object)
+    const id = setInterval(() => {
+      // Check if object does not exist anymore or if it was destroyed by vue
+      // eslint-disable-next-line
+      if (ref.deref()?._isDestroyed) {
+        this.deregisterListener()
+        clearInterval(id)
+      }
+    }, 1000)
+  }
```
rafaellehmkuhl: Can we somehow move the register/deregister/listener/object logic to a separate file, to be commonly used by all stores?

patrickelectric (changes requested):
We could just squash the commits on the necessary places to merge and encapsulate the fetch logic in another PR. I'll open an issue to register this task.

patrickelectric (approved):

PR #798, opened 2022-02-17
More udev rules

khancyr:
Add some  more  rules on udev:
make  modemmanager ignore autopilot boards :  really  usefull when you are using some 4/5G modems
make  the same for USB-UART converter, make them write enable by default for all user (maybe  do  the same for autopilot?)


patrickelectric (approved):

patrickelectric:
@khancyr thanks!

Issue #784, opened 2022-02-08
bug: Clean install never completes unless you background bootstrap


matt-bathyscope:
### Bug description

When running `install.sh` on a clean install of Raspberry Pi OS the installation script never completes unless you background the execution of the bootstrap container.

### Steps to reproduce

1. Install Raspberry Pi OS image on a new micro SD card
2. Run the installation script
3. Script never proceeds beyond the `docker run ...` [here](https://github.com/bluerobotics/BlueOS-docker/blob/master/install/install.sh#L130) because of the infinite loop [here in bootstrap.py](https://github.com/bluerobotics/BlueOS-docker/blob/master/bootstrap/bootstrap/bootstrap.py#L234)

### Primary pain point(s)

You have to manually background the execution of the bootstrap Docker container because of the infinite loop `bootstrap.py` that prints `core is already running, waiting for it to stop...` forever.

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

matt-bathyscope:
Output snippet from running `installer.sh`:

```
Going to install companion-docker version master.
Downloading and installing udev rules.
Disabling automatic Link-local configuration in dhcpd.conf.
Downloading bootstrap
master: Pulling from bluerobotics/companion-bootstrap
20cc49690d90: Pull complete 
8dd7fdf9ae3c: Pull complete 
da9874a7d833: Pull complete 
a44ce8d2fc5a: Pull complete 
450d05e80c18: Pull complete 
Digest: sha256:93b9877da3e32b4b88ce3fa6b6b3f5b5938118fbefd77917a0f1d4a618c1cee9
Status: Downloaded newer image for bluerobotics/companion-bootstrap:master
docker.io/bluerobotics/companion-bootstrap:master
WARNING: The requested image's platform (linux/arm/v7) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
unable to read startup.json file ([Errno 2] No such file or directory: '/root/.config/bootstrap/startup.json'), reverting to defaults...
Done
Starting bluerobotics/companion-core
Core started
Done
Could not talk to version chooser for 63.13048458099365: HTTPConnectionPool(host='localhost', port=80): Max retries exceeded with url: /version-chooser/v1.0/version/current (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0xf703bd10>: Failed to establish a new connection: [Errno 111] Connection refused'))
Reseting startup.json...
[...]

PR #812, opened 2022-02-24
CI: Implement native deployment of .img files

Williangalvani:
This builds images from scratch using our install install scripts. This currently requires a self-hosted armv7 runner.

It will save artifacts for branches, tags, and manually-dispatched actions. the image generation starts only after the docker build workflow is finished, as it consumes the docker image built there.

To-do:

 - [x] set startup.json to current thingy
 - [x] deploy tags
 - [x] allow forcing workflow run
 - [x] Tidy-up
 - [ ] ~alternative url for fetching files? fetch locally?~
 - [x] split into commits 
 - [x] add instructions to setup the raspberry pi runner (just install docker and the runner)
 - [ ] ~name artifacts~ https://github.com/actions/upload-artifact/issues/231
 - [x] only run after images are deployed to dockerhub
 - [x] check if expand_fs.sh is required at all (nope, not required)
 - [x] use BlueOS instead of companion
 - [x] fix rf-kill

for a next pr:
-  make it run on amd64 properly ( iptables issue)

there's an artifact here:
https://github.com/Williangalvani/companion-docker/runs/5325689537?check_suite_focus=true


deploy/pimod/companion.Pifile
```diff
@@ -1,14 +1,22 @@
...
-# Replace current companion version and update bootstrap to start it
-RUN sudo docker images
-RUN sudo docker rmi --force bluerobotics/companion-core:1.0.0.beta1
-RUN sudo docker pull bluerobotics/companion-core:1.0.0.beta4
-RUN sed -i 's/1.0.0.beta1/1.0.0.beta4/g' /root/.config/companion/bootstrap/startup.json
+RUN touch /boot/ssh
+
+RUN bash -c '
+EXPAND="/usr/bin/expand_fs.sh" &&
+sed -i "\%$EXPAND%d" /etc/rc.local || echo "sed failed to remove expand_fs entry in /etc/rc.local"
```
Williangalvani, line 13: this may be not required, as we never actually boot the image in CI...

Williangalvani: yep. not required.

install/install.sh
```diff
@@ -93,10 +101,28 @@ echo "Checking for docker."
...
-    echo "Please report this problem."
-    exit 1
+    echo "Are in in CI? trying dind."
 )
 systemctl enable docker
+if [ $RUNNING_IN_CI -eq 1 ]
+then
+
+
+set -x && \
```
patrickelectric, line 111: there are extra spaces here on the beginning and the indentation level is wrong. 

Can't you also use:
```
set -x
    code...
set +x 
``` 
?

patrickelectric, line 111: A comment explaining what is going on may help (what is dind, who is dind, why dind is important, what dind is doing and etc).

install/install.sh
```diff
@@ -93,10 +101,28 @@ echo "Checking for docker."
...
+
+    dind dockerd $DOCKER_EXTRA_OPTS &
+      while(! docker info > /dev/null 2>&1); do
+        echo "==> Waiting for the Docker daemon to come online..."
+        sleep 1
+      done
+    alias docker=dind
+fi
+
+sudo usermod -aG docker pi
```
patrickelectric, line 125: Do we want to give normal user docker access ?

install/install.sh
```diff
@@ -5,6 +5,9 @@ VERSION="${VERSION:-master}"
 REMOTE="${REMOTE:-https://raw.githubusercontent.com/${GITHUB_REPOSITORY}}"
 ROOT="$REMOTE/$VERSION"
 
+DIND_COMMIT="52379fa76dee07ca038624d639d9e14f4fb719ff"
+curl -fL -o /usr/local/bin/dind "https://raw.githubusercontent.com/moby/moby/${DIND_COMMIT}/hack/dind" && chmod +x /usr/local/bin/dind
```
patrickelectric, line 9: We could do it inside the CI check.

install/install.sh
```diff
@@ -151,12 +151,15 @@ sed -i '$ a noipv4ll' /etc/dhcpcd.conf
...
 echo "Downloading bootstrap"
 COMPANION_BOOTSTRAP="bluerobotics/companion-bootstrap:master" # We don't have others tags for now
+BLUEOS_CORE="bluerobotics/companion-core:$VERSION" # We don't have a stable tag yet
+
 docker pull $COMPANION_BOOTSTRAP
-# Start bootstrap for the first time to fetch the other images and allow docker to restart it after reboot
-docker run \
+docker pull $BLUEOS_CORE
+# Create blueos-bootstrap container
+docker create \
```
patrickelectric, line 159: What is the problem with docker run here to do the register done by https://github.com/bluerobotics/BlueOS-docker/pull/812/commits/b0449894a1330fff32c0af6bf0f736ea96361430 ?

.github/workflows/deploy_native.yml
```diff
@@ -0,0 +1,63 @@
...
+              echo "NEED=true" >> "$GITHUB_ENV"
+          else
+              echo "NEED=false" >> "$GITHUB_ENV"
+          fi
+      -
+        name: Pimod Build
+        if: env.NEED == 'true'
+        # We use our own pimod as upstream doesn't provide armv7 images
+        run: |
+          wget https://raw.githubusercontent.com/williangalvani/pimod/master/pimod.sh && chmod +x pimod.sh
```
patrickelectric, line 40: We should open a PR in pimod and request for armv7.

patrickelectric (changes requested):

.github/workflows/deploy_native.yml
Williangalvani: idk about that. we are kind of mis-using it from what I understand. what it does here could be done by chroot itself. I chose to keep it because we can work on top of it to get amd64 working, plus the PiFile is extremely useful.

install/install.sh
Williangalvani: this is just cleaner, as we don't need to kill bootstrap or have it kill itself in the first run.
this fixes https://github.com/bluerobotics/BlueOS-docker/issues/784, btw

install/install.sh
Williangalvani: good question. I'll try to run it whitout this line and see what happens

Williangalvani:
I'll squash the new commits before merging.

Williangalvani:
there's a working artifact here:
https://github.com/Williangalvani/companion-docker/actions/runs/1928402458

install/boards/bcm_27xx.sh
```diff
@@ -3,7 +3,7 @@
 echo "Configuring BCM27XX board (Raspberry Pi 4).."
 
 VERSION="${VERSION:-master}"
-REMOTE="${REMOTE:-https://raw.githubusercontent.com/bluerobotics/companion-docker}"
+REMOTE="${REMOTE:-https://raw.githubusercontent.com/${GITHUB_REPOSITORY}}"
```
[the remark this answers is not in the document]
patrickelectric: The problem here is that if this variable is not set, like, someone is calling `install.sh` the variable will not exist. There should be a default value. `REMOTE="${REMOTE:-https://raw.githubusercontent.com/${GITHUB_REPOSITORY:-bluerobotics/companion-docker}}"`
Local variables are also not shared between scripts, so if you set a default `GITHUB_REPOSITORY` in install.sh, the variable will not be visible for any other script that is invoked by it.

.github/workflows/deploy_native.yml
```diff
@@ -0,0 +1,66 @@
...
+      matrix:
+        platforms: ["linux/arm/v7"]
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+
+      -
+        name: Set env SHOULD_RUN
+        # sets a variabel that means we should run on branches/tags and manual dispatchs only. (no PRs)
```
patrickelectric, line 27: Can't you use only `github.event.issue.pull_request` ? and use directly inside the if configuration.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only

install/boards/bcm_27xx.sh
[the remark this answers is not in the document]
patrickelectric: image

patrickelectric (changes requested):

patrickelectric (commented):
image

patrickelectric (approved):

PR #822, opened 2022-02-28
core: frontend: Improve main menu for small and big screens 

patrickelectric:
Peek 28-02-2022 18-08

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -12,6 +12,7 @@
       >
         <v-app-bar-nav-icon
           id="hamburguer-menu-button"
+          :style="{visibility: drawer ? 'hidden' : 'visible'}"
           color="white"
           @click="drawer = true"
         />
@@ -211,7 +212,7 @@ export default Vue.extend({
 
   data: () => ({
     settings,
-    drawer: false,
+    drawer: undefined as boolean|undefined,
     drawer_running_tour: false,
     backend_offline: false,
     menus,
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -4,6 +4,7 @@
       flat
     >
       <v-app-bar
+        app
         rounded="0"
         :color="app_bar_color"
         :extended="extend_toolbar"
@@ -46,8 +47,8 @@
 
     <v-navigation-drawer
       v-model="drawer"
-      absolute
-      temporary
+      app
+      fixed
       @input="drawerEvent"
     >
       <v-container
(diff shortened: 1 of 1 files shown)


core/frontend/src/App.vue
```diff
@@ -41,97 +36,116 @@
...
+      class="fill-height"
+      no-gutters
     >
-      <v-container
-        elevation="0"
-        class="d-flex justify-center"
+      <v-navigation-drawer
+        id="navigator-drawer"
+        permanent
+        :expand-on-hover="is_small_screen"
```
Williangalvani, line 46:
```suggestion
        :expand-on-hover="$vuetify.breakpoint.mdAndDown"
```

Williangalvani: we shold use breakpoints where we can:
https://vuetifyjs.com/en/features/breakpoints/#thresholds

Williangalvani (changes requested):

core/frontend/src/App.vue
patrickelectric: Good catch

patrickelectric:
There is still a problem where when the settings/power/feedback options are open while the screen is small, on the button is clicked, the navbar lost focus, goes to the minimized size and the dialog that just opened get closed.

rafaellehmkuhl:
I would like to use the mini-variant collapsed for medium screens, and hiding the drawer only for smaller ones, but the problem in this case is that all the power-like buttons and the git info have to be hidden when the drawer is not hovered, and the mini-variant API apparently has a bug where the sync property, which allows us to listen for hovering, is not two-way-binded, and when used we cannot force the mini-variant just for some specific screen size. I'm making a minimum-code example to reproduce this bug and try to get some help from the vuetify team.

I ended up not using the mini-variant and just collapsing the drawer when the screen is medium-down and making it fixed on bigger screens. This behavior leaves more space to the main content.

Behavior was tested and adapted to make the tour work without problems.

ezgif-2-b47c9e41b8

core/frontend/src/App.vue
```diff
@@ -211,7 +212,7 @@ export default Vue.extend({
 
   data: () => ({
     settings,
-    drawer: false,
+    drawer: undefined as boolean|undefined,
```
patrickelectric, line 214: What is the reason to `drawer` to be an union and not a simple boolean ?

rafaellehmkuhl: With this patch, the drawer can start as both open or close, depending on the user screen size. As we don't know how it is at start, we make it initial value undefined, and the v-model will take care of updating at mounting time.

PR #847, opened 2022-03-04
core: ardupilot manager: use BETA firmware for Navigator and SITL

Williangalvani:
as weird as it may seem, we have more control over betas than dev versions.


patrickelectric (approved):
image

PR #848, opened 2022-03-04
CI: Add releases for tags and update code style

patrickelectric:


.github/workflows/test-and-deploy.yml
```diff
@@ -227,16 +227,32 @@ jobs:
             --workdir=/files \
             --platform ${{ matrix.platforms }} williangalvani/pimod:latest pimod.sh deploy/pimod/blueos.Pifile
 
+      # TODO: add GITHUB_REF_NAME after https://github.com/actions/upload-artifact/issues/231 is fixed
+      # name: blueos-${{ env.GITHUB_REF_NAME }}.zip
+      -
+        name: Zip image
```
Williangalvani: isn't this going to generate a zip-in-a-zip?
current artifact is already a zip

patrickelectric: That appears to be the case:
"[One of the consequences of this limitation is that if a zip is uploaded during a workflow run and then downloaded from the UI, there will be a double zip created.](https://github.com/actions/upload-artifact#permission-loss)"

Williangalvani (approved):

Issue #850, opened 2022-03-04
core: services: commander: Allow usage of arbitrary linux commands


patrickelectric:
Allow docker to send arbitrary linux commands to main linux

Williangalvani:
related to https://github.com/bluerobotics/BlueOS-docker/issues/175

PR #853, opened 2022-03-04
Fix shutdown/reboot

### LINKED ISSUE #563: core: frontend: loading icon does not appear when clicking reboot

patrickelectric:
docker: patrickelectric/commander_power

Fix #563 
Helps #850

### DIFF
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -1,9 +1,10 @@
 #! /usr/bin/env python3
 import logging
 import shutil
+import subprocess
 from enum import Enum
 from pathlib import Path
-from typing import Any, Callable
+from typing import Any
 
 import appdirs
 import uvicorn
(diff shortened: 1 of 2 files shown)


core/services/commander/main.py
```diff
@@ -34,6 +35,25 @@ class ShutdownType(str, Enum):
     POWEROFF = "poweroff"
 
 
+def run_command(command: str) -> subprocess.CompletedProcess:
+    user = "pi"
+    password = "raspberry"
```
rafaellehmkuhl, line 40: It would be good to have this somewhere on the installation script and import the config for both from some fixed place.

rafaellehmkuhl (changes requested):

core/services/commander/main.py
patrickelectric: We are already doing the same thing for filebrowser, we can fix the user/login configuration in a future PR.

rafaellehmkuhl: k

Williangalvani (approved):

PR #872, opened 2022-03-17
components: health: Fix heartbeat icon to use only vehicle heartbeat

### LINKED ISSUE #838: bug: Hearbeat icon beating fine when only a camera is alive
### Bug description

image

image

### Steps to reproduce

1. Upgrade Navigator firmware to Dev (which is broken)
2. Ardupilot-manager will fail to start the firmware and enter a restart loop
3. Watch the heartbit icon and Mavlink inspector

### Primary pain point(s)

User thinks everything is fine when it isn't.

### Prerequisites

patrickelectric:
Fix #838 

### DIFF
--- a/core/frontend/src/components/health/HealthTrayMenu.vue
+++ b/core/frontend/src/components/health/HealthTrayMenu.vue
@@ -46,7 +46,7 @@
           mdi-lightning-bolt
         </v-icon>
         <v-icon
-          v-if="heartbeat_age() < 3000"
+          v-if="heartbeat_age() < time_limit_heartbeat"
           class="px-1"
           :color="`rgba(255,255,255,${0.4 + (1000-heartbeat_age())/1000}`"
           title="MAVLink heartbeats arriving as expected"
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/health/HealthTrayMenu.vue
```diff
@@ -143,12 +149,20 @@ export default Vue.extend({
...
+    heartbeat_age(): number | undefined {
+      const heartbeat = mavlink_store_get(mavlink, 'HEARTBEAT') as MavlinkMessage
+      const last_date = heartbeat.timestamp
       if (last_date === undefined) {
-        return 5000
+        this.last_date_diff = 5000
+        return this.last_date_diff
+      }
+
+      if (heartbeat?.messageData.header.system_id !== 1 && heartbeat?.messageData.header.component_id !== 1) {
```
rafaellehmkuhl, line 160: Shouldn't it be `if x or y`? As I understand a vehicle has both system id and component id equal to 1, right?

rafaellehmkuhl (changes requested):

core/frontend/src/components/health/HealthTrayMenu.vue
patrickelectric: Yep

patrickelectric: good catch, fixed!

Williangalvani (approved):

PR #877, opened 2022-03-18
Allow frontend to set system time

### LINKED ISSUE #74: Add a mechanism for syncing clock time with topside

patrickelectric:
Fix #74 

docker: patrickelectric/blueos-core:set_time

### DIFF
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -2,6 +2,7 @@
 import logging
 import shutil
 import subprocess
+import time
 from enum import Enum
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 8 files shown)


core/services/commander/main.py
```diff
@@ -87,6 +88,18 @@ async def command_host(response: Response, command: str, i_know_what_i_am_doing:
         return {"message": f"{error}"}
 
 
+@app.post("/set_time", status_code=status.HTTP_200_OK)
+@version(1, 0)
+async def set_time(response: Response, unix_time_seconds: int, i_know_what_i_am_doing: bool = False) -> Any:
+    unix_time_seconds_now = int(time.time())
+    if abs(unix_time_seconds_now - unix_time_seconds) < 30 * 60:
```
Williangalvani, line 95: can you think of any reasons not to have a lower number here? 30 minutes is enough time to turn off the rov and move it somewhere else, which would result in some logs with the wrong time. not a great issue, but slightly inconvenient

patrickelectric: Any time may be arbitrary, what you suggest ? 5min ?

core/services/commander/main.py
```diff
@@ -87,6 +88,18 @@ async def command_host(response: Response, command: str, i_know_what_i_am_doing:
...
+        return {"message": f"External time ({unix_time_seconds}) is close to internal time ({unix_time_seconds_now})."}
```
Williangalvani, line 96: add a ", not updating." in the end.

core/services/commander/main.py
patrickelectric: Updated to 5min.

Williangalvani (commented):
Looks good. any idea if we should be worried about timezones?
or is everything utc?

any tips to test it?

patrickelectric:
> Looks good. any idea if we should be worried about timezones? or is everything utc?
> 
> any tips to test it?

`getTime` return time in `unix time format`, that is [by default UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_ecmascript_epoch_and_timestamps). the _date command_ accepts unix time with the `@`, that is also UTC.

core/services/commander/main.py
patrickelectric: Done.

Williangalvani (approved):

Issue #680, opened 2021-11-22
Allow changing companion.local to something else


patrickelectric:
Users can have multiple vehicles on the same network, we should allow them to change the mdns name.

rafaellehmkuhl:
This use-case came earlier then expected haha

patrickelectric:
We should have a single docker to take care of mdns/avahi for us.

rafaellehmkuhl:
I would also add the possibility of giving a "pretty" name to each companion, and put that on some easy place (maybe together with the git info?). This way users can identify each ROV by how they call them already (Rufus? Steve? BlueWale?).

rafaellehmkuhl:
From an external discussion with @patrickelectric and @Williangalvani came the idea of making `companion.local` point to a "companion select page" instead of the regular frontend. This page would be the same for all companions and would list all companions available on the network, with their respective mdns as links to their proper frontends.

This way we avoid other problems like one companion fetching data from the other because of the common `companion.local` address.

Williangalvani:
from my tests, it looks like a restart of the avahi service is required in order to change the hostname. You can change services on-the-fly, but changing the hostname requires a restart. so we may really need that to be on it's own container, or somehow restart it in the host (if we just just bind the avahi config into core)

Williangalvani:
Changing `host-name=` in /etc/avahi/avahi-daemon.conf followed by a system restart seems to be enough for updating the published hostname.

patrickelectric:
This is now possible

PR #878, opened 2022-03-18
Core: create Beacon

### LINKED ISSUE #360: mdns provides docker ip address for outside services :eyes:
We should somehow blacklist the docker network interface

### LINKED ISSUE #472: Avahi is publishing ipv6 addresses
this can be fixed my setting ipv6=no in `/etc/avahi/avahi-daemon.conf`

### LINKED ISSUE #849: core: Add mdns service
- [ ] Allow us to have more control over mdns
- [ ] We should disable avahi in our image
- [ ] Do not use IP address of docker #360

Williangalvani:
image

This is a very configurable mDNS alternative to avahi.

fix #472 
fix #360 
fix #849 

lays the groundwork for #680 

image is live at williangalvani/blueos-core:beacon

I'll go through this one more time to tidy up and add comments/logging/better exception handling

### DIFF
--- a/core/services/beacon/main.py
+++ b/core/services/beacon/main.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+import argparse
+import asyncio
+import logging
+import pathlib
+import socket
+from typing import Any, Dict, List, Optional
+
+import psutil
+from commonwealth.settings.manager import Manager
+from loguru import logger
+from zeroconf import IPVersion
+from zeroconf.asyncio import AsyncServiceInfo, AsyncZeroconf
+
+from settings import ServiceTypes, SettingsV1
+
+
+class AsyncRunner:
+    def __init__(self, ip_version: IPVersion, interfaces: List[str]) -> None:
+        self.ip_version = ip_version
+        self.aiozc: Optional[AsyncZeroconf] = None
+        self.interfaces: List[str] = interfaces
+        self.services: List[AsyncServiceInfo] = []
+
+    def add_services(self, service: AsyncServiceInfo) -> None:
+        logger.info("Adding services:")
+        logger.info(service)
+        self.services.append(service)
+
+    async def register_services(self) -> None:
+        self.aiozc = AsyncZeroconf(ip_version=self.ip_version, interfaces=[self.interfaces])  # type: ignore
+        tasks = [self.aiozc.async_register_service(info, cooperating_responders=True) for info in self.services]
+        background_tasks = await asyncio.gather(*tasks)
+        await asyncio.gather(*background_tasks)
+        logger.info("Finished registration, press Ctrl-C to exit...")
+
+    async def unregister_services(self) -> None:
+        assert self.aiozc is not None
+        tasks = [self.aiozc.async_unregister_service(info) for info in self.services]
+        background_tasks = await asyncio.gather(*tasks)
+        await asyncio.gather(*background_tasks)
+        await self.aiozc.async_close()
+
+    def __eq__(self, other: Any) -> bool:
+        return {str(service) for service in self.services} == {str(service) for service in other.services} and set(
+            self.interfaces
+        ) == set(other.interfaces)
+
+    def __repr__(self) -> str:
+        return f"Runner on {self.interfaces}, serving {[service.name for service in self.services]}."
+
+
+class Beacon:
+    def __init__(self) -> None:
+        self.runners: Dict[str, AsyncRunner] = {}
+        self.manager = Manager("Beacon", SettingsV1)
+        # manager still returns "valid" settings even if file is absent, so we check for the "default" field
+        # TODO: fix after https://github.com/bluerobotics/BlueOS-docker/issues/880 is solved
+        if self.manager.settings.default is None:
+            logger.warning("No configuration found, loading defaults...")
+            current_folder = pathlib.Path(__file__).parent.resolve()
+            default_settings_file = current_folder / "default-settings.json"
+            logger.debug("loading settings from ", default_settings_file)
+            self.manager.settings = self.manager.load_from_file(SettingsV1, default_settings_file)
+            self.manager.save()
+        self.settings = self.manager.settings
+        self.service_types = self.load_service_types()
+
+    def load_service_types(self) -> Dict[str, ServiceTypes]:
+        """
+        load services from settings as a dictionary
+        """
+        services = {}
+        for service in self.settings.advertisement_types:
+            services[service.name] = service
+        return services
+
+    def create_async_service_infos(
+        self, interface: str, service_name: str, domain_name: str, ip: str
+    ) -> AsyncServiceInfo:
+        """
+        Create A list of AsyncServiceInfo() for the given interface and service
+        Each domain results in a new service
+        """
+        service = self.service_types[service_name]
+        try:
+            return AsyncServiceInfo(
+                f"{service.name}.{service.protocol}.local.",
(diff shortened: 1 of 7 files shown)


Williangalvani:
no idea of how I closed this -.-

core/services/beacon/settings.py
```diff
@@ -0,0 +1,110 @@
...
+            return str(self.ip)
+        except OSError:
+            pass
+
+        if "ips[" in self.ip:
+            index = int(re.findall(r"ips\[(\d+)]", self.ip)[0])
+            if not index:
+                index = 0
+            return str(self.get_phys()[index].address)
+        return str(self.get_phys()[0].address)
```
patrickelectric, line 52:
```suggestion
        index = 0
        if "ips[" in self.ip:
            matches = re.findall(r"ips\[(\d+)]", self.ip)
            if len(matches) != 0:
                index = int(match)
        return str(self.get_phys()[index].address)
```
This is a safest alternative.

core/services/beacon/settings.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def get_ip(self) -> bytes:
+        """
+        returns the ip as 4 bytes
+        """
+        return socket.inet_aton(self.get_ip_str())
```
patrickelectric, line 58: Be aware this function can trow an exception and we are not catching, it may be a problem for IPv6 only networks or any crazy thing.

core/services/beacon/settings.py
```diff
@@ -0,0 +1,110 @@
...
+    name = StringField()
+    domain_names = ListField(item_type=str)
+    advertise = ListField(item_type=str)
+    ip = StringField()
+
+    def get_phys(self) -> psutil._common.snicaddr:
+        """
+        return Physical interface from psutil
+        """
+        return psutil.net_if_addrs()[self.name]
```
patrickelectric, line 34: What happens if the interface does not exist anymore ? Like if we remove an ethernet adapter.

core/services/beacon/main.py
```diff
@@ -0,0 +1,144 @@
...
+        await asyncio.gather(*background_tasks)
+        await self.aiozc.async_close()
+
+
+class Beacon:
+    def __init__(self) -> None:
+        self.runners: List[AsyncRunner] = []
+        self.manager = Manager("Beacon", SettingsV1)
+        self.manager.load()
+        # manager still returns "valid" settings even if file is absent, so we check for the "default" field
```
patrickelectric, line 52: Wait that appears to be an issue, can you open one and add an example to replicate it ?
Can you also tag such issue on this comment with a TODO: ?

patrickelectric (changes requested):

core/services/beacon/settings.py
Williangalvani: this code runs from checking the available interfaces, so it can't fail right here. That does open the question of how python-zeroconf handles it, though, which I haven't tested yet

core/services/beacon/settings.py
Williangalvani: yep, much better.

core/services/beacon/settings.py
Williangalvani: that said, I don't think we need to support hot-plugging of ethernet adapters.

core/services/beacon/settings.py
patrickelectric: you can also remove the first if, since the regex will catch it

core/services/beacon/main.py
Williangalvani: https://github.com/bluerobotics/BlueOS-docker/issues/880

core/services/beacon/settings.py
Williangalvani: added try/except blocks when creating runners and individual services

patrickelectric (changes requested):
I tested it from docker and it appears that is using the mac address as IP address

image

Williangalvani:
> I tested it from docker and it appears that is using the mac address as IP address
> 
> image

This should be fixed in the latest image. as a side-effect, it now handles changes in settings/interfaces

core/services/beacon/settings.py
```diff
@@ -0,0 +1,117 @@
...
+
+        index = 0
+        matches = re.findall(r"ips\[(\d+)]", self.ip)
+        if len(matches) != 0:
+            index = int(matches[0])
+        address = str(self.get_phys()[index].address)
+        if address.count(":") > 0:
+            # this interface returned the mac address instead of ip, abort
+            raise InvalidIpAddress("Got a mac address instead of ip address")
+        return str(address)
```
patrickelectric, line 59: You can simplify the logic and use a general check with `inet_aton`
```suggestion        
        address = str(self.ip)
        # Check for 'ips[n]'
        matches = re.findall(r"ips\[(\d+)]", self.ip)
        if len(matches) != 0:
            index = int(matches[0])
            address = str(self.get_phys()[index].address)
        # Validate ip
        try:
            socket.inet_aton(address)
            return address
        except OSError:
            raise InvalidIpAddress(f"Invalid IP address: {address}")
        return address
```

core/services/beacon/settings.py
```diff
@@ -0,0 +1,117 @@
...
+    name = StringField()
+    domain_names = ListField(item_type=str)
+    advertise = ListField(item_type=str)
+    ip = StringField()
+
+    def get_phys(self) -> psutil._common.snicaddr:
+        """
+        return Physical interface from psutil
+        """
+        return psutil.net_if_addrs()[self.name]
```
patrickelectric, line 38: You can use [AF_INET](https://docs.python.org/3/library/socket.html#socket.AF_INET) as `family` argument to avoid the mac address and IPv6.

patrickelectric (changes requested):

install/network/avahi.sh
```diff
@@ -22,6 +22,13 @@ AVAHI_SERVICE_PATH="/etc/avahi/services"
 )
 curl -fsSL $CONFIGURE_NETWORK_PATH/blueos.service > "${AVAHI_SERVICE_PATH}/blueos.service"
 
+AVAHI_DAEMON_CONFIG_PATH="cat /etc/avahi/avahi-daemon.conf"
+[ ! -e "${AVAHI_DAEMON_CONFIG_PATH}" ] && (
+    echo "Avahi daemon config file does not exist: ${AVAHI_DAEMON_CONFIG_PATH}"
+    exit 1
+)
+sed -i "s/#host-name=foo/host-name=blueos-avahi/g" ${AVAHI_DAEMON_CONFIG_PATH}
+
```
patrickelectric, line 31: This appears to be a bit weird, why are we failing if avahi is not available since is not required for the system to work (since now we have beacon).
Also, the `cat` will make all other lines to not work.
The correct would be:

```suggestion
AVAHI_DAEMON_CONFIG_PATH="/etc/avahi/avahi-daemon.conf"
AVAHI_HOST_NAME="blueos-avahi"
[ ! -d "${AVAHI_DAEMON_CONFIG_PATH}" ] && (
    echo "Avahi daemon config file found, changing avahi host name to: ${AVAHI_HOST_NAME}"
    sed -i "s/#host-name=/host-name=blueos-avahi/g" ${AVAHI_DAEMON_CONFIG_PATH}
) || echo "Avahi daemon config file not found in ${AVAHI_DAEMON_CONFIG_PATH}"
```

This also remove the old "foo" name to make the command more future proof.

patrickelectric (changes requested):

patrickelectric (approved):

Issue #601, opened 2021-10-19
core: frontend: Check for new versions when the webpage is open for the first time 


patrickelectric:

patrickelectric:
Done in #887

Issue #599, opened 2021-10-19
core: frontend: Notify user when a new version is available to update


patrickelectric:
We should also have a checkbox/combobox on settings to allow the user to subscribe on:
    - Cutting edge: The user will be notified when a new beta or stable is available, the newer one wins 
    - Bleeding edge: Everytime that a new version (master, beta, stable) is available, the newer one wins
        - Probably the same as master 

By default, if the checkbox is not selected, the user should be notified:
    - If on beta: About new betas until it reaches a new stable release
    - If on stable: Only about stable releases

patrickelectric:
Done in #887

PR #879, opened 2022-03-19
core: frontend: utils: Add version_chooser

patrickelectric:
Helps with #599 and #601 

docker: `patrickelectric/new_version_chooser_of_post_apocalyptic_armageddon`

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionChooser.vue
+++ b/core/frontend/src/components/version-chooser/VersionChooser.vue
@@ -176,8 +176,10 @@ import Vue from 'vue'
 
 import settings from '@/libs/settings'
 import { Dictionary } from '@/types/common'
-import { Version, VersionsQuery } from '@/types/version-chooser'
+import { Version, VersionsQuery, VersionType } from '@/types/version-chooser'
 import back_axios from '@/utils/api'
+// Version Chooser Utils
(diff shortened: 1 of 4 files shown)


core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -297,70 +257,35 @@ export default Vue.extend({
...
-        this.latest_stable = this.getLatestStable()
-      })
+
+      await VCU.loadAvailableVersions(this.selected_image)
+        .then((versions_query) => {
+          this.available_versions = versions_query
+        })
+        .finally(() => {
+          this.loading_images = false
+          this.available_versions = VCU.sortImages(this.available_versions)
```
Williangalvani, line 270: is this lines required? they are [already sorted by loadAvailableVersions](https://github.com/bluerobotics/BlueOS-docker/pull/879/files#diff-50ff488bb99812e1025436fc7d2a5701694d2523a2174814071dc9cf81173245R113)

patrickelectric: Well, better safe than sorry. We may change the utils to be not sorted by default. We can also make it not sorted (but there is no advantage for that).

Williangalvani (approved):

PR #884, opened 2022-03-21
frontend: video: Allow multiple IPs for each stream

### LINKED ISSUE #867: UI: Allow more than one IP on stream creation
`mavlink-camera-manager` supports it, but our frontend doesn't (validation kicks in).

Maybe a plus button besides the IP input box would be cool.

rafaellehmkuhl:
image

Fix #867

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStream.vue
+++ b/core/frontend/src/components/video-manager/VideoStream.vue
@@ -22,8 +22,13 @@
                 <tr>
                   <td>{{ stream.video_and_stream.stream_information.configuration.encode }}</td>
                 </tr>
-                <tr>
-                  <td>{{ stream.video_and_stream.stream_information.endpoints[0] }}</td>
+                <tr
+                  v-for="(endpoint, index) in stream.video_and_stream.stream_information.endpoints"
+                  :key="index"
+                >
+                  <td>
+                    {{ endpoint }}
+                  </td>
                 </tr>
                 <tr>
                   <td>{{ source_path }}</td>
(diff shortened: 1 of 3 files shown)


Williangalvani (changes requested):
Nice. Can you add a title/popover to the + button saying "Add new stream"?

Also, we need more validation. IIRC we only support multiple endpoints for UDP, and we need to ensure that the user doesn't try to mix rtsp and udp endpoints:

image

rafaellehmkuhl:
> Nice. Can you add a title/popover to the + button saying "Add new stream"?
> 
> Also, we need more validation. IIRC we only support multiple endpoints for UDP, and we need to ensure that the user doesn't try to mix rtsp and udp endpoints:
> 
> image

Good catch. I wrongly assumed the behavior would be the same for RTSP and just tested UDP.

So to get it correctly:
- Can we have multiple IPs for UDP?
- Are we limited to one RTSP IP per stream?
- Can we have multiple multiple UDPs and one RTSP for the same stream?

@patrickelectric or @joaoantoniocardoso maybe can answer this better.

patrickelectric:
> 

An UDP stream can have multiple UDP endpoint, a RTSP stream should have a single endpoint, you can't mix UDP and RTSP.

patrickelectric (changes requested):
The style is a bit weird with the plus icon on the side of endpoint when there is multiple endpoints, a button bellow the Stream endpoint would be better.
image

We could also only enable the "Add" button if it's an UDP endpoint.

rafaellehmkuhl:
> The style is a bit weird with the plus icon on the side of endpoint when there is multiple endpoints, a button bellow the Stream endpoint would be better. image
> 
> We could also only enable the "Add" button if it's an UDP endpoint.

image

Personally, I didn't like it. The plus sign is very intuitive and don't bloat the UI. @Williangalvani @joaoantoniocardoso what do you think?

rafaellehmkuhl:
> > The style is a bit weird with the plus icon on the side of endpoint when there is multiple endpoints, a button bellow the Stream endpoint would be better. image
> > We could also only enable the "Add" button if it's an UDP endpoint.
> 
> image
> 
> Personally, I didn't like it. The plus sign is very intuitive and don't bloat the UI. @Williangalvani @joaoantoniocardoso what do you think?

From external voting, we keep the plus sign button.

patrickelectric (approved):

PR #885, opened 2022-03-21
core: Add global exception handler

rafaellehmkuhl:
- Use a global exception handler, so exceptions not caught are all logged and return a default HTTP status-code of 500 with the exception message appended;
- Adapt the frontend of the services to consider this new standard.

With this new approach, we guarantee every not-handled exception to be in the logs, making debug easier, both for frontend and local. Error messages more specific than "returned status code of 500" will start to come.

### DIFF
--- a/core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
+++ b/core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
@@ -39,7 +39,7 @@ export default Vue.extend({
         .catch((error) => {
           autopilot.setAvailableEndpoints([])
           if (error === backend_offline_error) { return }
-          const message = `Could not fetch available MAVLink endpoints: ${error.message}`
+          const message = error.response.data.detail ?? error.message
           notifications.pushError({ service: autopilot_service, type: 'AUTOPILOT_ENDPOINT_FETCH_FAIL', message })
         })
     },
@@ -56,7 +56,7 @@ export default Vue.extend({
         .catch((error) => {
           autopilot.setAvailableBoards([])
           if (error === backend_offline_error) { return }
-          const message = `Could not fetch available Ardupilot boards: ${error.message}`
+          const message = error.response.data.detail ?? error.message
           notifications.pushError({ service: autopilot_service, type: 'AUTOPILOT_BOARDS_FETCH_FAIL', message })
         })
     },
@@ -72,7 +72,7 @@ export default Vue.extend({
         .catch((error) => {
           autopilot.setCurrentBoard(null)
           if (error === backend_offline_error) { return }
-          const message = `Could not fetch current Autopilot board: ${error.message}`
+          const message = error.response.data.detail ?? error.message
           notifications.pushError({ service: autopilot_service, type: 'AUTOPILOT_BOARD_FETCH_FAIL', message })
         })
     },
--- a/core/services/ardupilot_manager/main.py
+++ b/core/services/ardupilot_manager/main.py
@@ -7,10 +7,10 @@ import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Set
 
-from commonwealth.utils.apis import PrettyJSONResponse
+from commonwealth.utils.apis import GenericErrorHandlingRoute, PrettyJSONResponse
 from commonwealth.utils.general import is_running_as_root
 from commonwealth.utils.logs import InterceptHandler, get_new_log_path
-from fastapi import Body, FastAPI, File, HTTPException, Response, UploadFile, status
+from fastapi import Body, FastAPI, File, HTTPException, UploadFile, status
 from fastapi.staticfiles import StaticFiles
 from fastapi_versioning import VersionedFastAPI, version
 from loguru import logger
@@ -40,6 +40,7 @@ app = FastAPI(
     default_response_class=PrettyJSONResponse,
     debug=True,
 )
+app.router.route_class = GenericErrorHandlingRoute
 logger.info("Starting ArduPilot Manager.")
 autopilot = ArduPilotManager()
 if not is_running_as_root():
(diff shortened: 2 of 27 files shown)


core/services/ardupilot_manager/main.py
```diff
@@ -88,34 +77,29 @@ def update_endpoints(endpoints: Set[Endpoint] = Body(...)) -> Any:
...
+async def install_firmware_from_url(url: str) -> Any:
     try:
         if not autopilot.current_board:
             raise RuntimeError("Cannot install firmware as there's no board running.")
         await autopilot.kill_ardupilot()
         autopilot.install_firmware_from_url(url, autopilot.current_board)
     except Exception as error:
-        response.status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
-        return {"message": f"{error}"}
+        raise error
```
patrickelectric, line 95: You don't need to raise error, you can remove the try block.

core/services/cable_guy/main.py
```diff
@@ -86,22 +80,16 @@ def configure_interface(interface: EthernetInterface = Body(...)) -> Any:
 @version(1, 0)
 def add_address(interface_name: str, mode: AddressMode, ip_address: Optional[str] = None) -> Any:
     """REST API endpoint to add an IP address to an ethernet interface."""
-    try:
-        manager.add_ip(interface_name, mode, ip_address)
-        manager.save()
-    except Exception as error:
-        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(error)) from error
+    manager.add_ip(interface_name, mode, ip_address)
+    manager.save()
```
patrickelectric, line 84: This was returning bad request and not internal error

patrickelectric (changes requested):

patrickelectric (changes requested):
CI is also failing

core/frontend/src/components/autopilot/AutopilotManagerUpdater.vue
```diff
@@ -39,7 +39,7 @@ export default Vue.extend({
         .catch((error) => {
           autopilot.setAvailableEndpoints([])
           if (error === backend_offline_error) { return }
-          const message = `Could not fetch available MAVLink endpoints: ${error.message}`
+          const message = error.response.data.detail ? error.response.data.detail : error.message
```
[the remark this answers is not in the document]
rafaellehmkuhl: Forgot the coalescing operator existed. Will use.

Regarding having a single way, it is like that, but we never have a guarantee that the error we receive is indeed from our backend and so that it will respect our standard. Imagine a network interruption occurs. The received error will not have a body, so we should deal with it.

core/services/ardupilot_manager/main.py
rafaellehmkuhl: We need the try/except/finally block to ensure we start the autopilot if everything goes wrong.

core/services/cable_guy/main.py
rafaellehmkuhl: It was because we didn't have a convention. Bad requests (400) should be returned only if we know that this error is caused by a faulty usage from the client (which we cannot, in this case).

patrickelectric (changes requested):
I just tested a simple error command and it's failing to provide the message:
```
curl -v -H "Content-Type: application/json" --request POST http://192.168.0.46:9100/v1.0/command/host\?command\='pwd'\&i_know_what_i_am_doing\=false
*   Trying 192.168.0.46:9100...
* Connected to 192.168.0.46 (192.168.0.46) port 9100 (#0)
> POST /v1.0/command/host?command=pwd&i_know_what_i_am_doing=false HTTP/1.1
> Host: 192.168.0.46:9100
> User-Agent: curl/7.82.0
> Accept: */*
> Content-Type: application/json
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< date: Tue, 22 Mar 2022 10:46:14 GMT
< server: uvicorn
< content-length: 13
< content-type: application/json
< 
* Connection #0 to host 192.168.0.46 left intact
{"detail":""}%                                                        
```

master returns:
```
curl -v -H "Content-Type: application/json" --request POST http://192.168.0.46:9100/v1.0/command/host\?command\='pwd'\&i_know_what_i_am_doing\=false
*   Trying 192.168.0.46:9100...
* Connected to 192.168.0.46 (192.168.0.46) port 9100 (#0)
> POST /v1.0/command/host?command=pwd&i_know_what_i_am_doing=false HTTP/1.1
> Host: 192.168.0.46:9100
> User-Agent: curl/7.82.0
> Accept: */*
> Content-Type: application/json
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< date: Tue, 22 Mar 2022 10:48:09 GMT
< server: uvicorn
< content-length: 75
< content-type: application/json
< 
* Connection #0 to host 192.168.0.46 left intact
{"detail":"Developer, you don't know what you are doing, command aborted."}% 
```

rafaellehmkuhl:
> curl -v -H "Content-Type: application/json" --request POST http://192.168.0.46:9100/v1.0/command/host\?command\='pwd'\&i_know_what_i_am_doing\=false

Fixed. New image in build.

rafaellehmkuhl:
@patrickelectric regarding this one, what do you think?

patrickelectric:
I'm inclined to hold master branch for now and only merge stuff that are clear safe for the 1.0.0 release, we can move that to 1.0.1, what you think ?

rafaellehmkuhl:
> I'm inclined to hold master branch for now and only merge stuff that are clear safe for the 1.0.0 release, we can move that to 1.0.1, what you think ?

I honestly don't see how could this be unsafe, but no problem.

patrickelectric (approved):
Great cleanup

PR #888, opened 2022-03-22
core: frontend: componenets: video-manager: Only show extra configuration under pirate mode

patrickelectric:
Extra configuration is a feature that is intended to be used by advanced users for now.

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -49,7 +49,9 @@
             :rules="[validate_required_field, is_valid_schema]"
           />
 
-          <v-expansion-panels>
+          <v-expansion-panels
+            v-if="settings.is_pirate_mode"
+          >
             <v-expansion-panel>
               <v-expansion-panel-header>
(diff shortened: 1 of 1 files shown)


rafaellehmkuhl:
I don't get it. Why adding an RGB stream is non-pirate and adding a thermal camera is pirate?

rafaellehmkuhl:
> > I don't get it. Why adding an RGB stream is non-pirate and adding a thermal camera is pirate?
> 
> Thermal is a hack way to have multiple streams in qgc, not a real thermal camera use

So it has nothing to do with thermal? Why do we call it that way?

patrickelectric:
> > > I don't get it. Why adding an RGB stream is non-pirate and adding a thermal camera is pirate?
> > 
> > 
> > Thermal is a hack way to have multiple streams in qgc, not a real thermal camera use
> 
> So it has nothing to do with thermal? Why do we call it that way?

That's is the official name on the mavlink spec

Williangalvani (approved):

rafaellehmkuhl (approved):
Strange but fair 😅 
I will open an issue to remember getting this back once we bypass this workaround.

PR #890, opened 2022-03-22
Add welcome tour

rafaellehmkuhl:
This is an initial implementation. We can progressively add more in-depth tours on each component/service as we want.
Image available at `patrickelectric/blueos-core:add-tour`

### DIFF
--- a/core/frontend/src/libs/settings.ts
+++ b/core/frontend/src/libs/settings.ts
@@ -33,6 +33,19 @@ class Settings {
   updateVersionUpdateNotificationTime() {
     settingsStore.updateVersionUpdateNotificationTime()
   }
+
+  // eslint-disable-next-line
+  run_tour_version(version: number) : Promise<void> {
+    return new Promise((resolve) => {
+      const store_tour_version = settingsStore.tour_version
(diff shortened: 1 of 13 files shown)


patrickelectric (changes requested):
The docker available is failing to run:
TypeError: Cannot read properties of undefined (reading 'welcomeTour')

core/frontend/src/libs/settings.ts
```diff
@@ -33,6 +33,19 @@ class Settings {
   updateVersionUpdateNotificationTime() {
     settingsStore.updateVersionUpdateNotificationTime()
   }
+
+  // eslint-disable-next-line
+  run_tour_version(version: number) : Promise<void> {
+    return new Promise((resolve) => {
+      const store_tour_version = settingsStore.tour_version
+      if (version === store_tour_version) {
+        throw new Error(`Tour version (${version}) is the same as in store (${store_tour_version}})`)
```
Williangalvani, line 42: do we really need to throw an exception here? this is throwing when opening the page after first time.

patrickelectric (approved):

Williangalvani (approved):

PR #900, opened 2022-03-23
core: frontend: components: common: DefaultTooltip: Add Arrrgh prefix in pirate mode

patrickelectric:

### DIFF
--- a/core/frontend/src/components/common/DefaultTooltip.vue
+++ b/core/frontend/src/components/common/DefaultTooltip.vue
@@ -6,12 +6,14 @@
     :left="position === 'left'"
     :activator="activator"
   >
-    <span>{{ content }}</span>
+    <span>{{ pirate_prefix }}{{ content }}</span>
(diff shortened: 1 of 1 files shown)


Williangalvani:
I"m afraid this is a bit too much... we should ask for more opinions thought

ES-Alexander (commented):
Lol, is there a screenshot of what this would look like? I'm not sure where this would appear.

Williangalvani:
> Lol, is there a screenshot of what this would look like? I'm not sure where this would appear.

there you go:

https://user-images.githubusercontent.com/4013804/159807437-b3ceab4b-e1fc-4522-8062-686e4ad03c9b.mp4

ES-Alexander:
Assuming these options are only available in pirate mode, I think a flag or crossbones symbol would be better :-)
The Arrrgh is amusing but is almost as much text as the thing it's annotating, so agreed with Willian that it's a bit much as is.

PR #910, opened 2022-03-24
core: version-chooser: fix jinja2 and itsdangerious versions

Williangalvani:


rafaellehmkuhl:
If it's a hotfix, go for it.

PR #914, opened 2022-03-24
core: version-chooser: fix displaying error

Williangalvani:
An error here means the data is incomplete, not necessarily absent.


Williangalvani:
needs testing to make sure both the notificator and version-chooser are working as expected.

patrickelectric (approved):

PR #916, opened 2022-03-25
core: frontend: improve logic for checking if backend is online in versionchooser

Williangalvani:
It will now properly wait for the backend to first go offline, and then wait for it to come back online

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionChooser.vue
+++ b/core/frontend/src/components/version-chooser/VersionChooser.vue
@@ -234,25 +234,77 @@ export default Vue.extend({
     this.loadCurrentVersion()
   },
   methods: {
-    async checkIfBackendIsOnline() {
-      this.waiting = true
-      await back_axios({
-        method: 'get',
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -234,25 +234,77 @@ export default Vue.extend({
...
+              clearInterval(interval)
+              resolve('backend went online')
+            }
+          })
+        }, 1000)
+      })
+    },
+
+    async waitForBackendToRestart(reload: boolean) {
+      this.waiting = true
```
patrickelectric, line 300: Just to be sure, there is no need to set waiting to false since the logic here is to reload the page, right ?

patrickelectric (approved):

core/frontend/src/components/version-chooser/VersionChooser.vue
Williangalvani: setting waiting to true shows up the overlay and spinner.

PR #922, opened 2022-03-25
core: commander: use 'reboot' instead of 'shutdown -r' to reboot

Williangalvani:
fixes an issue where the pi is shutting down instead of rebooting.

image will be up at williangalvani/blueos-core:reboot when CI finishes.

### DIFF
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -112,7 +112,7 @@ async def shutdown(response: Response, shutdown_type: ShutdownType, i_know_what_
     try:
         hold_time_seconds = 5
         if shutdown_type == ShutdownType.REBOOT:
-            output = run_command(f"(sleep {hold_time_seconds}; sudo shutdown --reboot -h now)&")
+            output = run_command(f"(sleep {hold_time_seconds}; sudo reboot)&")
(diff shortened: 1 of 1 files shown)


core/services/commander/main.py
```diff
@@ -115,7 +115,7 @@ async def shutdown(response: Response, shutdown_type: ShutdownType, i_know_what_
             output = run_command(f"(sleep {hold_time_seconds}; sudo shutdown --reboot -h now)&")
             logger.debug(f"reboot: {output}")
         elif shutdown_type == ShutdownType.POWEROFF:
-            output = run_command(f"(sleep {hold_time_seconds}; sudo shutdown --poweroff -h now)&")
+            output = run_command(f"(sleep {hold_time_seconds}; sudo reboot)&")
```
patrickelectric, line 118: This is replacing power off and not rebooting, is this correct ?

Williangalvani: lol it is not

patrickelectric (approved):

ES-Alexander (approved):
Assuming this has been tested and confirmed to solve the issue, seems fine :-)

Williangalvani:
tested and it works.

PR #924, opened 2022-03-25
Factory

### LINKED ISSUE #917: Add a "Factory" image in place of master in startup.json, make it impossible for users to delete it.
This will be a "factory reset" version

Williangalvani:
Fix #917 

~Do we need to populate startup.json manually now? so it uses current branch instead of 'factory'~

### DIFF
--- a/bootstrap/startup.json.default
+++ b/bootstrap/startup.json.default
@@ -1,6 +1,6 @@
 {
     "core": {
-        "tag": "master",
+        "tag": "factory",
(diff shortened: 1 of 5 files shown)


Williangalvani:
we should copy over the default file to the host using `docker cp <containerId>:/file/path/within/container /host/path/target` and then use sed to apply the correct version to it.

patrickelectric (approved):

PR #925, opened 2022-03-25
core: services: pardal: Add root page

patrickelectric:
Fix Available Services 
image


Williangalvani (approved):
fell free to merge after fixing CI

Issue #457, opened 2021-09-01
logs: Save Pixhawk/Serial logs on BlueOS


rafaellehmkuhl:
Right now Pixhawk logs are not saved on Companion, so users don't have access to it trough Log Browser.

DAYS_TO_DEVELOP:2

patrickelectric:
@Williangalvani what is the status of this issue ? We did some telemetry changes on #1226 ?

ES-Alexander:
I tried setting `LOG_BACKEND_TYPE` to MAVLink and seemingly got some very inconsistent behaviour from the Log Browser:

https://user-images.githubusercontent.com/25898329/212253723-b2a1eb45-9493-4b94-b057-e118aad21ee7.mov

Regardless of how well the live-stream-to-autopilot approach works, I still think it's worth the mavftp approach being implemented too, since if someone uses lots of CPU or memory for other things then maybe they don’t want to stream logs live, but they may want to be able to fetch them off the autopilot without needing to open their enclosure and access the logs via QGC.

patrickelectric:
Already done via tlog, but will have full support to binlogs with mavftp mavftp #926

PR #926, opened 2022-03-25
add Mavftp lib/tool

Williangalvani:
Still needs some work.
This will help  #457

thanks @ES-Alexander for the help =]


core/services/mavftp/mavftp.py
```diff
@@ -0,0 +1,680 @@
+# As this file is basically copied over from https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_ftp.py
```
patrickelectric, line 1: Can't we just install it as a module with a bootstrap script ? Over duplicating the code.

core/services/mavftp/main.py
```diff
@@ -0,0 +1,154 @@
...
+        parent_dir = ("/" + "/".join(path.split("/")[:-1])).replace("//", "/")
+        logger.debug(f"cache miss, asking for {parent_dir}")
+        self.readdir(parent_dir)
+        if path not in self.files:
+            raise FuseOSError(errno.ENOENT)
+        return self.files[path]
+
+    def read(self, path: str, size: int, offset: int, _fh: int = 0) -> Optional[bytes]:
+        buf = self.ftp.read_sector(self.fix_path(path), offset, size)
+        # logger.info(f"read result: {buf}")
```
patrickelectric, line 56: lost debug code

core/services/mavftp/main.py
```diff
@@ -0,0 +1,154 @@
...
+            return path[1:]
+        return path
+
+    def getattr(self, path: str, _fh: int = 0) -> Any:
+        path_fixed = self.fix_path(path)
+        logger.info(f"Fuse: getattr {path_fixed}")
+        if path_fixed in self.files:
+            return self.files[path_fixed]
+
+        parent_dir = ("/" + "/".join(path.split("/")[:-1])).replace("//", "/")
```
patrickelectric, line 47: can you add a comment about what is happening here ?

core/services/mavftp/main.py
```diff
@@ -0,0 +1,154 @@
...
+            for path in base_file_paths
+        }
+        self.ftp = FTPModule(mav)
+
+    def fix_path(self, path: str) -> str:
+        if path.startswith("/@"):
+            return path[1:]
+        return path
+
+    def getattr(self, path: str, _fh: int = 0) -> Any:
```
patrickelectric, line 41: can we have a better name than _fh ?

core/services/mavftp/main.py
```diff
@@ -0,0 +1,154 @@
...
+                    "st_mtime": int(time.time()),  # Current time
+                    "st_atime": int(time.time()),  # Current time
+                }
+            ret[item.name] = new_item
+            new_path = path if path.endswith("/") else path + "/"
+            self.files[new_path + item.name] = new_item
+        self.files[path] = {"st_mode": (0o46766), "st_size": 0}
+        return ret
+
+    def create(self, path: str, fi: int = 0) -> None:
```
patrickelectric, line 87: what is fi ? Can we have a better name ?

core/services/mavftp/main.py
```diff
@@ -0,0 +1,154 @@
...
+        logger.error(f"Fuse: chown {path}")
+        raise FuseOSError(errno.EROFS)
+
+    def truncate(self, path: str) -> None:
+        logger.info(f"Fuse: truncate {path}")
+        raise FuseOSError(errno.EROFS)
+
+
+if __name__ == "__main__":
+    # logging.basicConfig(level=logging.DEBUG)
```
patrickelectric, line 129: lost debug message

patrickelectric (changes requested):

core/services/mavftp/main.py
joaoantoniocardoso: is this literally `f"/{path}/../"` sanitized?

core/services/mavftp/mavftp.py
Williangalvani: the original code doesn't really have the functionality we need. but as the goal is to eventually move it up to ardupiot, perhaps I could keep it in a different repo and use a bootstrap file. thoughts?

patrickelectric: wait, but is this file the original code or not ? what was changed ? I believe that the best way to track it is to use bootstrap and a branch of ardupilot with the changes that you want to apply on their remote, so we can track the changes and also the PR.

core/services/mavftp/main.py
Williangalvani: these are mostly the exact definition from the examples in fusepy, but I agree they are not descriptive, I'll change to file_handler

core/services/mavftp/mavftp.py
Williangalvani: this file is derived from the original code. A lot of the original code is gone, it is not organized in classes, linted, typed, and provides a beter interface to use.

I'm not sure this should live in the main ardupilot repo at all. maybe at pymavlink/tools? I think that makes sense

core/services/mavftp/main.py
Williangalvani: oops, this is one of the things I meant to cleanup before opening it!

JoaoMario109 (approved):

core/services/ardupilot_manager/autopilot_manager.py
```diff
@@ -407,6 +407,15 @@ async def start_mavlink_manager(self, device: Endpoint) -> None:
...
             ),
+            Endpoint(
+                name="MavFTP",
+                owner=self.settings.app_name,
+                connection_type=EndpointType.UDPClient,
+                place="127.0.0.1",
+                argument=14555,
+                persistent=True,
+                protected=True,
+            ),
```
patrickelectric, line 418: Can't we use one of our internal endpoints already ?

PR #927, opened 2022-03-25
frontend: Cache all but HTML

### LINKED ISSUE #920: frontend often tries to load spinninglogo after backend is down, so it doesn't show at all
image

rafaellehmkuhl:
Cache all files apart from HTML documents.
Only HTML documents do not receive an SHA and need to be re-fetched to guarantee we update assets correctly on version-updates.

Fix #920


Williangalvani:
well, it is cached!
image

core/tools/nginx/nginx.conf
```diff
@@ -164,8 +164,10 @@ http {
         location / {
             root /home/pi/frontend;
             try_files $uri $uri/ /index.html;
+        }
+
+        location ~* /.(html|css|js)$ {
             expires -1;
-            add_header Cache-Control no-store;
```
Williangalvani, line 168: I suspect we need this. it seems to be caching files it shouldn't:
image

rafaellehmkuhl: Good catch. Thought the "expires" was enough. Rebuilding.

Williangalvani (changes requested):
I believe you can drop the first commit now,
there's also a whitespace change, but I like this approach.

Williangalvani:
I can't get this to cache at all. Am I doing something wrong?

Williangalvani (approved):
Nevermind. looks good!
image

PR #931, opened 2022-03-28
Improve cable-guy 

### LINKED ISSUE #766: core: cable_guy: Error when starting
image

The report also says that the service just died when started 

version beta 11

rafaellehmkuhl:
- Refactor IP creation approach, to better reflect what is really happening behind the curtains
- Allow custom DHCP server configurations (base for hotspot)
- Allow multiple simultaneous DHCP servers (base for hotspot)
- Ignore invalid interfaces upfront (probably fixes #766)

image
image

With the new approach, the user has basically three options:
1. Add a static IP
2. Ask for a dynamic IP
3. Enable a DHCP server on an existing static IP

If there's a DHCP server running, the user can disable it without removing the static IP.

PS: Right now we are putting a limit of one DHCP server for each interface, but there's no hard limit for that. If we want to remove this limitation in the future, we should just check if a static IP (or any on its subnet) don't have an associated server already.

Image available at `rafaellehmkuhl/blueos-core`.


core/frontend/src/components/ethernet/InterfaceCard.vue
```diff
@@ -112,6 +151,37 @@ export default Vue.extend({
           notifications.pushError({ service: ethernet_service, type: 'ETHERNET_ADDRESS_DELETE_FAIL', message })
         })
     },
+    async triggerForDynamicIP(): Promise<void> {
+      ethernet.setUpdatingInterfaces(true)
+
+      await back_axios({
+        method: 'post',
+        url: `${ethernet.API_URL}/dynamic_ip`,
+        timeout: 10000,
```
Williangalvani, line 160: 10 seconds is a pretty long time...

rafaellehmkuhl: Hue
This should take around 1-2 seconds usually (the interface is restarted, basically), but I'm basically using at least 10 seconds on all timeouts because of network issues, mostly which I recon happening on wifi connections.
I don't remember though if those tests were done before or after the introduction of the backend status checker.

Williangalvani:
I ran into some issues:
image
image
image

rafaellehmkuhl:
> I ran into some issues: image image image

Congratulations! You've discovered a hidden bug!

image

Fixed 😆

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -3,13 +3,18 @@
...
 from loguru import logger
 
 
 class Dnsmasq:
-    def __init__(self, config_path: pathlib.Path) -> None:
+    def __init__(self, config_path: pathlib.Path, interface: str) -> None:
         self._subprocess: Optional[Any] = None
 
+        if interface not in psutil.net_if_stats():
+            raise ValueError(f"Interface '{interface}' not found.")
```
patrickelectric, line 16: It would be nice to have a list of valid interfaces on the error message.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -60,6 +63,8 @@ def command_list(self) -> List[Union[str, pathlib.Path]]:
             self.binary(),
             "--no-daemon",
             f"--interface={self._interface}",
+            f"--dhcp-range={self._ipv4_network_prefix()}.100,{self._ipv4_network_prefix()}.200,255.255.255.0,24h",
```
patrickelectric, line 65: It would be nice if such magical numbers could exist inside a variable with a good name and a comment to explain what such numbers are for.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -59,14 +54,25 @@ def is_valid_config(self) -> bool:
             return False
 
     def command_list(self) -> List[Union[str, pathlib.Path]]:
+        """List of arguments to be used in the command line call.
+        Refer to https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html for details about each argument."""
```
patrickelectric, line 59: nice comment

core/services/cable_guy/api/manager.py
```diff
@@ -385,3 +350,39 @@ def get_interface_info(self, interface_name: str) -> InterfaceInfo:
         """
         interface = self.get_interface_ndb(interface_name)
         return InterfaceInfo(connected=interface.carrier != 0, number_of_disconnections=interface.carrier_down_count)
+
+    def _is_ip_on_interface(self, interface_name: str, ip_address: str) -> bool:
+        interface = self.get_interface_by_name(interface_name)
+        try:
```
patrickelectric, line 356:
```suggestion
        return any(True for address in interface.addresses if address.ip == ip_address)
```

core/services/cable_guy/main.py
```diff
@@ -137,3 +137,4 @@ def trigger_dynamic_ip_acquisition(interface_name: str) -> Any:
     server = Server(config)
 
     loop.run_until_complete(server.serve())
+    manager.stop()
```
patrickelectric, line 140: This should be not necessary since it will be called by __del__.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
rafaellehmkuhl: It's difficult to find well written documentations like this one.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
rafaellehmkuhl: Done

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
rafaellehmkuhl: Agree. Done.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -91,5 +96,13 @@ def is_running(self) -> bool:
     def interface(self) -> str:
         return self._interface
 
+    @property
+    def ipv4_gateway(self) -> IPv4Address:
+        return self._ipv4_gateway
+
+    def _ipv4_network_prefix(self) -> str:
+        network_digits = str(self._ipv4_gateway).split(".")[:3]
```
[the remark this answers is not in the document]
rafaellehmkuhl: Nice! Done

core/services/cable_guy/main.py
rafaellehmkuhl: Agree.

rafaellehmkuhl:
I've implemented the suggested changes and a new image is building, but I will try to make a better usage of the `ipaddress` library regarding lease-range and subnet-mask validation. While that, the PR will be in draft mode.

rafaellehmkuhl:
> I've implemented the suggested changes and a new image is building, but I will try to make a better usage of the `ipaddress` library regarding lease-range and subnet-mask validation. While that, the PR will be in draft mode.

I'm now happy with the implementation. It's less specific, allows for basically any IPv4 network configuration, has better type-checking and at the same time is easy to use.

patrickelectric (approved):

patrickelectric:
I'll let this in hold until we don't have another production release to avoid any uncaught bugs.

PR #934, opened 2022-03-29
core: beacon: Log loguru output on file

rafaellehmkuhl:
Beacon was not outputting its logging to a file. Fix that.

image

### DIFF
--- a/core/services/beacon/main.py
+++ b/core/services/beacon/main.py
@@ -8,11 +8,14 @@ from typing import Any, Dict, List, Optional
 
 import psutil
 from commonwealth.settings.manager import Manager
+from commonwealth.utils.logs import get_new_log_path
 from loguru import logger
 from zeroconf import IPVersion
 from zeroconf.asyncio import AsyncServiceInfo, AsyncZeroconf
 
 from settings import ServiceTypes, SettingsV1
+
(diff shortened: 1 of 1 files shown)


core/frontend/vue.config.js
```diff
@@ -1,5 +1,5 @@
 process.env.VUE_APP_BUILD_DATE = new Date().toLocaleString()
-const SERVER_ADDRESS = 'http://blueos.local/'
+const SERVER_ADDRESS = 'http://192.168.1.102/'
```
Williangalvani, line 2: oops

core/services/beacon/main.py
```diff
@@ -53,7 +56,7 @@ def __repr__(self) -> str:
 class Beacon:
     def __init__(self) -> None:
         self.runners: Dict[str, AsyncRunner] = {}
-        self.manager = Manager("Beacon", SettingsV1)
+        self.manager = Manager(SERVICE_NAME, SettingsV1)
```
patrickelectric, line 59: The service name is not the same as before, can that be a problem ?

rafaellehmkuhl: No. Talked to @Williangalvani about that and he approved the change.

patrickelectric (approved):

PR #947, opened 2022-04-04
Endpoint editing

### LINKED ISSUE #775: core: frontend: endpoints: Improve buttons and edit logic
The current edit logic can be improve:
image

We could make the current icons as buttons:
image

It should be also possible to edit fields and options for each endpoint without deleting and recreating.

rafaellehmkuhl:
- Allow clicking on the status icon (lamp) to quickly enable/disable the endpoint
- Allow complete endpoint edit on the pencil button (only the endpoint name is still not editable as we search for them by the name, not the hash, on the backend) 
- Add default values for the endpoint

image

Fix #775 

Image available in 10 minutes.


Williangalvani (changes requested):
Works nicely.

But can we repĺace the lightbulb with a [switch](https://vuetifyjs.com/en/components/switches/) ?
That is a LOT more intuitive. Right now the only way to tell that the lightbulb is clickable is by hovering it.

patrickelectric:
> Works nicely.
> 
> But can we repĺace the lightbulb with a [switch](https://vuetifyjs.com/en/components/switches/) ? That is a LOT more intuitive. Right now the only way to tell that the lightbulb is clickable is by hovering it.

Totally agree with that.

rafaellehmkuhl:
> Works nicely.
> 
> But can we repĺace the lightbulb with a [switch](https://vuetifyjs.com/en/components/switches/) ? That is a LOT more intuitive. Right now the only way to tell that the lightbulb is clickable is by hovering it.

Yup! Will take a look!

rafaellehmkuhl:
New look with switch:

image

core/frontend/src/components/autopilot/EndpointCard.vue
```diff
@@ -179,7 +176,7 @@ export default Vue.extend({
         })
     },
     async toggleEndpointEnabled(): Promise<void> {
-      this.updated_endpoint.enabled = !this.updated_endpoint.enabled
+      await sleep(500)
```
patrickelectric, line 179: Wait, why this sleep is necessary ?

patrickelectric (changes requested):
The PR also needs to be rebased over master.

core/frontend/src/components/autopilot/EndpointCard.vue
rafaellehmkuhl: It's just so the user can see the switch toggle animation 😄 
Without it, the spinning logo will kick in as soon as the user clicks the switch.

patrickelectric: It would be nice to have a comment so, this is not clear when reading the code.

rafaellehmkuhl: Yup. Adding.

rafaellehmkuhl:
@patrickelectric rebased and update, so the user has a hint that he cannot edit the endpoint name (instead of being allowed by the frontend but receiving an error from the backend):

image

core/frontend/src/components/autopilot/EndpointCard.vue
patrickelectric: @rafaellehmkuhl still missing comment.

patrickelectric:
The last docker is from 3 days ago, is there a a new version ?

rafaellehmkuhl:
> The last docker is from 3 days ago, is there a a new version ?

Just left to do market. Will build a new one when back home.

rafaellehmkuhl:
> The last docker is from 3 days ago, is there a a new version ?

New image being built. Should be ready in 15 minutes.

With the Notifier, while rebasing this PR I noticed it could be useful integrating the error-alert on it, so I did it on #974. This PR uses the commits from #974, so should wait for it to be merged.

patrickelectric (approved):

patrickelectric:
@rafaellehmkuhl please rebase to merge.

PR #960, opened 2022-04-06
core: beacon: remove blueos, add companion-wifi to mdns domains

Williangalvani:


core/services/beacon/default-settings.json
```diff
@@ -15,7 +15,7 @@
       },
       {
         "name": "wlan0",
-        "domain_names": ["companion", "blueos", "blueos-wifi"],
```
patrickelectric, line 18: `companion` is used temporarily for retrocompatibility and should be removed soon.
`companion.local` is also redirected to `blueos.local` to 'teache' the user that the new address should be used. The patch should only remove "blueos" from the wifi publishment.

patrickelectric (changes requested):

patrickelectric (approved):

patrickelectric:
@Williangalvani do you remember why we removed it ?

Williangalvani:
> @Williangalvani do you remember why we removed it ?

yes. this was done in order to prevent connections from going through the wifi network instead of the wired one. which can happen if the same mdns address is broadcasted on both interfaces.

PR #966, opened 2022-04-06
core: beacon: change settings to V2

Williangalvani:
This removes blueos.local from both default domain and wlan0.

This means random ethernet adapters plugged into usb will not get any broadcasts by default.

We should discuss this in tomorrow's call.

### DIFF
--- a/core/services/beacon/settings.py
+++ b/core/services/beacon/settings.py
@@ -125,3 +125,30 @@ class SettingsV1(settings.BaseSettings):
         )
         self.interfaces.append(new)
         return new
+
+
+class SettingsV2(SettingsV1):
+    VERSION = 2
+
+    def __init__(self, *args: str, **kwargs: int) -> None:
+        super().__init__(*args, **kwargs)
+        self.VERSION = SettingsV2.VERSION
+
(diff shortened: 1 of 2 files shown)


core/services/beacon/settings.py
```diff
@@ -125,3 +125,28 @@ def get_interface_or_create_default(self, name: str) -> Interface:
...
+        self.VERSION = 2
+
+    def migrate(self, data: Dict[str, Any]) -> None:
+        if data["VERSION"] == SettingsV2.VERSION:
+            return
+
+        if data["VERSION"] < SettingsV2.VERSION:
+            super().migrate(data)
+
+        data["default"]["domain_names"] = [domain for domain in data["default"]["domain_names"] if domain != "blueos"]
```
patrickelectric, line 144: I believe that we should stay with `blueos.local` for cable ethernet connections only, since this is a reliable connection source for `blueos.local`, different of wireless that may have problems.

patrickelectric, line 144: I believe that we should stay with `blueos.local` for cable ethernet connections only, since this is a reliable connection source for `blueos.local`, different of wireless that may have problems.

patrickelectric (approved):

PR #971, opened 2022-04-08
core: nginx: fix 2770 -> 80 redirect

### LINKED ISSUE #632: core: nginx: consider serving the web interface at :2770, or at least redirect it to :80
people consistently try to use that

Williangalvani:
this redirects 192.168.2.2:2770 to 192.168.2.2.
ideally we should redirect it to "blueos.local" or whatever the current name is, but I don't think we can fetch that name from nginx yet...

We could use beacon in the frontend to choose a domain and redirect to it

fix #632

### DIFF
--- a/core/tools/nginx/nginx.conf
+++ b/core/tools/nginx/nginx.conf
@@ -34,7 +34,7 @@ http {
         listen 2770;
 
         location / {
-            rewrite ^/(.*)$ http://(blueos|companion).local redirect;
+            rewrite ^/(.*)$ http://$host redirect;
         }
     }
 


Williangalvani:
> > this redirects 192.168.2.2:2770 to 192.168.2.2. ideally we should redirect it to "blueos.local" or whatever the current name is, but I don't think we can fetch that name from nginx yet...
> > We could use beacon in the frontend to choose a domain and redirect to it
> > fix #632
> 
> Will this break companion.local ?

nah, this only makes anything on port 2770 to the same thing at port 80
the previous patch that changed this line was probably a mistake. as it made 192.168.2.2:2770 redirect to `http://(blueos|companion).local` which isn't valid at all (see related issue)

patrickelectric (approved):

PR #973, opened 2022-04-08
More rules

Williangalvani:
includes and project were required by the "await-thenable" rule.
This needs a bit more work to fix the errors.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -50,6 +50,7 @@ module.exports = {
     'no-param-reassign': 'off',
     'no-shadow': 'off',
     'no-useless-constructor': 'off',
+    'require-await': 'error',
     semi: ['error', 'never'],
     'simple-import-sort/exports': 'error',
     'simple-import-sort/imports': 'error',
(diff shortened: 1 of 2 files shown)


rafaellehmkuhl:
From what I see on the lint errors, the 'require-await' rule demands we use the await syntax on functions typed to return promises?

Williangalvani:
> From what I see on the lint errors, the 'require-await' rule demands we use the await syntax on functions typed to return promises?

kind of. I think it means "you only need to declare a function as async if it uses await syntax". promise.then() can run just fine on non-async function

rafaellehmkuhl:
> > From what I see on the lint errors, the 'require-await' rule demands we use the await syntax on functions typed to return promises?
> 
> kind of. I think it means "you only need to declare a function as async if it uses await syntax". promise.then() can run just fine on non-async function

Makes sense!

patrickelectric:
@Williangalvani can you rebase this ?

patrickelectric:

PR #976, opened 2022-04-08
open links in new tabs

NickNothom:
Just something that was bugging me. 
I guess `"rel=noopener"` [is not necessary anymore.](https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable)


patrickelectric (approved):
Thanks for the PR @NickNothom, since it's your first PR I'm going to merge it, but we usually use prefixes for the git commits, you can check the history for examples. This helps us to find and filter commits more easily.

PR #980, opened 2022-04-14
Specify ardupilot errors

### LINKED ISSUE #975: Be more specific with endpoint errors
If the user tries to create an endpoint that already exists, the backend will correctly raise, but does not add any message saying that the error is because the endpoint already exists.

Other similar errors should be reviewed.

rafaellehmkuhl:
Error received for duplicated name, before and after:

image
image

Error received for duplicated parameters, before and after:
image
image

Fix #975

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/logs.py
+++ b/core/libs/commonwealth/commonwealth/utils/logs.py
@@ -45,3 +45,13 @@ def get_new_log_path(service_name: str) -> Path:
     # Returned log path are service-specific and store datetime information
     datetime_now = datetime.now().strftime("%m-%d-%Y_%H:%M:%S")
     return service_log_folder.joinpath(f"logfile_{datetime_now}.log")
+
+
+def stack_trace_message(error: BaseException) -> str:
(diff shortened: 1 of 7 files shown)


core/libs/commonwealth/commonwealth/utils/logs.py
```diff
@@ -45,3 +45,15 @@ def get_new_log_path(service_name: str) -> Path:
     # Returned log path are service-specific and store datetime information
     datetime_now = datetime.now().strftime("%m-%d-%Y_%H:%M:%S")
     return service_log_folder.joinpath(f"logfile_{datetime_now}.log")
+
+
+def stack_trace_message(error: BaseException) -> str:
+    """Get string containing joined messages from all exceptions in stack trace, beggining with the most recent one."""
+    message = str(error)
+    sub_error = error.__cause__
+    while True:
```
patrickelectric, line 54: While sub_erroe is not None ?

patrickelectric (changes requested):
Nice!

patrickelectric (approved):

PR #982, opened 2022-04-14
Improve exception handling

rafaellehmkuhl:


core/services/cable_guy/api/manager.py
```diff
@@ -274,9 +271,7 @@ def remove_ip(self, interface_name: str, ip_address: str) -> None:
             interface_index = self._get_interface_index(interface_name)
             self.ipr.addr("del", index=interface_index, address=ip_address, prefixlen=24)
         except Exception as error:
-            error_msg = f"Cannot delete IP '{ip_address}' from interface {interface_name}. {error}"
-            logger.error(error_msg)
-            raise ValueError(error_msg) from error
+            raise RuntimeError(f"Cannot delete IP '{ip_address}' from interface {interface_name}.") from error
```
patrickelectric, line 274: Is it possible to visualize such errors also via the logger interface ?

core/services/wifi/main.py
```diff
@@ -60,13 +64,7 @@ async def scan() -> Any:
...
         return available_networks
     except BusyError as error:
-        logger.error("Failed to scan networks.")
-        logger.exception(error)
-        raise HTTPException(status_code=status.HTTP_425_TOO_EARLY, detail=str(error)) from error
-    except FetchError as error:
-        logger.error("Failed to scan networks.")
-        logger.exception(error)
-        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(error)) from error
+        raise StackedHTTPException(status_code=status.HTTP_425_TOO_EARLY, error=error) from error
```
patrickelectric, line 67: We are not dealing with two different erros here ? FetchError appears as 500 on the original code.

core/services/cable_guy/api/manager.py
rafaellehmkuhl: Yes, it will be there, because of the stacked_trace_message.

core/services/wifi/main.py
patrickelectric: Ping @rafaellehmkuhl

rafaellehmkuhl: This change will continue to catch BusyError as usual, returning a 425, while the FetchError will be treated as before (not catched thus returning a 500 error from the GenericHandler).

rafaellehmkuhl: This change will continue to catch BusyError as usual, returning a 425, while the FetchError will be treated as before (not catched thus returning a 500 error from the GenericHandler).

patrickelectric (approved):

PR #984, opened 2022-04-14
core: Add initial 3rd party integration

patrickelectric:
This follows our documentation described in the wiki page
To test:
create the following file in a path: `register_service`
content:
```js
{
    "name":"Extension name",
    "description":"A small description about the service",
    "icon":"mdi-alpha-p-box",
    "company":"Blue Robotics",
    "version":"1.2.0",
    "webpage":"https://github.com/bluerobotics/BlueOS-docker",
    "api":""
 }
```
start simple python http server:
`python3 -m http.server 9992`

docker: patrickelectric/blueos-core:add_3rd_party_integration

### DIFF
--- a/core/frontend/src/types/helper.ts
+++ b/core/frontend/src/types/helper.ts
@@ -1,7 +1,18 @@
+export interface ServiceMetadata {
+    name: string
+    description: string
+    icon: string
+    company: string
+    version: string
+    webpage: string
+    api: string
+}
+
 export interface Service {
     valid: boolean
     title: string
     documentation_url: string
     versions: Array<string>
     port: number
+    metadata?: ServiceMetadata
 }
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -2,7 +2,7 @@
 
 import http
 from pathlib import Path
-from typing import Any, List
+from typing import Any, List, Optional
 
 import psutil
 import requests
@@ -22,12 +22,23 @@ API_CANDIDATE_URLS = ["/docs.json", "/openapi.json", "/swagger.json"]
 HTML_FOLDER = Path.joinpath(Path(__file__).parent.absolute(), "html")
 
 
+class ServiceMetadata(BaseModel):
+    name: str
+    description: str
+    icon: str
+    company: str
+    version: str
+    webpage: str
+    api: str
+
+
 class ServiceInfo(BaseModel):
     valid: bool
     title: str
     documentation_url: str
     versions: List[str]
     port: int
+    metadata: Optional[ServiceMetadata]
 
 
 class Helper:
@@ -51,6 +62,15 @@ class Helper:
         # If not valid web server, documentation will not be available
         if not info.valid:
             return info
+
+        # Check for service description metadata
+        try:
+            with requests.get(f"http://127.0.0.1:{port}/register_service", timeout=0.2) as response:
+                if response.status_code == http.HTTPStatus.OK:
+                    info.metadata = ServiceMetadata.parse_obj(response.json())
+        except Exception:
+            # This should be avoided by the first try block, but better safe than sorry
+            pass
 
         for documentation_path in DOCS_CANDIDATE_URLS:
             try:
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -69,7 +69,7 @@
       >
         <v-divider />
         <v-list
-          v-for="(menu, i) in menus"
+          v-for="(menu, i) in computed_menu"
           :key="i"
           class="pa-0"
           nav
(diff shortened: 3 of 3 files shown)


rafaellehmkuhl:
The service discovery appears to be working correctly, with the extensions menu opening when I serve a new extension, but the extension name apparently is not being caught:

image

patrickelectric:
> The service discovery appears to be working correctly, with the extensions menu opening when I serve a new extension, but the extension name apparently is not being caught:
> 
> image

Did you restart the core docker or helper service ? It caches the last result.

rafaellehmkuhl:
> > The service discovery appears to be working correctly, with the extensions menu opening when I serve a new extension, but the extension name apparently is not being caught:
> > image
> 
> Did you restart the core docker or helper service ? It caches the last result.

My steps:

1. Created the `register_service` file, equal to your example (with name equal to "extension name")
2. Served it with python `http.server`
3. Everything working as intended!
4. Stop the server
5. Extensions menu disappear, great!
6. Edited the `register_service` file with another extension name
7. Served it again
8. Menu reappears, but menu name is as before, without the changes

Looking at the `web_services` request, it appears there's some caching going on, as it is getting the same info as before the changes, for all the fields:

image

rafaellehmkuhl:
> > The service discovery appears to be working correctly, with the extensions menu opening when I serve a new extension, but the extension name apparently is not being caught:
> > image
> 
> Did you restart the core docker or helper service ? It caches the last result.

Humn, I see now.

Can we prevent this caching somehow?
It seems counter-intuitive that we need to refresh the helper service/core container to catch the extension changes.

(this can be addressed in another PR, for time concerns)

patrickelectric:
> > > The service discovery appears to be working correctly, with the extensions menu opening when I serve a new extension, but the extension name apparently is not being caught:
> > > image
> > 
> > 
> > Did you restart the core docker or helper service ? It caches the last result.
> 
> Humn, I see now.
> 
> Can we prevent this caching somehow? It seems counter-intuitive that we need to refresh the helper service/core container to catch the extension changes.
> 
> (this can be addressed in another PR, for time concerns)

This is done by our time cache decorator, it should update after some time.

rafaellehmkuhl:
> 

It doesn't appear to be it. The helper cache is 1 minute. It's been more than 10 minute here, and it's still the same.

Will restart core to see how it goes.

rafaellehmkuhl:
> > 
> 
> It doesn't appear to be it. The helper cache is 1 minute. It's been more than 10 minute here, and it's still the same.
> 
> Will restart core to see how it goes.

When core is restarted, the changes are catch.

core/services/helper/main.py
```diff
@@ -52,6 +63,15 @@ def detect_service(port: int) -> ServiceInfo:
         if not info.valid:
             return info
 
+        # Check for service description metadata
+        try:
+            with requests.get(f"http://127.0.0.1:{port}/register_service", timeout=0.2) as response:
+                if response.status_code == http.HTTPStatus.OK:
+                    info.metadata = ServiceMetadata.parse_obj(response.json())
+        except Exception:
```
Williangalvani, line 71: shouldn't pylint catch this as a broad-exception?

we should at least log the exception here

Williangalvani (approved):
working fine here, but I think we should log that exception.

image

core/services/helper/main.py
patrickelectric: > shouldn't pylint catch this as a broad-exception?
> 
> we should at least log the exception here

We have it disabled: https://github.com/patrickelectric/companion-docker/blob/8ab788c945a1f656c2deedaaddf0096fb29fab21/pyproject.toml#L55

But I agree that we should log the issue.

rafaellehmkuhl: It should. Strange. And I can't find any `# pylint: ignore` blocks anywhere.

patrickelectric:
> working fine here, but I think we should log that exception.
> 
> image

nice icon for dvl :rofl:

core/frontend/src/types/helper.ts
```diff
@@ -1,7 +1,18 @@
...
+}
+
 export interface Service {
     valid: boolean
     title: string
     documentation_url: string
     versions: Array<string>
     port: number
+    metadata?: ServiceMetadata
 }
```
rafaellehmkuhl, line 18: Regarding this, how should one creating a service behave in respect to `name` vs `title`, `valid`, `webpage` vs `api` vs `documentation_url`, for example?

patrickelectric: Service is a structure that we use to help us easily identify services with the helper scraping logic. the metadata is the real information provided by the running service. 
We should rely on the metadata to provide all the necessary information for the frontend. Service was only a quick way to make it work without having this concept finished.
Remember that this is still not a supported thing and we can rework everything around this concept before have official support. The PR only allows the metadata access for people and services that are willing to have it working right now and accept the API break in the future.

rafaellehmkuhl (approved):

PR #986, opened 2022-04-15
core: nmea-injector: Make settings persistent between sessions

### LINKED ISSUE #708: core: NMEA-Injector does not save settings

rafaellehmkuhl:
Fix #708 

Image available.

### DIFF
--- a/core/services/nmea_injector/TrafficController.py
+++ b/core/services/nmea_injector/TrafficController.py
@@ -6,11 +6,13 @@ from typing import Dict, List, Tuple, Union
 
 import pynmea2
 from commonwealth.mavlink_comm.MavlinkComm import MavlinkMessenger
+from commonwealth.settings.manager import Manager
 from loguru import logger
 from pydantic import BaseModel, conint
 
 from exceptions import UnsupportedSocketKind
 from MavlinkNMEA import MavlinkGpsInput, parse_mavlink_from_sentence
+from settings import NmeaInjectorSettingsSpecV1, SettingsV1
 
 
 class SocketKind(str, Enum):
@@ -33,6 +35,21 @@ class NMEASocket(BaseModel):
 
     def __hash__(self) -> int:
         return hash(str(self))
+
+    @staticmethod
+    def from_settings_spec(settings_spec: NmeaInjectorSettingsSpecV1) -> "NMEASocket":
+        return NMEASocket(
+            kind=settings_spec.kind,
+            port=settings_spec.port,
+            component_id=settings_spec.component_id,
+        )
+
+    def to_settings_spec(self) -> NmeaInjectorSettingsSpecV1:
+        return NmeaInjectorSettingsSpecV1(
+            kind=self.kind,
+            port=self.port,
+            component_id=self.component_id,
+        )
 
 
 class TcpNmeaProtocol(asyncio.Protocol):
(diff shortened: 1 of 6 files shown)


core/services/nmea_injector/TrafficController.py
```diff
@@ -78,6 +88,12 @@ class TrafficController:
 
     def __init__(self) -> None:
         self._socks: Dict[NMEASocket, Union[asyncio.AbstractServer, asyncio.BaseTransport]] = {}
+        self._settings_manager = Manager("NmeaInjector", SettingsV1)
```
patrickelectric, line 91: we are using lowercase for our settings name.

core/services/nmea_injector/TrafficController.py
```diff
@@ -96,6 +112,10 @@ async def add_sock(self, sock: NMEASocket) -> None:
         else:
             raise UnsupportedSocketKind(f"Got {sock.kind}. Expected one of: {[kind.value for kind in SocketKind]}.")
         self._socks[sock] = server_socket
+        settings_spec = NmeaInjectorSettingsSpecV1.from_spec(sock)
```
patrickelectric, line 115: It would be better to have such method encapsulate over a free function or inside the `NMEASocket` class, the reason behind this is that each settings version will need to implement such function unnecessary, like V1, V2, v3, so every time that we update the settings such method will need to exist.
If we move it to a free function or inside the `NMEASocket` class, the function will only need to exist in a single place and be updated when the settings version change, over duplicating such method over and over for each settings.

core/services/nmea_injector/settings.py
```diff
@@ -0,0 +1,42 @@
...
+from commonwealth.settings import settings
+
+
+class NmeaInjectorSettingsSpecV1(pykson.JsonObject):
+    kind = pykson.StringField()
+    port = pykson.IntegerField()
+    component_id = pykson.IntegerField()
+
+    @staticmethod
+    def from_spec(spec: "NmeaInjectorSpec") -> "NmeaInjectorSettingsSpecV1":  # type: ignore
```
patrickelectric, line 13: This should be `NMEASocket` when reading `async def add_sock(self, sock: NMEASocket) -> None:`

image

core/services/nmea_injector/settings.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __eq__(self, other: object) -> Any:
+        if isinstance(other, NmeaInjectorSettingsSpecV1):
+            return self.kind == other.kind and self.port == other.port
+        return False
+
+
+class SettingsV1(settings.BaseSettings):
+    VERSION = 1
+    specs = pykson.ObjectListField(NmeaInjectorSettingsSpecV1)
```
patrickelectric, line 28: isn't it possible to remove specs and have the `NmeaInjectorSettingsSpecV1` properties here directly ?

patrickelectric (changes requested):

core/services/nmea_injector/TrafficController.py
rafaellehmkuhl: Agree. Was following the same pattern as the Bridget implementation, but this seems better.

core/services/nmea_injector/settings.py
rafaellehmkuhl: I don't think so, since this is where we store the multiple individual settings.

patrickelectric (approved):

rafaellehmkuhl:
@patrickelectric I'm waiting for [the image build](https://github.com/rafaellehmkuhl/companion-docker/runs/6063157999?check_suite_focus=true) to test with the modifications.

patrickelectric:
> @patrickelectric I'm waiting for [the image build](https://github.com/rafaellehmkuhl/companion-docker/runs/6063157999?check_suite_focus=true) to test with the modifications.

I tested and eveything appears to be working. 
Nice work!

PR #988, opened 2022-04-18
frontend: video: Add warning when creating streams from 'Fake source'

### LINKED ISSUE #781: core: frontend: Make clear that using video fake sources consumes a lot of CPU
### Current behaviour

When a video link with fake source is created, it can consume *a lot* of CPU. More specificaly, a 1440px@60hz uses about 3/4 of total CPU power.

image

We should put a warning when users create video links with fake sources, stating that it consumes a lot of CPU and that a good idea is to use a small resolution link.

Maybe we should also allow fake videos only under Pirate Mode.

### Prerequisites

rafaellehmkuhl:
image

Fix #781


patrickelectric (approved):
image

PR #1002, opened 2022-04-26
core: frontend: Make it 100% mobile friendly

### LINKED ISSUE #499: core: frontend: Make everything mobile friendly
Here is a list of known problems with the interface when running on mobiles, that can be updated/checked on-demand:
- [x] Buttons misalignment on FirmwareManager 
- [x] Endpoint cards do not scale well (cannot read most of its parameters)
- [x] Video device image gets off-screen
- [x] Video cards do not scale well (most stream parameters get hidden)
- [x] System-info page does not scale at all
- [x] Version-card info get hidden
- [x] Pirate mode phrase is still taking more than one line on app-bar

DAYS_TO_DEVELOP:5

### LINKED ISSUE #985: Kernel message colors on dark mode are not great
image

rafaellehmkuhl:
This PR aims to make BlueOS 100% mobile-friendly, this is, to support smaller screens.

As a reference, we use the classic size of the Samsung Galaxy S8+, which is the smallest smartphone size available at the Chromium dimensions inspector.

On the process of making the frontend mobile-friendly, several components received medium to large refactors, and some other minors like color-cohesion were solved.

There's a major change here, were Blue Robotic's default blue is used as the primary color of BlueOS. This should be discussed. Another good possibility is using `#08c`(current app bar color) as primary.

Fix #499
Fix #985

Before and after samples:

System-information:

Endpoints:

Video:

Version-chooser:


core/frontend/src/assets/colors/vuetify.ts
```diff
@@ -0,0 +1,7 @@
+export const PRIMARY = '#1976D2'
```
patrickelectric, line 1: what is the difference between this colors and the ones defines by default ?

core/frontend/vue.config.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
   transpileDependencies: ['vuetify', 'vuex-module-decorators'],
   pwa: {
     name: 'BlueOS',
-    themeColor: '#08c',
+    themeColor: '#2699D0',
```
patrickelectric, line 63: Should it have a comment pointing to the BR_BLUE color ? Should we define such colors in `.js` to allow the usage under both `.ts` and `.js` ?

core/frontend/src/App.vue
```diff
@@ -20,20 +20,9 @@
...
           <span v-if="settings.is_pirate_mode">
-            <span class="hidden-sm-and-down">Ahoy matey! You're running</span>
-            <v-icon>
-              mdi-flag-variant
-            </v-icon>
-            <v-icon>
-              mdi-skull-crossbones
-            </v-icon>
+            <v-icon>mdi-skull-crossbones</v-icon>
             Pirate Mode
```
patrickelectric, line 24: Maybe we should add `Running Pirate Mode` or `Pirate Mode Activated`, to make it explicit that the user has done something to have the `Pirate Mode` there.

core/frontend/src/assets/colors/vuetify.ts
rafaellehmkuhl: They are the same. I'm adding them here to allow for easier usage, as Vuetify (as far as I could check) doesn't export them (as 'primary', 'secondary', etc) anywhere.

core/frontend/vue.config.js
rafaellehmkuhl: Maybe. I had some trouble importing them as `.ts`, but with `.js` I think there will be no problem.

core/frontend/src/App.vue
rafaellehmkuhl: I'm of the opinion that the `Pirate Mode` text is enough. Anything more than that will also force the text to be significantly smaller to fit here, compromising layout/legibility.

patrickelectric: Valid point.

rafaellehmkuhl: What do you think about making the text black and bold?

image

patrickelectric: I think we can stay with the current approach without the bold. And see how it goes with SMS.

core/frontend/vue.config.js
patrickelectric: To make sure, you are checking that, right ?

rafaellehmkuhl: Completely forgot about that. Will take a look right now.

ES-Alexander (approved):
Not familiar enough with frontend stuff to properly review it, but the pictures look like a great improvement! 😃

core/frontend/vue.config.js
rafaellehmkuhl: More complicated than expected, as `vue.config.js` still runs on `commonjs`, but done.

rafaellehmkuhl:
@patrickelectric rebased. Just waiting for CI.

rafaellehmkuhl:
> image I haven't tested it, it seems great, my only concern is if this density is usable for touch users.
> 
> One option is to move the two floating icons a little upward, to get the second (the trash bin) away from the slider.

It's usable, but not great, indeed.

I tried getting the buttons up, but it got very strange. Let me try a new approach.

patrickelectric (approved):
@rafaellehmkuhl be free to merge it

rafaellehmkuhl:
@joaoantoniocardoso did some changes. What do you think?

image

joaoantoniocardoso:
> @joaoantoniocardoso did some changes. What do you think?
> 
> image

Nice, it seems much better!

PR #1005, opened 2022-05-03
core: versionchooser: cleanup dangling docker images when starting up

Williangalvani:
we should have a way of cleaning up until https://github.com/aio-libs/aiodocker/issues/696 is closed.


rafaellehmkuhl (approved):
Just needs the CI fix and I would also leave an issue open to track the aio lib change.

patrickelectric (approved):
Be free to merge if the developers don't fix the issue

PR #1006, opened 2022-05-05
core: nmea-injector: Cover it with tests

### LINKED ISSUE #557: core: nmea-injector: Create test with nmeasim

rafaellehmkuhl:
Uses [nmeasim](https://pypi.org/project/nmeasim/) to generate fake NMEA data.

~Unfortunatelly nmeasim does not have support for GNS messages, nor I could find much information about it. @Williangalvani @jaxxzer do you have any idea about this sentence type? If we have a sample message, I can add a hard-coded test to it.~ Added tests for GNS sentence type.

Fix #557


patrickelectric:
@rafaellehmkuhl maybe this ?
https://receiverhelp.trimble.com/alloy-gnss/en-us/NMEA-0183messages_GNS.html

rafaellehmkuhl:
> @rafaellehmkuhl maybe this ? https://receiverhelp.trimble.com/alloy-gnss/en-us/NMEA-0183messages_GNS.html

Cool! Will try with those.
I was trying to access those trimble docs some time ago, but they were not loading. Probably my internet.

rafaellehmkuhl:
@patrickelectric added tests for GNS type. This required some additional changes, the came for good. 

The tests are now separated by mavlink's GPS_INPUT field, so the test structure doesn't change with new sentence types being added.

patrickelectric (changes requested):
with the settings:
image

Master:
image

With: 
image


Results in:
image


I would also recommend tests to settings load.

rafaellehmkuhl:
> with the settings: image
> 
> Master: image
> 
> With: image
> 
> Results in: image
> 
> I would also recommend tests to settings load.

Just to be sure: Is this happening if you change to any other image or just when using the 'add_more_tests'?

rafaellehmkuhl:
@patrickelectric problem fixed and tests to settings added.

patrickelectric (approved):

PR #1008, opened 2022-05-06
frontend: firmware: Remove info for 'official' tag, since it was removed

rafaellehmkuhl:


ES-Alexander (approved):
Unsure whether we should add something like "Latest one is recommended." to the end of the Stable explanation, since we no longer specify a recommended one, but it could also be fine as is - up to you :-)

patrickelectric (approved):

PR #1013, opened 2022-05-17
Add cosmos concept

patrickelectric:
The idea here is to add a helper library that extend the types that already exist.
This is being used inside my cockpit work to make it more simpler and readable.

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -246,7 +246,7 @@ export default Vue.extend({
         })
 
       // No extensions, return the default menus
-      if (submenus.length === 0) {
+      if (submenus.isEmpty()) {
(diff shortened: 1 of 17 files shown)


rafaellehmkuhl (approved):

rafaellehmkuhl:
I think this helper is helpful nonetheless. We can gain a lot regarding maintainability by introducing some helpers and standards here and there.

About the string helpers specifically, I took a look to see if there's any maintained library that does what we are aiming (like what we use for datetimes) and found [underscore.string](https://github.com/esamattis/underscore.string). It has a `isBlank` function that tests some other cases as well. Maybe worth checking.

Williangalvani:
nice

PR #1014, opened 2022-05-17
ArdupilotManager: launch ardusub with gdbserver if debug symbols are present

### LINKED ISSUE #717: Add some debugging capabilities for Ardupilot
we could detect a debug binary and set up at least gdb server when launching. This part should be straight forward.
We could also display something in the frontend so the user knows they're using a debug binary.

Williangalvani:
This means a gdb server will be opened on port 5555 for debugging. ardupilot doesn't start until a debugger is attached.

While I would like to have something in the GUI signalling that we are in debug mode, there is 0% chance that someone will randomly stumble upon a debug build of ardupilot and upload it not knowing what they are doing...

Depends on https://github.com/bluerobotics/companion-docker-base/pull/14

Fix #717

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -4,6 +4,7 @@ import subprocess
 from copy import deepcopy
 from typing import Any, List, Optional, Set
 
+import magic
 import psutil
 from commonwealth.mavlink_comm.VehicleManager import VehicleManager
 from commonwealth.utils.Singleton import Singleton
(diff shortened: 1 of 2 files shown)


Williangalvani:
> Wait, we can't merge that since core with gdbserver is not available, right ?

yep.

Williangalvani:
it will require tagging a new base and switching to it

patrickelectric (commented):
just to make sure, the binaries provided by ardupilot does not come with debug symbols right ?

Williangalvani:
> just to make sure, the binaries provided by ardupilot does not come with debug symbols right ?

Nope,
Fun fact: regular binaries are 2 MB, debug binaries are 20MB

patrickelectric:
Is this still in draft ?

Williangalvani:
while this is not perfect. I think it is good to go in. it shouldn't harm users anyway. unless they find wild debug binary somewhere.

It is already useful for testing, despite a couple of edge cases

PR #1018, opened 2022-05-23
Update to use new bridges 0.8.1

patrickelectric:
- Allow multiple clients connected 
- Allow usage of by-path and by-id inputs


rafaellehmkuhl:
From the changes, I'm correct to assume the previous expected behavior was the UDP to auto-disconnect, and so the service settings does not need to be changed, and from now on the new bridges will by default not auto-disconnect?

Williangalvani (approved):

patrickelectric:
I'll wait for JM confirmation that is working for their use case

PR #1019, opened 2022-05-24
Add hotspot functionality

### LINKED ISSUE #307: core: Add always-available wifi hotspot

rafaellehmkuhl:
~This branch had it development stalled for some months to focus on other priorities. I'm pushing it as it was.
Last time I've checked, it was already working, but had several code improvements to be made. Once the CI builds the image, I'm going to test it to confirm.
I will put some work to finish it in the next weeks.~

image

Hotspot is now working, in parallel with wifi-client-behavior.

The main use-case for this feature is for the user to be able to connect to the Raspberry board if cabled-ethernet is not working or no known network is nearby, and from the frontend to debug its board and/or connect to a known wifi network.

By default, the hotspot is enabled, and it also auto-starts itself whenever there's no connection to any known wifi network for more than the watchdog timeout. We call the latter smart-hotspot.

If a user is on the hotspot, access the frontend and from there successfully connect to a known network, the hotspot will be disabled.

Default credentials are:
- SSID: BlueOS (FIRST_SIX_SERIAL_CHARS)
- Password: blueosap

The credentials can be changed by the user from the frontend interface.
The hotspot can also be enabled/disabled from the frontend.
The user can also scan a dynamically generated QR code to connect to the hotspot.

Testing this was somewhat painful, as most network-related features, but in the end I think it works as expected for the main use-case. I invite anyone reviewing this to also test it. Further tests to use the hotspot as a regular long-run connection were not performed, as this is not our goal right now, but can be something to be done in the future.

~PS1: For now, users that want this feature will need to perform a full image refresh or to manually edit the DHCP client file as explained [here](https://github.com/bluerobotics/BlueOS-docker/pull/1019#issuecomment-1136351839). @patrickelectric also mentioned he wants to do something on that to allow for ease upgrade.~

~PS2: @Williangalvani I didn't add the `uap0` interface to the `blueos-wifi` beacon settings as I didn't know if Beacon would behave correctly in the case where the interface does not exist (when the hotspot is disabled). Let me know if it is ok to add it there.~

~PS3: There's no persistence for the hotspot credentials yet, so if someone changes the SSID and password, they will be reset on the next power cycle. I'm opening the PR for review but planning to add persistence before merging it.~

Fix #307

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/general.py
+++ b/core/libs/commonwealth/commonwealth/utils/general.py
@@ -1,5 +1,23 @@
 import os
 
+from loguru import logger
+
 
 def is_running_as_root() -> bool:
     return os.geteuid() == 0
+
+
+def device_id() -> str:
+    try:
+        with open("/sys/firmware/devicetree/base/serial-number", "r", encoding="utf-8") as f:
+            return f.read()
+    except Exception as error:
+        logger.exception(f"Could not get device's serial-number. {error}")
+
+    try:
+        with open("/etc/machine-id", "r", encoding="utf-8") as f:
+            return "".join(f.read().split())
+    except Exception as error:
+        logger.exception(f"Could not get device's machine-id. {error}")
+
+    raise ValueError("Could not get device id.")
--- a/core/services/wifi/Hotspot.py
+++ b/core/services/wifi/Hotspot.py
@@ -1,0 +1,371 @@
+import hashlib
+import pathlib
+import shlex
+import shutil
+import subprocess
+import tempfile
+import time
+from enum import Enum
+from ipaddress import IPv4Address
+from typing import Any, Callable, List, Optional
+
+import psutil
+from commonwealth.utils.DHCPServerManager import Dnsmasq as DHCPServerManager
+from commonwealth.utils.general import device_id
+from loguru import logger
+from pyroute2 import IW, IPRoute
+
+from typedefs import WifiCredentials
+
+
+class HostapdFrequency(str, Enum):
+    """Valid hostapd frequency modes."""
+
+    HW_2_4 = "g"  # Hostapd id for 2.4 GHz mode
+    HW_5_0 = "a"  # Hostapd id for 5.0 GHz mode
+
+    @staticmethod
+    def mode_from_channel_frequency(frequency: int) -> "HostapdFrequency":
+        return HostapdFrequency.HW_2_4 if int(frequency / 100) == 24 else HostapdFrequency.HW_5_0
+
+
+class HotspotManager:
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        base_interface: str,
+        ipv4_gateway: IPv4Address,
+        ap_interface_name: str = "uap0",
+        ap_ssid: Optional[str] = None,
+        ap_passphrase: Optional[str] = None,
+    ) -> None:
+        self.iw = IW()
+        self.ipr = IPRoute()
+
+        self._ap_interface_name = ap_interface_name
+
+        try:
+            dev_id = device_id()
+        except Exception:
+            dev_id = "000000"
+        hashed_id = hashlib.md5(dev_id.encode()).hexdigest()[:6]
+        self._ap_ssid = ap_ssid or f"BlueOS ({hashed_id})"
+
+        self._ap_passphrase = ap_passphrase or "blueosap"
+
+        self._subprocess: Optional[Any] = None
+
+        if base_interface not in psutil.net_if_stats():
+            raise ValueError(f"Base interface '{base_interface}' not found.")
+        self._base_interface = base_interface
+
+        self._ipv4_gateway = ipv4_gateway
+
+        self._include_interface_on_dhcpcd()
+
+        self._dhcp_server: Optional[DHCPServerManager] = None
+
+        binary_path = shutil.which(self.binary_name())
+        if binary_path is None:
+            raise ValueError("Hostapd binary not found on system's PATH.")
+
+        self._binary = pathlib.Path(binary_path)
+        assert self.is_binary_working()
+
+    @staticmethod
+    def binary_name() -> str:
+        return "hostapd"
+
+    def binary(self) -> pathlib.Path:
+        return self._binary
+
+    def set_credentials(self, credentials: WifiCredentials) -> None:
+        logger.debug(f"Changing hotspot ssid to '{credentials.ssid}' and passphrase to '{credentials.password}'.")
+        self._ap_ssid = credentials.ssid
+        self._ap_passphrase = credentials.password
+
+    @property
+    def credentials(self) -> WifiCredentials:
+        return WifiCredentials(ssid=self._ap_ssid, password=self._ap_passphrase)
+
+    def is_binary_working(self) -> bool:
+        try:
+            subprocess.check_output([self.binary(), "-h"])
+            return True
+        except subprocess.CalledProcessError as error:
+            if error.returncode == 1:
+                return True
+            logger.error(f"Invalid binary: {error}")
+            return False
+
+    def base_interface_channel_frequency(self) -> int:
+        wireless_interfaces = self.iw.get_interfaces_dict()
+        if not self._base_interface in wireless_interfaces:
+            raise RuntimeError("Could not find base interface.")
+        last_channel = -1
+        time_last_channel_change = time.time()
+        time_start_searching = time.time()
+        while True:
+            current_channel = int(wireless_interfaces[self._base_interface][3])
+            if current_channel != last_channel:
+                time_last_channel_change = time.time()
+                last_channel = current_channel
+            seconds_in_same_channel = time.time() - time_last_channel_change
+            seconds_searching = time.time() - time_start_searching
+            if seconds_in_same_channel > 2:
+                return current_channel
+            if seconds_searching > 15:
+                raise RuntimeError("Could not find base interface channel. Timeout exceeded.")
+
+    def desired_channel_frequency(self) -> int:
+        return self.base_interface_channel_frequency()
+
+    def _create_virtual_interface(self) -> None:
+        logger.debug("Deleting virtual access point interface (if exists).")
+        wireless_interfaces = self.iw.get_interfaces_dict()
+        if self._ap_interface_name in wireless_interfaces:
+            interface_index = int(self.ipr.link_lookup(ifname=self._ap_interface_name)[0])
+            self.iw.del_interface(interface_index)
+        self._reach_condition_or_timeout(
+            lambda self: self._ap_interface_name not in self.iw.get_interfaces_dict(),
+            "Could not delete virtual interface. Timeout exceeded.",
+        )
+
+        logger.debug("Creating virtual access point interface.")
+        # pylint: disable=consider-using-with
+        subprocess.Popen(
+            shlex.split(f"iw dev {self._base_interface} interface add {self._ap_interface_name} type __ap")
+        )
+        # Following 2 lines are an alternative I could not get to work since its docs are not very clear
+        # base_interface_index = int(self.ipr.link_lookup(ifname=self._base_interface)[0])
+        # self.iw.add_interface(ifname=self._ap_interface_name, iftype="ap_vlan", dev=base_interface_index)
+        self._reach_condition_or_timeout(
+            lambda self: self._ap_interface_name in psutil.net_if_stats(),
+            "Could not create virtual interface. Timeout exceeded.",
+        )
+
+        logger.debug("Starting virtual access point interface.")
+        # pylint: disable=consider-using-with
+        subprocess.Popen(shlex.split(f"ifconfig {self._ap_interface_name} up"))
+        # Following 2 lines are an alternative I could not get to work since its docs are not very clear
+        # virtual_interface_index = int(self.ipr.link_lookup(ifname=self._ap_interface_name)[0])
+        # self.ipr.link("set", index=virtual_interface_index, state="up")
+        self._reach_condition_or_timeout(
+            lambda self: self._ap_interface_name in psutil.net_if_stats() and psutil.net_if_stats()[self._ap_interface_name][0],  # fmt: skip
+            "Could not start virtual interface. Timeout exceeded.",
+        )
+
+    def command_list(self) -> List[str]:
+        return shlex.split(f"{self.binary()} {self.config_path()}")
+
+    def start(self) -> None:
+        logger.info("Starting hotspot.")
+        try:
+            self._create_temp_config_file()
+            self._create_virtual_interface()
+            # pylint: disable=consider-using-with
+            if not self.is_running():
+                self._subprocess = subprocess.Popen(self.command_list(), shell=False, encoding="utf-8", errors="ignore")
+                time.sleep(3)
+                if not self.is_running():
+                    exit_code = self._subprocess.returncode
+                    raise RuntimeError(f"Failed to initialize Hostapd ({exit_code}).")
+            if not self._dhcp_server:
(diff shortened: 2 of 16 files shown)


rafaellehmkuhl:
Fixing branch to pass CI tests and generate test image.

rafaellehmkuhl:
@jaxxzer taking a look at the code, I realized this branch was in fact WIP, and not yet tested. The one that is tested and working is being built [here](https://github.com/rafaellehmkuhl/companion-docker/runs/6580264817?check_suite_focus=true).

patrickelectric:
> @jaxxzer taking a look at the code, I realized this branch was in fact WIP, and not yet tested. The one that is tested and working is being built [here](https://github.com/rafaellehmkuhl/companion-docker/runs/6580264817?check_suite_focus=true).
> 
> EDIT: image already available for test at `rafaellehmkuhl/add-hotspot-messy-but-tested`
> 
> To use it without needing to reflesh the SD card, one needs to manually add the following to the end of the `/etc/dhcpcd.conf` file and then restart the system:
> 
> ```
> interface uap0
>     static ip_address=192.168.3.2/24
>     nohook wpa_supplicant
> ```
> 
> Let me know if you need any help.

We could add just a single line on the `/etc/dhcpcd.conf` that includes our BluOS.conf file,
check `The include statement` in https://linux.die.net/man/5/dhcpd.conf

rafaellehmkuhl:
> @rafaellehmkuhl let me know once this is ready on BlueOS side, I can take care of the OS integration with the dhcpd configuration related to existent installations.

I'm on test phase, so I expect it to be done till friday.
About the configuration for existing installations, how are you thinking to approach?

patrickelectric:
> > @rafaellehmkuhl let me know once this is ready on BlueOS side, I can take care of the OS integration with the dhcpd configuration related to existent installations.
> 
> I'm on test phase, so I expect it to be done till friday. About the configuration for existing installations, how are you thinking to approach?

Check my previous comment:
https://github.com/bluerobotics/BlueOS-docker/pull/1019#issuecomment-1183102210

We can just include BlueOS configuration file on the `dhcpd.conf` with commander

rafaellehmkuhl:
> > > @rafaellehmkuhl let me know once this is ready on BlueOS side, I can take care of the OS integration with the dhcpd configuration related to existent installations.
> > 
> > 
> > I'm on test phase, so I expect it to be done till friday. About the configuration for existing installations, how are you thinking to approach?
> 
> Check my previous comment: [#1019 (comment)](https://github.com/bluerobotics/BlueOS-docker/pull/1019#issuecomment-1183102210)
> 
> We can just include BlueOS configuration file on the `dhcpd.conf` with commander

I think you mixed `dhcpd` (server) with `dhcpcd` (client). The latter does not have an "include" statement. Anyway, we should be able to directly modify the `dhcpcd.conf` with commander, right?

core/services/wifi/Hotspot.py
```diff
@@ -0,0 +1,339 @@
...
+            interface_index = int(self.ipr.link_lookup(ifname=self._ap_interface_name)[0])
+            self.iw.del_interface(interface_index)
+        time_start_delete = time.time()
+        while True:
+            if self._ap_interface_name not in self.iw.get_interfaces_dict():
+                time.sleep(0.3)
+                break
+            if time.time() - time_start_delete > 5:
+                raise RuntimeError("Could not delete virtual interface. Timeout exceeded.")
+            time.sleep(0.01)
```
patrickelectric, line 126: Since this code replicates a lot, I would recommend to create a function for it.
```python
def loop_raise_check(wait_condition, raise_message):
    time_start_start = time.time()
    while True:
        if wait_condition():
            break
        if time.time() - time_start_start > 5:
            raise RuntimeError(message)
        time.sleep(0.5)
```

core/services/wifi/Hotspot.py
```diff
@@ -0,0 +1,339 @@
...
+
+        logger.debug("Creating virtual access point interface.")
+        # pylint: disable=consider-using-with
+        subprocess.Popen(
+            shlex.split(f"iw dev {self._base_interface} interface add {self._ap_interface_name} type __ap")
+        )
+        # Following 2 lines are an alternative I could not get to work
+        # base_interface_index = int(self.ipr.link_lookup(ifname=self._base_interface)[0])
+        # self.iw.add_interface(ifname=self._ap_interface_name, iftype="ap_vlan", dev=base_interface_index)
```
patrickelectric, line 135: Do you know why ? Can you add in the comments ?

core/services/wifi/Hotspot.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def _create_temp_config_file(self) -> None:
+        logger.info(f"Saving temporary hostapd config file on {self.config_path()}")
+        with open(self.config_path(), "w", encoding="utf-8") as f:
+            f.write(self.hostapd_config())
+
+    @staticmethod
+    def device_serial() -> str:
+        try:
+            with open("/proc/cpuinfo", "r", encoding="utf-8") as f:
```
patrickelectric, line 243: Since the device serial information is not something that will exist in different boards or computers under cpuinfo, it's better to rely on: `/etc/machine-id`.
https://www.freedesktop.org/software/systemd/man/machine-id.html

core/services/wifi/WifiManager.py
```diff
@@ -314,7 +323,23 @@ async def auto_reconnect(self, seconds_before_reconnecting: float) -> None:
...
                 await self.wpa.send_command_reconnect()
+                self.enable_hotspot(True)
                 networks_reenabled = True
+
+    def enable_hotspot(self, enable: bool, ssid: Optional[str] = None, passphrase: Optional[str] = None) -> None:
+        if ssid is not None:
+            self._hotspot.set_ssid(ssid)
+        if passphrase is not None:
+            self._hotspot.set_passphrase(passphrase)
+        if enable:
```
patrickelectric, line 342: To make it more clear and easy to use, I would recommend to use a function just to disable/stop and remove the `enable` argument of the function

core/services/wifi/main.py
```diff
@@ -208,5 +208,5 @@ def hotspot_state() -> Any:
     config = Config(app=app, loop=loop, host="0.0.0.0", port=9000, log_config=None)
     server = Server(config)
 
-    loop.create_task(wifi_manager.auto_reconnect(10.0))
```
patrickelectric, line 193: Ouch, that's a lot of time.

core/services/wifi/main.py
```diff
@@ -208,5 +208,5 @@ def hotspot_state() -> Any:
...
+    loop.create_task(wifi_manager.auto_reconnect(60))
```
patrickelectric, line 211: Ouch, that's a lot of time. 
Can we adjust this value based if hotspot is running or not ?

patrickelectric (changes requested):
Some general comments and requests, I still need to test the docker image.

install/install.sh
```diff
@@ -162,6 +162,12 @@ sed -i '/noipv4ll/d' /etc/dhcpcd.conf
 # add noipv4ll
 sed -i '$ a noipv4ll' /etc/dhcpcd.conf
 
+tee -a /etc/dhcpcd.conf << END
+interface uap0
+    static ip_address=192.168.3.2/24
+    nohook wpa_supplicant
+END
```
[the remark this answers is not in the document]
patrickelectric: Based on your [last comment](https://github.com/bluerobotics/BlueOS-docker/pull/1019#issuecomment-1190183614), we can let right this for now.

core/services/wifi/Hotspot.py
joaoantoniocardoso: I'll leave here an interesting article about ids: http://0pointer.de/blog/projects/ids.html

core/services/wifi/Hotspot.py
```diff
@@ -0,0 +1,339 @@
...
+
+class HostapdFrequency(str, Enum):
+    """Valid hostapd frequency modes."""
+
+    HW_2_4 = "g"  # Hostapd id for 2.4 GHz mode
+    HW_5_0 = "a"  # Hostapd id for 5.0 GHz mode
+
+    @staticmethod
+    def mode_from_channel_frequency(channel: int) -> "HostapdFrequency":
+        return HostapdFrequency.HW_2_4 if str(channel).startswith("2") else HostapdFrequency.HW_5_0
```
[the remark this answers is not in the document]
rafaellehmkuhl: The hostapd input is the channel, not the frequency.

core/services/wifi/Hotspot.py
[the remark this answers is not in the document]
rafaellehmkuhl: This takes about 0.02-0.2 seconds, so it will take 20 iterations at most. I will put it to 0.1 seconds to make it less intensive.

core/services/wifi/main.py
rafaellehmkuhl: This is just a fallback, so we don't need to care much about it.

core/services/wifi/Hotspot.py
```diff
@@ -0,0 +1,339 @@
...
+class HotspotManager:
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        base_interface: str,
+        ipv4_gateway: IPv4Address,
+        ap_interface_name: str = "uap0",
+        ap_ssid: Optional[str] = None,
+        ap_passphrase: str = "blueosap",
```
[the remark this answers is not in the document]
rafaellehmkuhl: It already is. We make the SSID `None` just because we calculate its value from a method (it's not a plain string). Will make both equal to improve readability though.

core/services/wifi/Hotspot.py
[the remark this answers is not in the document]
patrickelectric: Ok, can you specify what are the inputs of channel ?
Are they under the 24XXMHz range ? Just to understand more the `str(channel).startswith("2")` logic, since it appears a bit weird looking only at the code and what it tried to accomplish.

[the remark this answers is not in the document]
rafaellehmkuhl: They are. In this case, the method argument had the wrong name. Fixed that. Changed the condition to use the `int(...) == 24`, although I imagine we won't have 2.5-2.9GHz networks anytime soon.

core/libs/commonwealth/commonwealth/utils/general.py
```diff
@@ -1,5 +1,23 @@
...
 
 def is_running_as_root() -> bool:
     return os.geteuid() == 0
+
+
+def device_id() -> str:
+    try:
+        with open("/sys/firmware/devicetree/base/serial-number", "r", encoding="utf-8") as f:
+            return f.read()
+    except Exception as error:
```
patrickelectric, line 14: This raise is weird, why we need to do it ? Can't we just fall back and try machine-id instead and have just a since raise in the end or return an optional ?

core/services/wifi/WifiManager.py
```diff
@@ -28,6 +30,7 @@ def connect(self, path: Any) -> None:
         self._updated_scan_results: Optional[List[ScannedWifiNetwork]] = None
         self._ignored_reconnection_networks: List[str] = []
         self.connection_status = ConnectionStatus.UNKNOWN
+        self._hotspot = HotspotManager("wlan0", IPv4Address("192.168.3.2"))
```
patrickelectric, line 33: I would recommend to move to `192.168.*133*.2`or something not close to the lower numbers, with that, we can avoid conflict with user network configuration inside the ROV in the future. @Williangalvani what you think ?

core/services/wifi/main.py
```diff
@@ -143,6 +143,24 @@ async def disconnect() -> Any:
     logger.info("Succesfully disconnected from network.")
 
 
+@app.post("/hotspot", summary="Enable/disable hotspot.")
+@version(1, 0)
+def toggle_hotspot(enable: bool, ssid: Optional[str] = None, password: Optional[str] = None) -> Any:
+    try:
+        wifi_manager.enable_hotspot(enable, ssid, passphrase=password)
```
patrickelectric: There is no more `enable` in `enable_hotspot`, how did the linters not get it ?

patrickelectric (changes requested):

core/services/wifi/WifiManager.py
joaoantoniocardoso: > I would recommend to move to `192.168.*133*.2`or something not close to the lower numbers, with that, we can avoid conflict with user network configuration inside the ROV in the future. @Williangalvani what you think ?

Usually, above 100 is used for dynamic IPs, so it might clash and get rejected.

joaoantoniocardoso: Why don't we ask for an IP? Wouldn't it be easier?

patrickelectric: > Why don't we ask for an IP? Wouldn't it be easier?

Hotspot should be something enabled by default. What we could add is an option to change the IP. But to do that, it may be necessary to change /etc/dhcpcd.conf file, and that's a pain..

core/services/wifi/main.py
rafaellehmkuhl: They got it. It's solved on the following commit. I will merge them when finished.

install/install.sh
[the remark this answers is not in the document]
patrickelectric: > Based on your [last comment](https://github.com/bluerobotics/BlueOS-docker/pull/1019#issuecomment-1190183614), we can let right this for now.

Did you check this: https://wiki.archlinux.org/title/dhcpcd#Hooks ?
It may be better for us to have a hook over editing the main conf file.
@rafaellehmkuhl what you think ?

core/libs/commonwealth/commonwealth/utils/general.py
rafaellehmkuhl: There's no raise here.

patrickelectric: My fault!

core/services/wifi/WifiManager.py
Williangalvani: Is there a reason not to end with .1? we only do .2 right now for... historical reasons. We should move away from this as we can.

As for the 192.168.3.* part, the only issue I can think of is that a user could have the same thought process and also setup a networking at (.2 +1), but I don't think it is likely for an user to have two different network

patrickelectric: > As for the 192.168.3.* part, the only issue I can think of is that a user could have the same thought process and also setup a networking at (.2 +1), but I don't think it is likely for an user to have two different networks in the same rov


for now, I would also go forward @joaoantoniocardoso previous comment:


> Usually, above 100 is used for dynamic IPs, so it might clash and get rejected.

rafaellehmkuhl: So `192.168.42.1`?

patrickelectric: Looks good.

core/services/wifi/main.py
rafaellehmkuhl: My bad. Committed that fix in the wrong commit.

patrickelectric (changes requested):
- hotspot should be enabled by default.
- It should have a persistent behavior across reboots. 
    - If it was 'on' on the last boot it should start 'on' on the current one

rafaellehmkuhl:
> * hotspot should be enabled by default.
> * It should have a persistent behavior across reboots.
>   
>   * If it was 'on' on the last boot it should start 'on' on the current one

Yep. It's the PS3 from the description. Will change that and PS2.

install/install.sh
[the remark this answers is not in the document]
patrickelectric: Maybe, should we bind `/etc/dhcpcd.conf` on the docker and to the necessary changes at runtime ?

patrickelectric (approved):

PR #1025, opened 2022-05-31
ardupilot-manager: Get stream from upload process

### LINKED ISSUE #500: core: frontend: Firmware flash: Add information about board and flash time

rafaellehmkuhl:
I was playing with streaming info from a (fast)API during the weekend and thought about implementing this old demand (#500) on BlueOS, which would also help users regarding the firmware upload problems.

For some reason, this doesn't work on BlueOS. I don't know if it's a problem with CORS, or maybe NGINX, or maybe the HTTP headers, but there's something blocking the operation, so instead of receiving the stream message by message, the frontend is receiving the entire stream (all messages) at once in the end of the HTTP response.

As it's not a priority, I will leave it here if someone in interested in trying fixing it. Maybe it's some silly problem I didn't realize. Maybe not.

Fix #500

### DIFF
--- a/core/frontend/src/components/autopilot/FirmwareManager.vue
+++ b/core/frontend/src/components/autopilot/FirmwareManager.vue
@@ -147,21 +147,25 @@
 
     <v-dialog
       v-model="show_install_progress"
-      hide-overlay
       persistent
-      width="300"
+      width="70%"
     >
       <v-card
         color="primary"
         dark
       >
-        <v-card-text>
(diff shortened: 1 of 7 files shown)


Williangalvani:
Nice.

Try using nginx settings similar to VersionChooser's:
https://github.com/bluerobotics/BlueOS-docker/blob/master/core/tools/nginx/nginx.conf#L157-L163

patrickelectric:

PR #1034, opened 2022-06-06
Update startup file

### LINKED ISSUE #414: bootstrap/update: allow update of binds in startup.json file

patrickelectric:
Fix #414


core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -0,0 +1,57 @@
...
+        if (k in dct and isinstance(dct[k], dict) and isinstance(merge_dct[k], dict)):  #noqa
+            dict_merge(dct[k], merge_dct[k])
+        else:
+            dct[k] = merge_dct[k]
+
+def main():
+    current_git_version = os.getenv('GIT_DESCRIBE_TAGS')
+    match = re.match(r'(?P<tag>.*)-(?P<commit_number>\d+)-(?P<commit_hash>[a-z0-9]+)', current_git_version)
+    tag, commit_number, commit_hash = match['tag'], match['commit_number'], match['commit_hash']
+    print(f"Running BlueOS: {tag=}, {commit_number=}, {commit_hash=}")
```
Williangalvani, line 37: was this data supposed to be consumed somwhere?

Williangalvani (commented):
The GIT_DESCRIBE thing seems to be a bit unrelated to the PR as it is, was it going to be consumed somewhere?

core/tools/blueos_startup_update/blueos_startup_update
patrickelectric: No, the idea is to make it available if version check is necessary inside BlueOS, as possible necessary in this script. Right now there is no way to know BlueOS version inside core.

Williangalvani (approved):

PR #1040, opened 2022-06-07
core: tools: scripts: bootstrap: Use absolute path for red-pill location.

joaoantoniocardoso:
Using relative path actually is relying on having "/" as the current working directory, which ~will~ might not be the case in future blueos-base docker images (eg. we may choose to use /home/pi as the default login path).

### DIFF
--- a/core/tools/scripts/bootstrap.sh
+++ b/core/tools/scripts/bootstrap.sh
@@ -3,5 +3,5 @@
 # Exit immediately if a command exits with a non-zero status
 set -e
 
-SCRIPTS_PATH=$(dirname "$0")
-cp $PWD/$SCRIPTS_PATH/red-pill /usr/bin/
+SCRIPTS_PATH="$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )";
+cp $SCRIPTS_PATH/red-pill /usr/bin/


core/tools/scripts/bootstrap.sh
```diff
@@ -3,5 +3,6 @@
 # Exit immediately if a command exits with a non-zero status
 set -e
 
-SCRIPTS_PATH=$(dirname "$0")
-cp $PWD/$SCRIPTS_PATH/red-pill /usr/bin/
+LOCAL_BINARY_PATH="/usr/bin/red-pill"
+
+cp /home/pi/tools/scripts/red-pill $LOCAL_BINARY_PATH
```
patrickelectric, line 8: You can rely that the source will be in this path, it's necessary to use dirname for relative access. Can you give an example where this fail ?

joaoantoniocardoso: I have set `WORKDIR=/home/pi` on the BlueOS-base's Dockerfile, then the old method expands as `cp /home/pi//home/pi/tools/scripts/red-pill /usr/bin/pi`, obviously wrong.

The `red-pill` will be at `/home/pi/tools/scripts/` because of `COPY tools /home/pi/tools` ([from core's Dockerfile](https://github.com/bluerobotics/BlueOS-docker/blob/master/core/Dockerfile#L19)).

joaoantoniocardoso: It could also be:
```suggestion
cd /
cp $PWD/$SCRIPTS_PATH/red-pill /usr/bin/
cd $OLDPWD
```

joaoantoniocardoso: yeah, $(dirname $(realpath $0)) does, but [this](https://stackoverflow.com/a/246128) looks better.

joaoantoniocardoso:
I've Tested this both with blueos-base:master and [blueos-base:gstreamer_1.20.2_with_webrtcsink](https://hub.docker.com/layers/blueos-base/joaoantoniocardoso/blueos-base/gstreamer_1.20.2_with_webrtcsink/images/sha256-05a0cbcb23a2707db104c70bdf56b6a0733d219e30f95edf17e56812637ca636?context=explore), looks fine.

patrickelectric (approved):

PR #1066, opened 2022-07-07
bridget: Fix some inconsistent logs

rafaellehmkuhl:
Logs were missing or inconsistent, making debug confuse.


patrickelectric (approved):
The commit message isn't clear compared with the change. How does this changes makes logging more consistent.

patrickelectric (changes requested):
Wrong button.

rafaellehmkuhl:
> The commit message isn't clear compared with the change. How does this changes makes logging more consistent. 

The "add" method logs that it's deleting the bridge.

PR #1076, opened 2022-07-15
Update bridges interface

patrickelectric:
- Use `/dev/serial/by-path` when available over `/dev/{device}`
- Docker: patrickelectric/blueos-core:update_bridges_interface

image
image


core/frontend/src/store/system-information.ts
```diff
@@ -237,6 +254,7 @@ const system_information: SystemInformationStore = getModule(SystemInformationSt
 
 system_information.fetchSystem()
 callPeriodically(system_information.fetchPlatform, 5000)
+callPeriodically(system_information.fetchSerial, 5000)
```
Williangalvani, line 257: not a big fan of these periodic calls... but we need a better solution for these anyway

Williangalvani (approved):

PR #1079, opened 2022-07-18
Use immutable with yarn

patrickelectric:
From yarn documentation:
> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.
> If you want to ensure yarn.lock is not updated, use --frozen-lockfile.

`--immutable` is the "new" waay to call `frozen-lockfile`, that's considered deprecated.

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Yarn install
         uses: borales/actions-yarn@v3.0.0
         with:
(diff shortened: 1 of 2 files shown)


Williangalvani (approved):

ES-Alexander:
Is this waiting on some testing or something, or can it be merged?

patrickelectric:
The documentation is a bit confuse but the results that we are currently having on master branch is already what we expect. We can close it for now and reopen if we have problems once moving to yarn3.

PR #1086, opened 2022-07-21
core: frontend: views: MainView: Improve welcome message

patrickelectric:


ES-Alexander (approved):

rafaellehmkuhl (approved):
Every single text that we Brazilians write to the user interface should be double-checked by native speakers.

PR #1089, opened 2022-07-22
Allow user to enable raspberry camera legacy mode

patrickelectric:
image

### DIFF
--- a/core/frontend/src/components/video-manager/VideoManager.vue
+++ b/core/frontend/src/components/video-manager/VideoManager.vue
@@ -45,23 +45,28 @@
         bottom
         right
         class="v-btn--example"
-        @click="openResetDialog"
+        @click="openSettingsDialog"
       >
-        <v-icon>mdi-restart-alert</v-icon>
+        <v-icon>mdi-cog</v-icon>
       </v-btn>
     </v-fab-transition>
 
     <v-dialog
       width="500"
-      :value="show_reset_dialog"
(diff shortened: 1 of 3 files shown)


joaoantoniocardoso:
Looks good, if it isn't too difficult, we might want to add a quick link to "restart now" in this dialogue:
image

patrickelectric:
> Looks good, if it isn't too difficult, we might want to add a quick link to "restart now" in this dialogue: image

I don't know if that's possible, @rafaellehmkuhl ?

rafaellehmkuhl:
> > Looks good, if it isn't too difficult, we might want to add a quick link to "restart now" in this dialogue: image
> 
> I don't know if that's possible, @rafaellehmkuhl ?

To add just the restart button, you would need to isolate it from the PowerMenu component, where it currently resides.
A faster option would be to just open the PowerMenu dialog, which is just a matter of importing the component and showing it.

patrickelectric:
> > > Looks good, if it isn't too difficult, we might want to add a quick link to "restart now" in this dialogue: image
> > 
> > 
> > I don't know if that's possible, @rafaellehmkuhl ?
> 
> To add just the restart button, you would need to isolate it from the PowerMenu component, where it currently resides. A faster option would be to just open the PowerMenu dialog, which is just a matter of importing the component and showing it.

I think that we can let as it's for now and rework that later to make a generic "restart is required" component.

joaoantoniocardoso (approved):

PR #1096, opened 2022-07-28
wifi-manager: Fix hotspot issues

rafaellehmkuhl:


patrickelectric:
Can you add in the PR body which issue this PR is fixing ?

core/services/wifi/WifiManager.py
```diff
@@ -106,6 +113,16 @@ def __dict_from_list(data: bytes) -> Dict[str, Any]:
 
         return output
 
+    @property
+    def hotspot(self) -> HotspotManager:
+        if self._hotspot is None:  # pylint: disable=access-member-before-definition
```
patrickelectric, line 118: What is the problem to define `self._hotspot` as None by default in the constructor ?

patrickelectric (changes requested):

core/services/wifi/WifiManager.py
rafaellehmkuhl: It is. Line 40.

patrickelectric: It's unnecessary inside a try block group, that's why you need the pylint disable comment, removing from the block will fix it.

patrickelectric:
@rafaellehmkuhl to make sure, will this fix the tagged issue ?
- Will this make the wifi service work if hotspot fails to start ?
- Will this enable the channels ? Will this make hotspot work with the channels disabled ?

From what I can see on the code, this only makes the wifi to work if something goes wrong with hotspot, and that great. But it appears that does not fix the hotspot issue when starting, making it unavailable on the first boot of the system (something that really needs to be fixed).

rafaellehmkuhl:
> @rafaellehmkuhl to make sure, will this fix the tagged issue ?
> 
> * Will this make the wifi service work if hotspot fails to start ?
> * Will this enable the channels ? Will this make hotspot work with the channels disabled ?
> 
> From what I can see on the code, this only makes the wifi to work if something goes wrong with hotspot, and that's great. But it appears that does not fix the hotspot issue when starting, making it unavailable on the first boot of the system (something that really needs to be fixed).

It does already.
I better explained it in the last software-meeting, but basically the wlan interface is configured in its first usage, so it fix itself already. The problem was that we were starting the hotspot before any wlan usage, so we were using it unconfigured. By adding a watchdog routine, everything is fixed seamlessly because it will try to start the hotspot again in a time the wlan is already configured.

patrickelectric (approved):

PR #1105, opened 2022-08-04
Fix multiple streams issue with QGC and mavlink-router

### LINKED ISSUE #1011: bug: cameras inconsistently appearing in QGC
### Bug description

Brought up in [this forum post](https://discuss.bluerobotics.com/t/disable-automatic-video-display-on-qgroundcontrol/12109):

> QGroundControl with BlueOS will automatically display the video streams ... [but] the video changing function is not very reliable and having 3 cameras on our setup, most of the time only 1 or 2 show up and its inconsistent with which camera shows and which don’t.

### Steps to reproduce

1. Start multiple streams in BlueOS
2. Check QGC for available ones
3. Repeat with inconsistent results [NEEDS CONFIRMATION]

### Primary pain point(s)

Makes it harder to effectively use multiple cameras with BlueOS, and encourages using external applications for streaming any/all video instead of QGC (which fragments the user experience).

### Prerequisites

patrickelectric:
Fix #1011

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -222,9 +222,9 @@ class ArduPilotManager(metaclass=Singleton):
             Endpoint(
                 "Internal Link",
                 self.settings.app_name,
-                EndpointType.UDPServer,
+                EndpointType.TCPServer,
                 "127.0.0.1",
-                14001,
+                5777,
                 persistent=True,
                 protected=True,
             ),
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -31,7 +31,7 @@ SERVICES=(
     'commander',"$SERVICES_PATH/commander/main.py"
     'nmea_injector',"$SERVICES_PATH/nmea_injector/nmea_injector/main.py"
     'helper',"$SERVICES_PATH/helper/main.py"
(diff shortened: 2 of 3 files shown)


joaoantoniocardoso (approved):
Great!

patrickelectric:
Needs more test and investigation

patrickelectric:
This still may break SITL since, from what I know, mavlink-router does not support multiple TCP servers, we should need to check if restarting the vehicle will not break the camera manager, since this is a TCP connection and we should handle disconnections.
@rafaellehmkuhl any tips ? @joaoantoniocardoso, please checkout

rafaellehmkuhl:
> This still may break SITL since, from what I know, mavlink-router does not support multiple TCP servers, we should need to check if restarting the vehicle will not break the camera manager, since this is a TCP connection and we should handle disconnections. @rafaellehmkuhl any tips ? @joaoantoniocardoso, please checkout

From what I remember, it doesn't support multiple TCP servers indeed.

patrickelectric:
> > This still may break SITL since, from what I know, mavlink-router does not support multiple TCP servers, we should need to check if restarting the vehicle will not break the camera manager, since this is a TCP connection and we should handle disconnections. @rafaellehmkuhl any tips ? @joaoantoniocardoso, please checkout
> 
> From what I remember, it doesn't support multiple TCP servers indeed.

So we may want to use 5760 in the end

Williangalvani:
superseeded by #1122

Issue #264, opened 2021-05-21
Ping service development tracker


Williangalvani:
- [x] Ping 1D
 - [x] Ping360
 - [x] Hotplugging
 - [ ] Mavlink interface for Ardusub
 - [ ] Support and advertise multiple devices
 - [ ] Web UI
 - [ ] Tests

ES-Alexander:
Can we add "support flashing Ping firmware"?

ES-Alexander:
Would be cool to allow Ping Sonars to be configured for either altitude measurement or obstacle awareness/avoidance, and ideally enable switching between two sonars on the same vehicle :-)

Raised in [this forum post](https://discuss.bluerobotics.com/t/2-ping-sonars-on-one-rov/11958).

PR #1111, opened 2022-08-05
Pingdriver

Williangalvani:
image at williangalvani/blueos-docker:pingdriver

what works:
 - [x] hotplugging
 - [x] bridges for ping360
-  [x] discovery of ethernet-enabled Ping360s
 - [x] bridges for ping1d
 - [x] mavlink driver for ping1d
    - [x] Toggling on/off
    - [x] Persistence using settings system
 - [x] new Navigator SVG from Daniel

image

Helps #264


core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,66 @@
...
+
+    available_networks = []
+    ips = []
+    for interface in stats.keys():
+        if interface in stats and getattr(stats[interface], "isup"):
+            available_networks.append(interface)
+
+    for interface in available_networks:
+        phys_list = list(filter(lambda address: address.family == socket.AF_INET, psutil.net_if_addrs()[interface]))
+        ips.extend([phys.address for phys in phys_list])
```
ES-Alexander, line 27: - avoid repeated `psutil` calls
- clean up with filtered generator expression
- avoid needless list creation

```suggestion
    network_interface_addresses = psutil.net_if_addrs()
    for interface in available_networks:
        ips.extend(phys.address for phys in network_interface_addresses[interface] if phys.family == socket.AF_INET)
```

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,66 @@
...
+    return list(set(ips))  # make sure there are not duplicated entries
+
+
+def remove_zeros(ip: str) -> str:
+    """
+    This removes the leading zeros return by the Ping360 firmware.
+    This functions transforms 192.168.015.016 into 192.168.15.16
+    """
+    new_ip = ".".join([str(int(i)) for i in ip.split(".")])
```
ES-Alexander, line 36: Use generator expression instead of list comprehension for single-use iterator
```suggestion
    new_ip = ".".join(str(int(i)) for i in ip.split("."))
```

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,67 @@
...
     def __init__(self) -> None:
         self.drivers: Dict[PingDeviceDescriptor, PingDriver] = {}
-        self.ping1d_current_port: int = 9090
-        self.ping360_current_port: int = 9092
+        self.ping1d_base_port: int = 9090
+        self.ping360_base_port: int = 9092
 
     def stop_driver_at_port(self, port: Serial) -> None:
         """Stops the driver instance running for port "port" """
         ping_at_port = list(filter(lambda ping: ping.port == port, self.drivers.keys()))
```
ES-Alexander, line 21: Clean up with list comprehension. Default dictionary iterator is already over the keys.
```suggestion
        ping_at_port = [ping for ping in self.drivers if ping.port == port]
```

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,67 @@
...
+            self.drivers[ping] = Ping360Driver(ping, -1)
+
+    def find_next_port(self, base_port: int, direction: int) -> int:
+        """
+        Find the next unused udp port, starts on base_port and increments/decrements
+        accordingly to direction
+        """
+        port = base_port
+        while udp_port_is_in_use(port):
+            port = port + direction
```
ES-Alexander, line 37:
```suggestion
            port += direction
```

core/services/ping/pingutils.py
```diff
@@ -36,12 +39,26 @@ class PingDeviceDescriptor:
...
 FW: v{self.firmware_version_major}.{self.firmware_version_minor}.{self.firmware_version_patch}
-port: {self.port.hwid}"""
+port: {self.get_hw_or_eth_info()}"""
+
+
+def udp_port_is_in_use(port: int) -> bool:
+    for conn in psutil.net_connections():
+        if conn.laddr.port == port and conn.type == socket.SocketKind.SOCK_DGRAM:
+            return True
+    return False
```
ES-Alexander, line 64:
```suggestion
    return any(conn.laddr.port == port and conn.type == socket.SocketKind.SOCK_DGRAM
               for conn in psutil.net_connections())
```

ES-Alexander (commented):
New UI is great!
Love the images that show where things are plugged in, and it'll be great to have Ping1D MAVLink messages again (one step closer to full Companion replacement). Also really cool that it detects Ping360s over ethernet. From a support standpoint this update should be super helpful - thanks! :-)

No bugs found, just some minor improvements and consideration points :-)

core/services/ping/pingutils.py
Williangalvani: I always forget about any/some :man_shrugging:

core/services/ping/main.py
```diff
@@ -40,6 +40,14 @@ def get_sensors() -> Any:
     return [PingDeviceDescriptorModel.from_descriptor(device) for device in ping_manager.devices()]
 
 
+@app.post("/sensors", status_code=status.HTTP_200_OK, summary="Set sensor settings.")
+@version(1, 0)
+def set_sensor(sensor_settings: dict[str, Any]) -> Any:
+    if "port" not in sensor_settings:
+        raise ValueError("'device' key is missing")
```
patrickelectric, line 47: The correct error message would be 'port', no ?

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,154 @@
...
+    def distance_message(time_boot_ms: int, distance: int, device_id: int) -> Dict[str, Any]:
+        return {
+            "type": "DISTANCE_SENSOR",
+            "time_boot_ms": time_boot_ms,
+            "min_distance": 20,
+            "max_distance": 5000,
+            "current_distance": int(distance),
+            "mavtype": {"type": "MAV_DISTANCE_SENSOR_LASER"},
+            "id": device_id,
+            "orientation": {"type": "MAV_SENSOR_ROTATION_PITCH_270"},
```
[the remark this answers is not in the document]
Williangalvani: that's being tracked here:
https://github.com/bluerobotics/BlueOS-docker/issues/264

core/frontend/src/menus.ts
```diff
@@ -16,6 +16,14 @@ const menus = [
         text: `General Autopilot settings. Allows you to start/stop Ardupilot (if using Navigator or SITL), or update
         the firmware of your flight controller.`,
       },
+      {
+        title: 'Pings',
+        icon: 'mdi-radar',
+        route: '/autopilot/pings',
+        advanced: false,
+        text: 'Ping devices discovery and management. This allows you to see all detected Ping devices connected to '
+              + 'either your Companion Computer or its local network',
```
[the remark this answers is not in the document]
Williangalvani: idk about this one. Companion computer is a concept that is not going anywhere anytime soon. But I do worry about confusing our users...

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,154 @@
...
+            "covariance": 0,
```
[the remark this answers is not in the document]
Williangalvani: I think the issue here was that QGC doesn't care for confidence, so our users were just seeing messed up data in the HUD.
That's why there's a check in line 64. 

I don't see any harm in sending the confidence when we do actually send the message, though.

core/services/ping/ping360_ethernet_prober.py
Williangalvani: nice catch

core/services/ping/pingmanager.py
Williangalvani: lol. the thought process behind that was... something. let's not get into details

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,154 @@
...
+                self.ping1d_io.connect(pingserver)
+                await asyncio.sleep(0.1)
+
+            # read data in from ping device
+            try:
+                data, _ = self.ping1d_io.recvfrom(4096)
+            except socket.error as exception:
+                # check if it's waiting for data
+                if exception.errno == errno.EAGAIN:
+                    continue
```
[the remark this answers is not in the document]
Williangalvani: true. Though a wise man once said "explicit is better than impicit". I do like having EAGAIN there so we know it happens and is being taking into consideration...

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,154 @@
...
+            "time_boot_ms": time_boot_ms,
+            "min_distance": 20,
+            "max_distance": 5000,
+            "current_distance": int(distance),
+            "mavtype": {"type": "MAV_DISTANCE_SENSOR_LASER"},
+            "id": device_id,
+            "orientation": {"type": "MAV_SENSOR_ROTATION_PITCH_270"},
+            "covariance": 0,
+            "horizontal_fov": 0,
+            "vertical_fov": 0,
```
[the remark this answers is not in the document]
Williangalvani: I think both should be set to this, assuming our beam is conical.
vertical and horizontal are probably the beam width in two different axis, right? :thinking:

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,154 @@
...
+            "quaternion": [0, 0, 0, 0],
+            "signal_quality": 0,
```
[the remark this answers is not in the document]
Williangalvani: done!

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,183 @@
...
+        return require('../../assets/img/devicePathHelper/rpi3b.jpg')
+      }
+      return ''
+    },
+    board_connector () : string {
+      const serial_port_path = this.serial_port_path as string
+      try {
+        return connector_map[serial_port_path]
+      } catch (error) {
+        console.log(error)
```
patrickelectric, line 98: This is not a log, this is an erro, use: `console.error`

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,183 @@
...
+      } else if (this.board_type === BoardType.Rpi4B) {
+        /* eslint-disable */
+        return require('../../assets/img/devicePathHelper/rpi4b.png')
+      } else if (this.board_type === BoardType.Rpi3B) {
+        /* eslint-disable */
+        return require('../../assets/img/devicePathHelper/rpi3b.jpg')
+      }
+      return ''
+    },
+    board_connector () : string {
```
patrickelectric, line 93: This should return an optional, avoid using empty strings for error handling

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,183 @@
...
+      const serial_port_path = this.serial_port_path as string
+      try {
+        return connector_map[serial_port_path]
+      } catch (error) {
+        console.log(error)
+      }
+      return ''
+    },
+    circle_class () : string {
+        return "circle " + this.board_connector
```
patrickelectric, line 103: This will fail if `board_connection` returns an empty string

core/frontend/src/components/ping/ping1d.vue
```diff
@@ -0,0 +1,115 @@
...
+    }
+  },
+  computed: {
+  },
+  mounted() {
+    this.user_desired_mavlink_driver_state = Boolean(this.device.driver_status.mavlink_driver_enabled)
+  },
+  methods: {
+    switch_loading() {
+      return this.user_desired_mavlink_driver_state !== !!this.device.driver_status.mavlink_driver_enabled
```
patrickelectric, line 94: Why not `Boolean(` ?

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,153 @@
...
+        self.tboot = time.time()
+        self.ping1d_io = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        self.ping1d_io.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        self.ping1d_io.setblocking(False)
+
+    def set_should_run(self, should_run: bool) -> None:
+        self.should_run = should_run
+
+    @staticmethod
+    def distance_message(time_boot_ms: int, distance: int, device_id: int, confidence: int) -> Dict[str, Any]:
```
patrickelectric, line 44: It would be nice to have the distance units in the variable name

core/frontend/vue.config.js
```diff
@@ -49,6 +49,9 @@ module.exports = {
       '^/network-test': {
         target: SERVER_ADDRESS,
       },
+      '^/ping': {
```
patrickelectric, line 52: but the commits convention that we are following, the commit body should contains `frontend:`.

core/frontend/src/types/frontend_services.ts
```diff
@@ -109,5 +109,5 @@ export const ping_service: Service = {
   name: 'Ping service',
   description: 'Service to manage ping devices',
   company: 'Blue Robotics',
-  version: '0.1.0',
+  version: '0.2.0',
```
patrickelectric, line 112: This should be squashed with 8eb54d7af3a2940d9cba9e74b1fd76a6de9c676b

patrickelectric, line 112: Avoid mixing python with frontend commits.

core/services/ping/pingmanager.py
```diff
@@ -27,15 +27,27 @@ async def register_ethernet_ping360(self, ping: PingDeviceDescriptor) -> None:
         if ping not in self.drivers:
             self.drivers[ping] = Ping360Driver(ping, -1)
 
+    def find_next_port(self, base_port: int, direction: int) -> int:
+        """
+        Find the next unused udp port, starts on base_port and increments/decrements
+        accordingly to direction
+        """
+        port = base_port
+        while udp_port_is_in_use(port):
```
patrickelectric, line 36: We should add a sleep here to avoid cpu lock if the `udp_port_is_un_use` always return a true somehow.

core/services/ping/pingmanager.py
```diff
@@ -27,15 +27,27 @@ async def register_ethernet_ping360(self, ping: PingDeviceDescriptor) -> None:
         if ping not in self.drivers:
             self.drivers[ping] = Ping360Driver(ping, -1)
 
+    def find_next_port(self, base_port: int, direction: int) -> int:
```
patrickelectric, line 30: Why not using port `0` to bind and get the port that the OS found for us ?

core/services/ping/pingutils.py
```diff
@@ -36,12 +39,25 @@ class PingDeviceDescriptor:
...
+    driver: Optional["PingDriver"]  # type: ignore
+
+    def get_hw_or_eth_info(self) -> str:
+        if self.port:
+            return str(self.port.device_path)
+        return str(self.ethernet_discovery_info)
 
     def __hash__(self) -> int:
-        return hash(self.port.hwid)
+        return hash(self.get_hw_or_eth_info())
```
patrickelectric, line 51: `ethernet_discovery_info` can be an empty string, in that case the function return value, that is used in around this PR with `==` with others variables and this hash function will give will have some problems, we should avoid or take care of that somehow, maybe moving `get_hw_or_eth_info` to an optional.

core/frontend/src/menus.ts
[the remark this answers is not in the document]
ES-Alexander: I think the idea was that when there's a separate flight controller, it's a companion computer, but since Navigator can't operate without the computer then "Onboard Computer" is a better term for general use (since it encompasses both options).

core/frontend/src/views/Pings.vue
```diff
@@ -0,0 +1,54 @@
...
+      class="d-flex pa-5"
+    >
+      <div
+        v-for="ping in ping_devices"
+        :key="ping.device_id"
+        class="pa-2"
+        style="min-height: 100%;"
+      >
+        <ping-card
+          v-if="ping.ping_type == 'Ping1D'"
```
patrickelectric, line 13: Create an enum on `ping.ts` to avoid string comparison.

patrickelectric (changes requested):
An initial review, will catch up later

core/services/ping/pingdriver.py
```diff
@@ -14,6 +14,8 @@ def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
         self.ping = ping
         self.port = port
         self.bridge: Optional[Bridge] = None
+        self.driver_status: Dict[str, Union[int, float, str, bool]] = {"udp_port": port}
```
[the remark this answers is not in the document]
Williangalvani: I'm leaving this generically on purpose as we'll have different settings for different sensors

core/services/ping/ping1d_mavlink.py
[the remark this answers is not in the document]
ES-Alexander: Lol, it's pretty ambiguous. That's definitely one interpretation that could be valid though...

core/services/ping/ping1d_mavlink.py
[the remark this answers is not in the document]
ES-Alexander: It could be in a comment, above the general `continue` for that loop?
Not a huge deal, I'm just not a fan of multiple branches that do exactly the same thing :-)

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,183 @@
...
+  '/dev/ttyAMA3': 'serial5',
+  '/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.3:1.0-port0': 'usb-top-left-pi4',
+  '/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0-port0': 'usb-bottom-left-pi4',
+  '/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.1:1.0-port0': 'usb-top-right-pi4',
+  '/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.2:1.0-port0': 'usb-bottom-right-pi4',
+  '/dev/serial/by-path/platform-3f980000.usb-usb-0:1.5:1.0-port0': 'usb-bottom-right-pi3', // RPI3
+  '/dev/serial/by-path/platform-3f980000.usb-usb-0:1.4:1.0-port0': 'usb-top-right-pi3', // RPI3
+  '/dev/serial/by-path/platform-3f980000.usb-usb-0:1.3:1.0-port0': 'usb-bottom-left-pi3', // RPI3
+  '/dev/serial/by-path/platform-3f980000.usb-usb-0:1.2:1.0-port0': 'usb-top-left-pi3', // RPI3
+}
```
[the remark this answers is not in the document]
Williangalvani: I am not sure, if the maps don't work, it should just not show anything (not even the (i) iicon)

core/services/ping/pingmanager.py
ES-Alexander: Haha, it's easy enough to forget sometimes. Or end up arriving at something like that in a roundabout way where you started with something more complex and whittled it down to an elegant and simple solution like the one here, without realising at the time that there's an operator for that :-P

core/services/ping/pingutils.py
ES-Alexander: `any` and `all` are quite useful. Same with the boolean operators (`or`, `and`). Short-circuiting things are oddly satisfying to me :-)

core/services/ping/pingmanager.py
Williangalvani: I prefer following some kind of order for now...

core/services/ping/main.py
```diff
@@ -40,6 +40,14 @@ def get_sensors() -> Any:
     return [PingDeviceDescriptorModel.from_descriptor(device) for device in ping_manager.devices()]
 
 
+@app.post("/sensors", status_code=status.HTTP_200_OK, summary="Set sensor settings.")
+@version(1, 0)
+def set_sensor(sensor_settings: dict[str, Any]) -> Any:
```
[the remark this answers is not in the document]
Williangalvani: the goal here is to have a somewhat abstract interface we can use for all sensors eventually, the goal is to have simple key:values for now, but allowing at least the basic types

core/services/ping/pingmanager.py
Williangalvani: also, we're starting at the ports checked by pingviewer for now. we still dont´t have the discovery protocol implemented for alternative ports

core/services/ping/pingutils.py
Williangalvani: ethernet_discovery_info should never be empty when `port is None`. I added a new commit for validating it

core/frontend/src/menus.ts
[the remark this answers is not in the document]
Williangalvani: ok, I'm convinced.

core/frontend/src/components/common/DevicePathHelper.vue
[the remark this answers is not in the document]
joaoantoniocardoso: I believe this whole by-path idea will work when it refers to the motherboard's native USB, but I don't think this will take care when someone uses some USB hub, we should test it.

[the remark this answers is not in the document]
joaoantoniocardoso: Well done @Williangalvani ! So, with a regex we maybe could have an indication of where the hub is connected to, clearly saying that we are talking about the hub, not the device, but it is way beyond what we want for now, so I leave it here just as something to be improved next time, maybe.

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,63 @@
...
+        ips.extend(phys.address for phys in network_interface_addresses[interface] if phys.family == socket.AF_INET)
+    return list(set(ips))  # make sure there are not duplicated entries
+
+
+def remove_zeros(ip: str) -> str:
+    """
+    This removes the leading zeros return by the Ping360 firmware.
+    This functions transforms 192.168.015.016 into 192.168.15.16
+    """
+    new_ip = ".".join(str(int(i)) for i in ip.split("."))
```
ES-Alexander, line 32:
```suggestion
    Removes leading zeros from sections of an IP address.
      e.g. transforms 192.168.015.016 into 192.168.15.16
    
    Leading zero IPs are returned by the Ping360 firmware.
    """
    new_ip = ".".join(str(int(section)) for section in ip.split("."))
```

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,68 @@
...
+            self.drivers[ping] = Ping360Driver(ping, None)
+
+    async def find_next_port(self, base_port: int, direction: int) -> int:
+        """
+        Find the next unused udp port, starts on base_port and increments/decrements
+        accordingly to direction
+        """
+        port = base_port
+        while udp_port_is_in_use(port):
+            port += direction
```
ES-Alexander, line 37: Makes it clear it could also be used with a step-size greater than 1.
```suggestion
    async def find_next_port(self, base_port: int, step: int) -> int:
        """
        Finds the next unused UDP port.
        Starts at 'base_port' and increments/decrements by 'step'.
        """
        port = base_port
        while udp_port_is_in_use(port):
            port += step
```

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,68 @@
...
+            await asyncio.sleep(0.1)
+        return port
+
+    async def launch_driver_instance(self, ping: PingDeviceDescriptor) -> None:
         """Launches a new driver instance for the PingDeviceDescriptor "ping" """
```
ES-Alexander, line 42:
```suggestion
        """Launches a new driver instance for the PingDeviceDescriptor "ping". """
```

core/services/ping/pingprober.py
```diff
@@ -11,19 +11,16 @@
 class PingProber:
     """PingProber is responsible for identifying Ping-enabled devices on serial ports"""
```
ES-Alexander, line 12:
```suggestion
    """PingProber is responsible for identifying Ping-enabled devices on serial ports. """
```

core/services/ping/pingprober.py
```diff
@@ -11,19 +11,16 @@
...
 
-    def __init__(self) -> None:
-        self.ping_found_callback: Callable[[PingDeviceDescriptor], None] = lambda x: None
-
-    def probe(self, port: SysFS) -> Optional[PingDeviceDescriptor]:
+    async def probe(self, port: SysFS) -> Optional[PingDeviceDescriptor]:
         """Attempts to communicate via Ping Protocol at port "port",
         calls on_ping_found when a ping device is found"""
```
ES-Alexander, line 16:
```suggestion
        """Attempts to communicate via Ping Protocol at port "port".
        Calls on_ping_found callback when a ping device is found. """
```

core/services/ping/pingutils.py
```diff
@@ -35,13 +40,30 @@ class PingDeviceDescriptor:
     firmware_version_major: int
     firmware_version_minor: int
     firmware_version_patch: int
-    port: SysFS
+    port: Optional[SysFS]
+    ethernet_discovery_info: Optional[str]
+    driver: Optional["PingDriver"]  # type: ignore
+
+    def __post_init__(self) -> None:
+        if not any([self.port, self.ethernet_discovery_info]):
```
ES-Alexander, line 48:
```suggestion
        if not (self.port or self.ethernet_discovery_info):
```

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,68 @@
...
+            port = await self.find_next_port(self.ping360_base_port, direction=+1)
+            driver = Ping360Driver(ping, port)
 
         self.drivers[ping] = driver
-        driver.start()
+        loop = asyncio.get_running_loop()
+        loop.create_task(driver.start())
 
     def devices(self) -> List[PingDeviceDescriptor]:
         return list(self.drivers.keys())
```
ES-Alexander, line 58:
```suggestion
        return list(self.drivers)
```

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,63 @@
...
+    available_networks = []
+    ips: List[str] = []
+    for interface in stats.keys():
+        if interface in stats and getattr(stats[interface], "isup"):
+            available_networks.append(interface)
+
+    network_interface_addresses = psutil.net_if_addrs()
+    for interface in available_networks:
+        ips.extend(phys.address for phys in network_interface_addresses[interface] if phys.family == socket.AF_INET)
+    return list(set(ips))  # make sure there are not duplicated entries
```
ES-Alexander, line 24:
```suggestion
    return set(ips)  # make sure there are not duplicated entries
```
If a `set` is being used to avoid duplicates, the order is already effectively random. Is the result of this being indexed into? If not (i.e. it's just being iterated over) then it can potentially be left as a `set`.

core/services/ping/portwatcher.py
```diff
@@ -35,21 +45,47 @@ def port_should_be_probed(self, port: SysFS) -> bool:
             return False
         return True
 
-    def probe_port(self, port: SysFS) -> None:
+    async def probe_port(self, port: SysFS) -> None:
         """Attempts to probe port "port" for up to MAX_ATTEMPTS times"""
```
ES-Alexander, line 49:
```suggestion
        """Attempts to probe "port" for up to MAX_ATTEMPTS times. """
```

core/services/ping/portwatcher.py
```diff
@@ -15,10 +16,19 @@ class PortWatcher:
     """Watches the Serial ports on the system, calls set_prober when a port is found,
     port_post_callback when a port is no longer present"""
```
ES-Alexander, line 17:
```suggestion
    """Watches the Serial ports on the system.
    Calls set_prober when a port is found, and port_post_callback when a port is no longer present. """
```

core/services/ping/portwatcher.py
```diff
@@ -35,21 +45,47 @@ def port_should_be_probed(self, port: SysFS) -> bool:
...
+                    firmware_version_minor=0,
+                    firmware_version_patch=0,
+                    ethernet_discovery_info=f"{ip}:12345",
+                    port=None,
+                    driver=None,
+                )
+            )
+
     async def start_watching(self) -> None:
         """Start watching for plugged/unplugged serial devices in the system"""
```
ES-Alexander, line 90:
```suggestion
        """Start watching for plugged/unplugged serial devices in the system. """
```

ES-Alexander (approved):
A few more clarity / consistency suggestions, but nothing major :-)

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,191 @@
...
+  },
+  computed: {
+    serial_port_path(): string {
+      /* returns the by-path path for the serial port if available */
+      const system_serial_ports: SerialPortInfo[] | undefined = system_information.serial?.ports
+      if (system_serial_ports === undefined || system_serial_ports.isEmpty()) {
+        return this.device as string
+      }
+      const port = system_serial_ports.find((a) => a.name === this.device)
+      return port?.by_path ?? this.device as string
```
patrickelectric, line 66: `return system_information.serial?.ports.?.find((a) => a.name === this.device) ?? this.device as string`

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,191 @@
...
+        return BoardType.Navigator
+      }
+      if (serial_port_path.includes('platform-3f980000')) {
+        return BoardType.Rpi3B
+      }
+      if (serial_port_path.includes('platform-fd500000')) {
+        return BoardType.Rpi4B
+      }
+      return BoardType.Unknown
+    },
```
patrickelectric, line 82:
```
switch (true) {
    case serial_port_path.includes('ttyAMA'):
    case serial_port_path.includes('ttyS0'):
        return BoardType.Navigator
    case ...
    
}
```

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,191 @@
...
+    board_image() : string {
+      switch (this.board_type) {
+        case BoardType.Navigator:
+          /* eslint-disable */
```
patrickelectric, line 86: It would be nice to know what eslint-disable is disabling, please specify the linter feature that you want to disable and not everything
```
// eslint-disable-next-line @typescript...
```

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,191 @@
...
+          return require('@/assets/img/devicePathHelper/navigator.jpg')
+        break;
```
patrickelectric, line 88: There is no need of break after return

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -0,0 +1,191 @@
...
+      if (system_serial_ports === undefined || system_serial_ports.isEmpty()) {
+        return this.device as string
+      }
+      const port = system_serial_ports.find((a) => a.name === this.device)
+      return port?.by_path ?? this.device as string
+    },
+    board_type() : BoardType {
+      /* Detects board type between navigator and Rpi4 */
+      // TODO: make sure this works with pi3 as well
+      const serial_port_path = this.serial_port_path as string // why is it needed?
```
patrickelectric, line 71: @rafaellehmkuhl do you know what the cast is necessary ?

core/frontend/src/components/ping/ping360.vue
```diff
@@ -0,0 +1,97 @@
...
+  data() {
+    return {
+      expand: false,
+    }
+  },
+  computed: {
+  },
+  methods: {
+    is_ethernet() {
+      return this.device.ethernet_discovery_info !== ''
```
patrickelectric, line 80: `!this.device.ethernet_discovery_info.isEmpty()`

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,151 @@
...
+    PING1D_DISTANCE_SIMPLE,
+    PING1D_PROFILE,
+    PING1D_SET_PING_INTERVAL,
+    PingMessage,
+    PingParser,
+)
+from commonwealth.mavlink_comm.MavlinkComm import MavlinkMessenger
+from loguru import logger
+
+## The minimum interval time for distance updates to the autopilot
```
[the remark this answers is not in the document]
patrickelectric: The variable suffix already says that is in seconds, we should avoid redundant information between code and comments.
code > comments ≥ whitespace
https://patrickelectric.work/blog/2020/how-to-rock-first-tips/

core/services/ping/ping1d_mavlink.py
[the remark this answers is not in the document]
patrickelectric: I'm inclined to agree with @ES-Alexander here, why `EAGAIN` is special over others that needs his special `continue` ?

core/services/ping/ping1d_mavlink.py
```diff
@@ -0,0 +1,151 @@
...
+                if tnow > last_distance_measurement_time + PING_INTERVAL_S * 10:
+                    logger.info("attempting reconnection...")
+                    self.ping1d_io.connect(pingserver)
+                    await asyncio.sleep(0.1)
+
+            # read data in from ping device
+            try:
+                data, _ = self.ping1d_io.recvfrom(4096)
+            except socket.error as exception:
+                # check if it's waiting for data
```
patrickelectric, line 129: We should log the exception here.

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,63 @@
...
+
+        server.sendto(message, ("255.255.255.255", 30303))
+        await asyncio.sleep(1)
+        try:
+            data, _ = server.recvfrom(1048)
+            decoded_message = data.decode("utf8")
+            logger.info(f"Found Ping360: {decoded_message}")
+            raw_ip = decoded_message.split("IP Address:- ")[1].strip()
+            found.append(remove_zeros(raw_ip))
+        except socket.timeout as e:
```
patrickelectric, line 61: `exception`

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+async def find_ping360_ethernet() -> List[str]:
+    """
+    Return a list of Ping360 IPs found in the connected ethernet interfaces
+    """
+    found = []
+    for ip in list_ips():
+        server = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+        server.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
```
patrickelectric, line 43: `optlen=1`

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,68 @@
...
+        if ping not in self.drivers:
+            self.drivers[ping] = Ping360Driver(ping, None)
+
+    async def find_next_port(self, base_port: int, direction: int) -> int:
+        """
+        Find the next unused udp port, starts on base_port and increments/decrements
+        accordingly to direction
+        """
+        port = base_port
+        while udp_port_is_in_use(port):
```
patrickelectric, line 36: We should limit the range that the port can be.
At least between 9090 and 9099, otherwise we could start having numbers totally outside the range if 3rd party programs are using it, without us having control to find the sensors with ping-viewer.
We should also start thinking about how ping-viewer is going to discover this devices, how that works currently ?

patrickelectric (changes requested):

core/frontend/src/types/ping.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface PingDevice {
+  ping_type: PingType,
+  device_id: number,
+  device_model: number,
+  device_revision: number,
+  firmware_version_major: number,
+  firmware_version_minor: number,
+  firmware_version_patch: number,
+  port: string,
+  driver_status: Dictionary<number|boolean|string>,
```
rafaellehmkuhl, line 17: As changed in the backend, it would be good to also have a DriverStatus interface definition in the frontend.

core/frontend/src/components/common/DevicePathHelper.vue
rafaellehmkuhl: Apparently because in line 66 it can return `undefined` if port is not defined.

rafaellehmkuhl (changes requested):
Nice work, really.

core/services/ping/ping1d_driver.py
```diff
@@ -1,7 +1,58 @@
...
 from pingdriver import PingDriver
 from pingutils import PingDeviceDescriptor
+from settings import Ping1dSettingsSpecV1, SettingsV1
+
+SERVICE_NAME = "ping"
 
 
 class Ping1DDriver(PingDriver):
-    def __init__(self, ping: PingDeviceDescriptor, port: int) -> None:
+    def __init__(self, ping: PingDeviceDescriptor, port: Optional[int]) -> None:
```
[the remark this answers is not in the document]
Williangalvani: This was just wrong. I think I need a vacation.
It is gone now

core/services/ping/ping1d_mavlink.py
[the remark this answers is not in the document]
Williangalvani: after thinking about this for a while, I realized the corrrect approach is to use `select` to check if there's anything to receive, instead of doing it blindly

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,63 @@
...
+def list_ips() -> List[str]:
+    """
+    Returns a list of IPs found on ethernet interfaces that are currently up
+    """
+    stats = psutil.net_if_stats()
+
+    available_networks = []
+    ips: List[str] = []
+    for interface in stats.keys():
+        if interface in stats and getattr(stats[interface], "isup"):
```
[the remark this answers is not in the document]
Williangalvani: good catch, I think this is leftover code from when I tried a different logic

core/services/ping/ping360_ethernet_prober.py
Williangalvani: that doesn't work. it complains that the functions doesnt take keyword arguments

core/services/ping/pingmanager.py
Williangalvani: I don't see the point in limiting it. Discovery needs to be improved in PingViewer.
As is it will work fine for one Ping1D and one Ping360.
We could either use the Ping360 discovery protocol, or ask Beacon to broadcast the devices using mDNS

core/frontend/src/components/ping/ping1d.vue
```diff
@@ -0,0 +1,115 @@
+<template>
+  <v-card style="min-width: 280px;">
+    <v-card-title class="justify-center pb-1">
+      <v-icon
+        x-large
+        class="ping1d"
+        color="primary"
```
patrickelectric, line 7: That will change if someone uses a custom color scheme, as distributors, we should enforce the same blue that we use.

core/services/ping/pingmanager.py
patrickelectric: Not limiting the range will result in a silent loop block where there is no condition to exit besides the OS saying that a port is available, if no port is available, this loop will never exit. But if you don't have strong feelings that such thing will never happen I'm ok with it.

core/services/ping/pingmanager.py
```diff
@@ -1,38 +1,69 @@
...
         driver: PingDriver
         if ping.ping_type == PingType.PING1D:
             logging.info("Launching ping1d driver")
-            driver = Ping1DDriver(ping, self.ping1d_current_port)
+            port = await self.find_next_port(self.ping1d_base_port, step=-1)
+            driver = Ping1DDriver(ping, port)
         elif ping.ping_type == PingType.PING360:
             logging.info("Launching ping360 driver")
-            driver = Ping360Driver(ping, self.ping360_current_port)
+            port = await self.find_next_port(self.ping360_base_port, step=+1)
```
patrickelectric, line 51: How the +1 and -1 logic will work with future sensors ?
As pointed in the idea over: https://github.com/bluerobotics/BlueOS-docker/pull/1111/files#r941320562

patrickelectric (commented):
Will check with the hardware to do the final review of the functionality

core/services/ping/pingmanager.py
Williangalvani: the only reason we needed `step` here was because they 9092 and 9090 are way to close, and the sensors ports would collide after we have two ping1Ds.

The step could be removed now that we have a udp_port_is_in_use(). but I would rather have a somewhat logical order, while we don't have proper discovery set up.

patrickelectric (changes requested):
- When there are no sensors, the screen appears black without content, we should have at least a No Ping Sensors or something, like we have for the camera manager. We may want to remove the entry point on the main menu if there are not sensors, but I'm not sure if that is user friendly, but since ping is not common in all ROVs, it may makes sense.
image
- After the sensor is connected, it took almost 90 seconds for the sensor start being used.
image
- Hot reload appears to work, but no more MAVLink messages are being sent 
image
- It failed to detect ping360 once connected via his native USB, the same cable works with ping-viewer
image
Connecting it back to the raspberry result in the following output
image
image

Williangalvani:
@patrickelectric 
We now have a message for "no devices detected"
The ping360 issue is on hardware, hot-plugging it (anywhere) is not possible as the usb implementation doesn't work:
Screenshot from 2022-08-25 13-36-29
in the screenshot it was detected after I power-cycled the ping360.

I cannot replicate the 90s to identify here, it takes ~15 seconds at most.

I made the driver more resilient to disconnections. It now tries to re-launch Bridges if something goes wrong, until PortWatcher tells it to quit.

core/services/ping/ping360_ethernet_prober.py
```diff
@@ -0,0 +1,69 @@
...
+import socket
+from typing import List, Set
+
+import psutil
+from loguru import logger
+
+
+def list_ips() -> Set[str]:
+    """
+    Returns a list of IPs found on ethernet interfaces that are currently up.
```
ES-Alexander, line 11:
```suggestion
def get_available_ips() -> Set[str]:
    """
    Returns a set of IPs found on ethernet interfaces that are currently up.
```

ES-Alexander (approved):
I'm short on brainpower at the moment so this wasn't a super in depth review, but great to see the logging has been added, and in general still looking good! :-)

One minor suggestion on naming correctness.

patrickelectric (approved):
image

PR #1112, opened 2022-08-08
core: bridges: Fix check for error message

patrickelectric:
I dream on a day where Python std will have type hints
I dream on a day without daily errors footprints
I dream on a day where Python will have a good documentation with types
I dream on a day with peace in my mind

Before:
image

After:
image

### DIFF
--- a/core/libs/bridges/bridges/bridges.py
+++ b/core/libs/bridges/bridges/bridges.py
@@ -29,8 +29,9 @@ class Bridge:
         self.process = Popen(shlex.split(command_line), stdout=DEVNULL, stderr=DEVNULL)
         time.sleep(1.0)
         if self.process.poll() is not None:
-            error = self.process.communicate()[1]
-            raise RuntimeError(f"Failed to initialize bridge: {self.process.returncode} - {error.decode('utf-8')}.")
+            _stdout, strerr = self.process.communicate()
+            error = strerr.decode("utf-8") if strerr else "Empty error"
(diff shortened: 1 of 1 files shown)


patrickelectric:
Report from: https://discuss.bluerobotics.com/t/ekf3-lane-switch-1-warning-when-armed/12693/6?u=patrickelectric

Williangalvani:
what is the output of bridges looking like right now?
I worry that this change could reintroduce the bug where things stop working because stdout gets full...
should we flush it once in a while?

patrickelectric:
> what is the output of bridges looking like right now? I worry that this change could reintroduce the bug where things stop working because stdout gets full... should we flush it once in a while?

Check the commit body of: https://github.com/bluerobotics/BlueOS-docker/pull/1112/commits/182c9678ea5d2ab9036f8bdf86893f78a32ecd10

Williangalvani:
Sounds good. still, we should make sure the time it takes now is range of "years", not "hours".

Williangalvani (approved):

patrickelectric:
> Sounds good. still, we should make sure the time it takes now is range of "years", not "hours".

There is no more outputs in bridges if you don't use verbose mode

Williangalvani:
> > Sounds good. still, we should make sure the time it takes now is range of "years", not "hours".
> 
> There is no more outputs in bridges if you don't use verbose mode

perfect =D

PR #1122, opened 2022-08-17
Update mavlink camera manager

joaoantoniocardoso:
docker image: `joaoantoniocardoso/blueos-core:update-mavlink-camera-manager`

- [x] Tested with pixhawk
- [x] Tested with SITL

Summary:
- RTSP is working again
- WebRTC is working again
- Improved QGC integration: multiple streams are working with QGC
- Logging to daily-rotating files (`/var/logs/mavlink-camera-manager/.`)

Issue #1052, opened 2022-06-15
Offer current IP for the user to be used on endpoint-related-UIs


rafaellehmkuhl:
This would be useful in both video-stream creation and autopilot endpoint creation.
This would save the user the hassle to go find it manually.

Williangalvani:
http://blueos.local:9111/v1.0/docs#/default/get_ip_ip_get

patrickelectric:
Fixed by #1124 #1125

PR #1125, opened 2022-08-25
core: frontend: make streamtype an enum and pre-fill stream urls

Williangalvani:
[Screencast from 25-08-2022 13:30:09.webm](https://user-images.githubusercontent.com/4013804/186720248-cc629741-692c-453a-946d-c90d6b082bca.webm)

depends on #1122
helps #1052

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -258,6 +258,9 @@ export default Vue.extend({
     vehicle_ip_address(): string {
       return beacon.nginx_ip_address
     },
+    sanitized_stream_name(): string {
+      return this.stream_name.replace(/[^a-z0-9]/gi, '_').toLowerCase()
+    },
   },
   methods: {
     validate_required_field(input: string | null): (true | string) {
@@ -318,13 +321,17 @@ export default Vue.extend({
         case StreamType.UDP:
           if (!this.stream_endpoints[index].includes('udp://')) {
             // Vue.set() forces the update of a nested property
-            Vue.set(this.stream_endpoints, index, `udp://${this.user_ip_address}:5600`)
+            Vue.set(this.stream_endpoints, index, `udp://${this.user_ip_address}:${5600 + index}`)
           }
           break
         case StreamType.RTSP:
           if (!this.stream_endpoints[index].includes('rtsp://')) {
             // Vue.set() forces the update of a nested property
-            Vue.set(this.stream_endpoints, index, `rtsp://${this.vehicle_ip_address}:8554/video0`)
+            Vue.set(
+              this.stream_endpoints,
+              index,
+              `rtsp://${this.vehicle_ip_address}:8554/video_${this.sanitized_stream_name}`,
+            )
           }
           break
         default:
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -47,6 +47,13 @@
             :key="index"
             class="d-flex justify-space-between align-center"
           >
+            <v-select
+              :items="[StreamType.UDP, StreamType.RTSP]"
+              :value="endpoint.startsWith('rtsp://') ? StreamType.RTSP : StreamType.UDP"
+              class="mr-10"
+              style="width:20%;"
+              @change="set_default_address_for_stream(index, $event)"
+            />
             <v-text-field
               v-model="stream_endpoints[index]"
               label="Stream endpoint"
@@ -118,8 +125,9 @@
 import Vue, { PropType } from 'vue'
 
 import settings from '@/libs/settings'
+import beacon from '@/store/beacon'
 import {
-  CreatedStream, Device, Format, FrameInterval, Size, StreamPrototype, VideoCaptureType,
+  CreatedStream, Device, Format, FrameInterval, Size, StreamPrototype, StreamType, VideoCaptureType,
   VideoEncodeType,
 } from '@/types/video'
 import { VForm } from '@/types/vuetify'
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
```diff
@@ -281,14 +300,35 @@ export default Vue.extend({
...
+        case StreamType.UDP:
+          if (!this.stream_endpoints[index].includes('udp://')) {
+            // Vue.set() forces the update of a nested property
+            Vue.set(this.stream_endpoints, index, `udp://${this.user_ip_address}:5600`)
+          }
+          break
+        case StreamType.RTSP:
+          if (!this.stream_endpoints[index].includes('rtsp://')) {
+            // Vue.set() forces the update of a nested property
+            Vue.set(this.stream_endpoints, index, 'rtsp://0.0.0.0:8554/video0')
```
joaoantoniocardoso, line 325: Instead of `0.0.0.0`, since [mavlink-camera-manager t3.7.0](https://github.com/bluerobotics/mavlink-camera-manager/releases/tag/t3.7.0) it should be the adapter's IP (e.g. `192.168.2.2` by default for the static ethernet IP)

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
```diff
@@ -47,11 +47,19 @@
...
+              :items="[StreamType.UDP, StreamType.RTSP]"
+              :value="endpoint.startsWith('rtsp://') ? StreamType.RTSP : StreamType.UDP"
+              class="mr-10"
+              style="width:20%;"
+              @change="set_default_address_for_stream(index, $event)"
+            />
             <v-text-field
               v-model="stream_endpoints[index]"
               label="Stream endpoint"
               placeholder="e.g. udp://192.168.2.2:5600 or rtsp://0.0.0.0:8554/video0"
```
joaoantoniocardoso, line 60:
```suggestion
              placeholder="e.g. udp://192.168.2.1:5600 or rtsp://192.168.2.2:8554/video0"
```

just to be more consistent with the default use case.

joaoantoniocardoso (changes requested):

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
Williangalvani: we have access to the interface on the other side, too. I'll update.

I wish we could have the camera manager do it though... so we could use it on all interfaces. We can do that later

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
```diff
@@ -281,14 +303,35 @@ export default Vue.extend({
...
+        case StreamType.UDP:
+          if (!this.stream_endpoints[index].includes('udp://')) {
+            // Vue.set() forces the update of a nested property
+            Vue.set(this.stream_endpoints, index, `udp://${this.user_ip_address}:5600`)
+          }
+          break
+        case StreamType.RTSP:
+          if (!this.stream_endpoints[index].includes('rtsp://')) {
+            // Vue.set() forces the update of a nested property
+            Vue.set(this.stream_endpoints, index, `rtsp://${this.vehicle_ip_address}:8554/video0`)
```
joaoantoniocardoso, line 328: I wonder if `video0` could be `video{i}` where `i` automatically increases by the number of RTSP streams or perhaps anything unique because it will, in the best scenario, only work the first time.

Williangalvani: yep! index sounds nice, I'll add this to the udp ones too
I was thinking about checking for ports in use in udp, but this is a lot more straight forward and can be done in no new lines, nice!

Williangalvani: image

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
```diff
@@ -274,6 +274,12 @@ export default Vue.extend({
       }
       return true
     },
+    no_repetitions(input: string): (true | string) {
+      if (this.stream_endpoints.filter((x) => x === input).length > 1) {
```
patrickelectric, line 285: `!this.stream_endpoints.filter((x) => x === input).isEmpty()`

joaoantoniocardoso (approved):
Great job in the ~satanization~ sanitization!

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
Williangalvani: it's checking for more than one, not more than zero

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
```diff
@@ -57,7 +57,7 @@
             <v-text-field
               v-model="stream_endpoints[index]"
               label="Stream endpoint"
-              placeholder="e.g. udp://192.168.2.2:5600 or rtsp://0.0.0.0:8554/video0"
+              placeholder="e.g. udp://192.168.2.1:5600 or rtsp://192.168.2.2:8554/video0"
```
rafaellehmkuhl, line 60: As you're making this major cleaning (👁️), would you mind declaring this address in some separate file and importing from it where it's been used? That would make sure we're not pointing to different places around the code.

rafaellehmkuhl (changes requested):

core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
Williangalvani: major cleaning ¬¬

I can take the placeholder out as it is no longer necessary...

ES-Alexander (approved):
LGTM, but I haven't reviewed the logic :-)

patrickelectric (approved):

PR #1139, opened 2022-09-02
beacon: Create settings v3

### LINKED ISSUE #1138: hotspot: MDNS for hotspot does not work
Accessing directly with the hotspot IP (192.168.42.1) everything works fine, but the MDNS (blueos-hotspot.local) does not.
Apparently I forgot to add a new version of the beacon settings.

Reported by @ES-Alexander

rafaellehmkuhl:
Add interface "uap0" (used for hotspot).

Fix #1138

### DIFF
--- a/core/services/beacon/settings.py
+++ b/core/services/beacon/settings.py
@@ -153,3 +153,28 @@ class SettingsV2(SettingsV1):
             logger.error(f"unable to update SettingsV1 to SettingsV2: {e}")
 
         data["VERSION"] = SettingsV2.VERSION
+
+
+class SettingsV3(SettingsV2):
+    VERSION = 3
+
+    def __init__(self, *args: str, **kwargs: int) -> None:
+        super().__init__(*args, **kwargs)
+        self.VERSION = SettingsV3.VERSION
+
+    def migrate(self, data: Dict[str, Any]) -> None:
+        if data["VERSION"] == SettingsV3.VERSION:
+            return
+
+        if data["VERSION"] < SettingsV3.VERSION:
+            super().migrate(data)
+
+        try:
+            if not any(interface["name"] == "uap0" for interface in data["interfaces"]):
+                data["interfaces"].append(
+                    {"name": "uap0", "domain_names": ["blueos-hotspot"], "advertise": ["_http"], "ip": "ips[0]"}
+                )
+        except Exception as e:
+            logger.error(f"unable to update SettingsV2 to SettingsV3: {e}")
+
+        data["VERSION"] = SettingsV3.VERSION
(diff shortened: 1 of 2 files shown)


rafaellehmkuhl:
Current problem:

image

core/services/beacon/settings.py
```diff
@@ -152,3 +152,28 @@ def migrate(self, data: Dict[str, Any]) -> None:
             logger.error(f"unable to update SettingsV1 to SettingsV2: {e}")
 
         data["VERSION"] = SettingsV2.VERSION
+
+
+class SettingsV3(SettingsV2):
```
patrickelectric, line 157: This should use V1 since V2 does not add extra fields. 
```suggestion
class SettingsV3(SettingsV1):
```

core/services/beacon/settings.py
```diff
@@ -152,3 +152,28 @@ def migrate(self, data: Dict[str, Any]) -> None:
...
+    def __init__(self, *args: str, **kwargs: int) -> None:
+        super().__init__(*args, **kwargs)
+        self.VERSION = SettingsV3.VERSION
+
+    def migrate(self, data: Dict[str, Any]) -> None:
+        if data["VERSION"] == SettingsV3.VERSION:
+            return
+
+        if data["VERSION"] < SettingsV3.VERSION:
+            super().migrate(data)
```
patrickelectric, line 169: Since the base class needs to be V1, we run the migration from V2.
```suggestion
            SettingsV2().migrate(data)
```

patrickelectric (changes requested):

core/services/beacon/settings.py
rafaellehmkuhl: Wait. Won't this jump V2 migration?

patrickelectric: Check my next comment.

rafaellehmkuhl: Got it. This is kinda hacky thought. In the end it is a bug that we are working around. Did you find out why this happens, or a fix in the library level? Writing the settings is a lot of work already. This problem will surely happen again.

patrickelectric: I have not gone deeply on the issue, I believe that the logic behind `extra_attributes` or `__get_fields_*` is failing to do the proper job, from deriving of a class without extra attributes deriving from a class that has no extra attributes while having a super.super class that contains `JsonObject`.
Since we are aiming to release a stable version of BlueOS and hotspot is a key feature, I would recommend the workaround of not deriving from classes that does not contains extra fields for now.
I'll investigate it a bit more.

rafaellehmkuhl:
Closing in favor of #1145

PR #1143, opened 2022-09-09
Add ruff

patrickelectric:


rafaellehmkuhl:
To better understand, we are keeping both `ruff` and `pylint`?

patrickelectric:
> To better understand, we are keeping both `ruff` and `pylint`?

Yes, the ideia is for now to fail fast with ruffs for things that are still connected with pylint, like the ignore comments around the code.
Little by little we could move away from pylint to ruff.

rafaellehmkuhl (approved):

rafaellehmkuhl:
Just [remove the import](https://github.com/bluerobotics/BlueOS-docker/runs/8273614682?check_suite_focus=true#step:5:22) and we are good to go.

joaoantoniocardoso (approved):

PR #1151, opened 2022-09-16
Create ExtensionView, show extensions in Iframe

Williangalvani:
iframe extensons

I don't know why it seems to be refreshing the page, though. thoughts?


rafaellehmkuhl (approved):
Looks good!
Couldn't find any info about the page reloading problem. Suspect It's something with the Vue Router configuration, but could not confirm.
Anyway, I think it's reasonable to merge this and open an issue for the reload problem, since it's a minor.

PR #1174, opened 2022-10-13
core: commonwealth: settings: Force settings files to be lowercase

patrickelectric:


Williangalvani (approved):

ES-Alexander (approved):
These modifications LGTM. I don't know if there are other files that should also be modified, but if not then this seems fine :-)

PR #1189, opened 2022-10-19
Allow displaying logs of running extensions 

Williangalvani:
https://user-images.githubusercontent.com/4013804/196813004-b88c0c5d-90dc-4df1-aba9-142ef481a6f0.mov

There's room for improvement, but I think this is a nice start.

I want to make it stream the data live, and interpret colors properly, in other PRs


ES-Alexander (approved):
Seems like a great idea! :-)
Keen to have this as is, and even more so once it's live and colourful in future :-)

PR #1191, opened 2022-10-20
core: helper: only cache successful probes

Williangalvani:
The goal here is to improve detection of Extensions, which sometimes get cached as an invalid service before they get up properly.

Not quite sure about this one. I Think the best thing to do would be to have a timeout. Say keep probing it for 1 minute, and then give up. but what if something takes longer than a minute to respond?
Also, what if something REALLY dislikes being probed?

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 import http
-from functools import cache
 from pathlib import Path
 from typing import Any, List, Optional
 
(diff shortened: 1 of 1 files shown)


core/services/helper/main.py
```diff
@@ -45,9 +44,12 @@ class ServiceInfo(BaseModel):
 class Helper:
     LOCALSERVER_CANDIDATES = ["0.0.0.0", "::"]
 
-    @staticmethod
-    @cache
-    def detect_service(port: int) -> ServiceInfo:
+    cache: Dict[int, ServiceInfo] = {}
+
+    def detect_service(self, port: int) -> ServiceInfo:
```
patrickelectric, line 49: The problem with this approach is that we are not going to hot-update information of the extension if an update happens or an extension was removed.

ES-Alexander: For tracking purposes, IIRC in our meeting patrick suggested using a `temporary_cache` here with a timeout of ~1 minute.

patrickelectric: I would recommend to also add a comment explaining the reason behind it, to hot reload services metadata without rebooting.

PR #1199, opened 2022-10-21
kraken: Add cpu and memory status

patrickelectric:
I'm without time to create the UI, but I did a quick look of how to extract the data and get more familiar with the docker API. 
The memory report only works with cgroups enabled in cmdline. It'll be fixed in our future BlueOS images, and users could run the if check that'll update the line.

### DIFF
--- a/core/frontend/src/views/ExtensionManagerView.vue
+++ b/core/frontend/src/views/ExtensionManagerView.vue
@@ -74,22 +74,60 @@
     <v-row>
       <v-col
         v-if="tab === 1"
-        class="pa-5"
+        class="pa-6"
       >
         <v-row
-          v-for="extension in installed_extensions"
-          :key="extension.name"
           dense
         >
           <v-col
-            class="pa-2"
+            v-for="extension in installed_extensions"
+            :key="extension.name"
+            class="pa-2 col-6"
           >
             <v-card>
-              <v-card-title>
-                {{ extension.name.replace('/', '') }}: <span style="color: grey;"> {{ extension.tag }}</span>
+              <v-card-title class="pb-0 ">
+                {{ extension.name.split('/')[1] }} <span class="ml-3" style="color: grey;"> {{ extension.tag }}</span>
               </v-card-title>
-              <v-card-text>
-                {{ getStatus(extension) }}
+              <span class="mt-0 mb-4 ml-4 text--disabled">{{ extension.name }}</span>
+              <v-card-text width="50%">
+                <v-simple-table>
+                  <template #default>
+                    <tbody>
+                      <tr>
+                        <td>Status</td>
+                        <td>{{ getStatus(extension) }}</td>
+                      </tr>
+                      <tr>
+                        <td>Memory usage</td>
+                        <td>
+                          <v-progress-linear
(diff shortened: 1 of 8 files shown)


Williangalvani (changes requested):
I think this should either not take an argument and return a list with stats for all containers, or be embedded into the /list endpoint.

My reasoning is that we already have a LOT of requests going on in BlueOS and I don't think having one more for each individual extension is a good idea...

That said, I don't mind doing these changes myself.

patrickelectric:
Be free to take it from here, I'll be working mainly on cockpit for some time.

core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -61,5 +83,20 @@ def main():
         time.sleep(10)
         sys.exit(1)
 
+    
+def update_cgroups() -> None:
```
patrickelectric, line 87: This should run inside `update_startup()`, previous `main()` where BlueOS `tag, commit_number, commit_hash` are available for necessary checks on the change functions.
You could also move the version check to be inside the new `main()` function and outside the `update_startup()`

core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -28,6 +28,28 @@ DELTA_JSON = {
 
 import collections
 
+import subprocess
+
+def run_command_in_host(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
```
patrickelectric, line 33: Since this is already used in `commander`, we may want to move it to `commonwealth`, avoid dupplicating the code.

core/frontend/src/views/ExtensionManagerView.vue
```diff
@@ -182,20 +220,55 @@ export default Vue.extend({
...
+          notifier.pushBackError('EXTENSIONS_METRICS_FETCH_FAIL', error)
+        })
+    },
+    getCpuUsage(extension: InstalledExtensionData): number {
+      const name = this.getContainerName(extension)?.replace('/', '')
+      if (!name) {
+        return NaN
+      }
+
+      return this.metrics[name]?.cpu
```
patrickelectric, line 263:
```suggestion
      return this.metrics[name]?.cpu ?? NaN
```

patrickelectric (commented):
I can't request changes in my own PR :lol:

core/frontend/src/views/ExtensionManagerView.vue
Williangalvani: doesnt seem to work 
Screenshot 2022-11-18 at 16 08 46

core/tools/blueos_startup_update/blueos_startup_update
Williangalvani: done

core/tools/blueos_startup_update/blueos_startup_update
Williangalvani: moved into main

Williangalvani (approved):
Very nice patch, love it 👀

patrickelectric (commented):
LGTM

PR #1204, opened 2022-10-28
core: frontend: components: MAVLinkInspector: Add filter for mavlink messages

patrickelectric:

### DIFF
--- a/core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
+++ b/core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
@@ -13,12 +13,17 @@
             class="mx-auto height-limited"
             max-height="700px"
           >
+            <v-text-field
+              v-model="message_filter"
+              class="ma-2"
+              label="Message"
+            />
             <v-list shaped>
               <v-list-item-group
                 v-model="selected_message_types"
                 multiple
               >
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
```diff
@@ -158,9 +163,17 @@ export default Vue.extend({
       messages_in_view: [] as any[],
       selected_message_types: [],
       detailed_message: null as (null | any),
+      message_filter: '',
     }
   },
   computed: {
+    filtered_messages(): string[] {
+      try {
+        return this.message_types.filter((name: string) => name.match(this.message_filter.trim()))
```
Williangalvani, line 172:
```suggestion
        return this.message_types.filter((name: string) => name.test(this.message_filter.trim()))
```

Maybe test() makes more sense here?

Williangalvani (changes requested):

core/frontend/src/components/mavlink-inspector/MavlinkInspector.vue
patrickelectric: `test` does not exist for `string`, only for `RegExp`. `match` was avoided to create a more simpler approach over creating a regex component.

Williangalvani (approved):

PR #1213, opened 2022-11-10
core: kraken: try launching extensions even if unable to pull them

Williangalvani:


ES-Alexander:
Does this mean the default behaviour for extensions is to automatically try to update when starting? I imagine some people could have security concerns with that if it's not possible to opt out of that behaviour without just not connecting BlueOS to the internet.

Regardless, am I correct in thinking this PR is effectively a bug-fix that enables running locally installed extensions without needing to be connected to the internet when the vehicle turns on?

patrickelectric (approved):

PR #1217, opened 2022-11-13
frontend: SystemInformationView: Fix kernel icon

patrickelectric:
Fix kernel icon after mdi update


ES-Alexander (approved):
LGTM - checked that the old one no longer exists in the icon list, and that the new one does :-)

Williangalvani (approved):

PR #1226, opened 2022-11-18
core: ArdupilotManager: support streaming dataflash logs with mavlink-router

Williangalvani:
needs testing.
needs LOG_BACKEND to be set to enable MAVLINK backend

it seemed to create some 0 bytes files in my tests, but the correct ones are generated as well. we could do a routing to erase these empty files..


Williangalvani:
I'd like some thoughts on this. This is very important functionality.
The only issue are the empty files randomly generated. I'm inclined to get this in for the functionality, maybe with an additional oneliner such as `find . -name "*.tif" -type f -size -1k -delete`

patrickelectric:
I got no files besides an empty .bin a single time, I tried to arm/disarm and enabling 

image

LOG_DISARMED and LOG_FILE_DSRMROT,is that the expected behavior ?

Williangalvani:
LOG_BACKEND_TYPE must be set to MAVLINK

just re-tested here and it seems to work

patrickelectric (approved):

PR #1237, opened 2022-11-25
core: components: parameter-editor: Fix display of bitmask values

### LINKED ISSUE #1234: bug: Some parameters appears as numbers and not as flags names on the UI
### Bug description

image

### Steps to reproduce

Check LOG_BACKEND_TYPE on the parameter editor

### Primary pain point(s)

It's not friendly which options are on

### Prerequisites

patrickelectric:
Fix #1234 

image

image

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterEditor.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterEditor.vue
@@ -5,9 +5,11 @@
   >
     <v-data-table
       v-if="finished_loading"
-      :headers="[{ text: 'Name', value: 'name' },
-                 { text: 'Description', value: 'description' },
-                 { text: 'Value', value: 'value' }]"
+      :headers="[
+        { text: 'Name', value: 'name' },
+        { text: 'Description', value: 'description' },
+        { text: 'Value', value: 'value', width: '20px' },
+      ]"
       :items="params"
       item-key="name"
       class="elevation-1"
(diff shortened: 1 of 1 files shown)


Williangalvani (changes requested):
We should also handle cases where bits not available in the metadata are set.
we could show each of them as an unkown entry?
"OPTION, OPTION2 , Bit 4, Bit 8". Idk.

patrickelectric:
> We should also handle cases where bits not available in the metadata are set. we could show each of them as an unkown entry? "OPTION, OPTION2 , Bit 4, Bit 8". Idk.

Done.

core/frontend/src/components/parameter-editor/ParameterEditor.vue
```diff
@@ -309,10 +311,29 @@ export default Vue.extend({
...
+      if (param.bitmask) {
+        const bitmask_result = []
+        // We check up to 64 bits to make sure that we are going to cover all possible bits
+        // Including the ones not listed in the bitmask
+        for (let v = 0; v < 64; v += 1) {
+          // eslint-disable-next-line no-bitwise
+          const bitmask_value = 1 << v
+          // eslint-disable-next-line no-bitwise
+          if (param.value & bitmask_value) {
+            bitmask_result.push(param.bitmask?.[v] ?? `Bit ${v + 1}`)
```
ES-Alexander, line 331: This could be ok as is, but if someone doesn't know what a bitmask is (just that the parameter has multiple options) it may be worth being more explicit that the value is unknown, not just what it is. Maybe something like
- `UNKNOWN (bit #)`
- `UNKNOWN(#)`
- `Unknown:#`
- `? (bit #)`

Williangalvani (approved):

PR #1246, opened 2022-12-01
core: components: parameter-editor: Fix check for no input

patrickelectric:

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterEditor.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterEditor.vue
@@ -194,7 +194,12 @@ export default Vue.extend({
     },
   },
   methods: {
-    isInRange(input: number): boolean | string {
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/parameter-editor/ParameterEditor.vue
```diff
@@ -194,7 +194,12 @@ export default Vue.extend({
     },
   },
   methods: {
-    isInRange(input: number): boolean | string {
+    isInRange(input: number | string): boolean | string {
+      // The input value is an empty string when the field is empty
+      if (typeof input === 'string' && input?.trim().length === 0) {
+        return 'No input'
```
Williangalvani, line 200: This is a silly complaint, but I would rather have a message saying something like.
"This should be a number between min and max"

Williangalvani (approved):

PR #1251, opened 2022-12-02
core: components: parameter-editor: Improve range information on parameter editor

### LINKED ISSUE #1212: core: frontend: parameters: improve edit UI/UX
### Current behaviour

- `Max` is shown before `Min`, which is unintuitive
- Horizontal spacing between Max and Min is small
- Vertical spacing between restrictions row and description row is small
- Increment is not considered

- [ ] Restrictions row should show something like
   **Min:** -0.5 &nbsp;&nbsp;&nbsp; **Max:** 3.5 &nbsp;&nbsp;&nbsp; **Increment:** 0.01
   - Not sure how to do the spacing, but I expect an increment is only ever present if there's also a range, so perhaps something like
      ```html
      <v-row v-if="edited_param.range">
        Min: {{ edited_param.range.low }}
        Max: {{ edited_param.range.high }}
        <span v-if="edited_param.increment">
          Increment: {{ edited_param.increment }}
        </span>
      </v-row>
      ```
- [ ] Vertical spacing should be increased between restrictions row and description
   - [ ] Possibly also put description first, so restrictions are shown directly above the interface that's changing the value
      - If description is first, I'm not sure if it's better to have `Reboot Required` above or below the description
- [ ] Ideally increment should be validated
   - Noting that [that may be a bit of effort](https://github.com/bluerobotics/BlueOS-docker/pull/806#issuecomment-1302122169), especially since values are floats, and some parameters (e.g. `AHRS_TRIM_*`) have ranges that mean increment can't just be checked directly with a modulo
   - It'd need to be something like `isclose((value - range_min) % increment, 0)`

### Prerequisites

patrickelectric:
Fix #1212

image

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterEditor.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterEditor.vue
@@ -56,15 +56,19 @@
           <v-container
             v-if="edited_param"
           >
-            <v-row v-if="edited_param.range">
-              Max: {{ edited_param.range.high }}
-              Min: {{ edited_param.range.low }}
-            </v-row>
             <v-row v-if="edited_param.rebootRequired">
               Reboot Required
             </v-row>
             <v-row>
               {{ edited_param.description }}
+            </v-row>
+            <v-row
+              v-if="edited_param.range"
+              class="pt-6"
+            >
+              Min: {{ edited_param.range.low }}
+              Max: {{ edited_param.range.high }}
+              Increment: {{ edited_param.increment ?? 0.01 }}
             </v-row>
             <v-row>
               <v-col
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/parameter-editor/ParameterEditor.vue
```diff
@@ -56,16 +56,20 @@
...
             <v-row>
               {{ edited_param.description }}
             </v-row>
+            <v-row
+              v-if="edited_param.range"
+              class="pt-6"
+            >
+              Min: {{ edited_param.range.low }}
+              Max: {{ edited_param.range.high }}
+              Increment: {{ edited_param.increment ?? 1 }}
```
Williangalvani, line 71: I don't think we should assume the increment is 1.
I'd rather just omit it if it is not set.

Imagine a situation where the range is smaller than 1. That would be awkward.

we could make also make the input step be (high-low)/100 in case increment isnt set

patrickelectric: That's already our default behaviour, changing to 0.01.

Williangalvani (approved):

PR #1254, opened 2022-12-02
parameter-editor: Allow saving parameters to file

patrickelectric:
File output example: [Submarine-4.2.0-DEV-20221202-150023.txt](https://github.com/bluerobotics/BlueOS-docker/files/10143320/Submarine-4.2.0-DEV-20221202-150023.txt)

image

PR #1255, opened 2022-12-04
Allow loading parameter from file

### LINKED ISSUE #1211: core: frontend: parameters: add file support
### Current behaviour

Parameters are currently only modifiable individually, and cannot be saved.

It would be useful to have options to save all current parameters (to a `.params` file), and load from file (with a subsequent vehicle reboot). This would enable the parameters page to fully replace parameter configuration and access via QGC.

- [ ] Probably best UI is buttons in a flex-box that takes part of the current search-bar space
   - [ ] That's also a good place to add a `refresh` button, to re-fetch parameters
- [ ] While tuning the aesthetics, it may be worth adding a bit extra padding above the "search" heading
   - it looks great at the moment when empty, but when a search has been made the "search" word moves up really close to the edge of the box, which looks a bit uncomfortable.

### Prerequisites

patrickelectric:
Needs to be merged after #1254
Fix #1211

### DIFF
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@mdi/font": "^6.9.96",
+    "@types/file-saver": "^2.0.5",
     "@types/lodash": "^4.14.175",
     "@types/marked": "^4.0.7",
     "@types/semver": "^7.3.9",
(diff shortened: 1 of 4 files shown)


Williangalvani (approved):
looks good, but I'd change the math.max thing before merging

PR #1256, opened 2022-12-05
Allow editing extensions and creating customs ones through the UI

Williangalvani:


core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,176 @@
...
+
+<script lang="ts">
+import Vue, { PropType } from 'vue'
+
+import Notifier from '@/libs/notifier'
+import { kraken_service } from '@/types/frontend_services'
+import { VForm } from '@/types/vuetify'
+import back_axios from '@/utils/api'
+
+import { InstalledExtensionData } from '../../types/kraken'
```
patrickelectric, line 79: use `@` and not relative reference

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,176 @@
...
+    extension() {
+      this.new_extension = this.extension
+    },
+  },
+  methods: {
+    closeDialog() {
+      this.new_extension = null
+    },
+    validate_identifier(input: string): (true | string) {
+      const re = /[A-Za-z]+[A-Za-z0-9]*\.[A-Za-z]+[A-Za-z0-9]*/
```
patrickelectric, line 119: `[A-Za-z]+[A-Za-z0-9]*` is weird, why not `[A-Za-z][A-Za-z0-9]*` ? The plus characters does not make sense to me.
Can you add an example if what this is validating ?

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,176 @@
...
+      if (re.test(input)) {
+        return true
+      }
+      return 'This field must container two words separated by a period. Numbers are allowed if not the first character'
+    },
+    validate_dockerhub(input: string): (true | string) {
+      const re = /[A-Za-z]+[A-Za-z0-9]*\/[A-Za-z]+[A-Za-z0-9]*/
```
patrickelectric, line 126: Can you add example for valid strings ?

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,176 @@
...
+      if (re.test(input)) {
+        return true
+      }
+      return 'This field must container two words separated by a forward slash'
+      + 'Numbers are allowed if not the first character. e.g example/docker1'
+    },
+    validate_name(input: string): (true | string) {
+      const re = /[A-Za-z]+[A-Za-z0-9 ]/
```
patrickelectric, line 134: The 3 three regex appears to be an abstraction over two string groups, if that is the case, we should split the regex in two groups and merge it for better readability of the proposal of each regex group.

patrickelectric (changes requested):

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
Williangalvani: yeah, in hindsight the + is meaningless. it is there so that the identifiers do not start with numbers

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
Williangalvani: I'll take the pluses out. this is based on dockerhub`s description:

A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

patrickelectric (changes requested):
image


It's failing to fetch for extensions

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,182 @@
...
+    },
+  },
+  methods: {
+    closeDialog() {
+      this.new_extension = null
+    },
+    validate_identifier(input: string): (true | string) {
+      // Identifiers should be two words separated by a period
+      // They can contain lower and uppercase characters, but cannot begin with a number
+      const re = /^[A-Za-z][A-Za-z0-9]*\.[A-Za-z]+[A-Za-z0-9]*/
```
patrickelectric, line 120: You can make it simpler by:
```
const name_validator = '[A-Za-z][A-Za-z0-9]+'
const regex = RegExp('^${name_validor}\.${name_validator}')
```

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,182 @@
...
+      if (re.test(input)) {
+        return true
+      }
+      return 'This field must container two words separated by a forward slash'
+      + 'Numbers are allowed if not the first character. e.g example/docker1'
+    },
+    validate_name(input: string): (true | string) {
+      // Lowercase, uppercase characters, and spaces
+      const re = /[A-Za-z0-9 ]/
+      if (re.test(input)) {
```
patrickelectric, line 141: A string that only contains spaces should not be valid, like: "      "

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,182 @@
...
+      const re = /[A-Za-z0-9_][A-Za-z0-9_.-]{0,127}/
+      if (re.test(input)) {
+        return true
+      }
+      return 'This field must container two words separated by a forward slash'
+      + 'Numbers are allowed if not the first character. e.g example/docker1'
+    },
+    validate_name(input: string): (true | string) {
+      // Lowercase, uppercase characters, and spaces
+      const re = /[A-Za-z0-9 ]/
```
patrickelectric, line 140: This regex only validate the first character, ' {}{}{}{}' is valid since the first character is a space.

patrickelectric (changes requested):

Williangalvani:
I updated the validation.

Your issue was that kraken was having issues with the updated settings format. I made it a big sturdier, but it will likely remove your currently installed extensions, now. I could go further, but beta features shouldn't need to deal with settings format changes anyway.

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,184 @@
...
+      }
+      return 'This field must container two words separated by a forward slash'
+      + 'Numbers are allowed if not the first character. e.g example/docker1'
+    },
+    validate_name(input: string): (true | string) {
+      if (input.trim().length === 0) {
+        return 'This field must not be empty'
+      }
+      if (input.length > 128) {
+        return 'This field must at most 128 characters long'
```
patrickelectric, line 144: It looks a bit weird, maybe: `This field must be no more than 128 characters long` or `This field should be not longer than 128 characters`

core/frontend/src/components/kraken/ExtensionCreationDialog.vue
```diff
@@ -0,0 +1,184 @@
...
+      }
+      return true
+    },
+    validate_permissions(input: string): (true | string) {
+      try {
+        JSON.parse(input)
+        return true
+      } catch {
+        return 'This field must be a valid Json'
```
patrickelectric, line 153: Maybe: `Invalid json format` or `This field is not a valid json format`

patrickelectric (changes requested):
Now it shows that there are no extensions installed, but I have zero tier on my rasp.

image

Williangalvani:
I changed the texts and made sure that if there are any running extensions, the dockers are stopped as well as it is uninstalled

PR #1290, opened 2022-12-20
deploy: pifile: bump pump to 1200M

Williangalvani:
potential fix for CI. (I hope)


Williangalvani:
> Was it tested ?

nope. I tried but I think we need to dedicated some time at getting rid of unnecessary hardcoded urls in the workflow.
even when I tried runing my branch, it still pumped for only 1gb

Williangalvani:
nevermind, I think I got it working now.
it is building

patrickelectric (approved):

Williangalvani:
working
weird though.

Running in chroot, ignoring request.
### PUMP 1200M; file system usage:
Filesystem           Size  Used Avail Use% Mounted on
/dev/mapper/loop0p2  2.8G  2.6G   45M  99% /tmp/tmp.RnDg3gRRSu

PR #1292, opened 2022-12-21
core: nmea_injector: add missing test dependency

Williangalvani:


Williangalvani:
this is only an issue testing locally. it is likely installed by another service in blueos

patrickelectric (approved):

PR #1293, opened 2022-12-21
WIP: add hdt support to NMEA injector

Williangalvani:
This does not work. it will require memory in order to send full messages. MAVLink doesn't support GPS messages with only YAW, and there doesn't seem to be any specific for yaw input.


Williangalvani:
Closing as we can now direct NMEA straight to the autopilot.

PR #1294, opened 2022-12-22
git: add node_modules to gitignore

Williangalvani:
This makes vscode happy


Williangalvani:
on the other hand, that folder should not exist at all...

PR #1300, opened 2023-01-05
core: frontend: Add BlueOS logo in main menu

patrickelectric:
image
image


ES-Alexander (commented):
This is nice, but I feel like it may as well wait for the actual logo, since that's nearly ready anyway.

PR #1301, opened 2023-01-05
core: frontend: Change frontend name to BlueOS

patrickelectric:
image
image
image


ES-Alexander (commented):
Would it be hard to do @rjehangir's suggestion of "[Page Title] - BlueOS"? If yes then this is still better than the current state, but if that's not too difficult then it would be preferred.

patrickelectric:
@ES-Alexander done

ES-Alexander (approved):

PR #1305, opened 2023-01-05
core: frontend: Remove tour steps if the network is already connected

patrickelectric:


ES-Alexander (approved):
This tour step may need to change more once it's possible to pass internet via the wired connection to the topside, but for now this filter seems ok.

If it's already possible to do a more general filter for "connected to internet" that may be better for future proofing.

PR #1317, opened 2023-01-11
core: frontend: Change pirate icon style to top bar

patrickelectric:
image

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -19,14 +19,6 @@
         <v-spacer />
         <span class="d-flex flex-column align-center">
           <backend-status-checker @statusChange="changeBackendStatus" />
-          <span
-            v-if="settings.is_pirate_mode"
-            class="black--text font-weight-bold"
-          >
-            <v-icon color="black">mdi-skull-crossbones</v-icon>
-            Pirate Mode
-            <v-icon color="black">mdi-skull-crossbones</v-icon>
-          </span>
         </span>
         <v-spacer />
         <beacon-tray-menu />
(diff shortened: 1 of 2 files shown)


ES-Alexander (approved):
I like this, although it may also be worth in general making this the place to turn pirate mode on/off (e.g. instead of making the icon disappear when pirate mode is off, just make it grey/crossed out with a diagonal "cancel" line, and then you can click the icon to swap modes).

patrickelectric:
> I like this, although it may also be worth in general making this the place to turn pirate mode on/off (e.g. instead of making the icon disappear when pirate mode is off, just make it grey/crossed out with a diagonal "cancel" line, and then you can click the icon to swap modes).

That's not easy if the icon does not exist. 
I would vote to merge at it's for now and work on the switch logic later.

ES-Alexander:
> That's not easy if the icon does not exist.

Not really sure what you mean by this - I was suggesting we have an icon that is always displayed, but changes depending on whether or not pirate mode is on, and can be clicked to swap in and out of pirate mode.

Anyway, we can merge for now if that's not easy to do.

patrickelectric:
> Not really sure what you mean by this - I was suggesting we have an icon that is always displayed, but changes depending on whether or not pirate mode is on, and can be clicked to swap in and out of pirate mode.

My point is how exactly do it, we are going to change the icon ? We are going to change the color ? What is the desired behavior of the UI

ES-Alexander:
I think changing icons is likely the most intuitive. Maybe swap between `pirate` (or `skull-crossbones`) and `emoticon-happy-outline` (or `emoticon-cool-outline`), with a tooltip that says "Click to {en/dis}able pirate mode."

joaoantoniocardoso:
@patrickelectric The hover/click actions are actually triggering the health icon

PR #1319, opened 2023-01-12
core: frontend: NetworkTestView: Improve text capitalization

patrickelectric:

### DIFF
--- a/core/frontend/src/views/NetworkTestView.vue
+++ b/core/frontend/src/views/NetworkTestView.vue
@@ -28,7 +28,7 @@
             <v-list-item>
               <v-list-item-icon>
                 <v-icon
-                  v-tooltip="'Download speed'"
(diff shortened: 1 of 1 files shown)


ES-Alexander (approved):
Would be good if we can also add explanations to the download and upload.

Not sure if we should change it to "Download speed (from vehicle to computer)" or add a "from vehicle to computer" tooltip to the Download speed thing (and "from computer to vehicle" for Upload speed).

patrickelectric:
> Would be good if we can also add explanations to the download and upload.
> 
> Not sure if we should change it to "Download speed (from vehicle to computer)" or add a "from vehicle to computer" tooltip to the Download speed thing (and "from computer to vehicle" for Upload speed).

I put just "from vehicle" and "to vehicle", computer may sound companion computer and local computer appears weird, "from vehicle" and "to vehicle" sounds like that is local based.

PR #1323, opened 2023-01-12
Improve parameter editor based on SMS feedback

patrickelectric:
Includes: https://github.com/bluerobotics/BlueOS-docker/pull/1322

https://user-images.githubusercontent.com/1215497/212071603-f6fbf5f7-87fa-4689-89a5-84a0e312623e.mp4

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterEditor.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterEditor.vue
@@ -59,6 +59,7 @@
       </template>
       <template #footer.prepend>
         <v-btn
+          v-tooltip="'Save all parameters to file'"
           :disabled="!finished_loading"
           color="primary"
           class="mr-5"
@@ -70,6 +71,7 @@
           <div>Save</div>
         </v-btn>
         <v-btn
+          v-tooltip="'Load parameters from file'"
           :disabled="!finished_loading"
           class="mr-5"
           color="primary"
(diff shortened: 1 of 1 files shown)


Williangalvani (approved):
lgtm.

Any thoughts on making the "Reboot required" text red?

patrickelectric:
> lgtm.
> 
> Any thoughts on making the "Reboot required" text red?

Red looks more as an error than a warning.

ES-Alexander (approved):
Looks great!

Could we also add tooltips to the Save and Load buttons for parameters files, just specifying "Save all parameters to a file" and "Load new parameters from a file"? :-)

I would prefer if the "Reboot Vehicle" button was in the header so that it follows you around other pages if you leave the parameters one, but not sure how hard that is to do.

patrickelectric:
> Could we also add tooltips to the Save and Load buttons for parameters files, just specifying "Save all parameters to a file" and "Load new parameters from a file"? :-)

Isn't Save and Load clear enough on the parameter page ?

ES-Alexander:
> Isn't Save and Load clear enough on the parameter page ?

At least for saving, it may not be clear whether they're saving all the current vehicle parameters, or just the ones currently shown in the table/display. Perhaps the tooltips should explicitly mention `.params` files, instead of just saying "a file".

patrickelectric:
> > Isn't Save and Load clear enough on the parameter page ?
> 
> At least for saving, it may not be clear whether they're saving all the current vehicle parameters, or just the ones currently shown in the table/display. Perhaps the tooltips should explicitly mention `.params` files, instead of just saying "a file".

Done.

patrickelectric:
We save it as "txt" since there is no rule for parameters files extensions, there a couple around.

PR #1327, opened 2023-01-12
core: frontend: VideoStreamCreationDialog: Add default name for stream

patrickelectric:
image

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -17,7 +17,7 @@
           <v-text-field
             v-model="stream_name"
             :counter="100"
-            label="Stream name"
+            label="Stream nickname"
             :rules="[validate_required_field]"
           />
 
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -158,7 +158,7 @@ export default Vue.extend({
       required: false,
       default() {
         return {
-          name: '',
+          name: `Stream ${this.device.source}`,
           encode: null,
           dimensions: null,
           interval: null,


joaoantoniocardoso:
Oh, since you are messing with this dialog, the endpoint helper only works when we press the "+" button.

patrickelectric:
> Oh, since you are messing with this dialog, the endpoint helper only works when we press the "+" button.

What ? Can you open an issue ? We can address that in another PR

ES-Alexander (changes requested):
I feel like including `/dev/video0` makes it seem potentially not arbitrary. If it's possible to have a placeholder of "My Video Stream" or something that seems like it would be clearer.

patrickelectric:
> I feel like including `/dev/video0` makes it seem potentially not arbitrary. If it's possible to have a placeholder of "My Video Stream" or something that seems like it would be clearer.

That will generate conflicts since the name can't be equal between streams, having the device name is safer.
We also don't have access to the number of video streams already used on the edit component to increase a index or something closer to it.
We can use 'Stream X', but that's also not helpful as name compared to device name.

ES-Alexander:
Ok, fair enough.
In that case can we change the field name to "Stream nickname"? :-)

patrickelectric:
> Ok, fair enough. In that case can we change the field name to "Stream nickname"? :-)

Is there a reason to use "nickname" over "name" ? it can add unnecessary internal changes on code and when integrating others systems, since name is what we use internally and in the API.

ES-Alexander:
> Is there a reason to use "nickname" over "name" ?

"Nickname" comes across as "this is what _you_ want to call it - only for human display/convenience". "Name" _could_ be something that the computer cares about, and having the device source included in the default name makes that seem more possible/likely (because that information isn't shown elsewhere in the configuration).

It's definitely possible to reason your way to the fact that if you're clicking on the "add stream" option for a given camera then the computer must already know the information it needs to about where that camera is and how it can connect to it, but a user shouldn't need to do that reasoning - it should just be unambiguous to start with :-)

patrickelectric:
> > Is there a reason to use "nickname" over "name" ?
> 
> "Nickname" comes across as "this is what _you_ want to call it - only for human display/convenience". "Name" _could_ be something that the computer cares about, and having the device source included in the default name makes that seem more possible/likely (because that information isn't shown elsewhere in the configuration).
> 
> It's definitely possible to reason your way to the fact that if you're clicking on the "add stream" option for a given camera then the computer must already know the information it needs to about where that camera is and how it can connect to it, but a user shouldn't need to do that reasoning - it should just be unambiguous to start with :-)

"Name" is how camera manager, camera manager rest api, QGC and the mavlnk protocol spec name the field used for the human name of the stream. But if you think that "nickname" is better, I just changed it.

PR #1333, opened 2023-01-12
Add mini pirate in submenu

patrickelectric:
image


Williangalvani:
can we make the icon even smaller so we can see it inside the red circle? it looks like a ghost to me, idk why.

patrickelectric:
> can we make the icon even smaller so we can see it inside the red circle? it looks like a ghost to me, idk why.

Done.

ES-Alexander (approved):

PR #1340, opened 2023-01-12
core: frontend: nmea injector: show supported NMEA sentences

### LINKED ISSUE #1157: NMEA Injector should list the supported NMEA sentences
https://github.com/bluerobotics/BlueOS-docker/blob/master/core/services/nmea_injector/nmea_injector/MavlinkNMEA.py#L61

Williangalvani:
fix #1157 
Screenshot 2023-01-12 at 20 22 59

### DIFF
--- a/core/frontend/src/components/nmea-injector/NMEAInjector.vue
+++ b/core/frontend/src/components/nmea-injector/NMEAInjector.vue
@@ -4,19 +4,23 @@
     elevation="0"
     class="mx-auto my-6 injectors-list"
   >
-  <v-card
-    class="mx-auto my-6"
-    width="500"
-    elevation="1"
-  >
-  <v-card-text>
-      <div class="text--primary">
-        The NMEA Injector receives NMEA via UDP or TCP and transforms it
-        into MAVLink data that is forwarded to the autopilot.
-        It currently supports the sentences GPGGA, GPRMC, GPGLL, and GPGNS.
-      </div>
-    </v-card-text>
-  </v-card>
+    <v-card
+      class="mx-auto my-6"
+      width="500"
+      elevation="1"
+    >
+      <v-card-text>
+        <div class="text--primary">
+          The NMEA Injector receives NMEA via UDP or TCP and transforms it
+          into MAVLink data that is forwarded to the autopilot.
+          It currently supports the sentences
+          <span v-html="createLink('GPGGA')"/>,
+          <span v-html="createLink('GPRMC')"/>,
+          <span v-html="createLink('GPGLL')"/>, and
+          <a href="https://receiverhelp.trimble.com/alloy-gnss/en-us/NMEA-0183messages_GNS.html">GPGNSS</a>
+        </div>
+      </v-card-text>
+    </v-card>
     <v-list
       v-if="are_nmea_sockets_available && !updating_nmea_sockets"
       style="background-color: transparent"
@@ -118,6 +122,9 @@ export default Vue.extend({
   methods: {
     openCreationDialog(): void {
       this.show_creation_dialog = true
+    },
+    createLink(sentence: string): string {
+      return `<a href="http://aprs.gids.nl/nmea/#${sentence.toLowerCase().substring(2)}">${sentence}</a>`
     },
   },
 })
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/nmea-injector/NMEAInjector.vue
```diff
@@ -4,6 +4,19 @@
...
+  <v-card
+    class="mx-auto my-6"
+    width="500"
+    elevation="1"
+  >
+  <v-card-text>
+      <div class="text--primary">
+        The NMEA Injector receives NMEA via UDP or TCP and transforms it
+        into MAVLink data that is forwarded to the autopilot.
+        It currently supports the sentences GPGGA, GPRMC, GPGLL, and GPGNS.
```
patrickelectric, line 16: Can we add hyperlinks to http://aprs.gids.nl/nmea/#gga ?

`http://aprs.gids.nl/nmea/#{message}`

Williangalvani: I wish there was a more oficial source for these sentences...

Williangalvani: maybe here instead

ES-Alexander: > I wish there was a more official source for these sentences...

Patrick's source views as less official/legit, but has more information. [This one](https://gpsd.gitlab.io/gpsd/NMEA.html#_gga_global_positioning_system_fix_data) seems somewhere in between?

That said, given NMEA is a proprietary standard the only actually "official" source we could provide is [the purchase link](https://www.nmea.org/nmea-0183.html). If an official source is a requirement then it's likely not worth including a link at all.

Perhaps also worth noting that people may be looking at this on a boat without internet access. There's some case to be made for links in BlueOS (wherever possible) to be internal, in which case perhaps we should be providing our own reference for specifically the messages we support.

ES-Alexander (approved):

core/frontend/src/components/nmea-injector/NMEAInjector.vue
patrickelectric: This same website is online for more than 20 years providing the same information, although official information is not available, that's not how open source works, we should trust people that provides it, having a source that we can point for users to checkout and understand is a better option.

> Perhaps also worth noting that people may be looking at this on a boat without internet access.

We have hyperlinks for online resources already being used in BlueOS.

> [This one](https://gpsd.gitlab.io/gpsd/NMEA.html#_gga_global_positioning_system_fix_data) seems somewhere in between?

The one that I'm referencing has a simple mark for each command, making the code to add the hyperlink super simple compared with the others.

```ts
function createLink(sentence: string): string {
    return `http://aprs.gids.nl/nmea/#{sentence.substring(2)}`
}
```

The others links will need to create an entire table to map the commands, making such feature not worth it codewise to maintain.

ES-Alexander: > This same website is online for more than 20 years providing the same information, although official information is not available, that's not how open source works, we should trust people that provides it, having a source that we can point for users to checkout and understand is a better option.

I don’t doubt that it’s a good source - it just _looks_ old and cobbled together, which may make some users uncomfortable trusting it. 

Since the NMEA spec is frustratingly paywalled I’m ok with having that site linked to, or even having no link, but I do agree with Willian’s general sentiment that it’d be _nice_ if there was an official source we could just reference directly, instead of needing to link to something external that’s based on information scraped together from elsewhere.

Williangalvani: > This same website is online for more than 20 years providing the same information

that is part of the issue. it _feels_ like the website has been online for 20 years with no updates whatsoever

patrickelectric (approved):

core/frontend/src/components/nmea-injector/NMEAInjector.vue
Williangalvani: I ended up using yours and a Trimble one, as neither supported all the sentences we do =|

ES-Alexander: Lol, the resource I linked to and the Trimble one you ended up using both support all our current messages. The Trimble one even has generateable links like the aprs one 😂

PR #1343, opened 2023-01-13
core: frontend: Autopilot: Add path to the record over using if checks in template

### LINKED ISSUE #1341: core: frontend: autopilot
### Current behaviour

#1308 moved all the autopilot data into a table except "Path".

Screenshot 2023-01-13 at 3 49 29 pm

"Path" should also be aligned in the table. I'm assuming that would be something like, but I don't know vue so there might be some other stuff required.
```vue
          <v-row
             v-if="current_board.path"
             class="ma-0"
             no-gutters
          >
            <v-col class="text-right ma-0 font-weight-bold">
              Path:
            </v-col>
            <v-col class="text-left ml-1">
              {{ current_board.path }}
            </v-col>
          </v-row>
```
It could also maybe just be added to the `autopilot_info` function instead.

### Prerequisites

patrickelectric:
Fix #1341
image

### DIFF
--- a/core/frontend/src/views/Autopilot.vue
+++ b/core/frontend/src/views/Autopilot.vue
@@ -23,13 +23,6 @@
               {{ value }}
             </v-col>
           </v-row>
-          <span
-            v-if="current_board.path"
-            class="font-weight-bold"
-          >
-            Path:
-          </span>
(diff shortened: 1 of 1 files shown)


Williangalvani (approved):
LGTM, but what is path again?

PR #1344, opened 2023-01-13
core: services: ardupilot_manager: Remove invalid log files

patrickelectric:


patrickelectric:
ping @Williangalvani @rafaellehmkuhl

Williangalvani:
what do we gain from leaving newer empty files?

patrickelectric:
> what do we gain from leaving newer empty files?

As discussed on the previous call, the idea is for the user to identify that the logs exist and something should be done about it. Removing empty logs that are newer may hide some potential problems.

Williangalvani (approved):

PR #1349, opened 2023-01-16
core: frontend: ReportMenu: Remove slack button

patrickelectric:


ES-Alexander:
Seems like the `settings` were only used for the slack option here, so the import is no longer needed (which is the failed frontend test).

PR #1356, opened 2023-01-17
core: frontend: VersionChooser: Add paginator for remote versions

### LINKED ISSUE #1110: core: frontend: version chooser: limit number of displayed versions
### Current behaviour

Local versions and remote versions display all versions in Pirate mode, which means it's necessary to fully scroll through all of the local versions to get to the remote versions, and through all of the remote versions to get to the manual upload option. 

There are already 17 remote versions in the `bluerobotics/blueos-core` docker hub, and we'll only be adding more as newer versions get released. When developing it's common to also get a variety of different local versions.

Ideally the local and remote versions boxes should be scrollable, and be set to a height that allows displaying something like 3-5 versions. This would make it easy to scroll down the side to get to other sections, but also possible to view a few of both local and remote versions while determining which new version to update to.

Another option would be to paginate the versions in the boxes, but that seems like excessive effort and probably a worse user experience (albeit potentially a better experience if there are enough versions that they start adding lag to load the information about all of them).

### Prerequisites

patrickelectric:
https://user-images.githubusercontent.com/1215497/212785064-aeda08bc-4c5f-4bc9-af5d-bc70cd3e531e.mp4

Fix #1110

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionChooser.vue
+++ b/core/frontend/src/components/version-chooser/VersionChooser.vue
@@ -85,7 +85,7 @@
           subtitle="Loading remote images..."
         />
         <version-card
-          v-for="image in available_versions['remote']"
+          v-for="image in paginatedComponents"
           :key="image.sha"
           :image="image"
           :remote="true"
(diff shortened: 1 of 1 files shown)


ES-Alexander (approved):

ES-Alexander:
Is it also worth paginating the installed versions, in case there are large numbers of them?

patrickelectric:
> Is it also worth paginating the installed versions, in case there are large numbers of them?

I believe that we should not enforce so manu versions installed on the same computer :laughing:

patrickelectric:
We can also do that in another PR

PR #1360, opened 2023-01-17
core: frontend: VersionChooser: Add text to inform users about pirate mode

patrickelectric:

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionChooser.vue
+++ b/core/frontend/src/components/version-chooser/VersionChooser.vue
@@ -9,6 +9,17 @@
         max-width="900"
         class="mx-auto my-12 pa-4"
       >
+        <v-alert
+          border="top"
+          colored-border
+          type="info"
+          elevation="2"
+        >
+          Turn on Pirate mode to view all available BlueOS versions, including previous installs
+          stored onboard, as well as past and current stable, beta, and development releases
+          downloadable from Blue Robotics.
+        </v-alert>
+
         <h2>Current Version</h2>
         <version-card
           v-if="current_version"


core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -9,6 +9,16 @@
...
         class="mx-auto my-12 pa-4"
       >
+        <v-alert
+          border="top"
+          colored-border
+          type="info"
+          elevation="2"
+        >
+          Pirate mode can be enabled to show all official unstable versions,
+          including beta and custom BlueOs releases from Blue Robotics.
```
ES-Alexander, line 19:
```suggestion
          Turn on Pirate mode to view all available BlueOS versions, including previous installs
          stored onboard, as well as past and current stable, beta, and development releases
          downloadable from Blue Robotics.
```

We may want to make "Turn on Pirate mode" into a hyperlink that turns it on, or that opens the thing that lets you turn it on.

Also, should this description still be shown if Pirate mode is already on? I'm undecided on my preference. We could potentially also toggle-change it to start with something like "Turn on Pirate mode to view" -> "This page shows" when Pirate mode is on.

ES-Alexander (changes requested):
Like the available services one, I'm not sure what this looks like on the page. If possible it would be good to have similar styling between the description here and on other pages (e.g. NMEA Injector).

core/frontend/src/components/version-chooser/VersionChooser.vue
patrickelectric: > Also, should this description still be shown if Pirate mode is already on?

There is a if on the parent component for it.



> We may want to make "Turn on Pirate mode" into a hyperlink that turns it on, or that opens the thing that lets you turn it on.

We can keep that in mind while we are refactoring the pirate-mode enable disable functionality, right now to do it we already need to rework the main component, so it's better to keep in a future PR of the main refactor.

ES-Alexander (approved):

PR #1361, opened 2023-01-17
core: frontend: availableServicesTable: Add text informing about the page

patrickelectric:

### DIFF
--- a/core/frontend/src/components/scanner/availableServicesTable.vue
+++ b/core/frontend/src/components/scanner/availableServicesTable.vue
@@ -1,61 +1,75 @@
 <template>
-  <v-simple-table class="helper-table">
-    <thead>
-      <tr>
-        <th>Port</th>
-        <th>Service Name</th>
-        <th>Webpage</th>
-        <th>API Documentation</th>
-        <th>Versions</th>
-      </tr>
-    </thead>
-    <tbody>
-      <tr
-        v-for="service in availableServices"
-        :key="service.port"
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/scanner/availableServicesTable.vue
```diff
@@ -1,61 +1,72 @@
...
-          >
+  <v-col>
+    <v-alert
+      border="top"
+      colored-border
+      type="info"
+      elevation="2"
+    >
+      Shows all available services running on BlueOS, including the service port,
+      name, webpage, REST API endpoint and swagger documentation per version.
```
ES-Alexander, line 10:
```suggestion
      Shows all available services running on BlueOS, including the service port,
      name, webpage, REST API endpoint, and swagger documentation per version.
      
      API endpoint links provide direct manual access to the APIs, and can be used
      to test how a service responds to a given request.
```

ES-Alexander (changes requested):
Basic explanation was good - I added a bit more to try to describe what someone could use that for.

I'm not sure what this looks like on the page. If possible it would be good to have similar styling between the description here and on other pages (e.g. NMEA Injector)

core/frontend/src/components/scanner/availableServicesTable.vue
patrickelectric: Done.

ES-Alexander (approved):

PR #1365, opened 2023-01-17
core: frontend: MAVLinkInspector: Fix components width and height

patrickelectric:
image


ES-Alexander (approved):
Still room for improvement (of course), but this looks much easier to use already :-)

PR #1366, opened 2023-01-17
Add miniatures in bridges menu

patrickelectric:
Add miniatures on bridges menu

image


ES-Alexander (approved):
Seems potentially a bit big, but it's probably ok (I just find it a bit weird to scroll through large selectable options, although the flip-side of that is that they're more visible, which is likely useful for people with small screens or poor eyesight).

I'm curious what happens if/when someone has a flight controller that isn't a Pixhawk or Navigator. It'd be good if we can document the setup/files required to add full support for a new flight controller.

Williangalvani (changes requested):
this seems to have broken the not-inline variant used in the ping page
Screenshot 2023-01-17 at 10 56 52

patrickelectric:
@Williangalvani check now

Williangalvani (approved):
This still has a couple of css issues, but this is good enough as we'll rework this soon.

PR #1370, opened 2023-01-17
core: frontend: yarn lint --fix

Williangalvani:


patrickelectric:
We should put it as CI test

Williangalvani:
> We should put it as CI test

we did, but we need to take the warnings down to 0

patrickelectric:
> > We should put it as CI test
> 
> we did, but we need to take the warnings down to 0

But we did end up adding a upper limit for warnings, no ?

PR #1372, opened 2023-01-17
New pirate mode UI

### LINKED ISSUE #1339: core: frontend: switch pirate mode via header icon
### Current behaviour

Changing into and out of pirate mode requires going into the settings from the sidebar, which is inconvenient.

#1317 added a pirate icon to the header to indicate whether pirate mode is currently active. It would be nice to be able to just click on that icon to switch modes.

>> @patrickelectric:
>> how exactly do it, we are going to change the icon ? We are going to change the color ? What is the desired behavior of the UI
>
>> @ES-Alexander: 
>> I think changing icons is likely the most intuitive. Maybe swap between `pirate` (or `skull-crossbones`) and `emoticon-happy-outline` (or `emoticon-cool-outline`), with a tooltip that says "Click to {en/dis}able pirate mode."

### Prerequisites

patrickelectric:
https://user-images.githubusercontent.com/1215497/212980960-e49c609c-fd97-4e83-bd15-4d6583d25f79.mp4

Fix #1339

### DIFF
--- a/core/frontend/src/components/app/PirateModeMenu.vue
+++ b/core/frontend/src/components/app/PirateModeMenu.vue
@@ -1,0 +1,50 @@
+<template>
+  <v-card
+    class="d-flex flex-column align-center pa-3"
+    outlined
+    width="300"
+  >
+    <v-alert
+      v-if="!settings.is_pirate_mode"
+      colored-border
+      type="info"
+      icon="mdi-skull-crossbones"
+      elevation="2"
+      text
+    >
+      Use Pirate Mode to show hidden pages and advanced settings.
+      Pirate powers should be used with care.
+    </v-alert>
+    <v-btn
+      @click="togglePirateMode"
+    >
+      {{ settings.is_pirate_mode ? "Disable Pirate Mode" : "Enable Pirate Mode" }}
+    </v-btn>
+  </v-card>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+
+import settings from '@/libs/settings'
+
+export default Vue.extend({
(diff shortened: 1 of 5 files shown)


core/frontend/src/components/app/PirateModeMenu.vue
```diff
@@ -0,0 +1,50 @@
...
+    <v-alert
+      v-if="!settings.is_pirate_mode"
+      colored-border
+      type="info"
+      icon="mdi-skull-crossbones"
+      elevation="2"
+      text
+    >
+      Use Pirate Mode to show hidden pages and advanced settings.
+      Note that it's an advanced mode and should be used with care.
```
ES-Alexander, line 16: Seems a bit verbose. Maybe something like this?
```suggestion
      Pirate powers should be used with care.
```

ES-Alexander (approved):
Generally looks good, and I like the little happy robot for non-pirate mode :-)

It's a bit weird to me that the right side of the notification box isn't aligned with the right side of the icon, but I'm not sure how hard that is to fix, and don't know if fixing that would cause issues on mobile by cramping the notification too far over to the left.

By the way, is this included in the tour, or will that be added separately?

core/frontend/src/components/app/PirateModeMenu.vue
patrickelectric: Done.

patrickelectric:
> By the way, is this included in the tour, or will that be added separately?

We can discuss that in a future PR, not sure about making this "invatable" for the user to mess up.

ES-Alexander (approved):

PR #1375, opened 2023-01-17
core: frontend: App: Improve internet connection text to be more direct

patrickelectric:

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -331,8 +331,7 @@ export default Vue.extend({
         },
         {
           target: '#tour-center-hook',
-          content: `As stated on our welcome card, one of the first things you should do on your first use of BlueOS
-          is to connect it to the internet.`,
+          content: 'Connect BlueOS to the internet to enable online functionalities.',
           filter_wifi_connected: true,
           params: {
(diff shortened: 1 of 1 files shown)


core/frontend/src/App.vue
```diff
@@ -331,8 +331,7 @@ export default Vue.extend({
         },
         {
           target: '#tour-center-hook',
-          content: `As stated on our welcome card, one of the first things you should do on your first use of BlueOS
-          is to connect it to the internet.`,
+          content: 'Connect BlueOS to the internet to unlock functionalities.',
```
ES-Alexander, line 334:
```suggestion
          content: 'Connect BlueOS to the internet to enable online functionalities.',
```
"unlock" comes across (to me) as "if you're offline _we_ block you from some things, but if you turn on the internet we'll allow you to do it", whereas the intent here is more that firmware and version downloads and stuff can't work without some form of online connectivity.

ES-Alexander (changes requested):

core/frontend/src/App.vue
patrickelectric: Done.

ES-Alexander (approved):

PR #1376, opened 2023-01-19
core: frontend: Add thumbnail to Video Device page

joaoantoniocardoso:
## Video preview

https://user-images.githubusercontent.com/5920286/213883220-34d10dbf-d929-4092-9ecc-a1dd58b1dd01.mp4

## Test image

Test image: [joaoantoniocardoso/blueos-core/update-mavlink-camera-manager-and-thumbnail](https://hub.docker.com/layers/joaoantoniocardoso/blueos-core/update-mavlink-camera-manager-and-thumbnail/images/sha256-19895be609e498c53300cd078348f2af5bbd42fa54da534f7aa00829e1962078?context=explore)

### DIFF
--- a/core/frontend/src/components/video-manager/VideoThumbnail.vue
+++ b/core/frontend/src/components/video-manager/VideoThumbnail.vue
@@ -1,0 +1,117 @@
+<template>
+  <v-container class="d-flex flex-column align-center justify-space-between">
+    <v-avatar
+      :height="height"
+      :width="width"
+      tile
+    >
+      <img
+        v-if="thumbnail?.status === 200"
+        :src="thumbnail?.source"
+      >
+      <span
+        v-else-if="not_available"
+        :style="{ border: '2px dashed' }"
+        class="text-caption"
+        style="opacity: 30%; padding: 20px"
+      >
+        Preview not<br>
+        available<br>
+        Add a stream<br>
+        to have one
+      </span>
+      <spinning-logo
+        v-else
+        size="10%"
+        subtitle="Fetching thumbnail..."
+      />
+    </v-avatar>
+  </v-container>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { VAvatar, VContainer } from 'vuetify/lib'
+
+import SpinningLogo from '@/components/common/SpinningLogo.vue'
+import type { Thumbnail } from '@/store/video'
+import video from '@/store/video'
+
+export default Vue.extend({
+  name: 'VideoThumbnail',
+  components: {
+    VAvatar,
+    VContainer,
+    SpinningLogo,
+  },
+  props: {
+    source: {
+      type: String,
+      required: true,
+    },
+    height: {
+      type: String,
+      default: 'auto',
+    },
+    width: {
+      type: String,
+      default: '280',
+    },
+    register: Boolean,
+  },
+  data() {
+    return {
+      thumbnail: undefined as undefined | Thumbnail,
+      interval: undefined as undefined | number,
+    }
+  },
+  computed: {
+    not_available(): boolean {
+      if (this.thumbnail?.status === 404) {
+        return true
+      }
+      return this.thumbnail?.source === undefined && this.thumbnail?.status !== 503
+    },
+  },
+  watch: {
+    register(newValue, oldValue) {
+      if (!newValue && oldValue) {
+        video.stopGetThumbnailForDevice(this.source)
+        if (this.interval !== undefined) {
+          clearInterval(this.interval)
+          this.interval = undefined
+        }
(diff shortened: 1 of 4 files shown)


rafaellehmkuhl:
IBAGE OR DIDNT HAPPEN

core/frontend/src/components/video-manager/VideoThumbnail.vue
```diff
@@ -0,0 +1,109 @@
...
+      :height="height"
+      :width="width"
+      tile
+    >
+      <img
+        v-if="thumbnail?.status === 200"
+        :src="thumbnail?.source"
+      >
+      <span
+        v-else-if="thumbnail?.status === 404 || (thumbnail?.source === undefined && thumbnail?.status !== 503)"
```
patrickelectric, line 13: It would be nice to have the ifs here as a properties for better readability of the ui description.

core/frontend/src/components/video-manager/VideoThumbnail.vue
```diff
@@ -0,0 +1,109 @@
...
+    register(newValue, oldValue) {
+      if (!newValue && oldValue) {
+        video.stopGetThumbnailForDevice(this.source)
+        if (this.interval !== undefined) {
+          clearInterval(this.interval)
+          this.interval = undefined
+        }
+      } else if (newValue && !oldValue) {
+        video.startGetThumbnailForDevice(this.source)
+        if (this.interval === undefined) {
```
patrickelectric, line 78: Why not clearing the interval if != of undefined in this case ?

patrickelectric (changes requested):
Besides that, everything looks fine to merge!

core/frontend/src/components/video-manager/VideoThumbnail.vue
joaoantoniocardoso: I don't see any case where we would need to clear the interval when `newValue && !oldValue`, can you explain better?

The overall idea here is that when `newValue && !oldValue`, we should start asking for thumbnails, and if the interval was already set, we'll just reuse it. 

When the opposite is true (`!newValue && oldValue`), then we need to stop asking for thumbnails, clearing the interval because we would no longer be updating the thumbnail.

I've tried to get rid of this setInterval by creating a computed, but it was not being updated, and I don't really understand how the store should communicate that that variable had changed.

core/frontend/src/components/video-manager/VideoThumbnail.vue
joaoantoniocardoso: done

core/frontend/src/components/video-manager/VideoThumbnail.vue
patrickelectric: My point is that is safer, it may think that there is no case where it may happen, but doing it early, would fix this issue and will not break the current logic, so the idea is: Its looks safer, so why not ?

joaoantoniocardoso: I agree with you, but didn't understand what 'clearing the interval `if != of undefined`' means in the context.

When it's `newValue && !oldValue`, we want to use the interval, creating it if hasn't been created yet. We definitely don't want to clear it, unless you meant something like this:

```
if (this.interval === undefined) {
  clearInterval(this.interval)
}
this.updateThumbnail()
this.interval = setInterval(() => {
  this.updateThumbnail()
}, 1000)
```

patrickelectric: I was suggesting:
```js
      } else if (newValue && !oldValue) {
        video.startGetThumbnailForDevice(this.source)
        this.updateThumbnail()
        this.interval = setInterval(() => {
          this.updateThumbnail()
        }, 1000)
```

joaoantoniocardoso: Okay, but I imagine this could also lead to a problem, 'cause if the interval was already created, then we'd have a 'dangling' setInterval, but we could:

```typescript
} else if (newValue && !oldValue) {
  video.startGetThumbnailForDevice(this.source)
  if (this.interval === undefined) { 
    clearInterval(this.interval) 
  }
  this.updateThumbnail() 
  this.interval = setInterval(() => {
    this.updateThumbnail()
  }, 1000)
}
```

vs (originally):

```typescript
} else if (newValue && !oldValue) {
  video.startGetThumbnailForDevice(this.source)
  this.updateThumbnail()
  if (this.interval !== undefined) { 
    this.interval = setInterval(() => {
      this.updateThumbnail()
    }, 1000)
  }
}
```

I see how it's better, but then I'd prefer to change in all the other parts too...

patrickelectric (approved):

PR #1384, opened 2023-01-22
core: versionchooser: Check if sha does not exist on remote

patrickelectric:
If a master version was updated it will noy exist on remote since it'll be replaced by the new one available.

Fix #1368 
Fix #1379 

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

PR #1387, opened 2023-01-22
Add suport for OCI manifests

### LINKED ISSUE #1368: Getting failed getting sha from DockerHub
image

### LINKED ISSUE #1379: bug: update not showing in version chooser
### Bug description

Update does not show for master
image

### Steps to reproduce

Check master and wait for a update

### Primary pain point(s)

The button should appear as update and not download it manually again 

### Prerequisites

Williangalvani:
We were getting `OCI manifest found, but accept header does not support OCI (Open Container Initiative) manifest` errors when trying to fetch latest master manifests. My guess is that we somehow are shipping the image (or at least the manifest) in a different container format now:
Screenshot 2023-01-22 at 14 06 10

The error message doesn't look particularly good. but I'm happy to have the additional data in there

fix #1368 
fix #1379

potentially replaces #1384 

tested both in an older master with local changes, and current master with local changes. it seems to work fine:

Screenshot 2023-01-22 at 14 09 19

solution was grabbed from https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/31127

### DIFF
--- a/core/services/versionchooser/utils/dockerhub.py
+++ b/core/services/versionchooser/utils/dockerhub.py
@@ -70,7 +70,7 @@ class TagFetcher:
         """Fetches the digest sha from a tag. This returns the image id displayed by 'docker image ls'"""
         header = {
             "Authorization": f"Bearer {self.last_token}",
-            "Accept": "application/vnd.docker.distribution.manifest.v2+json",
+            "Accept": "application/vnd.docker.distribution.manifest.v2+json,application/vnd.oci.image.manifest.v1+json",
         }
         async with aiohttp.ClientSession() as session:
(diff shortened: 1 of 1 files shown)


patrickelectric (changes requested):
Update appears to be not working

image

Williangalvani:
the update button only shows up for the same tags. it compares the sha and if it is diferrent:
https://github.com/bluerobotics/BlueOS-docker/blob/master/core/frontend/src/components/version-chooser/VersionChooser.vue#L370-L378

that's why I mentioned the "local changes". I launched an older master in blueos and put the new code there to test.

Williangalvani:
oh. you probably took longer than 2 min with version chooser stopped and bootstrap reverted you to factory. btw

patrickelectric (approved):

PR #1388, opened 2023-01-23
More linting

Williangalvani:


patrickelectric (approved):
We should also decrease the counter when possible for the CI linter

PR #1395, opened 2023-01-25
core: frontend: autopilot: parameters: Add  to Array<Parameter>

patrickelectric:
Allow the usage of:

`parameters.get("SERVO(\d+)_FUNCTION")`

it'll return a list of available parameters if it exists


Williangalvani:
superseded by #1396

PR #1347, opened 2023-01-16
core: frontend: start peripherals setup

Williangalvani:
Screenshot 2023-01-16 at 18 57 32

PR #1396, opened 2023-01-25
core: frontend: store: autopilot: Add getters for parameters

patrickelectric:
A replace for #1395

As discussed on the dentist

### DIFF
--- a/core/frontend/src/store/autopilot.ts
+++ b/core/frontend/src/store/autopilot.ts
@@ -28,6 +28,17 @@ class AutopilotStore extends VuexModule {
 
   metadata_loaded = false
 
+  get parameter() {
+    return (name: string): Parameter | undefined => this.parameters.find((parameter) => parameter.name === name)
+  }
+
+  get parameterRegex() {
+    return (pattern: string): Parameter[] => {
+      const tester = new RegExp(pattern)
+      return this.parameters.filter((parameter) => tester.test(parameter.name))
+    }
+  }
+
   @Mutation
   reset(): void {
     this.parameters = []


Williangalvani:
I think you need to declare things slightly differently in order for these to work as vuex getters.
check https://championswimmer.in/vuex-module-decorators/pages/core/getters.html and https://github.com/bluerobotics/BlueOS-docker/pull/1347/commits/cd4ba0435befd646edcdb8b9cae9106dd0cea263

It returns a lambda that takes the parameters

patrickelectric:
@Williangalvani check now

core/frontend/src/store/autopilot.ts
```diff
@@ -28,6 +28,18 @@ class AutopilotStore extends VuexModule {
 
   metadata_loaded = false
 
+  get parameter() {
+    // eslint-disable-next-line max-len
+    return (name: string): Parameter | undefined => this.parameters.filter((parameter) => parameter.name === name).first()
```
Williangalvani, line 33:
```suggestion
    return (name: string): Parameter | undefined => this.parameters.find((parameter) => parameter.name === name)
```

Williangalvani (changes requested):
I think we should have a "filter" option that returns a list.  This is useful in cases such as "Get me all the SERVOXX_FUNCTION" parameters"

patrickelectric:
> I think we should have a "filter" option that returns a list. This is useful in cases such as "Get me all the SERVOXX_FUNCTION" parameters". I'm not sure if it should be the regex one in this pr or new one that returns multiple elements

The regex on returns multiple objects

core/frontend/src/store/autopilot.ts
patrickelectric: Nice

patrickelectric:
> > I think we should have a "filter" option that returns a list. This is useful in cases such as "Get me all the SERVOXX_FUNCTION" parameters". I'm not sure if it should be the regex one in this pr or new one that returns multiple elements
> 
> The regex on returns multiple objects

Ops, my fault, fixing it

Williangalvani (approved):

PR #1404, opened 2023-01-27
CI: build armv8 64bit images

Williangalvani:
just creating a draft to check if CI finds some errors for us

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -94,7 +94,7 @@ jobs:
       matrix:
         docker: [bootstrap, core]
         project: [blueos]
-        platforms: ["linux/arm/v7"]
+        platforms: ["linux/arm/v7,linux/arm64/v8"]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -121,7 +121,7 @@ jobs:
 
           echo "docker_image=${DOCKER_IMAGE}" >> $GITHUB_OUTPUT
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
-          echo "buildx_args=--platform ${{ matrix.platforms }} \
+          echo "buildx_args=\
             --build-arg GIT_DESCRIBE_TAGS=$(git describe --tags --long) \
             --build-arg VUE_APP_GIT_DESCRIBE=$(git describe --long --always --dirty --all) \
             --cache-from 'type=local,src=/tmp/.buildx-cache' \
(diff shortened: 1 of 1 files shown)


Williangalvani:
it doesnt show a manually built version on version chooser for me, which probably means the [architecture filter](https://github.com/Williangalvani/blueos-docker/blob/master/core/services/versionchooser/utils/dockerhub.py#L104) in there is workind.

Williangalvani:
the tag for this branch doesnt show here though: https://hub.docker.com/layers/williangalvani/blueos-core/v8-goes-brrrrrr/images/sha256-706ba408ecf8d24cfc12a9091713006eca1c9ea5284a5cec079a383965f28363?context=explore

I suspect the one that finishes last is overwriting the first one, in CI

.github/workflows/test-and-deploy.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         docker: [bootstrap, core]
         project: [companion]
         new_project: [blueos]
-        platforms: ["linux/arm/v7"]
+        platforms: ["linux/arm/v7", "linux/arm64/v8"]
```
Williangalvani, line 98:
```suggestion
        platforms: ["linux/arm/v7,linux/arm64/v8"]
```

We'll need to run both in the same action, apparently. I'll do some reading on docs.

Williangalvani:
@patrickelectric thoughts?

The issue here is that we cant do  the  `--output` line for two different images at the same time.

I think we could maybe do something like a `.join()` somewhere. or maybe we can use --output for two lines, but I couldn't find how to do it.

Williangalvani:
nevermind me, [it seems to be possible](https://docs.docker.com/engine/reference/commandline/buildx_build/#output), let me troubleshoot it a bit further

patrickelectric:
You can use a matrix and the name of the architecture in the artifact

patrickelectric:
> > architecture
> 
> using a matrix causes 2 independent pushes to docker hub, where the latest one overwrites the first one

ouch, we can build both together and each individually for the artifacts, it should use the cache from both builds and just create it for us.

Williangalvani:
> > > architecture
> > 
> > 
> > using a matrix causes 2 independent pushes to docker hub, where the latest one overwrites the first one
> 
> ouch, we can build both together and each individually for the artifacts, it should use the cache from both builds and just create it for us.

That is what the current change does. apparently we could use the "tar" output, but that doesn't seem to be "docker image specification compliant"

patrickelectric (approved):
I did some attempts to parallelize it but not ended well, I did updates and now appears to be fine.

patrickelectric:
Be free to merge.

PR #1409, opened 2023-01-28
Improve BlueOS logo

patrickelectric:
- Use new full logo

image

image


ES-Alexander (approved):
Looks fine to me :-)

My only note would be that companies re-skinning BlueOS will likely want to replace that main logo location with their own, in which case I'm not sure what the best location is for having a BlueOS logo (assuming we care about making it clear that the software running is based on BlueOS).

patrickelectric:
> Looks fine to me :-)
> 
> My only note would be that companies re-skinning BlueOS will likely want to replace that main logo location with their own, in which case I'm not sure what the best location is for having a BlueOS logo (assuming we care about making it clear that the software running is based on BlueOS).

We are still discussing it and how to accomplish custom skins in a better way without building BlueOS.
Extensions or zip folders may be a solution.

ES-Alexander:
> We are still discussing it and how to accomplish custom skins in a better way without building BlueOS.
> Extensions or zip folders may be a solution.

Yes, I'm expecting having a "theme" type of extension will be the most easily shareable. Longer term it would be good to have a theme editor built into BlueOS, so people can customise the theme from within BlueOS and then save and choose to share their theme automatically.

Regardless, that's unrelated to whether or not people will want to change the main logo, and doesn't necessarily help us determine where the best place is for a secondary BlueOS logo (assuming we want one) in the case of someone applying a custom theme.

PR #1414, opened 2023-01-28
Improve deviceid_decoder

patrickelectric:


Williangalvani:
idk exactly how I messed that up. but in my original it makes sense:
https://github.com/ArduPilot/UAVLogViewer/blob/master/src/components/widgets/DeviceIDViewer.vue#L152-L158

IIRC, UAVCAN and EAHRS do not provide the ID, so I did that in order to not show the wrong device

patrickelectric:
@Williangalvani check now

Williangalvani (approved):

PR #1417, opened 2023-01-30
core: services: ardupilot_manager: Add try block in remove_old_logs

patrickelectric:
iterdir can fail


Williangalvani (approved):
I assume you considered using is_dir() and decided otherwise.

Issue #1073, opened 2022-07-13
install: remove dependence on `pi` user


ES-Alexander:
Raised in #1072, and relevant because:
- not every OS has `pi` as the default user (e.g. Ubuntu)
- support for non-Raspberry Pi boards will likely require support for alternative user names

Should either:
- change script to use `$USER` or `$USERNAME` instead of `pi`, or
- set a `target_user` variable (or similar) at the top which defaults to `pi` (or `$USER`) and allows specifying something else with a command-line argument

Jollerprutt:
The Raspberry Pi foundation has [announced](https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/) they are dropping pi as the default user, and that there will no longer be any default username.
So this issue will soon become more urgent as people start using the newer images.

ES-Alexander:
Will also need to change [commonwealth](https://github.com/bluerobotics/BlueOS-docker/blob/master/core/libs/commonwealth/commonwealth/utils/commands.py#L5)

ES-Alexander:
> support for non-Raspberry Pi boards...

[Case in point](https://discuss.bluerobotics.com/t/jetson-nano-as-onboard-computer/14016/20).

PR #1419, opened 2023-01-31
create "userdata" folder for allowing customizations

Williangalvani:
Currently works with custom css files located in the host's [/home/pi/userdata/](http://blueos.local:7777/file-browser/files/home/pi/userdata/)user_style.css`
image

The screenshot uses these updated variables:
```
:root {
    --v-primary-base: #dd8603 !important;
    --v-info-base: #dd8603 !important;
    }
```
The full list  (courtesy of somewhere in chrome's dev tools) is here:  (upload when chrome is happier)

I intend to use the same or a similar mechanism to allow overriding images/models.

### DIFF
--- a/bootstrap/startup.json.default
+++ b/bootstrap/startup.json.default
@@ -29,6 +29,10 @@
             "/var/logs/blueos": {
                 "bind": "/var/logs/blueos",
                 "mode": "rw"
+            },
+            "/usr/blueos/userdata": {
+                "bind": "/usr/blueos/userdata",
+                "mode": "rw"
             }
         },
         "privileged": true
--- a/core/tools/nginx/nginx.conf
+++ b/core/tools/nginx/nginx.conf
@@ -187,6 +187,15 @@ http {
             try_files $uri $uri/ /index.html;
         }
 
+        location /userdata {
+            root /usr/blueos;
+            autoindex on;
+            # disable cache to improve developer experience
+            # this should have very little impact for users
+            expires -1;
+            add_header Cache-Control no-store;
+        }
+
         location ~* \.html$ {
             root /home/pi/frontend;
             try_files $uri $uri/ /index.html;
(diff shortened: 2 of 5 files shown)


bootstrap/startup.json.default
```diff
@@ -29,6 +29,10 @@
             "/var/logs/blueos": {
                 "bind": "/var/logs/blueos",
                 "mode": "rw"
+            },
+            "/home/pi/userdata": {
+                "bind": "/home/pi/userdata",
```
patrickelectric, line 34: We should not bind the host computer with the username, we should use the root folder or a better path, maybe `/usr/blueos`

ES-Alexander: Agreed with @patrickelectric, especially for future consistency with #1073

Williangalvani:
from meeting: 
try loading a  "model-override.glb", if it is not available, fallback to frame-specific

change to theme-style.css

Williangalvani:
add support for default firmware, default parameters.

patrickelectric (commented):
nice!

Williangalvani:
> nice!

thanks, I'm pushing some other stuff for the vehicle setup in here, but i'll probably remove it to add it in another pr later on. I'm just making sure everything works the way I expect first

ES-Alexander:
> change to theme-style.css

Is this still planned in this PR?

core/tools/nginx/nginx.conf
```diff
@@ -187,6 +187,14 @@ http {
             try_files $uri $uri/ /index.html;
         }
 
+        location /userdata {
+            root /usr/blueos;
+            autoindex on;
+            # do we want these next two lines?
+            expires -1;
+            add_header Cache-Control no-store;
```
joaoantoniocardoso, line 195: It makes sense to disable cache, makes development easier... And I don't think it would impact that much on loading performance.

Williangalvani:
> > change to theme-style.css
> 
> Is this still planned in this PR?

updated

Williangalvani:
ping @patrickelectric

PR #1427, opened 2023-02-01
core: frontend: do not compare non-semver tags to semver tags

### LINKED ISSUE #1423: Version Chooser: can't update - "Invalid version: master"
Can't update from [heads/master-0-gbddcf49](https://github.com/bluerobotics/blueos-docker/tree/bddcf49)

This is a very _thin_ bug - if you are in master and not in pirate mode, the Version Chooser keeps the spinning wheel as if it would be capable of updating from it, but apparently, it doesn't know 'master':

image

Williangalvani:
fix #1423


core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -330,6 +330,12 @@ export default Vue.extend({
       if (!this.current_version) {
         return ''
       }
+      try {
+        const test = new SemVer(this.current_version.tag)
+      } catch {
+        // if we are on a random, non-SemVer tag, we cant compare it to a SemVer tag.
+        return ''
+      }
```
joaoantoniocardoso, line 338: The compiler seems skeptical about the need for this try-catch

joaoantoniocardoso (approved):
Tested, and I can confirm that it solves the problem :)

just make the compiler happy and you're good to go

joaoantoniocardoso (changes requested):
Actually, there is a missing detail here:
- When we are at the latest master, the spinner keeps spinning forever
image

core/frontend/src/components/version-chooser/VersionChooser.vue
Williangalvani: got rid of it

Williangalvani:
updated and rebased, no spinner for me:
Screenshot 2023-02-02 at 12 27 56

patrickelectric (approved):

joaoantoniocardoso (approved):
Confirmed :)

Issue #1434, opened 2023-02-03
Mavlink2Rest lib fails silently if socket doesnt exist


Williangalvani:
If socket is null here, the function fails silently.
https://github.com/bluerobotics/BlueOS-docker/blob/master/core/frontend/src/libs/MAVLink2Rest/index.ts#L108

Williangalvani:
fixed by #1462

PR #1437, opened 2023-02-03
core: frontend: mavlink2rest: recreate websocket if needed

Williangalvani:


core/frontend/src/libs/MAVLink2Rest/index.ts
```diff
@@ -104,6 +104,11 @@ class Mavlink2RestManager {
    *
    */
   sendMessageViaWebsocket(message: any): void {
+    if (!this.socket) {
+      this.socket = this.createSocket(this.baseUrl)
+      this.socket.onopen = () => this.sendMessageViaWebsocket(message)
+      return
```
patrickelectric, line 110: That's a weird case, for this to be the case, the constructor would not be called at all, because that is how `createSocket` happens.

How `socket` gets undefined since the usage of the class itself makes it exist ?
And if it's closed, https://github.com/bluerobotics/BlueOS-docker/pull/1437/files#diff-df3bf14ffcc5f9af5bdd02a4c9f2ad4d7f85532d34d2ce3b095e8d0bafa7892aR82 should take effect.

Williangalvani: `probeBaseUrlCandidates` is failiting on my end in the `fetch(asHtto)` step, which means `setBaseUrl` is never called.

patrickelectric: Doesn't it make more sense to fix it under `probeBaseUrlCandidates` ? Where if the probe is failing it reeschedule to run again ? Or add a fallback with `this.createSocket(this.baseUrl)` ? That's the same thing that you are doing.

Williangalvani: this already cost me more time than I wish it did, I'll come back to it later.

joaoantoniocardoso:
Is this the fix for https://github.com/bluerobotics/BlueOS-docker/issues/1434?

Williangalvani:
> Is this the fix for #1434?

yes. are you running into it too?

PR #1439, opened 2023-02-03
core: frontend: deal with error in probing baseUrlCandidates

Williangalvani:


Williangalvani:
This actually fails to catch the CORS error.
So does surrounding the fetch with try/catch...

PR #1451, opened 2023-02-07
Add 404 on router

patrickelectric:
image


joaoantoniocardoso (approved):
Just notice here that making a catch-all route will result in the 404 page having a `200` status code response, which works in our case, but for online websites usually is not the best idea. Probably we would have to have a catch-all on the server instead, maybe redirecting it to a `/404` route :)

patrickelectric:
This is just for the router on BlueOS, we should add one in nginx.

PR #1453, opened 2023-02-08
Decrease the warning number from 36 to 22

patrickelectric:


patrickelectric:
The only warning messages are from MAVLink messages:

image


I'm planning to fix that soon.

core/frontend/src/components/speedtest/Graph.vue
```diff
@@ -71,10 +71,11 @@ export default Vue.extend({
     },
   },
   methods: {
-    updateSeries(series: Data) {
+    async updateSeries(series: Data) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const chart = this.$refs.chart as any
```
joaoantoniocardoso, line 76: We could define a `chart` type to prevent using `any`

patrickelectric: So, I did try creating a d.ts but without luck, the typescript was just failing to find it at all.

joaoantoniocardoso (approved):

PR #1458, opened 2023-02-08
Add External/Internal and priority to compass overview

Williangalvani:
Screenshot 2023-02-08 at 15 40 07

### DIFF
--- a/core/frontend/src/components/vehiclesetup/overview/OnboardSensors.vue
+++ b/core/frontend/src/components/vehiclesetup/overview/OnboardSensors.vue
@@ -43,7 +43,9 @@
               :key="compass.param"
             >
               <td><b>{{ compass.deviceName ?? 'UNKNOWN' }}</b></td>
-              <td>Compass</td>
+              <td>
+                {{ compass_description[compass.param] }}
+              </td>
               <td>{{ compass.busType }} {{ compass.bus }}</td>
               <td>{{ `0x${compass.address}` }}</td>
               <td>{{ compass_is_calibrated[compass.param] ? 'Calibrated' : 'Needs Calibration' }}</td>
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/vehiclesetup/overview/OnboardSensors.vue
```diff
@@ -96,6 +98,40 @@ export default Vue.extend({
...
+                priority = '1st'
+                break
+              case 2:
+                priority = '2nd'
+                break
+              case 3:
+                priority = '3rd'
+                break
+              default:
+                priority = 'Unused'
```
patrickelectric, line 122: Should we use [ordinal](https://www.npmjs.com/package/ordinal) ? Os it'll not be greater than 3 ?

core/frontend/src/utils/deviceid_decoder.ts
```diff
@@ -162,5 +163,6 @@ export default function decode(device: string, devid: number): deviceId {
     bus,
     address: toHex(address),
     devtype,
+    paramValue: devid,
```
patrickelectric, line 166: isn't that the same as param.value ?

core/frontend/src/components/vehiclesetup/overview/OnboardSensors.vue
Williangalvani: Ardupilot doesn't support more than 3 compasses in use

core/frontend/src/utils/deviceid_decoder.ts
Williangalvani: param here is a string  :sweat_smile:

patrickelectric: facepalm-really

patrickelectric (approved):

PR #1461, opened 2023-02-09
core: create filesystem shortcutes/symlinks

Williangalvani:
Have you ever had to tell users where to find some specific data in the filesystem?
Worry no more! With the new, revamped "shortcuts 2000", your path-finding days are over!

We now offer a revolutionary way of directing users to the correct folders:

Screenshot 2023-02-09 at 13 40 51

And that is not all.  If you review it in the next 2 hours, you get it for absolutely FREE! that's right, "free" as in "free beer" AND "free" as in "freedom" !!!!11!!eleven!!

image at williangalvani/blueos-core:shortcutes


joaoantoniocardoso (approved):
:1st_place_medal: for the PR description

patrickelectric (approved):

PR #1467, opened 2023-02-13
core: commander: Add environment_variables endpoint

patrickelectric:


Williangalvani:
I wonder if these could be used to expose things we dont want exposed, such as CI variables

patrickelectric:
> I wonder if these could be used to expose things we dont want exposed, such as CI variables

You already have access to it via the terminal

joaoantoniocardoso:
> > I wonder if these could be used to expose things we dont want exposed, such as CI variables
> 
> You already have access to it via the terminal

I believe env variables set for creating the docker wouldn't be accessible from inside the docker if we don't deliberately pass them with `ENV` or as args.

joaoantoniocardoso (approved):

ES-Alexander:
> I believe the ... wouldn't be accessible from inside the docker ...

I may be misunderstanding this, but given we can run `red-pill` from the terminal in the docker to access the base operating system, everything outside the docker is technically already accessible from inside the docker is it not? At least I imagine that's the case from a privacy/security standpoint, except for things that require elevating privileges.

joaoantoniocardoso:
@ES-Alexander 

* Inside the docker:
image

* Outside the docker:
image

ES-Alexander:
> * Inside the docker:
> ...
> * Outside the docker:
>  ...

Yes, but my point was that we have a tool inside the docker that allows you to access the outside, so from a security standpoint anything we can access via the web terminal that doesn't require putting in a password is also accessible to an arbitrary program running in the docker.

joaoantoniocardoso:
> > * Inside the docker:
> >   ...
> > * Outside the docker:
> >   ...
> 
> Yes, but my point was that we have a tool inside the docker that allows you to access the outside, so from a security standpoint anything we can access via the web terminal that doesn't require putting in a password is also accessible to an arbitrary program running in the docker.

I agree, and that security breach exists since we added the commander (`/command/host`). I think the commander should have a restrictive CORS and only allows access from the same address, but it has no relation with environment variables.

Williangalvani:
I was actually thinking of CI, where our dockerhub credentials were avaialable, but github does require us to allow CI to run for new contributors, and also I think now that is only provided to the deployment action

Williangalvani (approved):

PR #1472, opened 2023-02-14
core: frontend: Fix settings and bug/report dialogues feel

### LINKED ISSUE #1470: core: frontend: sidebar settings widget has a weird bottom margin
Screenshot 2023-02-14 at 6 31 54 pm

joaoantoniocardoso:
Closes #1470

image

image


ES-Alexander:
Thanks for looking at this :-)
The issue wasn't a "feel" problem - the margin extended outside the box so it looked like there was a second window being covered by the front part (you can see the semi-transparent bit in my screenshot in the issue).

I think feeling wise I prefer the tighter padding that's currently being used - the padding in your screenshots seems a bit excessive (to me anyway).

joaoantoniocardoso:
> Thanks for looking at this :-) The issue wasn't a "feel" problem - the margin extended outside the box so it looked like there was a second window being covered by the front part (you can see the semi-transparent bit in my screenshot in the issue). It's only an issue on the Settings window, not the bug report one.
> 
> I think feeling wise I prefer the tighter padding that's currently being used - the padding in your screenshots seems a bit excessive (to me anyway).

Hmm, I made them match the Shutdown menu. If we want fewer margins, then I would like to change the Shutdown too

ES-Alexander:
> Hmm, I made them match the Shutdown menu. If we want smaller margins, then I would like to change the Shutdown too

I'd be fine with tighter padding on the shutdown window :-)
My focus is just not having a clearly broken UI element, which is why I reported it :-)

joaoantoniocardoso:
@ES-Alexander take a look now

ES-Alexander (approved):
LGTM, although I also don't know what was causing the margin issue before, so🤞🏻that this fixes it :-)

If you're after some tooltip suggestions:

Button | tooltip suggestion
---|---
Reset Settings | Restores BlueOS services to default configurations
Remove Services Log Files | Frees up space on the SD card
With GitHub | Track changes, contribute, and get notified when fixed
On Blue Robotics Forum | Discuss ideas with the community

joaoantoniocardoso:
> I also don't know what was causing the margin issue before

The wrong margin seems to be that Vue gets the wrong height of a card when it's not inside a container.

I added the suggested tooltips :)

ES-Alexander:
@patrickelectric can we merge this? :-)

patrickelectric (approved):

PR #1500, opened 2023-02-20
core: frontend: components: SystemCondition: Filter out docker binds in disks

patrickelectric:
There is no easy way to do it for now using sysinfo library and update linux2rest, hopefully more time or something better may be possible in the future.

### DIFF
--- a/core/frontend/src/components/system-information/SystemCondition.vue
+++ b/core/frontend/src/components/system-information/SystemCondition.vue
@@ -86,14 +86,16 @@ export default Vue.extend({
       const [used_disk_space, total_disk_space] = get_space_disk(main_disk)
 
       const disk_text = disks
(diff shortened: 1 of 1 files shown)


Williangalvani:
this seems to be doing something to the disks instead of ipv6...

joaoantoniocardoso:
> this seems to be doing something to the disks instead of ipv6...

@patrickelectric you might have mistakenly kept the wrong part of the https://github.com/bluerobotics/BlueOS-docker/commit/45b58c5324fa7aa092ca2db6e3cbccf18dfbe9f9 commit..?

patrickelectric:
@Williangalvani @joaoantoniocardoso I just fixed

joaoantoniocardoso (approved):
I didn't test it but LGTM

PR #1504, opened 2023-02-21
github: ci: Update BlueOs docker image artifact name

patrickelectric:


joaoantoniocardoso (commented):
Just to understand, why are we changing it?

patrickelectric:
> Just to understand, why are we changing it?

Because the release name does not imply that it's a docker image.

https://github.com/bluerobotics/BlueOS-docker/releases/tag/1.1.0-beta.16

joaoantoniocardoso (approved):
makes sense, thanks

Issue #1503, opened 2023-02-21
Manual update appears to be not working on beta 16


patrickelectric:

Williangalvani:
I am unable to replicate.
I tried both in master and beta.16. both worked as expected.
bootstrap reported no downtime other than ~15 s for switching versions.
version chooser also had no relevant output in the logs.

maybe this is related to boostrap version? I am on 1.0.1:
```
pi@blueos:~ $ docker image ls | grep bootstrap
bluerobotics/blueos-bootstrap          1.0.1           34887e76b5d5   10 months ago   86.6MB
```

joaoantoniocardoso:
> I am unable to replicate. I tried both in master and beta.16. both worked as expected. bootstrap reported no downtime other than ~15 s for switching versions. version chooser also had no relevant output in the logs.
> 
> maybe this is related to boostrap version? I am on 1.0.1:
> 
> ```
> pi@blueos:~ $ docker image ls | grep bootstrap
> bluerobotics/blueos-bootstrap          1.0.1           34887e76b5d5   10 months ago   86.6MB
> ```

I'm also unable to replicate it, from what version you had experience that @patrickelectric ?

Williangalvani:
it seems we are running out of memory when extracting on pi 3... we'll try increasing the available swap.

ES-Alexander:
We tried updating to `blueos-bootstrap:master`, and increasing swap, which may have helped, but didn't fully resolve the issue.

Williangalvani:
I still believe the issue is memory. I think we should switch https://aiodocker.readthedocs.io/en/latest/_modules/aiodocker/images.html#DockerImages.import_image for https://docker-py.readthedocs.io/en/1.5.0/api/#import_image_from_file

PR #1505, opened 2023-02-21
core: tools: nginx: Increase timeout for version-chooser to process image

patrickelectric:
Helps #1503


patrickelectric:
ping @Williangalvani

PR #1506, opened 2023-02-21
Allow download system log

### LINKED ISSUE #1047: core: frontend: Add page to download logs
### Current behaviour

Logs are available in `/var/logs/blueos`.
The files can be accessed and downloaded from the file manager.

We could have a page to list all logs for the user per services maybe and let him choose to download or download everything automatically.

### Prerequisites

patrickelectric:
Fix #1047


joaoantoniocardoso (approved):
My only concern is about this:

image

PR #1507, opened 2023-02-21
Add system log folder size

### LINKED ISSUE #1487: Remove logs tool could show logs size before deletion

patrickelectric:
image

Fix #1487

### DIFF
--- a/core/frontend/src/components/app/SettingsMenu.vue
+++ b/core/frontend/src/components/app/SettingsMenu.vue
@@ -13,19 +13,18 @@
       <v-icon>mdi-cog</v-icon>
     </v-btn>
     <v-dialog
+      class="pa-2"
       width="fit-content"
       :value="show_dialog"
       @input="showDialog"
     >
-      <v-card>
-        <v-card-title class="align-center">
-          Settings
-        </v-card-title>
+      <v-card class="mx-auto pa-2 flex-column">
+        <v-card>
+          <v-card-title class="align-center">
+            Settings
+          </v-card-title>
 
-        <v-divider />
-
-        <v-container class="pa-2">
-          <v-card-actions class="flex-column">
+          <v-container class="pa-2 align-center">
             <v-btn
               v-tooltip="'Restores BlueOS services to default configurations'"
               class="ma-2"
@@ -36,7 +35,15 @@
               </v-icon>
               Reset Settings
             </v-btn>
+          </v-container>
 
+          <v-divider />
+
+          <v-card-title class="align-center">
+            System Log Files ({{ log_folder_size }})
+          </v-card-title>
+
+          <v-card-actions class="flex-row">
             <v-btn
               v-tooltip="'Download log for all services in BlueOS'"
               class="ma-2"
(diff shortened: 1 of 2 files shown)


patrickelectric:
> Maybe add a confirmation before removing the files?

There is a issue just for that, it's outside the scope of this one

joaoantoniocardoso (commented):
I'm testing with `yarn serve`, and it's showing nothing inside the parenthesis, even though I downloaded a 21MB zipped file of logs.

patrickelectric:
> I'm testing with `yarn serve`, and it's showing nothing inside the parenthesis, even though I downloaded a 21MB zipped file of logs: image

That changes the backend, you need to use my remote and download a new version

joaoantoniocardoso (changes requested):
image

It's reporting the wrong size, seems to be getting the size when the system startup maybe..? Or cached?
I intentionally added a 1GB file and the reported size didn't change.
edit: Refreshing the page makes it report 1GB, which is a good-enough approximation.

Also, the UI looks odd here:
image

patrickelectric:
@joaoantoniocardoso check the latest version

joaoantoniocardoso (changes requested):
Hmm, not quite there yet, I believe:

1. The cached size still exists.

2. There's a fundamental problem here - if we delete the logs that our services are writing their logs into, we'll lose all the next log lines until they decide to create a new one.

IMHO the simplest solution would be to use `lsof` to discover and exclude them from the result of a recursive deletion algorithm.

3. About the UI:
image

It has two cards with different spacings, inside a white block, and it's using a card title to convey information, instead of text. This looks quite odd compared to our other dialogs, which recently had their look/feel uniformized:

image

Maybe try to use only one card (inside a container) and separators, and put the size information as text, removing the necessity to have two cards to separate the contents.

Another option is to create a card inside a container, and two cards inside that card, with more spacing around them, like in the video widget (it might be what you are doing, then add more space to look less odd.. But I still think the previous options is better):

image

patrickelectric:
@joaoantoniocardoso can you take a look now ?

joaoantoniocardoso (changes requested):
1. UI looks perfect now:
image
 
2. Services stop logging when we delete the logs and they only restart logging after a restart (or probably after their rotation period). As I said above, we should avoid deleting the logs files in use.

3. The size only changes when I reload the tab or when I click on the 'delete' button.

4. By deleting the entire `services` folder, we create an incompatibility with the expected behavior from the File Browser.

See 3 and 4 in the video below:
weird_behavior_and_compatibility

patrickelectric:
> 3\. [not good] The size only changes when I reloaded the tab or when I clicked on the 'delete' button.

Well, that's expected, is the only two cases where the size will change while the user is in the page. Calculating the size can be pretty expensive, we should avoid when necessary.

patrickelectric:
> * [not good] The size only changes when I reloaded the tab or when I clicked on the 'delete' button.
> * [not good] By deleting the entire `services` folder, we create an incompatibility with the expected behavior from the File Browser.

Both of this issues are unrelated from this PR, they exist on master branch.

joaoantoniocardoso (approved):
You probably meant #1523 :laughing:

PR #1508, opened 2023-02-21
core: frontend: check both tag and repository for showing current image

Williangalvani:
testing is welcome


patrickelectric (approved):
It works

image

PR #1514, opened 2023-02-23
core: frontend: Change icon and tooltip for SDP URL copy button

joaoantoniocardoso:
image

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStream.vue
+++ b/core/frontend/src/components/video-manager/VideoStream.vue
@@ -38,8 +38,8 @@
           <v-btn v-tooltip="'Download to file'" class="ma-1 elevation-1" small @click="downloadSDPFile">
             <v-icon>mdi-download</v-icon>
           </v-btn>
-          <v-btn v-tooltip="'Copy to clipboard'" class="ma-1 elevation-1" small @click="copySDPFileURL">
-            <v-icon>mdi-content-copy</v-icon>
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

Williangalvani:
I feel like content-copy makes more sense. I'm pretty sure I've seen it used for link plenty of times
https://pictogrammers.com/library/mdi/icon/content-copy/

joaoantoniocardoso:
> I feel like content-copy makes more sense. I'm pretty sure I've seen it used for link plenty of times https://pictogrammers.com/library/mdi/icon/content-copy/

Icon names are not always the best guess for their use, but would the [`content-copy`](https://pictogrammers.com/library/mdi/icon/content-copy/) be better suited than [`file-link`](https://pictogrammers.com/library/mdi/icon/file-link/) in this case?

patrickelectric:
@joaoantoniocardoso @Williangalvani what is the status of this PR ?

joaoantoniocardoso:
@patrickelectric I was waiting for @Williangalvani to show examples or further reasoning against the icon change.

ES-Alexander (approved):
LGTM :-)

Willian, if we were _showing_ the file url here then I'd agree that `content-copy` would make more sense (since there would be clear content being copied), but since it just says "File URL" with a download button then I think the `file-link` makes more sense in this case.

Williangalvani (approved):
yeah, I dont have THAT strong feelings about this

Issue #1296, opened 2023-01-03
Add webpage/system telemetry to BlueOS


patrickelectric:
Right now we don't have information about our users besides the surveys in our forum.
Adding telemetry in BlueOS could automatically inform us about a series of key factors, such as: BlueOS version, vehicle type, number of missions and more.

Williangalvani:
What do we want:
 - Basic system information:
   - Host Hardware
     - Sd card size/storage status
     - Detected peripherals
     - memory size
   - Ardupilot firmware
   - Cpu temperature Peak
 - BlueOS uptime
 - Ardupilot uptime
 - BlueOS  version
 
 Good to have:
   - Website interactions

patrickelectric:
* Water temperature (thruster design)
* Maximum and medium depth (products)

patrickelectric:
@voorloopnul what is the status ?

PR #1522, opened 2023-03-01
Introducing: Major tom and Ground Control

Williangalvani:
Major Tom will be responsible for sporadically sending data back to Ground Control.

Ground control will run on a cloud (which is weird to think about it) and will live in different repository.

Once every 5 minutes, Major Tom will gather some data and try to send it back if there is internet connectivity.

We can get this smaller version in for the stable release, and then work our way up. As the backend can live independently, it can be very bare for now.

helps #1296

### DIFF
--- a/core/services/major_tom/major_tom.py
+++ b/core/services/major_tom/major_tom.py
@@ -1,0 +1,82 @@
+import asyncio
+from typing import Any, Dict, List
+
+import requests
+from commonwealth.settings.manager import Manager
+from loguru import logger
+
+from session_tracker import SessionTracker
+from settings import SettingsV1
+from system_data import SystemDataGatherer
+
+SERVICE_NAME = "major_tom"
+TELEMETRY_VERSION = 1
+
+
+class MajorTom:
+    """
+    Consolidates data from SessionTracker and SystenDataGatherer in one place and sends it to GroundControl
+    """
+
+    last_session = Dict[str, Any]
+
+    def __init__(self) -> None:
+        self.system_data_gatherer = SystemDataGatherer()
+        self.settings_manager = Manager(SERVICE_NAME, SettingsV1)
+        self.sessions_folder = self.settings_manager.config_folder / "sessions"
+        self.sessions_folder.mkdir(parents=True, exist_ok=True)
+        self.session_tracker = SessionTracker(self.sessions_folder)
+        self.settings = self.settings_manager.settings
+        self.remote = self.settings.remote
+
+    def phone_home(self, vehicle_data: Dict[str, Any], system_data: Dict[str, Any]) -> int:
+        """
+        Sends data to GroundControl
+        """
+        blacklist = self.settings_manager.settings.blacklist
+        filtered_vehicle_data = {key: value for key, value in vehicle_data.items() if key not in blacklist}
+        filtered_system_data = {key: value for key, value in system_data.items() if key not in blacklist}
+        try:
+            req = requests.post(
+                f"{self.remote}/v1.0/telemetry/send",
+                json={
+                    "version": TELEMETRY_VERSION,
+                    "vehicle_data": filtered_vehicle_data,
+                    "system_data": filtered_system_data,
+                },
+                timeout=5,
+            )
+        except Exception as e:
+            logger.debug(f"Failed to phone home: {e}")
+            return 523  # 523: Origin Unreachable
+        return req.status_code
+
+    def get_blacklist(self) -> Any:
+        return self.settings_manager.settings.blacklist
+
+    def set_blacklist(self, blacklist: List[str]) -> List[str]:
+        self.settings_manager.settings.blacklist = blacklist
+        self.settings_manager.save()
+        return self.settings_manager.settings.blacklist
+
+    def get_remote(self) -> Any:
+        return self.settings_manager.settings.blacklist
+
+    def set_remote(self, remote: str) -> List[str]:
+        self.settings_manager.settings.remote = remote
+        self.settings_manager.save()
+        return self.settings_manager.settings.remote
+
+    async def run(self) -> None:
+        while True:
+            try:
+                await self.session_tracker.update_session()
+            except Exception as e:
+                logger.warning(f"Failed to update flight session stats: {e}")
+                # walrus?
+            last_session = self.session_tracker.get_oldest_session()
+            system_data = self.system_data_gatherer.get_system_data()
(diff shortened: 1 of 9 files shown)


Williangalvani:
Next: websocket to gather mavlink data

Williangalvani:
I'll still be working on it today, but I think it is ready for a initial review so we can expedite things

Williangalvani:
current (bare-bones) GroundControl is locate here:
https://github.com/bluerobotics/GroundControl

patrickelectric:
@Williangalvani let me know once it's ready for review

core/services/major_tom/system_data.py
```diff
@@ -0,0 +1,61 @@
...
+        return {"serial": serial_number, "model": model}
+
+    def getSystemData(self) -> Dict[str, Any]:
+        data: Dict[str, Any] = {}
+        data["disk"] = self.getDiskUsage()
+        data["memory"] = self.getMemoryUsage()
+        data["usb"] = self.getUsbDevices()
+        data["thermal"] = self.getTemps()
+        data["cpu"] = self.getCpuInfo()
+        return data
```
patrickelectric, line 61: All functions here are static methods

core/services/major_tom/system_data.py
```diff
@@ -0,0 +1,61 @@
...
+                if info:
+                    dinfo = info.groupdict()
+                    dinfo["device"] = f"/dev/bus/usb/{dinfo.pop('bus')}/{dinfo.pop('device')}"
+                    devices.append(dinfo)
+        return devices
+
+    def getTemps(self) -> Any:
+        return psutil.sensors_temperatures()
+
+    def getCpuInfo(self) -> Dict[str, Any]:
```
patrickelectric, line 40: This does not follow the python standard, should use snake case.

core/services/major_tom/system_data.py
```diff
@@ -0,0 +1,61 @@
...
+    def getMemoryUsage(self) -> Any:
+        return psutil.virtual_memory()
+
+    def getUsbDevices(self) -> List[Dict[str, Any]]:
+        device_re = re.compile(
+            r"Bus\s+(?P<bus>\d+)\s+Device\s+(?P<device>\d+).+ID\s(?P<id>\w+:\w+)\s(?P<name>.+)$", re.I
+        )
+        df = subprocess.check_output("lsusb", encoding="utf-8")
+        devices = []
+        for i in df.split("\n"):
```
patrickelectric, line 28: what is `i` ? an index ? an interface ?

core/services/major_tom/system_data.py
```diff
@@ -0,0 +1,61 @@
...
+            if i:
```
patrickelectric, line 29: Why not filtering before using the list ?

core/services/major_tom/system_data.py
```diff
@@ -0,0 +1,61 @@
...
+        return results
+
+    def getMemoryUsage(self) -> Any:
+        return psutil.virtual_memory()
+
+    def getUsbDevices(self) -> List[Dict[str, Any]]:
+        device_re = re.compile(
+            r"Bus\s+(?P<bus>\d+)\s+Device\s+(?P<device>\d+).+ID\s(?P<id>\w+:\w+)\s(?P<name>.+)$", re.I
+        )
+        df = subprocess.check_output("lsusb", encoding="utf-8")
```
patrickelectric, line 26: This can raise.

core/services/major_tom/system_data.py
```diff
@@ -0,0 +1,61 @@
...
+            except Exception as error:
+                print(f"unable to read disk usage: {error}")
+        return results
+
+    def getMemoryUsage(self) -> Any:
+        return psutil.virtual_memory()
+
+    def getUsbDevices(self) -> List[Dict[str, Any]]:
+        device_re = re.compile(
+            r"Bus\s+(?P<bus>\d+)\s+Device\s+(?P<device>\d+).+ID\s(?P<id>\w+:\w+)\s(?P<name>.+)$", re.I
```
patrickelectric, line 24: Can you give an example of what exactly will this match as a comment ?

core/services/major_tom/major_tom.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        blacklist = self.settings_manager.settings.blacklist
+        filtered_vehicle_data = {key: value for key, value in vehicle_data.items() if key not in blacklist}
+        filtered_system_data = {key: value for key, value in system_data.items() if key not in blacklist}
+        try:
+            req = requests.post(
+                f"{self.remote}/v1.0/telemetry/send",
+                json={
+                    "version": TELEMETRY_VERSION,
+                    "vehicle_data": filtered_vehicle_data,
```
patrickelectric, line 44: We should use a structure that we will be sure that will not change or will match with ground control, since both need to have the same API, we need to enforce that somehow.

core/services/major_tom/major_tom.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self) -> None:
+        self.system_data_gatherer = SystemDataGatherer()
+        self.settings_manager = Manager(SERVICE_NAME, SettingsV1)
+        self.sessions_folder = self.settings_manager.config_folder / "sessions"
+        self.sessions_folder.mkdir(parents=True, exist_ok=True)
+        self.session_tracker = SessionTracker(self.sessions_folder)
+        self.settings = self.settings_manager.settings
+        self.remote = self.settings.remote
+
+    def phoneHome(self, vehicle_data: Dict[str, Any], system_data: Dict[str, Any]) -> int:
```
patrickelectric, line 32: We should have a better structure than a dict to ensure that we are not going to break the API shared with ground control.

core/services/major_tom/session_tracker.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def writeSession(self, session: SessionRegistry) -> None:
+        filename = f"{self.sessions_folder}/registry-{session.session_start}.txt"
+        print(f"writing session to file '{filename}'")
+        with open(filename, "w", encoding="utf-8") as f:
+            f.write(json.dumps(asdict(session)))
+
+    def getOldestSession(self) -> Dict[Any, Any]:
+        try:
+            oldest_file = min(glob.glob(f"{self.sessions_folder}/registry-*.txt"), key=os.path.getctime)
```
patrickelectric, line 130: Why not having oldest_file as a class variable ?

DevinNorgarb:
10/10 for PR title @Williangalvani 💯 😄

PR #1527, opened 2023-03-02
core: frontend: components: ParameterEditorDialog: Add check for no options in parameter

### LINKED ISSUE #1526: bug: Parameter editor seems to be broken
### Bug description

Since `BlueOS - 1.1.0-beta.16`, from what I can tell:
- clicking numerical parameter rows fails to open the edit window
- clicking dropdown parameter rows opens the edit window but sometimes doesn't show the current value in the editor until the actual dropdown has been opened
    - it seems to be fine when accessing the editor window via the `Vehicle Setup / PWM Outputs` page
- clicking bitmask parameter rows seems to work fine

### Steps to reproduce

1. Update to beta16 or later
2. Find a numerical parameter
3. Click it to edit
4. Nothing happens
5. Find a parameter with fixed options
6. Click it to edit
7. It might show no value in the dropdown

### Primary pain point(s)

It makes the parameter editor largely unusable for editing.

### Prerequisites

patrickelectric:
Peek 02-03-2023 10-20

Fix #1526

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterEditorDialog.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterEditorDialog.vue
@@ -43,7 +43,7 @@
                   />
                 </template>
                 <v-autocomplete
-                  v-else-if="!custom_input && Object.entries(param?.options).length > 10"
+                  v-else-if="!custom_input && Object.entries(param?.options ?? []).length > 10"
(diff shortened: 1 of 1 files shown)


ES-Alexander (approved):
LGTM - does this also fix the dropdowns sometimes showing as blank, or is that a separate / unrelated issue?

PR #1528, opened 2023-03-02
system-information: firmware: Add up-to-date and latest

patrickelectric:
Peek 02-03-2023 14-40


Williangalvani:
> does it need internet access? have you tried it with the pi offline?


I just tested. works fine.

Williangalvani (approved):

PR #1536, opened 2023-03-08
bootstrap: bootstrap: Revert to factory if getting NotFound error

patrickelectric:
From:
```
/bootstrap/bootstrap.py:175: UserWarning: Error trying to pull an upda
te image: 404 Client Error for http+docker://localhost/v1.42/images/create?tag=patrick_1.0.2&fromImage=bluerobotics%2Fblueos-core: Not Found ("manifest for blu
erobotics/blueos-core:patrick_1.0.2 not found: manifest unknown: manifest unknown")
  warn(f"Error trying to pull an update image: {error}")
```

### DIFF
--- a/bootstrap/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap/bootstrap.py
@@ -167,7 +167,7 @@ class Bootstrapper:
         if not self.image_is_available_locally(image_name, image_version):
             try:
                 self.pull(component_name)
-            except docker.errors.ImageNotFound:
+            except docker.errors.NotFound:
                 warn("Image not found, reverting to default...")
                 self.overwrite_config_file_with_defaults()
                 return False


patrickelectric:
> @patrickelectric After this, if someone deliberately deletes their current running docker, it reverts to the factory, right?

Probably, if fixes the problem when a docker does not exist locally.

Williangalvani:
this is weird. the error mentions line 175, which is in the seconds exception block, for APIError.

maybe the issue is actually that we are not reverting to defaults if start() fails [here](https://github.com/bluerobotics/BlueOS-docker/blob/master/bootstrap/bootstrap/bootstrap.py#L242-L243)?

patrickelectric:
> this is weird. the error mentions line 175, which is in the second exception block, for APIError.
> 
> maybe the issue is actually that we are not reverting to defaults if start() fails [here](https://github.com/bluerobotics/BlueOS-docker/blob/master/bootstrap/bootstrap/bootstrap.py#L242-L243)?

`NotFound` inherits `APIError`, `ImageNotFound` inherits `NotFound`, but when `NotFound` is triggered it goes to `APiError` over `ImageNotfound`.

Example:
```
class APIError(Exception):
    pass

class DockerException(Exception):
    pass

class NotFound(APIError):
    pass

class ImageNotFound(NotFound):
    pass

try:
    raise NotFound()
except ImageNotFound as error:
    print(f"NotFound: {error}")
except APIError as error:
    print(f"APIError: {error}")
    
try:
    raise NotFound()
except (ImageNotFound, NotFound) as error:
    print(f"NotFound: {error}")
except APIError as error:
    print(f"APIError: {error}")
```

Williangalvani (approved):

PR #1551, opened 2023-03-21
Fix bootloader date format

### LINKED ISSUE #1550: Bootloader version is not formatted correctly
Beta 17
image

patrickelectric:
image
Fix #1550

### DIFF
--- a/core/frontend/src/components/system-information/Firmware.vue
+++ b/core/frontend/src/components/system-information/Firmware.vue
@@ -128,11 +128,12 @@ export default Vue.extend({
       const latest = []
 
       for (const line of lines) {
-        if (!line.includes(':')) {
+        const result = line.splitOnce(':')
+        if (result === undefined) {
           continue
         }
 
-        let [key, value] = line.split(':')
+        let [key, value] = result
         key = key.trim()
         value = value.trim()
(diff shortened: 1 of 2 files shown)


rafaellehmkuhl (changes requested):
It still looks very strange.
It starts with date, then theres time, then there's date again (year).

patrickelectric:
> It still looks very strange. It starts with date, then theres time, then there's date again (year).

I'm not sure if I follow, can you be more explicit ?
This fix the current format that's broken, where it only shows until the first `:`

rafaellehmkuhl:
> > It still looks very strange. It starts with date, then theres time, then there's date again (year).
> 
> I'm not sure if I follow, can you be more explicit ? This fix the current format that's broken, where it only shows until the first `:`

Previously, the output was incomplete, but it was just a date, so not confusing.

Right now the output is "day-of-the-week day month hour:minute timezone year (epoch)".

We should have something more straightforward like "day-of-the-week day month yeart timezone hour:minute (epoch)"

patrickelectric:
@rafaellehmkuhl check the latest version

rafaellehmkuhl (approved):

PR #1569, opened 2023-03-30
core: tools: mavlink2rest: Update to latest version

patrickelectric:
This is a on-going test to be sure that API compatibility will not break, there are some tests on mavlink2rest to check some basic stuff, but I'm going to test more on BlueOS and see if something breaks.

Ref: https://github.com/mavlink/rust-mavlink/pull/164


patrickelectric:
this turns all byte arrays into arrays of u8 over arrays of chars, breaking float value and parameters info

PR #1573, opened 2023-04-04
core: frontend: SettingsMenu: Wait for log_folder_size longer

patrickelectric:
There is no reason to wait only 4 seconds, we can wait longer since the calculation time is based on the amount of files that exist.


patrickelectric:
Thanks @rafaellehmkuhl

PR #1580, opened 2023-04-06
core: ardupilotmanager: fix support for autopilots on non-usb serial ports

Williangalvani:
This works in my head. needs testing.

### DIFF
--- a/core/services/ardupilot_manager/flight_controller_detector/Detector.py
+++ b/core/services/ardupilot_manager/flight_controller_detector/Detector.py
@@ -69,7 +69,9 @@ class Detector:
         unique_serial_devices: List[SysFS] = []
         for port in sorted_serial_ports:
             # usb_device_path property will be the same for two serial connections using the same USB port
-            if port.usb_device_path not in [device.usb_device_path for device in unique_serial_devices]:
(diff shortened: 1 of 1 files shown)


ES-Alexander:
The code here seems fine to me, although we may want to add some logging for which boards have been detected.

From a cleanliness standpoint, `Detector.detect_serial_flight_controllers` should be a `classmethod` not a `staticmethod`, but perhaps that's out of scope for this PR since it'd change the existing code?

Williangalvani:
so... this will not work as we don't actually try to talk mavlink to identify the boards. (we only check for usb pid/vid)
perhaps we really need an override option?

ES-Alexander:
> perhaps we really need an override option?

Not sure how annoying it'd be to implement, but it presumably would be quite useful for people if there was a config file of 'extras' where each line just specifies a device path or something to try to connect to?

patrickelectric:
We could just add a function here that sets one
https://github.com/bluerobotics/BlueOS-docker/blob/master/core/services/ardupilot_manager/flight_controller_detector/Detector.py#L62

patrickelectric:

Issue #1466, opened 2023-02-13
bug: VEHICLE id hardcoded to 1 (one)


mat2022:
### Bug description

Hey all,
I have to deploy a setup involving multiple vehicle. I am usingPi4/Navigator/BlueOS/custom ArduRover.
I think I found a “bug” in blueOS:
when you define `sysid_thismav` to something different from one (as I understand has to be the case in multi-vehicle setup), some strange things happen:
- If you use BlueOS 1.1.0 beta 14, you get `VEHICLE_TYPE_FETCH_FAIL` and `UPDATE_TIME_FAIL` notification all the time.
- If you use BlueOS 1.1.0 beta 11, you get `VEHICLE_TYPE_FETCH_FAIL` and `FIRM_FETCH_FAIL` notification all the time.
- If you use the latest (older) stable release of BlueOS which is 1.0.1, you get no notification at all (great!).
BUT unfortunately, you still get the heartbeat lost (red broken heart) whereas there are plenty of heartbeat messages exchanged.
- In all cases, as long as you define `sysid_thismav` to something not equal to one, you get the red broken heart, “heartbeat lost”.

I could find something in `core/frontend/src/components/health/HealthTrayMenu.vue` line `208`:
what happens if `system_id` is not one ? like in a multi-vehicle setup ?
```.ts
mounted() {
    mavlink2rest.startListening('HEARTBEAT').setCallback((message) => {
      if (message?.header.system_id !== 1 || message?.header.component_id !== 1) {// <-- this is line 208
        return
      }
```

more specifically this `system_id` is used in different places:
command: 
```shell
grep -inr system_id
```
gives:
```shell
1 core/start-blueos-core:8:MAV_SYSTEM_ID=1
2 core/start-blueos-core:30:    'mavlink2rest',"mavlink2rest --connect=udpin:127.0.0.1:14000 --server 0.0.0.0:6040 --system-id $MAV_SYSTEM_ID --component-id $MAV_COMPONENT_ID_ONBOARD_COMPUTER4"
3 core/frontend/src/types/mavlink.ts:4:    system_id: number,
4 core/frontend/src/components/health/HealthTrayMenu.vue:155:      if (message?.header.system_id !== 1 || message?.header.component_id !== 1) {
5 core/frontend/src/libs/MAVLink2Rest/index.ts:101:        system_id: 255,
[...]

PR #1581, opened 2023-04-06
Allow using system ids other than 1

Williangalvani:
This goes on the assumption that the vehicle is only one system with multiple components.
it relies on mavlink2rest to detect the most recent message received and locks on to that ID

Relevant to #1466

### DIFF
--- a/core/frontend/src/components/health/HealthTrayMenu.vue
+++ b/core/frontend/src/components/health/HealthTrayMenu.vue
@@ -137,6 +137,7 @@
 import Vue from 'vue'
 
 import mavlink2rest from '@/libs/MAVLink2Rest'
+import { MavType } from '@/libs/MAVLink2Rest/mavlink2rest-ts/messages/mavlink2rest-enum'
 import mavlink from '@/store/mavlink'
 import system_information from '@/store/system-information'
 import { RaspberryEventType } from '@/types/system-information/platform'
@@ -205,7 +206,7 @@ export default Vue.extend({
   },
   mounted() {
     mavlink2rest.startListening('HEARTBEAT').setCallback((message) => {
-      if (message?.header.system_id !== 1 || message?.header.component_id !== 1) {
+      if (!this.is_vehicle(message?.message.mavtype.type) || message?.header.component_id !== 1) {
         return
       }
 
@@ -215,6 +216,28 @@ export default Vue.extend({
   methods: {
     heartbeat_age(): number {
       return new Date().valueOf() - this.last_heartbeat_date.getTime()
+    },
+    is_vehicle(type: string) {
+      return [
+        MavType.MAV_TYPE_FIXED_WING,
+        MavType.MAV_TYPE_VTOL_DUOROTOR,
+        MavType.MAV_TYPE_VTOL_QUADROTOR,
+        MavType.MAV_TYPE_VTOL_TILTROTOR,
+        MavType.MAV_TYPE_VTOL_RESERVED4,
+        MavType.MAV_TYPE_VTOL_RESERVED5,
+        MavType.MAV_TYPE_QUADROTOR,
+        MavType.MAV_TYPE_COAXIAL,
+        MavType.MAV_TYPE_HELICOPTER,
+        MavType.MAV_TYPE_HEXAROTOR,
+        MavType.MAV_TYPE_OCTOROTOR,
+        MavType.MAV_TYPE_TRICOPTER,
+        MavType.MAV_TYPE_GROUND_ROVER,
+        MavType.MAV_TYPE_SURFACE_BOAT,
+        MavType.MAV_TYPE_SUBMARINE,
+        // The next two ones are not defined?
+        'MAV_TYPE_VTOL_FIXEDROTOR',
+        'MAV_TYPE_VTOL_TAILSITTER',
+      ].includes(type)
     },
   },
 })
--- a/core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
+++ b/core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
@@ -1,16 +1,20 @@
 import asyncio
 import json
 import os
+import re
 import time
+from datetime import datetime, timezone
 from typing import Any, Dict, Optional
 
 import aiohttp
+from loguru import logger
 
 from commonwealth.mavlink_comm.exceptions import (
     FetchUpdatedMessageFail,
     MavlinkMessageReceiveFail,
     MavlinkMessageSendFail,
 )
+from commonwealth.mavlink_comm.typedefs import MavlinkVehicleType
 
 
 class MavlinkMessenger:
@@ -21,6 +25,7 @@ class MavlinkMessenger:
         self.m2r_address = "localhost:6040"
 
     def set_system_id(self, system_id: int) -> None:
+        logger.info(f"system_id set to: {system_id}")
         self.system_id = system_id
 
     def set_component_id(self, component_id: int) -> None:
@@ -38,10 +43,22 @@ class MavlinkMessenger:
     def m2r_rest_url(self) -> str:
         return f"http://{self.m2r_address}/mavlink"
 
+    async def get_all_mavlink(self) -> Any:
+        request_timeout = 1.0
+        async with aiohttp.ClientSession() as session:
+            try:
+                async with session.get(self.m2r_rest_url, timeout=request_timeout) as response:
+                    if not response.status == 200:
+                        raise MavlinkMessageReceiveFail(f"Received status code of {response.status}.")
+                    message = await response.json()
+            except asyncio.exceptions.TimeoutError as error:
+                raise MavlinkMessageReceiveFail(f"Request timed out after {request_timeout} second.") from error
+        return message
+
     async def get_mavlink_message(
-        self, message_name: Optional[str] = None, vehicle: Optional[int] = 1, component: Optional[int] = 1
+        self, message_name: Optional[str] = None, vehicle: Optional[int] = None, component: Optional[int] = 1
     ) -> Any:
-        request_url = f"{self.m2r_rest_url}/vehicles/{vehicle}/components/{component}/messages"
+        request_url = f"{self.m2r_rest_url}/vehicles/{vehicle or self.system_id}/components/{component}/messages"
         if message_name:
             request_url += f"/{message_name.upper()}"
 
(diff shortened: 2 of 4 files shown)


core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -51,28 +67,58 @@ async def get_mavlink_message(
...
+        for vehicle_id, vehicle in json_data["vehicles"].items():
+            # skip vehicles with id 2 (mavlink-router?) and 255 (common GCS)
+            if vehicle_id in {"2", "255"}:
+                continue
+
+            for component in vehicle["components"].values():
+                if "HEARTBEAT" in component["messages"]:
+                    message = component["messages"]["HEARTBEAT"]
+                    # we are looking for vehicles, not GCSs
+                    if message["mavtype"]["type"] == "MAV_TYPE_GCS":
```
ES-Alexander, line 93: Gimbal, ADSB, and 30+ [MAV_TYPE](http://mavlink.io/en/messages/common.html#MAV_TYPE)s should probably also be excluded, and likely Onboard Computer as well.

Williangalvani: yep, updated

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -21,6 +25,7 @@ def __init__(self) -> None:
         self.m2r_address = "localhost:6040"
 
     def set_system_id(self, system_id: int) -> None:
+        logger.info(f"system_id detected as: {system_id}")
```
ES-Alexander, line 28:
```suggestion
        logger.info(f"system_id set to: {system_id}")
```
This is a setter function - there's no obvious context in this function as to why the value being set has been "detected" rather than just set as a user preference.

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -51,28 +68,58 @@ async def get_mavlink_message(
...
+                    if not MavlinkVehicleType(message["mavtype"]["type"]).is_actually_a_vehicle():
+                        continue
+                    last_update_str = message["status"]["time"]["last_update"]
+                    last_update_str = re.sub(r"(\.\d{6})\d+", r"\1", last_update_str)
+                    last_update = datetime.fromisoformat(last_update_str)
+
+                    if last_update > most_recent_timestamp:
+                        most_recent_timestamp = last_update
+                        most_recent_vehicle_id = vehicle_id
+        return int(most_recent_vehicle_id or 1)
```
ES-Alexander, line 103:
```suggestion
        if most_recent_vehicle_id:
            logger.debug(f'{most_recent_vehicle_id = } (detected)')
            return int(most_recent_vehicle_id)
        else:
            logger.debug('no vehicle ID detected - using default (1)')
            return 1
```
Non-essential, but could come in handy at some point.

core/frontend/src/components/health/HealthTrayMenu.vue
```diff
@@ -205,7 +205,7 @@ export default Vue.extend({
   },
   mounted() {
     mavlink2rest.startListening('HEARTBEAT').setCallback((message) => {
-      if (message?.header.system_id !== 1 || message?.header.component_id !== 1) {
+      if (message?.message.mavtype.type === 'MAV_TYPE_GCS' || message?.header.component_id !== 1) {
```
ES-Alexander, line 208: I'm not certain what this `mounted` function is for. Is it supposed to indicate that the vehicle is connected to a topside computer, or is it supposed to show that an autopilot is connected and running well, or something else?

Does the current non-zero component ID check mean that being connected to the onboard camera manager counts as being mounted?

ES-Alexander (commented):
Mostly looks good to me, although I haven't tested it.

I've left a couple of comments / suggestions :-)

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -51,28 +68,58 @@ async def get_mavlink_message(
                 async with session.get(request_url, timeout=request_timeout) as response:
                     if not response.status == 200:
                         raise MavlinkMessageReceiveFail(f"Received status code of {response.status}.")
+                    # if message is "None", try re-detecting systemid
+                    if await response.text() == "None":
```
patrickelectric, line 72: I'm worried about this. Correct me if I'm wrong, but looking at the core it appears that the user does not have control over the `system id` that it gona be used with this detection logic.

If that's the case, this can result in problems, where the user restart BlueOS with multiple vehicle on the network, or it going to connect in a vehicle without knowing wich one, since it can change every time. The user should enforce the `system id`.

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -51,28 +68,58 @@ async def get_mavlink_message(
...
+        most_recent_vehicle_id = None
+        for vehicle_id, vehicle in json_data["vehicles"].items():
+            # skip vehicles with id 2 (mavlink-router?) and 255 (common GCS)
+            if vehicle_id in {"2", "255"}:
+                continue
+
+            for component in vehicle["components"].values():
+                if "HEARTBEAT" in component["messages"]:
+                    message = component["messages"]["HEARTBEAT"]
+                    # we are looking for vehicles, not GCSs
```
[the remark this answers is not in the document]
patrickelectric:
```suggestion
                    # Filter everything that is not a vehicle
```

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -51,28 +68,58 @@ async def get_mavlink_message(
...
                 raise MavlinkMessageReceiveFail(f"Request timed out after {request_timeout} second.") from error
 
         return message
 
+    async def get_most_recent_vehicle_id(self) -> int:
+        json_data = await self.get_all_mavlink()
+        most_recent_timestamp = datetime.min.replace(tzinfo=timezone.utc)
+        most_recent_vehicle_id = None
+        for vehicle_id, vehicle in json_data["vehicles"].items():
+            # skip vehicles with id 2 (mavlink-router?) and 255 (common GCS)
```
patrickelectric, line 86: What ? How mavlink-router is using system-id 2 ?

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
```diff
@@ -51,28 +68,58 @@ async def get_mavlink_message(
...
+                continue
+
+            for component in vehicle["components"].values():
+                if "HEARTBEAT" in component["messages"]:
+                    message = component["messages"]["HEARTBEAT"]
+                    # we are looking for vehicles, not GCSs
+                    if not MavlinkVehicleType(message["mavtype"]["type"]).is_actually_a_vehicle():
+                        continue
+                    last_update_str = message["status"]["time"]["last_update"]
+                    last_update_str = re.sub(r"(\.\d{6})\d+", r"\1", last_update_str)
```
patrickelectric, line 97: Can you add a comment about what the regex is doing ? This is not clear.

patrickelectric (changes requested):

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
Williangalvani: very good point, the detected thing should be where it is actually detected.

core/frontend/src/components/health/HealthTrayMenu.vue
Williangalvani: this makes the Heartbeat icon beat. This change makes it beat with anything sending a heartbeat that is not a GCS.
We should filter for all valid types here, too.

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
Williangalvani: The user has no control over it, but it is not enough to have multiple vehicles in the network, someone will need to actually create a misconfigured mavlink endpoint pointing to this blueos instance instead of the topside.

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
Williangalvani: honestly I'm not sure of who this is, but since logging is involved, I assumed mavlink-router:
image

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
Williangalvani: I'm leaning to this approach as it handles the user simply changing SYSID_THISMAV graciously.

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
Williangalvani: since the code is checking for vehicle type and heartbeat messages, that code was redundant, so I removed it.
I'd still like to know who ID 2 is, though

patrickelectric: https://github.com/mavlink-router/mavlink-router/blob/3b48da112545e1683c776c6131ff410425e8a978/src/binlog.cpp#L35
https://github.com/mavlink-router/mavlink-router/blob/3b48da112545e1683c776c6131ff410425e8a978/src/logendpoint.h#L30

patrickelectric: Please add a comment explaining the findings

core/frontend/src/components/health/HealthTrayMenu.vue
ES-Alexander: Ahh, I think I misunderstood the logic - your description clarified what it's actually doing, thanks :-)

That said, I don't think the heartbeat icon should beat if only the camera manager is present, so it should only be beating if there's a valid autopilot connected. Presumably that's what you meant by filtering for valid types?

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
patrickelectric: Ok, we can continue as it is so.

patrickelectric (changes requested):

patrickelectric:
There are some points where the `system id` is not defined globally.
Other extensions will fail, services that are not based on python may have the same logic, but identify different `system ids`.
What should we do with mavlink2rest, that needs the `system id` when starting from the command line: https://github.com/bluerobotics/BlueOS-docker/blob/537aa6d9f840e5af7b3489d872ef03e691056d98/core/start-blueos-core#L39

Isn't safer and simpler to use the `MAV_SYSTEM_ID` and use bootstrap to define such variable before starting all dockers ? With that there is a defined and global value for it where the user can configure the system. We could also do that once the parameter `SYSID_THISMAV` is changed by the user in BlueOS.

Williangalvani:
> Isn't safer and simpler to use the `MAV_SYSTEM_ID` and use bootstrap to define such variable before starting all dockers ? With that there is a defined and global value for it where the user can configure the system. We could also do that once the parameter `SYSID_THISMAV` is changed by the user in BlueOS.

Things get messy as the parameter can be changed in multiple places, such as QGC.
But I'm thinking of actually doing both:

- Have a interface where the user can change the environment variable and SYSID_THISMAV at the same time. This will be a reboot required for blueos due to using environment variables
- At the same time, keep the automatic detection, additionally pushing a notification in case it gets triggered, so the user can fix things. Note that this only gets triggered after something goes on with getting data from the configured MAV_SYSTEM_ID

core/libs/commonwealth/commonwealth/mavlink_comm/MavlinkComm.py
Williangalvani: I removed the code, but yeah 🤦

patrickelectric: Ref: https://github.com/mavlink-router/mavlink-router/issues/392

core/frontend/src/components/health/HealthTrayMenu.vue
Williangalvani: done

Williangalvani:
I'd like to get this in as is, then as I get the dialog in #1582, I intend to introduce some other vehicle-related configuration in there, as the expected system id

patrickelectric (approved):

PR #1585, opened 2023-04-07
core: create Bag of Holding

### LINKED ISSUE #1583: We need a service for storing small, unrelated pieces of information in the backend.
Or do we?

I'm currently missing it for storing the current image for both logo and vehicle for the sidebar:

image

Williangalvani:
This allow storing key/value pairs as json in the backend, which get stored into a json file.

fix #1583

*No AIs were harmed in writing this code

### DIFF
--- a/core/services/bag_of_holding/main.py
+++ b/core/services/bag_of_holding/main.py
@@ -1,0 +1,92 @@
+#! /usr/bin/env python3
+import json
+import logging
+from pathlib import Path
+from typing import Any, Dict
+
+import appdirs
+import dpath
+import uvicorn
+from commonwealth.utils.apis import GenericErrorHandlingRoute
+from commonwealth.utils.logs import InterceptHandler, init_logger
+from fastapi import Body, FastAPI, HTTPException
+from fastapi.responses import HTMLResponse, JSONResponse
+from fastapi_versioning import VersionedFastAPI, version
+from loguru import logger
+from pydantic import BaseModel
+
+SERVICE_NAME = "bag_of_holding"
+FILE_PATH = Path(appdirs.user_config_dir("bag of holding", "db.json"))
+
+logging.basicConfig(handlers=[InterceptHandler()], level=0)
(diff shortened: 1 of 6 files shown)


core/services/bag_of_holding/main.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@app.post("/set")
+@version(1, 0)
+async def set_value(item: KeyValue) -> JSONResponse:
+    db = read_db()
+    db[item.key] = item.value
+    write_db(db)
+    return JSONResponse(content={"status": "success"})
+
```
patrickelectric, line 70: I would recommend to allow reading and setting values from nested paths:

```python
# Check https://pypi.org/project/dpath/

@app.post("/write/{path}")
@version(1, 0)
async def write_data(path: str, data: DataModel):
    current_data = read_db()
    dpath.util.new(current_data, path, data.data)
    write_db(current_data)
    return JSONResponse(content={"status": "success"})

@version(1, 0)
@app.get("/get/{path}")
async def read_data(path: str):
    current_data = read_db()
    try:
        result = dpath.util.get(current_data, path)
    except KeyError:
        raise HTTPException(status_code=400, detail="Invalid path")
    return JSONResponse(content={"value": db[key]})
```

Williangalvani: great idea!

Williangalvani:
@ES-Alexander where should docs for this go? Do we have docs for the individual services?

patrickelectric (approved):

ES-Alexander:
> where should docs for this go? 
The docs should likely be in the Available Services section of the Advanced Usage page. It may also be worth mentioning / linking to them in the extensions docs as things people might want to use to help with their integrations.

> Do we have docs for the individual services?

We don't currently have docs for individual services, but I think it would be good if we add some for at least the services that developers are likely to want to make use of in their extensions and the like. At this stage that seems like specifically Commander and Bag of Holding, and maybe Beacon?

PR #1582, opened 2023-04-07
Allow setting vehicle name and mDNS from the frontend

Williangalvani:
https://user-images.githubusercontent.com/4013804/230667364-2208bdcf-2c05-4b37-99e1-fa76f450e575.mov

requires #1585

PR #1589, opened 2023-04-08
Add wizard

patrickelectric:
Needs ~~#1582~~ and #1585
Peek 08-04-2023 16-01

Add wizard setup that opens on the first time that BlueOS shows
Sets vehicle name and mdns name to be used

### DIFF
--- a/core/frontend/src/components/wizard/Wizard.vue
+++ b/core/frontend/src/components/wizard/Wizard.vue
@@ -1,0 +1,207 @@
+<template>
+  <v-dialog
+    v-model="should_open"
+    transition="dialog-top-transition"
+    width="600"
+    height="400"
+    persistent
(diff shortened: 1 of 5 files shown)


core/frontend/src/components/wizard/Wizard.vue
```diff
@@ -0,0 +1,209 @@
...
+              + `${error.message ?? error.response?.data}.`
+          })
+
+        return
+      }
+      this.configuration_failed = true
+      this.wait_configuration = false
+    },
+    setupROV() {
+      console.log('ROV setup')
```
Williangalvani, line 186: lost comment

Williangalvani (approved):
looks good, I guess we can get rid of that console.log though

PR #1595, opened 2023-04-09
core: frontend: components: DHCPServerDialog: Fix if check

patrickelectric:
If was checking for something that does not exist, add correct computed method to check status


Williangalvani (approved):
was CI no catching this? it even shows up red in vscode here, weird.

PR #1601, opened 2023-04-10
Bag: Add store

patrickelectric:


joaoantoniocardoso (approved):
I'm not much into typescript async, but this refactoring LGTM

PR #1602, opened 2023-04-10
Bag: Improve access

patrickelectric:


joaoantoniocardoso:
I'm a bit lost here. What/how is the `payload`?

patrickelectric:
> I'm a bit lost here. What/how is the `payload`?

payload is a json

core/services/bag_of_holding/main.py
```diff
@@ -55,17 +55,24 @@ def write_db(data: Dict[str, Any]) -> None:
...
 @version(1, 0)
 async def write_data(path: str, payload: dict[str, Any] = Body(...)) -> JSONResponse:
+    logger.debug(f'Write path: {path}, {json.dumps(payload)}')
     current_data = read_db()
-    dpath.new(current_data, path, payload, separator=".")
+    if path == 'root':
+        current_data = payload
+    else:
+        dpath.new(current_data, path, payload, separator=".")
     write_db(current_data)
```
joaoantoniocardoso, line 64: So.. Here we are replacing the entire db without checking if it's structure is correct. I can't evaluate the implications because I'm not aware of the rest of the code, I believe you judgement

joaoantoniocardoso: like.. a `write_data("/root")` call would erase the db

patrickelectric: I have created an endpoint to only do that, making it more explicit

core/services/bag_of_holding/main.py
```diff
@@ -52,21 +54,30 @@ def write_db(data: Dict[str, Any]) -> None:
...
 
-@app.get("/get/{path}")
+@app.get("/get/{path:path}")
 @version(1, 0)
 async def read_data(path: str) -> JSONResponse:
+    logger.debug(f"Get path: {path}")
     current_data = read_db()
     try:
-        result = dpath.get(current_data, path, separator=".")
+        result = dpath.util.search(current_data, path)
```
Williangalvani, line 80: I'm not sure I like this, what is the purpose of using search? I think if we have something like
```
{
  "potato: {
    "a": 1,
    "b": 2
    }
}

and we ask for "potato/a" we should just get "1", instead I'm getting:

```
{
  "potato: {
    "a": 1
    }
}

```

patrickelectric: Check now

image

PR #1609, opened 2023-04-11
Commander: create store

patrickelectric:


core/frontend/src/store/commander.ts
```diff
@@ -0,0 +1,185 @@
...
+        const message = `Could not set time: ${error.response?.data ?? error.message}.`
+        notifier.pushError('COMMANDER_SET_TIME_FAIL', message, true)
+      })
+  }
+
+  async shutdown(shutdownType: ShutdownType): Promise<void> {
+    return back_axios({
+      method: 'post',
+      url: `${this.API_URL}/shutdown`,
+      timeout: 5000,
```
joaoantoniocardoso, line 64: Note: was `2000`

core/frontend/src/store/commander.ts
```diff
@@ -0,0 +1,185 @@
...
+        const message = `Could not execute shutdown: ${error.message ?? error.response?.data}.`
+        notifier.pushError('COMMANDER_SHUTDOWN_FAIL', message, true)
+      })
+  }
+
+  async getRaspiCameraLegacy(): Promise<boolean | undefined> {
+    return back_axios({
+      method: 'get',
+      url: `${this.API_URL}/raspi_config/camera_legacy`,
+      timeout: 5000,
```
joaoantoniocardoso, line 90: Note: was `3000`

core/frontend/src/store/commander.ts
```diff
@@ -0,0 +1,185 @@
...
+        notifier.pushError('COMMANDER_GET_CAMERA_LEGACY_FAIL', message, true)
+        return undefined
+      })
+  }
+
+  async setRaspiCameraLegacy(enable = true): Promise<boolean> {
+    return back_axios({
+      method: 'post',
+      url: `${this.API_URL}/raspi_config/camera_legacy`,
+      timeout: 5000,
```
joaoantoniocardoso, line 108: Note: was `3000`

joaoantoniocardoso (approved):
Cool, gained awareness with additional push notifications, and more timeout, which doesn't hurt.

core/frontend/src/store/commander.ts
patrickelectric: The time was updated to make it more robust

Williangalvani (approved):
This doesnt quite feel like a store, it is more a lib? 

> At the center of every Vuex application is the store. A "store" is basically a container that holds your application state. There are two things that make a Vuex store different from a plain global object:

We don't quite have a state in there.

But this is huge improvement anyway, so feel free to merge as is =]

PR #1610, opened 2023-04-11
Do not show Wizard when BlueOS is down

patrickelectric:


core/frontend/src/components/wizard/Wizard.vue
```diff
@@ -210,8 +210,8 @@ export default Vue.extend({
     },
   },
   async mounted() {
-    const wizard = await bag.getData('wizard')
-    if (wizard?.version !== WIZARD_VERSION) {
+    const wizard = await bag.getData('wizard2')
+    if (wizard !== undefined && wizard?.version !== WIZARD_VERSION) {
```
Williangalvani, line 214: what is the logic here? is undefined for "unable to reach backend"? I think it deserves at least a comment

core/frontend/src/components/wizard/Wizard.vue
```diff
@@ -210,8 +210,8 @@ export default Vue.extend({
     },
   },
   async mounted() {
-    const wizard = await bag.getData('wizard')
-    if (wizard?.version !== WIZARD_VERSION) {
+    const wizard = await bag.getData('wizard2')
```
Williangalvani, line 213: why changing to "wizard2"?

Williangalvani (changes requested):

core/frontend/src/components/wizard/Wizard.vue
patrickelectric: Oh, that was code to test the feature

Williangalvani:
fixes #1620 ?

PR #1619, opened 2023-04-14
Add initial https support 

### LINKED ISSUE #1618: BlueOS lacks HTTPS, which prevents several browser APIs to be used
Without a secure context, features like recording the screen and accessing USB/HID devices, among others, do not work.

patrickelectric:
Fix #1618

image

A palliative solution until we create a better way


patrickelectric:
@Williangalvani if we end up using extension paths over extension TCP ports in the URL, will it fix the problem ?
BTW, your diff is based over this PR or standalone ?

patrickelectric:

Issue #1624, opened 2023-04-17
We should support 3rd party ArduPilot binaries


patrickelectric:
We may have our own manifest version of ArduPilot manifest file format. 
Or deal with it with userdata.

PR #1625, opened 2023-04-17
Support non-navigator linux boards (initially Argonot)

### LINKED ISSUE #1623: ardupilot-manager: The service is currently hardcoded to support only Navigators when it comes to Linux boards

rafaellehmkuhl:
With this small change, we can support running Ardupilot Linux-based firmwares for boards other than the Navigator.

This was motivated to support the Argonot system, from SymbyTech (a request from @DevinNorgarb).

The only downside for now is that when there's no firmware installed for that board, we will use the one we have stored for Navigator.

The PR is in draft mode until we confirm the sensors used to identify this board.

Fix #1623.

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -15,6 +15,7 @@ from loguru import logger
 from exceptions import (
     ArdupilotProcessKillFail,
     EndpointAlreadyExists,
+    NoDefaultFirmwareAvailable,
     NoPreferredBoardSet,
 )
 from firmware.FirmwareManagement import FirmwareManager
@@ -173,6 +174,10 @@ class ArduPilotManager(metaclass=Singleton):
                 self.firmware_manager.install_firmware_from_file(
                     pathlib.Path("/root/blueos-files/ardupilot-manager/default/ardupilot_navigator"),
                     board,
+                )
+            else:
+                raise NoDefaultFirmwareAvailable(
+                    f"No firmware installed for '{board.platform}' and no default firmware available. Please install the firmware manually."
                 )
 
         firmware_path = self.firmware_manager.firmware_path(self._current_board.platform)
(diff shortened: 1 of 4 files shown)


patrickelectric:
I believe that would be more future proof to do what we currently do by default, and if there is a binary under userdata, run the binary if its valid.

rafaellehmkuhl:
> I believe that would be more future proof to do what we currently do by default, and if there is a binary under userdata, run the binary if its valid.

But in this case, we wouldn't offer the firmware download/update nor the board identification features to those.

rafaellehmkuhl:
> > > I believe that would be more future proof to do what we currently do by default, and if there is a binary under userdata, run the binary if its valid.
> > 
> > 
> > But in this case, we wouldn't offer the firmware download/update nor the board identification features to those.
> 
> That'll be more related to: #1624, on the idea of the manifest file

Didn't quite get it. Why not use the ardupilot one if it exists?

rafaellehmkuhl:
> > > > > I believe that would be more future proof to do what we currently do by default, and if there is a binary under userdata, run the binary if its valid.
> > > > 
> > > > 
> > > > But in this case, we wouldn't offer the firmware download/update nor the board identification features to those.
> > > 
> > > 
> > > That'll be more related to: #1624, on the idea of the manifest file
> > 
> > 
> > Didn't quite get it. Why not use the ardupilot one if it exists?
> 
> 1. 3rd party companies when releasing a product will rarely do a firmware update, that's the most common use case.
> 2. If we rely only in ArduPilot manifest we and others will be stuck with it, the code review process, the release cycle and more. Custom binaries will be heavily expensive to maintain over it, and new binaries may not be supported by the 3rd party companies or validated, making the release cycle out of the companies to the hand of the ArduPilot organization.
> 3. Having our own manifest will help 3rd party companies to release their own binaries and have total control over it, including providing not open source versions of ArduPilot firmwares.

I agree we should support this alternative. I just don't get why not to do both.

patrickelectric:
@DevinNorgarb was it tested ?

PR #1511, opened 2023-02-22
Add custom banner for BlueOS

patrickelectric:
In discussion

[userdata.zip](https://github.com/bluerobotics/BlueOS-docker/files/10803916/userdata.zip)

image

PR #1636, opened 2023-04-21
Improve vehicle sidebar interface

patrickelectric:
This is based over @rjehangir proposal:
https://github.com/bluerobotics/BlueOS-docker/pull/1511#issuecomment-1458864096

**Current:**
image
image

**New:**
image

image

Ref suggestion:
image

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -53,7 +53,7 @@
       <v-divider />
       <v-container
         elevation="0"
-        class="d-flex justify-center align-center"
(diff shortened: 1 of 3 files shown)


Williangalvani (approved):
I don't quite like the blue color when in dark mode:

Screenshot 2023-04-21 at 13 36 47

The vehicle icon also feels a bit small to me, but this was the initial concept, so we can merge it and discuss this further later on.

rjehangir:
I agree on both of those points. The vehicle icon should be a little larger and the text should be white.

PR #1638, opened 2023-04-21
Allow user to download a png image with the boundaries of the 3D model

patrickelectric:
Helpful to draw schemes, set custom vehicle image on the UI and more.

image

### DIFF
--- a/core/frontend/src/components/vehiclesetup/viewers/GenericViewer.vue
+++ b/core/frontend/src/components/vehiclesetup/viewers/GenericViewer.vue
@@ -2,6 +2,7 @@
   <div>
     <model-viewer
       v-if="model_path"
+      id="modelviewer"
       ref="modelviewer"
       :src="model_override_path || model_path"
       :auto-rotate="autorotate"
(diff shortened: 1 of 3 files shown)


core/frontend/src/components/vehiclesetup/viewers/GenericViewer.vue
```diff
@@ -30,6 +31,18 @@
...
+
+      <v-btn
+        id="image-download-btn"
+        class="mt-6"
+        fab
+        x-small
+        @click="download"
+      >
+        <v-icon>
+          mdi-content-save
```
Williangalvani, line 43: maybe mdi-camera instead?
Screenshot 2023-04-21 at 17 25 01

joaoantoniocardoso (approved):

Williangalvani (approved):
I still prefer the camera icon, but nbd

ES-Alexander:
> I still prefer the camera icon, but nbd

I'm with @Williangalvani on this - the camera icon makes it more clear that you're saving an image, rather than the whole model.

patrickelectric:
documented in: https://github.com/bluerobotics/ardusub-zola/pull/92

PR #1669, opened 2023-05-05
frontend: Use round edges and color gradient based on Elisa feedback

patrickelectric:
image

image

image

image

image

image


ES-Alexander (approved):
The background looks great :-)

Per our discussion, the service cards are acceptable, and not worth putting heaps of time in since we're planning to remove them soonish anyway once we've got a real dashboard.

Nice work! :-)

PR #1689, opened 2023-05-19
core: frontend: views: VehicleSetupView: Improve code and visual

patrickelectric:
now:
image

before:
image

### DIFF
--- a/core/frontend/src/views/VehicleSetupView.vue
+++ b/core/frontend/src/views/VehicleSetupView.vue
@@ -1,26 +1,29 @@
 <template>
   <v-container fluid>
     <v-tabs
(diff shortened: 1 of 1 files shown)


ES-Alexander:
All the other pages have the blue background coming through instead of having a flat base colour - does it perhaps make more sense to make the container for the peripherals at the bottom transparent, and then just change the styling of the tabs bar at the top to better fit with the rest of the UI elements?

Also, the Extensions Manager page also has tabs, so will need to be updated to have consistent styling with whatever we do for the Vehicle Setup view :-)

ES-Alexander (approved):

PR #1691, opened 2023-05-19
core: frontend: componentes: PingInfo: Do not show if there is no device

patrickelectric:
now:
image

before:
image

### DIFF
--- a/core/frontend/src/components/vehiclesetup/overview/PingInfo.vue
+++ b/core/frontend/src/components/vehiclesetup/overview/PingInfo.vue
@@ -1,5 +1,8 @@
 <template>
-  <v-card class="ma-2 pa-2 pinginfo">
+  <v-card
+    v-if="pings.length"
+    class="ma-2 pa-2 pinginfo"
+  >
     <v-card-title class="justify-center">
       Ping Sensors
     </v-card-title>


ES-Alexander:
This is inconsistently applied logic, so we should discuss what we want here. Are Pings special compared to things like lights or the gripper (I personally think Ping stuff belongs in an extension, so I could see an argument for "yes" on that grounds at least), or should we generally hide things if there aren't any of them available?

I think my favoured approach would be:
1. if the software has the capability to detect something, it should have a box, with a clear indication if the thing is not detected/configured (à la "Not configured" for the current gripper display)
    - I consider "capability" to include
        - a configuration option (or combination) that tends to imply the usage of a particular device (like ArduSub params for lights1/2 and the gripper)
        - a built in service that can actually detect a device (like for the camera manager, and currently for Ping devices)
        - an installed device-integration extension (like for a DVL/USBL, and perhaps in future for Ping devices)
   - This then pulls double duty as showing both
       1. what is currently connected, as well as
       2. what the vehicle thinks it's capable of handling
2. if there is no capability to detect it / assume it's connected, don't show a box for it

patrickelectric:
> Pings special compared to things like lights or the gripper

Yes, ping has its own service to be detected. 

My point is that, things that are automatically detected and integrated in BlueOS could be dynamically show or not, like Ping here. Meanwhile, things that need user intervention on the vehicle/or are default, should show as "Not configured".

joaoantoniocardoso:
Orrrr.... instead of hiding it, we could use that space to subtle advertise it, like `Not installed, click >here< to see how to get one`, linking to our store :smile:

Williangalvani:
I agree it is better to say something along the lines of "No devices detected" rather than omitting the panel completely. Omitting it completely could lead users to think that "ping support is not installed", which is very relevant as there are thoughts of turning it into an extension.

PR #1694, opened 2023-05-22
core: services: helper: Ping BlueOS for simple telemetry system

patrickelectric:
We don't have time to have a full telemetry system for now
But that should be enough for us to have an idea


voorloopnul:
:heart:

joaoantoniocardoso (approved):

Issue #1570, opened 2023-03-30
Support Param diffs


Williangalvani:
Ardupilot now reports both Default and Current parameter values. we can leverage this to show only changed parameters.

Williangalvani:
These are available via mavftp on newer firmware

Issue #1525, opened 2023-03-02
core: frontend: provide a comparison of parameters when loading from file


ES-Alexander:
### Current behaviour

Loading parameters from a file currently saves them into the vehicle firmware, but does not provide an indication as to which (if any) parameters are being changed. This makes it unnecessarily inconvenient to understand which changes are being applied.

### Expected or desired behaviour

It would be helpful to provide some form of diff window when loading parameters, to show all the loaded parameters that differ from those already on the vehicle. Ideally this would be like QGC's comparison table, which allows checking/unchecking the parameters individually to determine which ones to include in the update (and which ones to leave as they are).

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

ES-Alexander:
This could potentially share some functionality with #1245

patrickelectric:
I believe that is already done

PR #1709, opened 2023-05-25
Add interface for loading parameters

### LINKED ISSUE #1060: bug: Download vehicle parameters when downloading SITL
### Bug description

Rover SITL can't do mission 

### Steps to reproduce

Use Rover SITL, open QGC, upload a mission, arm and set to auto
360 spin

### Primary pain point(s)

It does not work

### Prerequisites

### LINKED ISSUE #1725: Add full BlueBoat support out of the box
BlueOS should be able to do the necessary configuration on the wizard to have full BlueBoat support and ready to use.

### LINKED ISSUE #1734: Add default parameters logic
Should handle BlueROV, BlueROV2, Blue Boat and SITL.
Including pixhawk and navigator.

### LINKED ISSUE #1750: we should be able to load qgc parameters in blueos

Williangalvani:
This add defaults parameters for BlueROV2, BlueBoat120, and Sitl (rover/sub).

https://github.com/bluerobotics/BlueOS-docker/assets/4013804/0654cb48-e8c5-4858-8526-f6a403cdfa80

I opted to have individual sets of parameters. The trickiest thing is the Pixhawk Servo assignments, which changes between standard and heavy, while also being different than Navigator.

I believe this will be a lot easier to maintain than having multiple individual files, one for each possible combination.

We _could_  use a code like this to actually generate the parameter files, but I believe this is more meaningful.

This does not try to fetch parameters from an online repository, though we _could_ do that later if we judge it necessary.

We could also have "maximum supported version" of firmware for blueos/parameters, but I am strongly against it, as the parameter changes in ArduSub _should_ never be so severe as to cause breakages. Also, gnome does this and I'm tired of manually updating the "supported version" of my extensions so I can enable them again.

The blueboat params are currently a full dump, which I expect to cleanup as soon as I'm cleared to do so.

The sensors Infos are the bottom were kept as they are useful to see parameter changes.

on a next PRs, I intend to add both a progress bar and an option to filter the table to show only the parameters that will be changed (instead of the whole set). We also still need an option to "reset to firmware defaults".

fix #1060 
fix #1734 
should fix #1725 
fix #1750

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterLoader.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterLoader.vue
@@ -76,7 +76,7 @@
           </template>
         </v-progress-linear>
       </v-card-text>
-      <v-card-text v-else-if="writing">
+      <v-card-text v-else-if="write_finished">
         <v-alert
           type="success"
         >
@@ -152,6 +152,9 @@ export default Vue.extend({
     },
     user_selected_params_length(): number {
       return Object.keys(this.user_selected_params).length
+    },
+    write_finished(): boolean {
+      return this.user_selected_params_length === 0
     },
   },
   watch: {
--- a/core/frontend/src/components/parameter-editor/ParameterLoader.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterLoader.vue
@@ -1,0 +1,318 @@
+<template>
+  <v-dialog
+    v-model="should_open"
+    width="fit-content"
+    max-width="80%"
+    @click:outside="done"
+  >
+    <v-card>
+      <v-card-title class="pt-6">
+        Loading Parameters
+      </v-card-title>
+      <v-card-text v-if="Object.keys(different_param_set).length !== 0">
+        <v-row class="virtual-table-row">
+          <v-col class="virtual-table-cell checkbox-cell">
+            <v-checkbox
+              ref="selectAllCheckbox"
+              v-model="select_all"
+              class="virtual-table-cell"
+              label="Write"
+              :disabled="writing"
+              :indeterminate="select_all === null"
+              @click="toggleSelectAll"
+            />
+          </v-col>
+          <v-col class="virtual-table-cell name-cell">
+            <strong>Name</strong>
+          </v-col>
+          <v-col class="virtual-table-cell">
+            <strong>Value</strong>
+          </v-col>
+          <v-col class="virtual-table-cell">
+            <strong>New Value</strong>
+          </v-col>
+        </v-row>
+        <!-- display all parameters in a concise table using virtual scroller -->
+        <v-virtual-scroll
+          :items="printable(different_param_set)"
+          height="300"
+          item-height="30"
+          class="virtual-table"
+        >
+          <template #default="{ item }">
+            <v-row class="virtual-table-row">
+              <v-col class="virtual-table-cell">
+                <v-checkbox
+                  v-model="param_checkboxes[item.name]"
+                  class="checkbox-label checkbox-cell"
+                  :disabled="writing"
+                />
+              </v-col>
+              <v-col class="virtual-table-cell name-cell">
+                {{ item.name }}
+              </v-col>
+              <v-col class="virtual-table-cell">
+                {{ printParam(item.current_value) }}
+              </v-col>
+              <v-col class="virtual-table-cell">
+                {{ printParam(item.value) }}
+              </v-col>
+            </v-row>
+          </template>
+        </v-virtual-scroll>
+        <v-progress-linear
+          v-if="user_selected_params_length > 0 && initial_size > 0"
+          slot="progress"
+          :color="error ? 'red' : 'blue'"
+          height="20"
+          :value="progress"
+        >
+          <template #default>
+            <strong
+              class="ml-5"
+            >
+              {{ user_selected_params_length }} to go
+            </strong>
+          </template>
+        </v-progress-linear>
+      </v-card-text>
+      <v-card-text v-else-if="writing">
+        <v-alert
+          type="success"
+        >
+          Parameters written successfully
+        </v-alert>
+      </v-card-text>
+      <v-card-text v-else>
+        No parameters to change. This means all the
+        parameters in this group are already set to the desired values
+      </v-card-text>
+      <v-card-text v-if="error">
+        <v-alert
+          type="error"
+        >
+          {{ error }}
+        </v-alert>
+      </v-card-text>
+      <v-card-actions class="justify-center">
+        <v-btn
+          :disabled="user_selected_params_length === 0 || writing"
+          class="ma-6 elevation-2"
+          x-large
+          color="primary"
+          @click="writeParams"
+        >
+          Write Parameters
+        </v-btn>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from 'vue'
+import { Dictionary } from 'vue-router/types/router'
+
+import mavlink2rest from '@/libs/MAVLink2Rest'
+import autopilot_data from '@/store/autopilot'
+
+export default Vue.extend({
+  name: 'ParameterLoader',
+  props: {
+    parameters: {
+      type: Object as PropType<Dictionary<number>> | undefined,
+      default: undefined,
+    },
+  },
+  data: () => ({
+    initial_size: 0,
+    param_checkboxes: {} as Dictionary<boolean>,
+    select_all: true as boolean | null,
+    retries: 0,
+    retry_interval: 0,
+    should_open: false,
+    error: false as boolean | string,
+    writing: false,
+  }),
+  computed: {
+    progress(): number {
+      return 100 - this.different_param_set_length / this.initial_size * 100
+    },
+    writeable_param_set(): Dictionary<number> {
+      return this.filterParamsByReadOnly(this.parameters)
+    },
+    different_param_set(): Dictionary<number> {
+      return this.filterParamsByValue(this.writeable_param_set)
+    },
+    user_selected_params(): Dictionary<number> {
+      return this.filterParamsBySelection(this.different_param_set)
+    },
+    different_param_set_length(): number {
+      return Object.keys(this.different_param_set).length
+    },
+    user_selected_params_length(): number {
+      return Object.keys(this.user_selected_params).length
+    },
+  },
+  watch: {
+    param_checkboxes: {
+      deep: true,
+      handler() {
+        this.updateSelectAllStatus()
+      },
+    },
+    parameters: {
+      handler(newval) {
+        if (Object.keys(this.writeable_param_set).length !== 0) {
+          this.should_open = true
+        }
+        this.updateParamCheckboxes(newval)
+      },
+    },
+  },
+  methods: {
+    done() {
+      this.$emit('done')
+      this.error = false
+      this.retries = 0
+      setTimeout(() => {
+        // delay this as the dialog is still open when the done event is emitted
+        // which couses the text to change during the close animation
+        this.writing = false
+      }, 500)
+      this.initial_size = 0
+      this.should_open = false
+    },
+    toggleSelectAll() {
+      const new_value = this.select_all === null ? true : this.select_all
+      for (const key of Object.keys(this.param_checkboxes)) {
+        this.$set(this.param_checkboxes, key, new_value)
+      }
+    },
+    printParam(value: number) {
+      try {
+        if (Math.abs(value) > 1e4) {
+          return value.toExponential(3)
+        }
+        if (Math.abs(value) < 0.01 && value !== 0) {
+          return value.toExponential(3)
+        }
+        return value.toFixed(2)
+      } catch {
+        return 'N/A'
+      }
+    },
+    printable(paramset: Dictionary<number>) {
+      return this.createPrintableParams(paramset)
+    },
+    writeParams() {
+      this.writing = true
+      this.initial_size = this.user_selected_params_length
+      this.writeSelectedParams()
+      this.retry_interval = setInterval(() => {
+        if (this.user_selected_params_length) {
+          this.retries += 1
+          if (this.retries > 5) {
+            clearInterval(this.retry_interval)
+            this.retries = 0
+            this.error = 'Failed to write some parameters. Please restart the vehicle and try again.'
+            return
+          }
+          this.writeSelectedParams()
+        } else {
+          clearInterval(this.retry_interval)
+        }
+      }, 1000)
+    },
+    writeSelectedParams() {
+      for (const [name, value] of Object.entries(this.user_selected_params)) {
+        this.writeParam(name, value)
+      }
+    },
+    writeParam(name: string, value: number) {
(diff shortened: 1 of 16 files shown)


ES-Alexander (approved):
This definitely seems handy! :-)

I imagine this would benefit from both #1525 and #1570.

The relevant docs (screenshot+description) should go in [the `advanced-usage#vehicle-setup` section of `BlueOS-devel-temp`](https://github.com/bluerobotics/ardusub-zola/blob/BlueOS-devel-temp/advanced-usage/index.md#vehicle-setup)

core/frontend/src/components/vehiclesetup/params/subParams.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const BLUEROV2_HEAVY = {
+  ...BASE_BLUEROV,
+  ATC_ANG_PIT_P: 6.0,
+  ATC_ANG_RLL_P: 6.0,
+  ATC_ANG_YAW_P: 6.0,
+  FRAME_CONFIG: 2,
+}
+
+function getSubParamSets(board: string | undefined, version: SemVer | undefined): Dictionary<Dictionary<number>> {
```
patrickelectric, line 61: what is the user case for undefined board ?

core/frontend/src/components/vehiclesetup/params/roverParams.ts
```diff
@@ -0,0 +1,915 @@
+import { SemVer } from 'semver'
+import { Dictionary } from 'vue-router/types/router'
+
+// This is from the QGC file as provided by Eldin
+// # Vehicle-Id Component-Id Name Value Type
+const bb120params = {
```
patrickelectric, line 6: are all this parameters necessary ? Wasn't the idea to not include everything ?

patrickelectric (changes requested):
The current approach looks ok-ish.

My problems are:
- How we are going to maintain it for multiple vehicles ?
- How we are going to follow this up for multiple versions of firmware ?
- Wow people are going to provide such things without dealing with BlueOS source code ?
- Will people need to go to setup after going to the wizard ? Can we move the funcionality to the wizard ?
- Do we fail if version / board are not defined ?

core/frontend/src/components/vehiclesetup/params/subParams.ts
Williangalvani: no use case case, it is just happens before the board type is loaded from the backend

patrickelectric: So the function should not accept this case.

core/frontend/src/components/vehiclesetup/params/roverParams.ts
Williangalvani: > The blueboat params are currently a full dump, which I expect to cleanup as soon as I'm cleared to do so.

I'm waiting for confirmation that I can clean these up from the bb120 team

Williangalvani:
> How we are going to maintain it for multiple vehicles ?
  
this already does? it supports both rover and sub, and 2 differnt vehicle setups for sub.
  For custom frames, we should allow loading custom ones from userdata
 

> How we are going to follow this up for multiple versions of firmware ?

Like [this](https://github.com/bluerobotics/BlueOS-docker/pull/1709/files#diff-63ed259d6e59fcad1a77ee7ec630312da81eb8d669c865b74cf877e14e2f0f54R83). There will never be a huge change in parameters, so this should work fine and handle incremental changes nicely.

> How we are going to follow this up for hardware revisions ?

I don't think there's an easy solution for this either way. We can't really detect hardware revisions, can we? only option I can think of is having the user specify it somewhere, then we can show the appropriate params.

> How people are going to provide such things without dealing with BlueOS source code ?

We can provide an option to load additional paramsets from userdata, using a well-defined structure

> Will people need to go to setup after going to the wizard ? Can we move the funcionality to the wizard ?

Yep! my goal is to add it there, too.

> Do we fail if version / board are not defined ?

If there are no parameter sets for the detected vehicle/board, we don't show any.

> This kinda of replicates what QGC does, and that's kind of scary.

I'm not super happy about it, but [this](https://github.com/ArduPilot/ardupilot/tree/master/Tools/Frame_params/Sub) is far worse to maintain as 80% of the params there are irrelevant, and a new set of files need to be generated for every firmware/hw/frame change

patrickelectric:
> > How we are going to maintain it for multiple vehicles ?
> 
> this already does? it supports both rover and sub, and 2 differnt vehicle setups for sub. For custom frames, we should allow loading custom ones from userdata

I'm talking about different or custom vehicles, like JM ROVs, Px4 params for custom 3rd party products and etc. 

> 
> > How we are going to follow this up for multiple versions of firmware ?
> 
> Like [this](https://github.com/bluerobotics/BlueOS-docker/pull/1709/files#diff-63ed259d6e59fcad1a77ee7ec630312da81eb8d669c865b74cf877e14e2f0f54R83). There will never be a huge change in parameters, so this should work fine and handle incremental changes nicely.

Yes. But that requires manual intervention on BlueOS, so each version of ArduPilot reflects in a new version of BlueOS, and that's not ok.

> 
> > How we are going to follow this up for hardware revisions ?
> 
> I don't think there's an easy solution for this either way. We can't really detect hardware revisions, can we? only option I can think of is having the user specify it somewhere, then we can show the appropriate params.

That's up to the user, that'll make it simpler, QGC does a good job on that making the user chooser the hardware that he's using as an option to configure the parameter. We could have automatic check, but manual is already a good step foward.

> 
> > How people are going to provide such things without dealing with BlueOS source code ?
> 
> We can provide an option to load additional paramsets from userdata, using a well-defined structure
> 

Can we use the same structure for our own parameters ? Will the current logic affect it ?


> > Will people need to go to setup after going to the wizard ? Can we move the funcionality to the wizard ?
> 
> Yep! my goal is to add it there, too.

Nice!

> 
> > Do we fail if version / board are not defined ?
> 
> If there are no parameter sets for the detected vehicle/board, we don't show any.

Good.

> 
> > This kinda of replicates what QGC does, and that's kind of scary.
> 
> I'm not super happy about it, but [this](https://github.com/ArduPilot/ardupilot/tree/master/Tools/Frame_params/Sub?rgh-link-date=2023-05-25T12%3A43%3A05Z) is far worse to maintain as 80% of the params there are irrelevant, and a new set of files need to be generated for every firmware/hw/frame change

I don't disagree that dealing with parameters as delta is a better idea. The problem is the logic after we have more versions and different configurations, and that's QGC problem.

I would be nice if we end up with two structures, one for specific or custom hardware for specific version of firmwares for the vehicle.
And another one for default parameters per vehicle type, per firmware version per vehicle.
With that we are going to have more files, but the changes will be much simpler to deal with.
As e.g:
    - Check firmware type / version
    - Load default parameters for a Sub with a minimum version support
    - Load default parameters for BlueROV2
    - Load parameters for hardware changes / hardware revision
    
There is no simpler way to fix that, but we should work with the one that will be easier to maintain and sustain, while making it easy for people to work on codewise. If we end up doing something now that is unsustainable or gets out of control, we are going to have more work to fix that later. 

It's ok to what we are currently doing now, but I don't expect to continue using it in the next year.

Also, supporting 3rd party vehicles / firmwares is a priority for BlueOS adoption.

Williangalvani:
> Load default parameters for a Sub with a minimum version support

The firmware defaults are sane enough that we don't need to replicate them in our own files

 >  I'm talking about different or custom vehicles, like JM ROVs, Px4 params for custom 3rd party products and etc.

My vision for how this goes so far:

We keep the structure on this PR, and additionally we check in userdata for a file structure like:
`<vehicle_type>/<board>/<version>/<any name>.param`
These could be the full param file, but we could also setup a scheme with version "inheritance" where the params for 4.1.1 include the params for 4.1, and the params for 4.1 include the params for 4.0.

These would be added in by distributors, possibly via extensions.


 > Yes. But that requires manual intervention on BlueOS source code, so each version of ArduPilot reflects in a new version of BlueOS, and that's not ok.
 
I'm not sure of how to get any smart logic going if we move these to an online repository, plus then the user would need to go online to get new parameters.

> but the changes will be much simpler to deal with.

I disagree

Williangalvani:
maybe we can get to a middle ground.
we move all files to a repository. but keep them separated logically.

we could also have metadata on the files to check which can be applied,
something like:
```
# name: BlueROV2 Heavy
# vehicle: ArduSub
# version: >=4.0
# board: Navigator
```

but then, could we merge all applicable files into one? should we?
we could also just list all of the files for the user, but that may result in a poorer user experience

patrickelectric:
> > Load default parameters for a Sub with a minimum version support
> 
> The firmware defaults are sane enough that we don't need to replicate them in our own files

That's not what I'm implying. But we should enforce at least have the option to enforce default firmware parameters before settings our own, to help the users with any parameter issue that he can find and in a clean / factory parameter value list.

> > I'm talking about different or custom vehicles, like JM ROVs, Px4 params for custom 3rd party products and etc.
> 
> My vision for how this goes so far:
> 
> We keep the structure on this PR, and additionally we check in userdata for a file structure like: `<vehicle_type>/<board>/<version>/<any name>.param` These could be the full param file, but we could also setup a scheme with version "inheritance" where the params for 4.1.1 include the params for 4.1, and the params for 4.1 include the params for 4.0.
> 
> These would be added in by distributors, possibly via extensions.

Yes, but we can also add that in a repository where any distributor could provide it, we can also download and store it locally to be available offline.

> > Yes. But that requires manual intervention on BlueOS source code, so each version of ArduPilot reflects in a new version of BlueOS, and that's not ok.
> 
> I'm not sure of how to get any smart logic going if we move these to an online repository, plus then the user would need to go online to get new parameters.
> 

Can't we just follow the folder structure that you just said ? 

> maybe we can get to a middle ground. we move all files to a repository. but keep them separated logically.
> 
> we could also have metadata on the files to check which can be applied, something like:
> 
> ```
> # name: BlueROV2 Heavy
> # vehicle: ArduSub
> # version: >=4.0
> # board: Navigator
> ```
> 
> but then, could we merge all applicable files into one? should we? we could also just list all of the files for the user, but that may result in a poorer user experience

Can't we have the rules being provided by the file or the folder structure ?
Like:
`ArduPilot/ArduSub/navigator/default.ts`
`ArduPilot/ArduSub/navigator/4.1.*/{standard, heavy, base}.ts`
And the content of `heavy.ts` (pseudo code):
```ts
import BASE from "base.ts"

const PARAMS = {
  BASE,
  ATC_ANG_PIT_P: 6.0,
  ATC_ANG_RLL_P: 6.0,
  ATC_ANG_YAW_P: 6.0,
  FRAME_CONFIG: 2,
}
```

`base.ts` would include `default.ts`, we could have versions and hardware revision logic encapsulated. 
We could compile the files and provide it for BlueOS to fetch online and make it available offline. BlueOS could also download compile. The important thing is to define the necessary rules, we can do what we are currently doing here, but structure in a place that is outside BlueOS source code, where anyone can colaborate without making the logic inside BlueOS more complex. 

This will take time, but it'll clear for BlueOS and future proof. No need for people to update BlueOS to fetch to support new versions, no need for 3rd party to change BlueOS source code, free BlueOS support to any vehicle if someone colaborate with the parameter repository. And, if someone want to have the full parameter file, the structure will still support it.

joaoantoniocardoso (commented):
The first few times I tried to upload the params of my Pixhawk I got this, even after a full reboot. Note that this video is for the first try, the freezing in my browser is probably just a coincidence:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/4bc45371-a551-4bbc-93ae-190a8b6af172

Then, I don't know what happened, but suddenly it started to work. So after all params were uploaded, I manually changed a few, uploaded the defaults again, and yeah, it's working :1st_place_medal:

I experienced the same behavior with SITL -- it failed to write the parameters:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/63d1ffe5-0974-4f64-9002-4c7c2ba661b3

Continuation after the reboot:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/c0f3d71e-f7fd-4b84-9e02-04afaab58403

Again with SITL but this time with Rover, also got into a similar case:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/7727acbf-e525-421c-a41b-86bf4e147f2f

---

A nitpicky thing is that the success dialogue is shown for a brief moment whenever we close the upload dialogue, you can check it looking for a green button on the next video:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/113c926c-8942-4822-99fa-cc6218d21ff0

joaoantoniocardoso:
> @joaoantoniocardoso that is due to the way ardupilot works, unfortunately. they have some "enable" parameters that "block" other params. that is why the restart is required. we could automate that, though.

Hmm, not sure if just one restart works, 'cause I tried that

joaoantoniocardoso:
> > Hmm, not sure if just one restart works, 'cause I tried that
> 
> possibly related to #1756 we were restarting the fw but not refetching the params

Hmm okay, if you rebase it and recreate your dockerhub image I can test it again

core/frontend/src/components/parameter-editor/ParameterLoader.vue
```diff
@@ -0,0 +1,306 @@
+<template>
+  <v-dialog
+    v-model="should_open"
+    width="fit-content"
+    max-width="80%"
+    @click:outside="done"
+  >
+    <v-card>
+      <v-card-title class="justify-center pt-6">
+        Loading Parameters: {{ selected_paramset_name }}
```
joaoantoniocardoso, line 10:
```suggestion
        Loading Parameters {{ selected_paramset_name }}
```

joaoantoniocardoso (commented):
* To fix: Params selection should be disabled when uploading:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/051d5012-12ff-4e42-a840-e32d7047278f


* To fix: The 'write parameters' button should be disabled instead of hidden when no params are selected:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/a6ce3771-326c-4fe6-939c-1837881eff6b

core/frontend/src/components/parameter-editor/ParameterLoader.vue
```diff
@@ -0,0 +1,321 @@
...
+            </strong>
+          </template>
+        </v-progress-linear>
+      </v-card-text>
+      <v-card-text v-else-if="write_finished">
+        <v-alert
+          type="success"
+        >
+          Parameters written successfully
+        </v-alert>
```
joaoantoniocardoso, line 84: This triggers whenever there's no diff in the parameters, meaning that it shows "parameters written successfully" if you've executed the upload once in the past:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/63ff1dc4-d98d-4d57-b9ca-735af2ebd03c

joaoantoniocardoso (commented):
Almost perfect for me

core/frontend/src/components/parameter-editor/ParameterLoader.vue
Williangalvani: I added a check for initial_size > 0 to check if it actually attempted to write anything

joaoantoniocardoso (approved):
Great!

PR #1712, opened 2023-05-25
frontend: widgets: Add first version

patrickelectric:
- Allow top bar widgets
- Allow user to drag and move it around
- Allow add / remove
- Add dynamic support

peek

### DIFF
--- a/core/frontend/src/store/settings.ts
+++ b/core/frontend/src/store/settings.ts
@@ -24,6 +24,8 @@ class SettingsStore extends VuexModule {
 
   tour_version = 0
 
+  user_top_widgets = [] as string[]
+
     @Mutation
   setDarkTheme(value: boolean): void {
     this.is_dark_theme = value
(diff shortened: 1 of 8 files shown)


core/frontend/src/store/settings.ts
```diff
@@ -93,7 +93,7 @@ class SettingsStore extends VuexModule {
      */
     static saveVariable<T>(name: string, value: T): void {
       // eslint-disable-next-line
-      window.localStorage.setItem(SettingsStore.settingsName(name), value as any)
+      window.localStorage.setItem(SettingsStore.settingsName(name), JSON.stringify(value))
```
joaoantoniocardoso, line 96: probably `value` here needs to have no hidden fields to be retrieved with no errors from the store... So why not change the type T to JSON to avoid the trap?

joaoantoniocardoso (approved):
Coooooool

core/frontend/src/store/settings.ts
patrickelectric: > probably `value` here needs to have no hidden fields to be retrieved with no errors from the store... So why not change the type T to JSON to avoid the trap?

Can you give an example of where that would fail ? 
Not sure that I follow your concern.

patrickelectric: As discussed the objects are simple and thay should be the case.

PR #1713, opened 2023-05-25
core: frontend: store: system-information: Check if backend is offline before warn

patrickelectric:


Williangalvani:
isn't 'backend_offline_error' thrown by `back_axios` ?
the function here is using regular `axios`

patrickelectric:
> isn't 'backend_offline_error' thrown by `back_axios` ? the function here is using regular `axios`

nice catch, fixed.

Williangalvani (approved):

PR #1728, opened 2023-05-26
Show correct name of image for an installed extension

rotu:
Previously, we used the second slash-separated path component instead of the last one for the image name.

This is incorrect as, e.g. in the below example. The name of the image identified by "docker.io/library/alpine" is "alpine", not "library"

image


patrickelectric (approved):
LGTM!

PR #1737, opened 2023-05-31
core: frontend: fix SemVer not being populated

### LINKED ISSUE #1736: FirmwareInfo.Version is a string instead of the declared SemVer object
image
image

Williangalvani:
fix #1736

### DIFF
--- a/core/frontend/src/components/autopilot/AutopilotManagerUpdater.ts
+++ b/core/frontend/src/components/autopilot/AutopilotManagerUpdater.ts
@@ -1,4 +1,5 @@
 import { AxiosResponse } from 'axios'
+import { SemVer } from 'semver'
 
 import Notifier from '@/libs/notifier'
 import autopilot from '@/store/autopilot_manager'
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):

joaoantoniocardoso (changes requested):
wait... haha let me see it again

core/frontend/src/components/autopilot/AutopilotManagerUpdater.ts
```diff
@@ -74,6 +75,8 @@ export async function fetchFirmwareInfo(): Promise<void> {
       url: `${autopilot.API_URL}/firmware_info`,
       timeout: 10000,
     })
+    // Version comes out as a string, let's turn it into a SemVer object
+    response.data.version = new SemVer(response.data.version)
```
joaoantoniocardoso, line 79: Is this the only place we need to make this conversion?

Williangalvani: it is the only one I found so far. I'm too deep in the params thing to look for others...

joaoantoniocardoso: I checked, and it is.

joaoantoniocardoso (approved):

PR #1746, opened 2023-06-06
core: frontend: wizard: add skip and disable options to wizard

Williangalvani:
Screenshot 2023-06-06 at 00 00 50

Skip is for the current session. Disable is until the next Wizard version change.

Not quire sure how I feel about this one 🤔

### DIFF
--- a/core/frontend/src/components/wizard/Wizard.vue
+++ b/core/frontend/src/components/wizard/Wizard.vue
@@ -57,6 +57,21 @@
                 @click="setupBoat"
               />
             </div>
+            <v-row class="pa-5">
+              <v-btn
+                color="warning"
+                @click="cancel"
+              >
+                Ask me later
+              </v-btn>
+              <v-spacer />
+              <v-btn
+                color="error"
+                @click="setWizardVersion(); cancel()"
+              >
+                Don't show again
+              </v-btn>
+            </v-row>
           </v-stepper-content>
 
           <v-stepper-content step="2">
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (changes requested):
I think we should at least explain the implications and how to revert the action, when possible.

maybe change "Skip" to "Do later", and require a confirmation for "disable", with a tooltip like "I want to do the system configuration manually by myself", with a dialogue explaining something like "Disabling is not recommended, and will require manual configuration. Be sure to visit the firmware and vehicle pages. To enable the wizard again, a full settings reset is necessary."

ES-Alexander:
Agreed with @joaoantoniocardoso that it's not very clear what these do at the moment, and that explanatory tooltips are a good idea (possibly with a (?) icon to make it clear tooltips are available). Maybe we could change the names to "Ask Me Later" and "Leave Me Alone"/"I Know What I'm Doing" or something? Also, "Disable Wizard" just sounds brutal - poor wizard... 😂 

I also think that having two vehicles and two buttons makes it seem a bit like the buttons correspond to the vehicles, so maybe we should put both buttons on the same side, or put a horizontal separator line between the vehicle selector and the buttons or something.

Williangalvani:
> I would go for "Ask me later" and "Don't show again". Those are widely used, so users are already used to them.

I like it.

rafaellehmkuhl (approved):

PR #1747, opened 2023-06-06
core: frontend: wizard: add 'other' vehicle

### LINKED ISSUE #1681: core: frontend: Add "Other" vehicle type to Wizard setup for users that are not setting up a BROV2 or BlueBoat
### Current behaviour

The wizard only shows options for BROV2 and BlueBoat, with no option to exit the wizard on first boot. This forces users who are not setting up either of these vehicles to select one just to move on from the Wizard.

A third vehicle type option of "Other" to account for everything else that is not an ROV or Boat.

### Prerequisites

Williangalvani:
image

Fix #1681

### DIFF
--- a/core/frontend/src/components/wizard/Wizard.vue
+++ b/core/frontend/src/components/wizard/Wizard.vue
@@ -56,6 +56,13 @@
                 class="model-button ma-2"
                 @click="setupBoat"
               />
+              <v-icon
+                v-tooltip="'Other Vehicle Setup'"
+                class="model-button ma-2 other-button"
+                @click="setupOther"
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
LGTM if we don't need to set up anything else for *others*

ES-Alexander (approved):
I like it, although I think it'd be useful for accessibility to have visible text labels instead of just tooltips (but not sure how much clutter that adds on small screens).

Williangalvani:
we can allow further configurations for "other" vehicles, but I think this is a decent start

PR #1748, opened 2023-06-06
core: frontend: wizard: only write wizard to bag after finishing all other steps

### LINKED ISSUE #1721: [wizard] Wizard should be disable only when succeeded
The last step of the wizard, which disables the wizard, should be done after all others were finished with success, otherwise, we can end up with a partially configured vehicle.

Example:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/ce2715c9-2b85-47c1-946e-26407e1e7deb

Williangalvani:
fix #1721


joaoantoniocardoso (approved):
makes sense, didn't tested

Williangalvani:
> makes sense, didn't test

I did =]

PR #1752, opened 2023-06-06
Set the default firmware during the wizard

### LINKED ISSUE #1729: [autopilot firmware] Default firmware for Sub is DEV, but should be.. the latest STABLE?
https://github.com/bluerobotics/BlueOS-docker/assets/5920286/e116c4a0-81a9-4bad-b5d2-8aa04dcee752

Williangalvani:
This will save the firmware downloaded during the wizard as the default fw in userdata.
we'll still fallback to a fw in the docker container

fix #1729 

image will be up at williangalvani/blueos-core : default_fw as soon as [this](https://github.com/Williangalvani/blueos-docker/actions/runs/5192546688/jobs/9362068931) finishes.

to test, make sure you are using a beta fw, then go through the wizard
after the wizard, you will have sub 4.1.0. switch to a beta again.
now click "restore default firmware" it should get you back to 4.1.0

### DIFF
--- a/core/frontend/src/components/autopilot/AutopilotManagerUpdater.ts
+++ b/core/frontend/src/components/autopilot/AutopilotManagerUpdater.ts
@@ -129,7 +129,7 @@ export async function availableFirmwares(vehicleType: Vehicle): Promise<Firmware
     })
 }
 
-export async function installFirmwareFromUrl(url: URL): Promise<void> {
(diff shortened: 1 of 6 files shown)


rafaellehmkuhl (approved):
Tested.
Everything working.
Thanks for the test instructions.

PR #1756, opened 2023-06-07
core: frontend: reset frontend parameters when flashing new firmware and restarting

### LINKED ISSUE #1742: Duplicated autopilot parameters
image

I was using SITL and changed to Navigator. Don't know if this was the reason, though.

Williangalvani:
should fix #1742


rafaellehmkuhl (approved):
Problem fixed for me.

PR #1765, opened 2023-06-08
Do not apply dwc2 OTG overlay to RPI3

### LINKED ISSUE #1757: Cameras freezing on RPI 3
On Raspberry Pi 3, USB camera streams (using video4linux) are constantly freezing in a non-periodic interval, as shown below:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/878eda94-1a8b-483b-98f9-69313059e76d

Facts:
- freezes on master
- freezes on stable
- doesn't freeze on the old companion
- GStreamer version doesn't matter (tested with 1.8, 1.18, 1.20, and 1.22)

More details on the [original post (internal slack) ](https://bluerobotics.slack.com/archives/C015VGZT5K6/p1685955211923179)

joaoantoniocardoso:
Fixes #1757 

- [x] Fixes running systems
  - [x] Fixes OTG messing with RPI3
  - [x] Moves all Navigator-related configurations to pi4 section
  - [x] Don't mess with user configurations (as long as they place an inline comment with the word `custom` on it)
  - [x] Backup files before patch them
- [ ] ~~Fixes new images~~

We should document the use of the 'custom' inline comment to protect the line from this script


core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -75,18 +75,54 @@ def update_cgroups() -> None:
...
+    sudo sed --in-place "{dwc2_overlay_match_pattern}" "{config_file}"
+    '''
     print(run_command(command, False))
 
-    command = '''sudo sh -c '\
-    grep -q "dtoverlay=dwc2" /boot/config.txt || (
-        echo "dtoverlay=dwc2" | sudo tee -a /boot/config.txt
-    )' '''
+    # Create a pi4 section at the end if it doesn't exist
+    cm4_section_match_pattern = '^\\[pi4\\].*$'
```
Williangalvani, line 88: why "cm4" ?

Williangalvani (commented):
what if instead of doing a lot of seds and greps, we load the file using cat and commander and then put it back there?

Then maybe we could have more legible code.
there are probably libraries we can leverage on to parse/edit the file in a better way, too.

core/tools/blueos_startup_update/blueos_startup_update
joaoantoniocardoso: > what if instead of doing a lot of seds and greps, we load the file using cat and commander and then put it back there?
> 
> Then maybe we could have more legible code. there are probably libraries we can leverage on to parse/edit the file in a better way, too.

I agree, I tried the parserconfig but this config file is incompatible because it doesn't present a header... Which if loading with `cat` we could add and later filter out... But before doing this workaround I'd take a look on other options :)

joaoantoniocardoso: > why "cm4" ?
Lol, should be `pi4`

ES-Alexander (approved):
Nice!

Tried it out, and confirmed that it updates the config file as specified, and fixes the camera issue (after rebooting, since the config runs on boot).

Before / After:
Screenshot 2023-06-16 at 12 14 14 pmScreenshot 2023-06-16 at 12 29 52 pm

joaoantoniocardoso:
@ES-Alexander @Williangalvani Well, now it's rebooting the pi.

Williangalvani:
can you update it to deal with the changes in #1806?

after it, my config file looked like this:

image

core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -43,58 +48,287 @@ def dict_merge(dct, merge_dct):
...
+
+    navigator_configs_with_match_patterns = [
+        ('enable_uart=1', '^enable_uart=.*'),
+        ('dtoverlay=uart1', '^dtoverlay=uart1.*'),
+        ('dtoverlay=uart3', '^dtoverlay=uart3.*'),
+        ('dtoverlay=uart4', '^dtoverlay=uart4.*'),
+        ('dtoverlay=uart5', '^dtoverlay=uart5.*'),
+        ('dtparam=i2c_vc=on', '^dtparam=i2c_vc=.*'),
+        ('dtoverlay=i2c1', '^dtoverlay=i2c1.*'),
+        ('dtparam=i2c_arm_baudrate=1000000', '^dtparam=i2c_arm_baudrate.*'),
```
joaoantoniocardoso, line 292: @Williangalvani here's the new config since #1806

patrickelectric:
Fix https://github.com/bluerobotics/BlueOS-docker/issues/1802

PR #1772, opened 2023-06-11
use .params and friends for saving/loading param files

Williangalvani:


ES-Alexander (approved):
I dislike that ArduPilot somehow has 3 different file extensions that all support multiple different file formats, but that's not our issue, we just have to deal with the repercussions.

We should add support for all the valid ArduPilot formats, as well as QGC's format, but that can happen in a separate PR

PR #1774, opened 2023-06-12
core: frontend: parameterEditorDialog: save when pressing enter

### LINKED ISSUE #1665: ParameterEditor: pressing "Enter" after editing a parameter causes the page to refresh without actually setting the parameter.
we may have missed a `preventDefault`

Williangalvani:
fix #1665

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterEditorDialog.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterEditorDialog.vue
@@ -31,7 +31,10 @@
               sm="6"
               md="6"
             >
-              <v-form ref="form">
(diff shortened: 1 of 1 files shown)


ES-Alexander (approved):
I'm not familiar with the syntax here, but I'm assuming you've tested it, and it's definitely a welcome feature :-)

Williangalvani:
> I'm not familiar with the syntax here, but I'm assuming you've tested it, and it's definitely a welcome feature :-)

the `.prevent`  part prevents the default behavior of submitting the form, which refreshed the browser. the new function runs in its place, writing the parameter =]

PR #1775, opened 2023-06-12
core: frontend: stop notifying ParametersTable misses, improve message

### LINKED ISSUE #1739: PARAM_SET_FAIL don't need to be reported in notifications
image

we should be able to just consume the parameter and create the entry in the parameter table, shouldn't we?

Williangalvani:
fix #1739


ES-Alexander (approved):
Possibly capitalise the first word of the message, but 🤷‍♂️

PR #1776, opened 2023-06-12
Fix two bugs on ParameterEditorDialog

### LINKED ISSUE #1771: parameter editor doesnt allow setting SERVOX_FUNCTION params to -1

Williangalvani:
fix #1771 and a another bug where we had to click outside of the textfield for the save button to get enabled.


ES-Alexander (approved):
Nice!
Maybe add a comment that the <0 check means it's disabled or something (if that's what it means), but up to you.

PR #1779, opened 2023-06-12
core: frontend: extension manager: show spinner in bars instead of as a large overlay

Williangalvani:
The large overlay blocked the buttons, which means it blocked event the "logs" button when something went wrong.
before:
image

later:
image


joaoantoniocardoso (approved):
That's a nice idea

PR #1782, opened 2023-06-12
core: frontend: wizard: move wifi connection to first step of wizard

### LINKED ISSUE #1781: wifi internet picking should be the first step
right now it is being done after picking the firmware version which doesn't work without internet

Williangalvani:
fix #1781

### DIFF
--- a/core/frontend/src/components/wizard/RequireInternet.vue
+++ b/core/frontend/src/components/wizard/RequireInternet.vue
@@ -1,26 +1,24 @@
 <template>
   <div>
-    <v-card
-      color="primary"
-      dark
-      class="mb-5"
-    >
-      <v-card-text
-        class="d-flex flex-column align-center subtitle font-weight-medium"
-      >
-        {{ status }}
-        <v-progress-linear
-          v-if="connected"
-          indeterminate
-          color="white"
-          class="mb-0"
-        />
-      </v-card-text>
+    <v-card elevation="0">
+      <v-stepper vertical>
+        <v-stepper-step
+          step="1"
+          :color="icon_color"
+          :complete-icon="icon"
+          :complete="true"
+          active
+          class="step-label"
+        >
+          {{ text }}
+        </v-stepper-step>
+      </v-stepper>
+      <WifiManager
+        v-if="!is_online && !checking"
+        :show-top-bar="false"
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso (commented):
LGTM, someone needs to test it

joaoantoniocardoso (changes requested):
UI is kinda broken: the fourth step is not on the top.

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/281e37e8-800d-4a9c-aeee-1a972c0fb504

Also, I think it should be something like:

```
# BlueOS Wizard
---
< Some introductory message like..: > Hello, this wizard will help you
 set up your vehicle for its first use! In order to fetch firmware and
 other data from the internet, please ensure that you have a wired
 internet connection, or click the button below to configure your WiFi connection.

[x] Trying to reach the internet via cable... Failed. 

[ask me later] [don't show again] [configure WiFi] [retry]
```

and then, the WiFi dialogue shows if you click there.

edit: We might then, remove the old 'welcome to BlueOS' message, would be redundant

Williangalvani:
@joaoantoniocardoso wired internet is going to be very rare. I don't think it should be mentioned. The wifi step should already be skipped automatically if internet is detected in any form.

I do like the idea of a welcome text, and I'll trim the steps so we only have 4

joaoantoniocardoso:
> @joaoantoniocardoso wired internet is going to be very rare. I don't think it should be mentioned. The wifi step should already be skipped automatically if internet is detected in any form.
> 
> I do like the idea of a welcome text, and I'll trim the steps so we only have 4

I see... But do we want it to be rare? If we are designing Cockpit to handle multiple clients, I think the most common setup would be the vehicle connected to a router, and then I can imagine (at least half) of the users would prefer to just have the internet cable to that router too, sharing between all computers and vehicle.

Williangalvani:
> I see... But do we want it to be rare? If we are designing Cockpit to handle multiple clients, I think the most common setup (for multiple clients) would be the vehicle connected to a router, and then I can imagine (at least half) of the users would prefer to just have the internet cable to that router too, sharing between all computers and vehicle.

I am confused, so all vehicles would be effectively wired together? :thinking:

joaoantoniocardoso (approved):

PR #1786, opened 2023-06-13
core: frontend: wizard: automate stepping

Williangalvani:


joaoantoniocardoso (approved):
Coooooool, seems much better

Issue #1783, opened 2023-06-13
core: frontend: log ardusub output


clydemcqueen:
### Current behaviour

ArduSub output is not saved, so there's no way to examine crash and printf output after the tmux session is stopped.

### Expected or desired behaviour

ArduSub output is saved to a log file that can be retrieved later.

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

PR #1789, opened 2023-06-13
Attempt at logging ardupilot's process output

Williangalvani:
attempt at #1783 

doesnt seem to work. maybe ardupilot is not using stdout?
nope, [it does in fact use stdout](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_HAL_Linux/ConsoleDevice.cpp#L39)


patrickelectric:

PR #1790, opened 2023-06-14
Bootstrap tweaks

Williangalvani:
- Fixes a bug where an effectively broken  image (williangalvani/blueos-core:fixotg) was not reverted to factory.
 - Reduces the verbosity
 - increases core timeout to 5 minutes, as that is a last-resort fix anyway and we REALLY dont want false positives.


joaoantoniocardoso (approved):
Good catch

Issue #1815, opened 2023-06-21
Can't load parameter file that was already previously selected


patrickelectric:
![peek](https://github.com/bluerobotics/BlueOS-docker/assets/1215497/ee400d24-5b74-4144-86a2-1295e6077f57)

PR #1816, opened 2023-06-21
core: frontend: ParameterEditor: Empty parameter file entry before the menu shows

patrickelectric:
Fix https://github.com/bluerobotics/BlueOS-docker/issues/1815


joaoantoniocardoso (approved):
I failed to follow the video, but the diff seems reasonable.

Issue #1539, opened 2023-03-13
Allow changing network priority 


patrickelectric:
When using a vehicle with multiple network interfaces: 4G, satellite, ethernet, wifi. We should allow the user to choose the priority between each one.

patrickelectric:
This appears to affect blue boat where internet access is not possible while the radio is on in some cases.

ES-Alexander:
It's be nice to also allow changing the priority [between wifi networks](https://www.raspberrypi.com/documentation/computers/configuration.html#adding-multiple-wireless-network-configurations). I'm not sure if that fits here or if it should be its own issue.

patrickelectric:
Yes, the idea is to show all network interfaces for the user to organize, exactly like macos.

ES-Alexander:
macOS standard network priority has “Wifi” as a single interface, which can be ordered relative to other interfaces like different ethernet ports.

Prioritisation of which wifi network to choose is handled separately, in an interface dedicated to just wifi selection, which is what I linked the equivalent of for raspberry pi. 

I’m not sure if you’re planning to “just” provide an interface for prioritisation between different network types (e.g. wifi/hotspot/usb-otg/Ethernet), or also allow prioritising between the known wifi networks.

patrickelectric:
> Prioritisation of which wifi network to choose is handled separately, in an interface dedicated to just wifi selection, which is what I linked the equivalent of for raspberry pi.

I'm not familiar to this interface, can you provide a screenshot ?
My idea is to show all interfaces.

ES-Alexander:
> I'm not familiar to this interface, can you provide a screenshot ?

It's apparently not possible anymore as of Ventura, but this is how it used to work:
https://www.howtogeek.com/532437/how-to-prioritize-wi-fi-networks-on-a-mac/

patrickelectric:
The network interface priority is changing after configuration

PR #1836, opened 2023-06-28
Allow user to change network interface prioprity

patrickelectric:
Adds a globe icon to show that internet access is available:

image

The same icon can be used to control network interface priority.

image

So.. The problem is that network priority is done by network route and not network and that makes things more confuse to show to the user.

```
root@blueos:~# route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.0.1     0.0.0.0         UG    202    0        0 eth0
0.0.0.0         192.168.86.1    0.0.0.0         UG    304    0        0 wlan0
172.18.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
192.168.0.0     0.0.0.0         255.255.255.0   U     202    0        0 eth0
192.168.2.0     0.0.0.0         255.255.255.0   U     0      0        0 eth0
192.168.3.0     0.0.0.0         255.255.255.0   U     0      0        0 usb0
192.168.86.0    0.0.0.0         255.255.255.0   U     304    0        0 wlan0
192.168.196.0   0.0.0.0         255.255.255.0   U     0      0        0 ztyxasoevg
```

So, a single interface can have multiple destinations with different priorities.
We could do the same thing as ifmetric and change the network priority for all routes.

```
root@blueos:~# ifmetric eth0 303
root@blueos:~# route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.0.1     0.0.0.0         UG    303    0        0 eth0
0.0.0.0         192.168.86.1    0.0.0.0         UG    304    0        0 wlan0
172.18.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
192.168.0.0     0.0.0.0         255.255.255.0   U     303    0        0 eth0
192.168.2.0     0.0.0.0         255.255.255.0   U     303    0        0 eth0
192.168.3.0     0.0.0.0         255.255.255.0   U     0      0        0 usb0
192.168.86.0    0.0.0.0         255.255.255.0   U     304    0        0 wlan0
192.168.196.0   0.0.0.0         255.255.255.0   U     0      0        0 ztyxasoevg
```

Is that correct ? How should we display the order based on the original version ?


rafaellehmkuhl:
Does the wireless interface appear on the Ethernet menu after this change?

patrickelectric:
> Does the wireless interface appear on the Ethernet menu after this change?

No, there's a new endpoint for all interfaces now.

core/frontend/src/components/app/InternetTrayMenu.vue
```diff
@@ -0,0 +1,68 @@
...
+        elevation="0"
+        color="transparent"
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon
+          color="white"
+          :title="tooltip"
+        >
+          {{ helper.has_internet ? 'mdi-earth' : 'mdi-earth-off' }}
```
ES-Alexander, line 23:
```suggestion
          {{ helper.has_internet ? 'mdi-web' : 'mdi-web-off' }}
```

core/tools/cable_guy/bootstrap.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+# Exit if something goes wrong
+set -e
+
+# Hostapd turns a network wireless interface into a wifi hotspot
+echo "Installing ifmetric."
+apt update
+apt install -y --no-install-recommends ifmetric
```
joaoantoniocardoso, line 9:
```suggestion
apt install -y --no-install-recommends ifmetric

```

joaoantoniocardoso (changes requested):
Some observations:
- Sometimes "Apply" does nothing to the route metrics.
- When "Apply" works, it always changes the value of the routes, even if the order is kept the same.
- Sometimes the order in the list is changed to the inverse of what is applied: I switch wlan0 to the second, then if I open the dialogue again, it's the first.
- The colors are very confusing to me, I have to think a lot to mentally keep track of what's happening when I've switched the position of two items. I recommend not using different colors, as the meaning should be the same as the written position. I think my little brain just expects the colors to mean each interface, not each position.
- Ipv6 metrics are not following Ipv4 metrics changes.

joaoantoniocardoso (commented):
From the previous observations, only one is still unfixed:
- Ipv6 metrics are not following Ipv4 metrics changes.

patrickelectric:
> From the previous observations, only one is still unfixed:
> 
> * Ipv6 metrics are not following Ipv4 metrics changes.

It should be working now, isn't it ?

joaoantoniocardoso:
> > From the previous observations, only one is still unfixed:
> > 
> > * Ipv6 metrics are not following Ipv4 metrics changes.
> 
> It should be working now, isn't it ?

nope, Ipv6 metrics are kept unchanged :(

patrickelectric:
> > > From the previous observations, only one is still unfixed:
> > > 
> > > * Ipv6 metrics are not following Ipv4 metrics changes.
> > 
> > 
> > It should be working now, isn't it ?
> 
> nope, Ipv6 metrics are kept unchanged :(

I believe that is a problem with the ifmetric and not this implementation, since we do not support IPv6 around our code base and this already fix the main issue, it should not be a holder.

joaoantoniocardoso (approved):
> > > > From the previous observations, only one is still unfixed:
> > > > 
> > > > * Ipv6 metrics are not following Ipv4 metrics changes.
> > > 
> > > 
> > > It should be working now, isn't it ?
> > 
> > 
> > nope, Ipv6 metrics are kept unchanged :(
> 
> I believe that is a problem with the ifmetric and not with this implementation, since we do not support IPv6 around our code base and this already fix the main issue, it should not be a holder.

I see, then just so we are aware, https://github.com/bluerobotics/BlueOS-docker/issues/1539 might not be solved for all the cases.

joaoantoniocardoso:
Just to be clear, the metrics changes can be done using `ip`, `ifmetric` is just a helper for ipv4.

joaoantoniocardoso:
And there's [this fork](https://github.com/vitek/ifmetrics), apparently with ipv6 support.

Issue #998, opened 2022-04-25
Core: Available services (helper) should prefer linking to nginx paths instead of ports


Williangalvani:
This will sort out CORS issues for people using our APIs in the frontend, and will avoid needing to remember random port numbers.

patrickelectric:
@Williangalvani just to be sure, this was not fixed by #1843, right ?

PR #1842, opened 2023-06-29
Create extension endpoints in nginx

### LINKED ISSUE #1401: It should be possible to have a hardlink to an extension
curently we use the port in the url, which is dynamically allocated.

Williangalvani:
image

fix #1401

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -630,12 +630,8 @@ export default Vue.extend({
       }
     },
     createExtensionAddress(service: Service): string {
-      if (service?.metadata?.new_page !== true) {
-        return `/extensions/${service.port}`
-      }
-
-      // The `//` force href to be used as absolute path and not as relative
-      return `//${window.location.hostname}:${service.port}`
+      const sanitized_name = service?.metadata?.sanitized_name
+      return `/extension/${sanitized_name}` ?? `/extensions/${service.port}`
     },
     setupCallbacks(): void {
       this.tourCallbacks = {
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -4,6 +4,8 @@ import asyncio
 import http
 import logging
 import os
+import re
+import subprocess
 from enum import Enum
 from pathlib import Path
 from typing import Any, Dict, List, Optional
@@ -58,6 +60,7 @@ class ServiceMetadata(BaseModel):
     route: Optional[str]
     new_page: Optional[bool]
     api: str
+    sanitized_name: Optional[str]
 
 
 class ServiceInfo(BaseModel):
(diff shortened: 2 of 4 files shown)


Williangalvani:
Downside of this approach is that it the extension/ endpoints will only be created after  a port scan is triggered.
We can make it run automatically every once in a while, I guess.

core/services/helper/main.py
```diff
@@ -159,6 +179,35 @@ async def check_internet_access() -> Dict[str, WebsiteStatus]:
         status_list = await asyncio.gather(*tasks)
         return {status.site.name: status for status in status_list}
 
+    @staticmethod
+    def setup_nginx_route(metadata: ServiceMetadata, port: int) -> bool:
+        name: str = re.sub(r"[^a-z0-9]", "", metadata.name.lower())
```
patrickelectric, line 184: We should store ServiceMetadata and avoid running this function to services already registered successfully.

core/services/helper/main.py
```diff
@@ -159,6 +179,35 @@ async def check_internet_access() -> Dict[str, WebsiteStatus]:
...
+        Path.mkdir(Path("/home/pi/tools/nginx/extensions/"), parents=True, exist_ok=True)
+        try:
+            with open(filename, "r", encoding="utf-8") as f:
+                if f.read() == text:
+                    return False
+        except Exception as e:
+            logging.info(f"file '{filename}' not found ({e}):, a new one will be created")
+        with open(filename, "w", encoding="utf-8") as f:
+            f.write(text)
+            print("file updated: ", filename)
```
patrickelectric, line 208: Should be using logging

patrickelectric (commented):
It's just missing access to the frontend now to the new urls.

ES-Alexander:
Nice! :-)

Relevant docs are likely just a comment saying how the links are generated - not sure if that fits best in `advanced-usage/#extensions-manager` or `development/extensions`

core/frontend/src/App.vue
```diff
@@ -630,12 +630,8 @@ export default Vue.extend({
       }
     },
     createExtensionAddress(service: Service): string {
-      if (service?.metadata?.new_page !== true) {
-        return `/extensions/${service.port}`
-      }
-
-      // The `//` force href to be used as absolute path and not as relative
-      return `//${window.location.hostname}:${service.port}`
+      const sanitized_name = service?.metadata?.name.toLowerCase().replace(/[^a-z0-9]/g, '')
```
patrickelectric, line 633: Since this is being done is helper, why not putting the path on the API ? With that we'll have less hardcoded references and we can deal with this path in others places, and avoiding doing regex around since the API will be the source of true.

patrickelectric:
Fix #998

Williangalvani:
This ends up introducing issues with absolute paths. we should probably use an iframe instead. That means we need to add routes dynamically to our vue router

Williangalvani:
this apparently will not be doable in nginx:
https://stackoverflow.com/questions/30350185/nginx-reverse-proxy-configuration-to-handle-absolute-paths

I'll close this as the alternative is a completely differnt approach

PR #1843, opened 2023-07-03
Create named links for extensions... again

### LINKED ISSUE #1401: It should be possible to have a hardlink to an extension

Williangalvani:
fix #1401


patrickelectric (commented):
Great approach!

patrickelectric (commented):
Great approach!

patrickelectric (approved):
Great approach!

PR #1850, opened 2023-07-05
Require manual mode for motor test

### LINKED ISSUE #1804: Motor tests when in Auto mode are dangerous
We should put the vehicle in Manual mode before starting the test, otherwise if there's a mission in the vehicle, the motors will spin.

From @raimullin14:

"Just a heads up. I was disarmed in auto mode on cockpit and when I armed it in this window it resumed the mission in cockpit and started running the thrusters".

image

joaoantoniocardoso:
Closes #1804

### DIFF
--- a/core/frontend/src/components/vehiclesetup/PwmSetup.vue
+++ b/core/frontend/src/components/vehiclesetup/PwmSetup.vue
@@ -132,7 +132,10 @@ import Vue from 'vue'
 import ParameterEditorDialog from '@/components/parameter-editor/ParameterEditorDialog.vue'
 import VehicleViewer from '@/components/vehiclesetup/viewers/VehicleViewer.vue'
 import mavlink2rest from '@/libs/MAVLink2Rest'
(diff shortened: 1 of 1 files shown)


Williangalvani (changes requested):
looks good, I just really dislike the hardcoded mode numbers there

core/frontend/src/components/vehiclesetup/PwmSetup.vue
```diff
@@ -269,6 +300,11 @@ export default Vue.extend({
         .map((_, i) => data[`servo${i + 1}_raw`])
     },
   },
+  watch: {
+    is_armed() {
+      this.desired_armed_state = this.is_armed
```
patrickelectric, line 305: please add a comment to explain the reason of the watch, since it's not clear that the idea is to track desired behavior change from others software like cockpit.

patrickelectric (changes requested):

patrickelectric (approved):

PR #1852, opened 2023-07-06
Fix beacon frontend

### LINKED ISSUE #1809: Frontend interface for beacon is not reflecting changes in the backend
image
image
image

joaoantoniocardoso:
Closes #1809

### DIFF
--- a/core/frontend/src/store/beacon.ts
+++ b/core/frontend/src/store/beacon.ts
@@ -80,6 +80,15 @@ class BeaconStore extends VuexModule {
 
   @Action
   async setHostname(hostname: string): Promise<boolean> {
+    hostname = hostname.trim().toLowerCase()
+    const regex = /^[a-zA-Z0-9-]+$/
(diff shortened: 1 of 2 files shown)


core/frontend/src/store/beacon.ts
```diff
@@ -33,7 +33,7 @@ class BeaconStore extends VuexModule {
 
   nginx_ip_address = ''
 
-  hostname: undefined | string = undefined
+  hostname = ''
```
patrickelectric, line 36: why is this an empty string ? with that there is no way for us to check if the value was proper set/valid or not.

joaoantoniocardoso: How is this different from the other properties?

patrickelectric: Well for IP consistency it makes sense.. but for valid check empty string is not the proper way.

patrickelectric (approved):

PR #1857, opened 2023-07-08
Optimize helper

### LINKED ISSUE #1855: [helper] Helper is eating CPU
It happens spontaneously when the system is being used, ~but it's easily replicable by reloading the BlueOS page 10 times in a row:~ correction: seems to happens at every each ~60 seconds

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/09303437-0f7c-4d15-a578-d4b519f3693e

This might be a source of instability.

joaoantoniocardoso:
## Changes

Most of the CPU was spent inside the HTTP requests done by `aiohttp` and `request`. By replacing them with the low-level `http.client`, both the speed and CPU usage were improved.

Closes #1855 

---

All development data and tools can be accessed [here](https://github.com/joaoantoniocardoso/BlueOS-docker/tree/new_helper_dev/core/services/helper/dev_benchmark)

## A/B test:
By doing several runs of the services, the port scanning is not idempotent, meaning that the list of available services can differ from one call to the other.

After analysis, it turns out that the only service that has different outputs is the Ping service (at port `9110`): sometimes it refuses to answer. Increasing the timeouts of the requests of the port_scan, the problem happens less.

Apart from this service, if the skipped ports are ignored, the results are equal both before and after this PR.

## Benchmark

The implementations are:
- `main_opt`: the code on this PR initially
- `main_master`: the code before this PR
- `main_master_with_skip_list`: the skip list patch applied to the code before this PR
- `main_opt2`: the code on this PR, but with [this suggestion](https://github.com/bluerobotics/BlueOS-docker/pull/1857#discussion_r1257488131)
- `main_httpx`: the code on [this other PR](https://github.com/bluerobotics/BlueOS-docker/pull/1858), using httpx to replace both aiohttp and requests
- `main_opt3`: the code on this PR at some point, adding the ethernal cache list suggested by `main_httpx` (disabled during these tests).
- `main_opt4`: the code on this PR at some point, with longer timeouts and 2 concurrent port scans.
- `main_opt4_cached`: the same as the `main_opt4` plus caches added for blueos services and an improved algorithm to decide what ports to scan.

### Speed
image

image

### CPU%
image

image

## Summary of improvements

* Speed of `check_internet_access` increased by **132x** on average (of cache hit + miss), and **1.99x** in the worst case (cache miss).
* Speed of `web_services` (the port scan) increased by **58x** on average (of cache hit + miss), and **1.55x** in the worst case (cache miss).
* CPU usage of `check_internet_access` decreased by **1.61x** on average (of cache hit + miss), and **1.71x** in the worst case (cache miss).
* CPU usage of `web_services` decreased by **1.36x** on average (of cache hit + miss), and increased by **0.97x** in the worst case (cache miss).
* Because of the increased timeouts on the port scan requests, it's more forgiven when services are too busy, thus it's more reliable.
* Because of the added and rebalancing in caches, it's more robust and responsive.

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -363,18 +363,36 @@ class Helper:
     @staticmethod
     @temporary_cache(timeout_seconds=10)
     def scan_ports() -> List[ServiceInfo]:
-        # Filter for TCP ports that are listen and can be accessed by external users (server in 0.0.0.0)
-        connections = (
-            connection
+        # Get TCP ports that are listen and can be accessed by external users (like server in 0.0.0.0, as described by the LOCALSERVER_CANDIDATES)
+        ports = {
+            connection.laddr.port
             for connection in psutil.net_connections("tcp")
             if connection.status == psutil.CONN_LISTEN and connection.laddr.ip in Helper.LOCALSERVER_CANDIDATES
-        )
+        }
 
-        # And check if there is a webpage available that is not us
-        # Use it as a set to remove duplicated ports
-        ports = set(connection.laddr.port for connection in connections)
-        services = [Helper.detect_service(port) for port in ports if port != PORT]
-        return [service for service in services if service.valid]
+        # If a known service is not within the detected ports, we remove it from the known services
+        if Helper.KEEP_BLUEOS_SERVICES_ALIVE:
+            Helper.KNOWN_SERVICES = {
+                service
+                for service in Helper.KNOWN_SERVICES
+                if service.port in ports or service.port in Helper.BLUEOS_SYSTEM_SERVICES_PORTS
+            }
+        else:
+            Helper.KNOWN_SERVICES = {service for service in Helper.KNOWN_SERVICES if service.port in ports}
+
+        # Filter out ports we want to skip, as well as the ports from services we already know, assuming the services don't change,
+        known_ports = {service.port for service in Helper.KNOWN_SERVICES}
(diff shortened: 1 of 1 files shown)


core/services/helper/main.py
```diff
@@ -121,94 +140,191 @@ class SpeedTestResult(BaseModel):
...
+        ]
 
         # And check if there is a webpage available that is not us
         # Use it as a set to remove duplicated ports
         ports = set(connection.laddr.port for connection in connections)
-        services = [Helper.detect_service(port) for port in ports if port != PORT]
-        return [service for service in services if service.valid]
+        services = [
+            Helper.detect_service(port) for port in ports if port != Helper.PORT and port not in Helper.SKIP_PORTS
+        ]
```
joaoantoniocardoso, line 271: We could experiment with doing this using the concurrent future executor here as well, which might shape the CPU usage as a higher but shorter peak, reducing the overall response time.

joaoantoniocardoso:
@patrickelectric @Williangalvani CI is failing due to code quality metrics, I accept suggestions on how to fix it.

ES-Alexander:
> CI is failing due to code quality metrics, I accept suggestions on how to fix it.

Not sure if you made fixes in your latest PR, but at the moment the only failed code quality stuff is two `pass` lines that aren't doing anything (because there is existing code in those blocks), so removing those two lines should fix the issue.

patrickelectric (approved):

PR #1858, opened 2023-07-10
core: services: helper: Improve speed and cpu usage

patrickelectric:
Replaces #1857 

Closes https://github.com/bluerobotics/BlueOS-docker/issues/1855


core/services/helper/main.py
```diff
@@ -154,28 +191,35 @@ def detect_service(port: int) -> ServiceInfo:
...
                                 if "swagger-ui" in response.text:
                                     info.versions += [path]
                 break
 
             except Exception:
                 # This should be avoided by the first try block, but better safe than sorry
                 break
+
+        if info.valid:
+            Helper.FOUND_BLUEOS_SERVICES[port] = info
```
joaoantoniocardoso, line 221: We don't want to cache all the services, only those that we know are from BlueOS:
```suggestion
        if info.valid and port in Helper.BLUEOS_PORTS:
            Helper.FOUND_BLUEOS_SERVICES[port] = info
```

joaoantoniocardoso (changes requested):
## A/B test:

There's a change in the algorithm to cache all known services, meaning that if a service dies, it keeps showing as an available option. 

Ignoring that, it passes the A/B test just like to main_opt and main_master, with the same observations written in [the other PR](https://github.com/bluerobotics/BlueOS-docker/pull/1857#issue-1795091170).

---

## Benchmark

Compared to http.client, the CPU usage with httpx doesn't look good, but because the change in the algorithm, the speed after the first request of 'web_services' is fantastic:

### Speed

image

To compare apples-to-apples, I reverted the change in the algorithm:
```diff
-        if port in Helper.FOUND_BLUEOS_SERVICES:
-            return Helper.FOUND_BLUEOS_SERVICES[port]
+        #if port in Helper.FOUND_BLUEOS_SERVICES:
+            #return Helper.FOUND_BLUEOS_SERVICES[port]
```

and the results are the following:

image

### CPU Usage

image

## Conclusion

In our case, httpx is worse than both aiohttp and requests, but the change in the algorithm looks promising, specially if combined with the lightweight of the `http.client` from the other PR.

PR #1873, opened 2023-07-13
bootstrap: add missing 'return False'  in is_running()

Williangalvani:
not tested yet, but a docker is up at  williangalvani/blueos-bootstrap:addreturn

### DIFF
--- a/bootstrap/test_bootstrap.py
+++ b/bootstrap/test_bootstrap.py
@@ -218,3 +218,12 @@ class BootstrapperTests(TestCase):  # type: ignore
         bootstrapper.run()
         self.mock_response.json.return_value = {"repository": ["core"]}
         assert bootstrapper.is_running("core")
+
+    @pytest.mark.timeout(10)
+    def test_bootstrap_start_invalid_json(self) -> None:
+        self.fs.create_file(Bootstrapper.DEFAULT_FILE_PATH, contents=SAMPLE_JSON)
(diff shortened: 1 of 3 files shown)


patrickelectric (approved):

bootstrap/test_bootstrap.py
```diff
@@ -160,23 +171,30 @@ def test_start_core_no_config(self) -> None:
...
+    def test_is_running(self) -> None:
         fake_client = FakeClient()
         bootstrapper = Bootstrapper(fake_client, FakeLowLevelAPI())
+
+        fake_client.set_active_dockers([])
         assert bootstrapper.is_running("core") is False
+
         fake_core = FakeContainer(Bootstrapper.CORE_CONTAINER_NAME)
         fake_client.set_active_dockers([fake_core])
-        assert bootstrapper.is_running("core")
```
Williangalvani, line 171: this line should not have been removed

Williangalvani: nevermind, I read the patch wrong =|

patrickelectric (changes requested):
It's restarting core every 12s
image

patrickelectric (approved):
After further testing I got it working, it was a problem with my binds

Williangalvani:
I am actually not so sure. I'll add a test for the timeout in another pr just to be sure

PR #1880, opened 2023-07-18
bootstrap: bootstrap: Add core definition

patrickelectric:
The `is_running` logic was broken. It was checking both frontend response and docker response, since version chooser was not online on the first seconds it was restarting it over and over again.

This PR breaks the logic of `is_running` to check only for docker containers, and a single function to check if `core` has version chooser running inside a timeout. 

Fix https://github.com/bluerobotics/BlueOS/issues/1876


Williangalvani (approved):
NICE!
(tested and works nicely)

PR #1887, opened 2023-07-25
core: frontend: App: NetworkInterfacePriorityMenu: Fix card size

patrickelectric:
Now:

image

After:

image


joaoantoniocardoso (approved):
Great, I saw that but forgot to fix it. Is the other tab (DNS) correct?

patrickelectric:
> Great, I saw that but forgot to fix it. Is the other tab (DNS) correct?

yep

PR #1902, opened 2023-07-26
core: services: cable_guy: Move interface priority from ipr to dhcpcd

### LINKED ISSUE #1539: Allow changing network priority

patrickelectric:
Fix #1539

Right now to have it working it's required to do a system reboot.
Doing dhcpcd service restart breaks our wpa_supplicant connection. #1958 was an attempt to fix that.

### DIFF
--- a/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
+++ b/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
@@ -105,9 +105,9 @@ export default Vue.extend({
         .then((response) => {
           const interfaces = response.data as EthernetInterface[]
           interfaces.sort((a, b) => {
-            if (!a.info) return 1
-            if (!b.info) return -1
-            return b.info.priority - a.info.priority
+            if (!a.info) return -1
+            if (!b.info) return 1
+            return a.info.priority - b.info.priority
           })
           this.interfaces = interfaces
         })
--- a/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
+++ b/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
@@ -95,6 +95,7 @@ export default Vue.extend({
           )
           this.close()
         })
+      await this.fetchAvailableInterfaces()
     },
     async fetchAvailableInterfaces(): Promise<void> {
       await back_axios({
--- a/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
+++ b/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
@@ -31,6 +31,13 @@
     <v-divider />
     <v-card-actions class="justify-center pa-2">
       <v-btn
+        color="primary"
+        @click="close()"
+      >
+        Cancel
+      </v-btn>
+      <v-spacer />
+      <v-btn
         color="success"
         :disabled="is_loading"
         @click="setHighestInterface()"
--- a/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
+++ b/core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
@@ -3,7 +3,7 @@
     <v-list>
       <v-card-subtitle class="text-md-center" max-width="30">
         Move network interfaces over
-        to change network access priority
+        to change network access priority.<br>Applied changes require a <b>system reboot</b>.
       </v-card-subtitle>
       <draggable v-model="interfaces">
         <v-card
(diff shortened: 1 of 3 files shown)


core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
```diff
@@ -3,7 +3,7 @@
     <v-list>
       <v-card-subtitle class="text-md-center" max-width="30">
         Move network interfaces over
-        to change network access priority
+        to change network access priority.<br>Applied changes require a <b>system reboot</b>.
```
Williangalvani, line 6: can we make it show the reboo-required icon we use for parameter changes? even in bold this is quite subtle.
a popup after saving saying "configurations will be applied on the next boot" should work well, too.

core/services/cable_guy/api/manager.py
```diff
@@ -396,71 +415,182 @@ def get_interfaces_priority(self) -> List[NetworkInterfaceMetric]:
...
+
+        List[NetworkInterfaceMetric]: A list of priority metrics for each interface.
         """
-        metric: NetworkInterfaceMetric
-        for metric in self.get_interfaces_priority():
-            if interface_name == metric.name:
-                return metric
+        lines = self._get_service_dhcpcd_content()
+        result = []
+        current_interface = None
```
Williangalvani, line 476: did you miss the type here? it looks like you missed the type, and that caused the errors in line 480 and 482 that you suppressed.

core/services/cable_guy/api/manager.py
```diff
@@ -396,71 +415,182 @@ def get_interfaces_priority(self) -> List[NetworkInterfaceMetric]:
...
+                sorted by priority to set, if values are undefined.
+        """
+        self._set_interfaces_priority_to_dhcpcd(interfaces)
+
+    def _set_interfaces_priority_to_dhcpcd(self, interfaces: List[NetworkInterfaceMetricApi]) -> None:
+        """Sets network interface priority..
 
         Args:
-            interfaces (List[NetworkInterfaceMetricApi]): A list of interfaces and their priority metrics.
+            interfaces (List[NetworkInterfaceMetricApi]): List of network interfaces
```
Williangalvani, line 547: the priority metrics are still there, why was the description changed?

core/services/cable_guy/api/manager.py
```diff
@@ -396,71 +415,182 @@ def get_interfaces_priority(self) -> List[NetworkInterfaceMetric]:
             else:
                 metric_dict[d["index"]] = d["metric"]
 
+        # Highest priority wins for ipr but not for dhcpcd, so we sort and revet the list
```
Williangalvani, line 418:
```suggestion
        # Highest priority wins for ipr but not for dhcpcd, so we sort and reverse the list
```

core/frontend/src/components/app/NetworkInterfacePriorityMenu.vue
patrickelectric: > can we make it show the reboo-required icon we use for parameter changes? even in bold this is quite subtle. Or maybe a popup after saving saying "configurations will be applied on the next boot", if that is easier.

Yes, but that idea is under https://github.com/bluerobotics/BlueOS/pull/1962, to be a global notifier.

Williangalvani (changes requested):
It is working well, nice. Just had some minor questions

Williangalvani:
btw, I tested with `ip route get <external ip>`
that shows what interface the packet would be routed through to get at the destination

core/services/cable_guy/api/manager.py
patrickelectric: Yes, this is unrelated, the source of the issue is from mypy.
https://github.com/python/mypy/issues/14987

joaoantoniocardoso:
Here it's not changing the metrics, and `ip route get` is always trying through the same interface (eth0)

patrickelectric:
> Here it's not changing the metrics (seeing using `ip -4 route` and `ip -6 route`), and `ip route get` is always trying through the same interface (eth0)

did you reboot ?

joaoantoniocardoso:
> > Here it's not changing the metrics (seeing using `ip -4 route` and `ip -6 route`), and `ip route get` is always trying through the same interface (eth0)
> 
> did you reboot ?

I did not. Oops

joaoantoniocardoso:
https://github.com/bluerobotics/BlueOS/assets/5920286/c33f34d1-d39f-4d3b-bc0b-eee8dc7f3ff1

joaoantoniocardoso (approved):

PR #1910, opened 2023-07-28
core: frontend: Fix video controls odd feedback when updating

joaoantoniocardoso:
No more confusing UI feedback when changing the video controls :tada: 

## Before

https://github.com/bluerobotics/BlueOS/assets/5920286/258e81d4-f869-465d-ac94-7e6584241493

## After

https://github.com/bluerobotics/BlueOS/assets/5920286/3889581e-0ddc-413e-b8e4-f4b4e18fa951

### DIFF
--- a/core/frontend/src/components/video-manager/VideoControlsDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoControlsDialog.vue
@@ -150,7 +150,10 @@ export default Vue.extend({
       await this.updateControlsValues(updated_controls)
     },
     async updateControlsValues(controls: Control[]): Promise<void> {
-      for await (const control of controls) {
+      const totalControls = controls.length
+      let completedControls = 0
+
+      for (const control of controls) {
         let value = 0
         if ('Menu' in control.configuration) {
           value = control.configuration.Menu.value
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/video-manager/VideoControlsDialog.vue
```diff
@@ -160,28 +169,36 @@ export default Vue.extend({
...
+              value,
+            },
+          })
+        } catch (error) {
           const message = `Could not update value on ${control.name} control: ${error}.`
           notifier.pushError('CONTROL_VALUE_UPDATE_FAIL', message)
-        })
-      }
+        } finally {
+          completedControls += 1
```
patrickelectric, line 187: How can `completedControls` be equal to `totalControls` if it only runs once ?

joaoantoniocardoso: Here, `finally` runs for every request, so with the pre-incremented `completedControls`, it will be equal to `totalControls` when all requests are finished.

patrickelectric: I missed the loop, thanks.

patrickelectric (approved):

PR #1928, opened 2023-08-03
core: frontend: device-path-helper: reduce default size

Williangalvani:
before:
image
after:
image

### DIFF
--- a/core/frontend/src/components/common/DevicePathHelper.vue
+++ b/core/frontend/src/components/common/DevicePathHelper.vue
@@ -80,7 +80,7 @@ export default Vue.extend({
     height: {
       type: String,
       required: false,
-      default: '500px',
+      default: '250px',
     },
   },
   data: () => ({


ES-Alexander:
If the problem is a Mac DPI thing then this fix might cause problems on Windows devices - we’ll need someone to confirm.

The Navigator images in the drop-down thingo are well-sized - we should check if there was something special applied there or if it’s just a set pixel scale.

core/frontend/src/components/common/DevicePathHelper.vue
```diff
@@ -80,7 +80,7 @@ export default Vue.extend({
     height: {
       type: String,
       required: false,
-      default: '500px',
+      default: '250px',
```
patrickelectric, line 83: I think that a better solution is to use `em` over `px`, since `em` is relative to the text size. But I remember a discussion with @rafaellehmkuhl about that a time ago.

patrickelectric:
Navigator appears smaller, but possible to understand. Can someone that does not have a 4k screen or high resolution check it out ?
image

core/frontend/src/components/common/DevicePathHelper.vue
Williangalvani: browser pixels are not real pixels. also I use 100% scaling on my linux setup (where these prints were taken)

patrickelectric (approved):
It's still big in a normal 1280 display, but it's ok.

PR #1932, opened 2023-08-03
core: frontend: do not return 80 if port is unknown

Williangalvani:
should fix the blueos-in-blueos issue

williangalvani/blueos-core:no-port-80

### DIFF
--- a/core/frontend/src/views/ExtensionView.vue
+++ b/core/frontend/src/views/ExtensionView.vue
@@ -33,8 +33,8 @@ export default Vue.extend({
         (service) => service?.metadata?.sanitized_name === this.$route.params.name,
       )[0] ?? undefined
     },
-    port(): number {
-      return this.service?.port ?? 80
+    port(): number | undefined {
+      return this.service?.port
     },
     service_path(): string {
       return `${window.location.protocol}//${window.location.hostname}:${this.detected_port}`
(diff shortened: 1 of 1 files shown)


core/frontend/src/views/ExtensionView.vue
```diff
@@ -27,10 +34,13 @@ export default Vue.extend({
     fullpage(): boolean {
       return this.$route.query.full_page === 'true'
     },
-    port(): number {
+    port(): number | undefined | null {
```
Williangalvani, line 37: is it obvious what it means? or should I add a comment?
number = port
undefined = we dont know (still loading)
null = nope

core/frontend/src/views/ExtensionView.vue
```diff
@@ -4,18 +4,25 @@
     :source="service_path"
     :class="fullpage ? 'fullpage' : ''"
   />
+  <SpinningLogo v-else-if="detected_port === undefined" />
```
patrickelectric, line 7: can we put v-else-if and v-else in the beggning of the component for better readability ?

Like:
`<v-else page-not-found/>`

patrickelectric (commented):
Check comment

core/frontend/src/views/ExtensionView.vue
Williangalvani: that doesn't seem to be valid syntax

patrickelectric (approved):

PR #1936, opened 2023-08-04
core: frontend: components: Firmware: Only show update button when the latest firmware is newer

patrickelectric:


Williangalvani (approved):
tricky to test, but code looks good.

PR #1945, opened 2023-08-04
core: services: versionchooser: Add more checks and debug information

patrickelectric:

### DIFF
--- a/core/services/versionchooser/utils/chooser.py
+++ b/core/services/versionchooser/utils/chooser.py
@@ -237,6 +237,8 @@ class VersionChooser:
                 core = await self.client.containers.get("blueos-core")  # type: ignore
                 if core:
                     await core.kill()
+                    result = await core.wait()  # type: ignore
+                    logger.info(f"Response after waiting for core to be killed: {result}")
                 return web.Response(status=200, text=f"Changed to version {image}:{tag}, restarting...")
 
             except KeyError:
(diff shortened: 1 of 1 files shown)


core/services/versionchooser/utils/chooser.py
```diff
@@ -236,6 +239,8 @@ async def set_version(self, image: str, tag: str) -> web.StreamResponse:
                 core = await self.client.containers.get("blueos-core")  # type: ignore
                 if core:
                     await core.kill()
+                    result = await core.wait()  # type: ignore
+                    logger.info(f"Response while waiting core to be killed: {result}")
```
Williangalvani, line 243:
```suggestion
                    logger.info(f"Response after waiting for core to be killed: {result}")
```

core/services/versionchooser/utils/chooser.py
```diff
@@ -183,6 +184,8 @@ async def set_bootstrap_version(self, tag: str) -> web.StreamResponse:
             await bootstrap.rename(backup_name)
             logger.info(f"Stop {self.bootstrap_name}")
             await bootstrap.stop()
+            result = await bootstrap.wait()  # type: ignore
+            logger.info(f"Response while waiting {self.bootstrap_name} to be stopped: {result}")
```
Williangalvani, line 188:
```suggestion
            logger.info(f"Response after waiting  for {self.bootstrap_name} to be stopped: {result}")
```

Williangalvani (changes requested):
small change. It sounds nit-picky,  but the current wording made me think there was a loop waiting for those things to happen, which doesn't seem to be the case

Williangalvani (approved):

PR #1946, opened 2023-08-04
core: frontend: create SelfHealthTest, add test for factory images

### LINKED ISSUE #1041: The user should be warned when bootstrap reverts to "Factory" version
You may not notice it if you don't look at version chooser, which is a large oversight on our part

Williangalvani:
image

fix #1041

### DIFF
--- a/core/frontend/src/components/health/SelfHealthTest.vue
+++ b/core/frontend/src/components/health/SelfHealthTest.vue
@@ -1,0 +1,60 @@
+<template>
+  <v-container>
+    <v-row
+      class="mb-6 mt-6"
+      justify="center"
+      no-gutters
+    >
+      <v-alert
+        v-if="is_running_factory"
+        border="top"
+        colored-border
+        type="warning"
+        elevation="2"
+        dismissible
+      >
+        You are running "factory" version.
+        This generally means something went wrong and the system reverted to this version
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/health/SelfHealthTest.vue
```diff
@@ -0,0 +1,60 @@
...
+        elevation="2"
+        dismissible
+      >
+        You are running "factory" version.
+        This generally means something went wrong and the system reverted to this version
+        in order to recover.
+        Please file an issue at <a
+          href="https://github.com/bluerobotics/BlueOS/issues"
+          target="_blank"
+        >BlueOS/issues</a>
```
joaoantoniocardoso, line 22: Here are the query parameters, we should, by default, use a template and at least fill current version information. I looked at that one software I commented about that creates an issue with the report already filled, and it was embedding the report as HTML in between `<detail>` tags, not uploading the file.

Docs: https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-url-query

Williangalvani:
I'll follow this up with improving logging in bootstrap, then we can link to the logs as well

patrickelectric (approved):

PR #1950, opened 2023-08-08
bootstrap: add logging

Williangalvani:
This moves bootstrap to loguru, and populates a file at the same place our other logs are

### DIFF
--- a/bootstrap/main.py
+++ b/bootstrap/main.py
@@ -4,15 +4,19 @@ import os
 import sys
 
 import docker
-
+from loguru import logger
 from bootstrap.bootstrap import Bootstrapper
 
+
 if __name__ == "__main__":
+    logger.add("/var/logs/blueos/services/bootstrap/bootstrap_{time}.log", enqueue=True, rotation="30 minutes")
     if os.environ.get("BLUEOS_CONFIG_PATH", None) is None:
-        print("Please supply the host path for the config files as the BLUEOS_CONFIG_PATH environment variable.")
-        print("Example docker command line:")
-        print(
-            "docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/.config/blueos:"
+        logger.info("Please supply the host path for the config files as the BLUEOS_CONFIG_PATH environment variable.")
+        logger.info("Example docker command line:")
+        logger.info(
+            "docker run -it -v /var/run/docker.sock:/var/run/docker.sock"
+            " -v $HOME/.config/blueos:"
+            " -v /var/logs/blueos:/var/logs/blueos"
             "/root/.config/blueos -e BLUEOS_CONFIG_PATH=$HOME/.config/blueos"
             "bluerobotics/blueos-bootstrap:master"
         )
--- a/bootstrap/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap/bootstrap.py
@@ -10,6 +10,7 @@ from warnings import warn
 
 import docker
 import requests
+from loguru import logger
 
 
 class Bootstrapper:
@@ -68,7 +69,7 @@ class Bootstrapper:
                     assert key in config[Bootstrapper.SETTINGS_NAME_CORE], f"missing key in json file: {key}"
 
         except Exception as error:
-            print(f"unable to read startup.json file ({error}), reverting to defaults...")
+            logger.error(f"unable to read startup.json file ({error}), reverting to defaults...")
             # Copy defaults over and read again
             Bootstrapper.overwrite_config_file_with_defaults()
             with open(Bootstrapper.DEFAULT_FILE_PATH, encoding="utf-8") as config_file:
@@ -137,7 +138,7 @@ class Bootstrapper:
             curses.echo()
             curses.nocbreak()
             curses.endwin()
-        print("Done")
+        logger.info("Done")
 
     def image_is_available_locally(self, image_name: str, tag: str) -> bool:
         """Checks if the image is already available locally"""
@@ -176,7 +177,7 @@ class Bootstrapper:
                 warn(f"Error trying to pull an update image: {error}")
                 return False
 
-        print(f"Starting {image_name}")
+        logger.info(f"Starting {image_name}")
         # Remove image if name already exist
         self.remove(component_name)
         try:
@@ -193,7 +194,7 @@ class Bootstrapper:
             self.overwrite_config_file_with_defaults()
             return False
 
-        print(f"{component_name} ({docker_name}) started")
+        logger.info(f"{component_name} ({docker_name}) started")
         return True
 
     def is_running(self, component: str) -> bool:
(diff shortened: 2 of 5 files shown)


bootstrap/main.py
```diff
@@ -4,15 +4,19 @@
 import sys
 
 import docker
-
+from loguru import logger
 from bootstrap.bootstrap import Bootstrapper
 
+
 if __name__ == "__main__":
+    logger.add("/var/logs/blueos/services/bootstrap/bootstrap.log", level="DEBUG", rotation="1 MB")
```
patrickelectric, line 12: Will this work if BlueOS rotator runs ?

Williangalvani: I have no idea. should I just take this out? I guess the other services don't any rotation by themselves, right?

patrickelectric: Let it run and check if there is a rotation and it's working fine, right now we have a default rotation of 10min:
https://github.com/bluerobotics/BlueOS/blob/master/core/libs/commonwealth/commonwealth/utils/logs.py#L63

And a zipper every hour:
https://github.com/bluerobotics/BlueOS/blob/master/core/start-blueos-core#L78

patrickelectric:
I just tested the last version and it appears to be not working:

image

patrickelectric:
> I just tested the last version and it appears to be not working:
> 
> image

How is the rotation is done exactly in this case ? For log_zipper, it'll zip the rotated logs.

patrickelectric:
> > > I just tested the last version and it appears to be not working:
> > > image
> > 
> > 
> > How is the rotation is done exactly in this case ? For log_zipper, it'll zip the rotated logs older than X, but this file from what I can see is never rotated.
> 
> it relies on log_zipper, too. but that is running on 1 hour intervals, apparently.

My point is, if log_zipper runs, bootstrap will not generate new logs.

Williangalvani:
> I just tested the last version and it appears to be not working:
> 

I was told we don't delete files while in use. There doesn't seem to be anything within loguru to deal with that.

I updated the code so it rotated every 10min, and the files are now timestamped, so they won't get reused.

joaoantoniocardoso:
> > I just tested the last version and it appears to be not working:
> 
> I was told we don't delete files while in use. There doesn't seem to be anything within loguru to deal with that.
> 
> I updated the code so it rotated every 10min, and the files are now timestamped, so they won't get reused.

That's true, all the other services have the same problem.

patrickelectric:
> > I just tested the last version and it appears to be not working:
> 
> I was told we don't delete files while in use. There doesn't seem to be anything within loguru to deal with that.

As pointed here https://github.com/bluerobotics/BlueOS/pull/1950#discussion_r1287459469.

In core, we rotate logs every 10 minutes, so the log being used will be T-10min.
log_zipper rotates logs older than 1h, So T-60min.
We don't zip open log files because they are too new to be zipped.




> 
> I updated the code so it rotated every 10min, and the files are now timestamped, so they won't get reused.

Williangalvani:
> > > I just tested the last version and it appears to be not working:
> > 
> > 
> > I was told we don't delete files while in use. There doesn't seem to be anything within loguru to deal with that.
> 
> As pointed here [#1950 (comment)](https://github.com/bluerobotics/BlueOS/pull/1950#discussion_r1287459469).
> 
> In core, we rotate logs every 10 minutes, so the log being used will be T-10min. log_zipper rotates logs older than 1h, So T-60min. We don't zip open log files because they are too new to be zipped.
> 
> > I updated the code so it rotated every 10min, and the files are now timestamped, so they won't get reused.
> 
> The codes says 10MB not 10min. https://github.com/bluerobotics/BlueOS/pull/1950/files#diff-4d1dd542128fe2565a2688dc51e6e6bcc45fea5d94a4c4a476ef917fdd4c278fR12

lol, I mean't MB but wrote minutes. I think this will result in 1 log per boot, which should be perfectly fine

patrickelectric:
> > > > I just tested the last version and it appears to be not working:
> > > 
> > > 
> > > I was told we don't delete files while in use. There doesn't seem to be anything within loguru to deal with that.
> > 
> > 
> > As pointed here [#1950 (comment)](https://github.com/bluerobotics/BlueOS/pull/1950#discussion_r1287459469).
> > In core, we rotate logs every 10 minutes, so the log being used will be T-10min. log_zipper rotates logs older than 1h, So T-60min. We don't zip open log files because they are too new to be zipped.
> > > I updated the code so it rotated every 10min, and the files are now timestamped, so they won't get reused.
> > 
> > 
> > The codes says 10MB not 10min. [#1950 (files)](https://github.com/bluerobotics/BlueOS/pull/1950/files#diff-4d1dd542128fe2565a2688dc51e6e6bcc45fea5d94a4c4a476ef917fdd4c278fR12)
> 
> lol, I mean't MB but wrote minutes. I think this will result in 1 log per boot, which should be perfectly fine

As said, if log_zipper runs, the bootstrap log file will be zipped, and bootstrap will not log anymore.

image
This is after log_zipper runs.

Williangalvani:
how's this being handled by the other services?

patrickelectric:
> how's this being handled by the other services?

https://github.com/bluerobotics/BlueOS/pull/1950#issuecomment-1673734257

Williangalvani:
> > how's this being handled by the other services?
> 
> [#1950 (comment)](https://github.com/bluerobotics/BlueOS/pull/1950#issuecomment-1673734257)

The longer answer is "We initiate loguru with a [custom LogRotator](https://github.com/bluerobotics/BlueOS/blob/master/core/libs/commonwealth/commonwealth/utils/logs.py#L11). That makes Loguru itself rotate the files every 10 minutes. this way, the open file can never be older than 10 minutes.

When log_zipper runs, it looks for [files last modified earlier than 60 minutes ago](https://github.com/bluerobotics/BlueOS/blob/master/core/services/log_zipper/main.py#L56).

This means setting Loguru to 60 minutes should work, right? so we don't need to add commonwealth to Bootstrap.

patrickelectric:
> > > how's this being handled by the other services?
> > 
> > 
> > [#1950 (comment)](https://github.com/bluerobotics/BlueOS/pull/1950#issuecomment-1673734257)
> 
> The longer answer is "We initiate loguru with a [custom LogRotator](https://github.com/bluerobotics/BlueOS/blob/master/core/libs/commonwealth/commonwealth/utils/logs.py?rgh-link-date=2023-08-10T19%3A12%3A09Z#L11). That makes Loguru itself rotate the files every 10 minutes. this way, the open file can never be older than 10 minutes.
> 
> When log_zipper runs, it looks for [files last modified earlier than 60 minutes ago](https://github.com/bluerobotics/BlueOS/blob/master/core/services/log_zipper/main.py?rgh-link-date=2023-08-10T19%3A12%3A09Z#L56).
> 
> This means setting Loguru to 60 minutes should work, right? so we don't need to add commonwealth to Bootstrap.

anything less than 60min should work, I would recommend to use between 30min and 10min.

patrickelectric (commented):
tested and working

patrickelectric (approved):

PR #1959, opened 2023-08-10
core: start-blueos-core: Use alphabetic order for services initialization

patrickelectric:


joaoantoniocardoso (changes requested):
I'll just oppose myself so we don't end up merging it without testing the startup strategies since we changed it already in #2020.

PR #1962, opened 2023-08-10
Notify user if BlueOS has no access to the host computer

### LINKED ISSUE #1960: Do not fail startup if the user changed default raspberry password
There are users that changed the raspberry password and the system works in 1.0.1 and from what I tested in beta.23;

We should be able to start BlueOS and alert the user about that in the frontend.

patrickelectric:
Fix #1960 

image
image

### DIFF
--- a/core/frontend/src/components/app/SystemCheckerTrayMenu.vue
+++ b/core/frontend/src/components/app/SystemCheckerTrayMenu.vue
@@ -1,0 +1,135 @@
+<template>
+  <v-menu
+    v-if="visible"
+    v-model="show_menu"
+    :close-on-content-click="false"
+    nudge-left="200"
+    nudge-bottom="25"
+  >
+    <template
+      #activator="{ on, attrs }"
+    >
+      <v-card
+        class="px-1"
+        elevation="0"
+        color="transparent"
+        v-bind="attrs"
+        v-on="on"
+      >
+        <v-icon
+          class="blinking"
+        >
+          mdi-alert
+        </v-icon>
+      </v-card>
+    </template>
+
+    <v-dialog v-model="show_menu" max-width="700px">
+      <v-card class="text-md-center">
+        <v-card-title>
+          Problems found
+        </v-card-title>
+        <v-card-text>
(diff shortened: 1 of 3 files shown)


Williangalvani:
while this fixes the issue, it doesn't fix the underlying issue with how we do things. I'll experiment with ssh keys, if that doesn't work or gets too complex or dangerous, we should get this in

patrickelectric:
> while this fixes the issue, it doesn't fix the underlying issue with how we use a hardcoded password. I'll experiment with ssh keys, if that doesn't work or gets too complex or dangerous, we should get this in

Keep in mind that stable is feature freeze, we are not merging new features.

core/frontend/src/components/app/SystemCheckerTrayMenu.vue
```diff
@@ -0,0 +1,135 @@
...
+
+      // From: https://github.com/kevinburke/sshpass/blob/master/main.c#L43
+      let human_message
+      switch (output.return_code) {
+        case 0: return undefined
+        case 1: human_message = 'Invalid arguments were provided'; break
+        case 2: human_message = 'Conflicting arguments were provided'; break
+        case 3: human_message = 'A runtime error occurred'; break
+        case 4: human_message = 'An error occurred while parsing'; break
+        case 5: human_message = 'The password provided is incorrect, it should be "raspberry"'; break
```
Williangalvani, line 102: I think we need a clearer message here, actually letting the user know what the issue was and how it was caused

```suggestion
        case 5: human_message = 'Login failed using the password "raspberry", was the "pi" user password changed? For the time being BlueOS requires the host computer username:password to be pi:raspberry'; break
```

I'm. actually not happy with my wording either, so I'll summon @ES-Alexander for help.

patrickelectric: We can merge at it's to test the functionality and fix the text later, the important thing here to test is the logic, the text can be fixed before tagging stable.

Williangalvani (approved):

PR #1964, opened 2023-08-10
Fix telemetry system

patrickelectric:
Fix https://github.com/bluerobotics/BlueOS/issues/1965

The current architecture is not working with port 80, HTTP. So the code was changed to support 443, HTTPS.


Williangalvani (approved):
Code looks fine, but an approval by @voorloopnul would be ideal. unless you can test yourself, then merge at will

Issue #952, opened 2022-04-06
bug: Default password is raspberry


patrickelectric:
### Bug description

We need to change the default password since it makes it easy for internet attacks 

### Steps to reproduce

sudo su, password is raspberry

### Primary pain point(s)

raspberry as default password is unsafe 

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

patrickelectric:
We could add a password as env variable to be accessible in BlueOS.
We could also have an alert on the frontend asking if the user changed the password.

patrickelectric:
beta.23 works if the users changes the password with minor bugs, but  beta.27 just breaks.

patrickelectric:
This appears to be fixed in latest master version, the user can change the default password after installing blueos.

Issue #854, opened 2022-03-04
Host OS username and password are hardcoded in some random places


rafaellehmkuhl:
It would be good to have this somewhere on the installation script and import the config for both from some fixed place.

_Originally posted by @rafaellehmkuhl in https://github.com/bluerobotics/BlueOS-docker/pull/853#discussion_r819738129_

PR #1968, opened 2023-08-11
Switch to using an ssh key for accessing the host computer

Williangalvani:
This adds the mount  `/home/pi/.ssh` into core.
if an id_rsa file already exists, it is used. if not, we create a new one.

we then switch commander to use the ssh key instead of the password.

The old sshpass behavior is kept as a fallback, as some things (cough blueos_startup_update) run before commander does the ssh setup.

helps #854 and #952

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/commands.py
+++ b/core/libs/commonwealth/commonwealth/utils/commands.py
@@ -1,7 +1,16 @@
 import subprocess
+from pathlib import Path
+
+from loguru import logger
 
 
-def run_command(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
+class KeyNotFound(Exception):
+    """Raised when the SSH key is not found."""
+
+
+def run_command_with_password(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
+    # attempt to run the command with sshpass
+    # used as a fallback if the ssh key is not found
     user = "pi"
     password = "raspberry"
 
@@ -21,3 +30,41 @@ def run_command(command: str, check: bool = True) -> "subprocess.CompletedProces
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
     )
+
+
+def run_command_with_ssh_key(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
+    # attempt to run the command with the ssh key
+    user = "pi"
+    id_file = f"/home/{user}/.ssh/id_rsa"
+    if not Path(id_file).exists():
+        raise KeyNotFound
+
+    return subprocess.run(
+        [
+            "sshpass",
+            "ssh",
+            "-i",
+            id_file,
+            "-o",
+            "StrictHostKeyChecking=no",
+            f"{user}@localhost",
+            command,
+        ],
+        check=check,
+        text=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+
+
+def run_command(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
+    # runs the given command on the host computer.
+    # we first try with the ssh key, which is the default behavior.
+    # we need to fallback to sshpass as some systems will try to call this function before the ssh key is generated.
+    # this is the case for the first boot of this image after updating.
+    # not including the sshpass step causes blueos_startup_update to fail hard. crashing BlueOS as a whole.
+    try:
+        return run_command_with_ssh_key(command, check)
+    except Exception as error:
+        logger.warning(f"Failed to run command with SSH key. {error}, trying with sshpass:\n{command}")
+        return run_command_with_password(command, check)
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -1,7 +1,9 @@
 #! /usr/bin/env python3
 import logging
 import os
+import shlex
 import shutil
+import subprocess
 import time
 from enum import Enum
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


core/services/commander/main.py
```diff
@@ -204,7 +206,25 @@ async def root() -> Any:
...
+        if not os.path.exists("/home/pi/.ssh/id_rsa.pub"):
+            subprocess.run(shlex.split("ssh-keygen -t rsa -f /home/pi/.ssh/id_rsa -q -N ''"), check=True)
+        # add id_rsa.pub to authorized_keys if not there already
+        if not os.path.exists("/home/pi/.ssh/authorized_keys"):
+            subprocess.run(shlex.split("touch /home/pi/.ssh/authorized_keys"), check=True)
+        with open("/home/pi/.ssh/authorized_keys", "r+", encoding="utf-8") as f:
+            if not any("ssh-rsa" in line for line in f):
+                with open("/home/pi/.ssh/id_rsa.pub", encoding="utf-8") as key_file:
+                    f.write(key_file.read())
+    except Exception as error:
```
patrickelectric, line 221: Will this work if there is already a ssh-rsa key configured by the user ?

core/libs/commonwealth/commonwealth/utils/commands.py
```diff
@@ -21,3 +34,35 @@ def run_command(command: str, check: bool = True) -> "subprocess.CompletedProces
...
+
+
+def run_command_with_ssh_key(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
+    # attempt to run the command with the ssh key
+    user = "pi"
+    id_file = f"/home/{user}/.ssh/id_rsa"
+    if not file_exists(id_file):
+        raise KeyNotFound
+
+    ssh_command = ["ssh", "-o", "StrictHostKeyChecking=no", "-i", id_file, f"{user}@localhost", command]
```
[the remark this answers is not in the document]
Williangalvani: we could, but the purpose of sshpass is to pass the text via "fake keyboard input", so I don't see the why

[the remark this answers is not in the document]
patrickelectric: > we could, but the purpose of sshpass is to pass the text via "fake keyboard input", so I don't see the why

The error code of both binaries are different, handling such errors on the frontend, without knowing which backend was used to communicate with the host computer, will be more problematic. We could fix that by adding which backend was used on the command answer to the front.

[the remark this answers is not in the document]
Williangalvani: > > we could, but the purpose of sshpass is to pass the text via "fake keyboard input", so I don't see the why
> 
> The error code of both binaries are different, handling such errors on the frontend, without knowing which backend was used to communicate with the host computer, will be more problematic. We could fix that by adding which backend was used on the command answer to the front.

good point, I'll switch to sshpass

core/services/commander/main.py
Williangalvani: yep, I deleted it and created one in the pi. it still worked fine

patrickelectric (changes requested):
red-pill is failing and asking for password
image

It appears that sshpass is also failing during blueos startup
image

Williangalvani:
> commander: image

I managed to stash my most recent changes instead of pushing them 🤡

Williangalvani:
latest changes stop creating a key in /home/pi/.ssh, and creates it in /roo/.config/.ssh instead. that means we don't interfere with the pi user. we still mount the pi's .ssh, but only for adding our key to authorized_keys

core/services/commander/main.py
```diff
@@ -204,7 +206,33 @@ async def root() -> Any:
...
+def setup_ssh() -> None:
+    # store the key in the docker .config volume
+    key_path = "/root/.config/.ssh/"
+    # authorized_keys is in the pi home directory, we just add our key to it
+    authorized_keys_path = "/home/pi/.ssh/authorized_keys"
+
+    try:
+        Path("key_path").mkdir(parents=True, exist_ok=True)
+        # check if id_rsa.pub exists, creates a new one if it doesnt
+        if not os.path.exists(Path(key_path) / "id_rsa.pub"):
```
patrickelectric, line 218: This should use the same Path object to avoid recreating it.

core/services/commander/main.py
```diff
@@ -204,7 +206,33 @@ async def root() -> Any:
...
+            subprocess.run(shlex.split(f"mkdir -p {key_path}"), check=True)
```
patrickelectric, line 219: Shouldn't this be done in line 216 ? I fixed.
`Path(key_path).mkdir(parents=True, exist_ok=True)`

core/services/commander/main.py
```diff
@@ -204,7 +206,33 @@ async def root() -> Any:
...
+            subprocess.run(shlex.split(f"ssh-keygen -t rsa -f {key_path}/id_rsa -q -N ''"), check=True)
+        # add id_rsa.pub to authorized_keys if not there already
+        if not os.path.exists(authorized_keys_path):
+            subprocess.run(shlex.split(f"touch {authorized_keys_path}"), check=True)
+        with open(authorized_keys_path, "r+", encoding="utf-8") as f:
+            with open(Path(key_path) / "id_rsa.pub", encoding="utf-8") as key_file:
+                my_key = key_file.read()
+                if not any(my_key in line for line in f):
+                    f.write(my_key)
```
patrickelectric, line 228: Need to ensure that the file has an empty line before adding it.

core/libs/commonwealth/commonwealth/utils/commands.py
```diff
@@ -1,7 +1,20 @@
...
+from loguru import logger
 
-def run_command(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
+
+class KeyNotFound(Exception):
+    """Raised when the SSH key is not found."""
+
+
+def file_exists(file_name: str) -> bool:
+    return Path(file_name).exists()
```
patrickelectric, line 12: This functions appears to be unnecessary to be used in a single place, and without much text to replace.

patrickelectric (changes requested):

core/services/commander/main.py
```diff
@@ -204,7 +206,33 @@ async def root() -> Any:
...
 
 
+def setup_ssh() -> None:
+    # store the key in the docker .config volume
+    key_path = "/root/.config/.ssh/"
+    # authorized_keys is in the pi home directory, we just add our key to it
+    authorized_keys_path = "/home/pi/.ssh/authorized_keys"
+
+    try:
+        Path("key_path").mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
Williangalvani: yep!

core/services/commander/main.py
Williangalvani: done

core/services/commander/main.py
Williangalvani: yep!

core/libs/commonwealth/commonwealth/utils/commands.py
Williangalvani: fair enough, I got rid of it.

core/services/commander/main.py
Williangalvani: as [empty lines are ignored](https://www.ibm.com/docs/en/zos/2.5.0?topic=daemon-format-authorized-keys-file), I just added a newline to keep to code short.

patrickelectric (approved):
nice!

PR #1973, opened 2023-08-11
Allow user to skip wizard configuration steps

### LINKED ISSUE #1947: [Wizard] Add "skip" ability to every step.
Can either be done with a button + timeout, or a checklist to select the steps in the beginning

patrickelectric:
Fix #1947 

https://github.com/bluerobotics/BlueOS/assets/1215497/0171e819-1dac-4ee8-8e38-fed9d96b9492

### DIFF
--- a/core/frontend/src/components/wizard/ActionStepper.vue
+++ b/core/frontend/src/components/wizard/ActionStepper.vue
@@ -1,19 +1,25 @@
 <template>
-  <v-stepper vertical>
-    <v-stepper-step
-      v-for="(config, index) in configurations"
-      :key="index"
-      :step="index + 1"
-      :color="getStepColor(config)"
-      :complete-icon="getStepIcon(config)"
-      :complete="true"
-      active
-      class="step-label"
-    >
-      {{ config.title }}
-      <small v-if="config.summary">{{ config.summary }}</small>
-      <small v-if="config.message" :color="getStepColor(config)">Error: {{ config.message }}</small>
-    </v-stepper-step>
+  <v-stepper vertical min-width="100%">
+    <div v-for="(config, index) in configurations" :key="index" class="step-container">
+      <v-stepper-step
+        :key="index"
+        :step="index + 1"
+        :color="getStepColor(config)"
+        :complete-icon="getStepIcon(config)"
+        :complete="true"
+        active
+        class="step-label"
+      >
+        {{ config.title }}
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/wizard/ActionStepper.vue
```diff
@@ -49,6 +59,9 @@ export default Vue.extend({
       return 'mdi-alert'
     },
     getStepColor(config: Configuration) {
+      if (config.skip) {
+        return 'black'
```
joaoantoniocardoso, line 63: This looks a bit odd... Can we add an 'X' or a '/' over the icon maybe?

joaoantoniocardoso (approved):
The black icon looks odd to me, but can be personal taste. 

Another thing is that the configuration step is happening on the "Done" step, but it's just another detail.

Despite that
image

patrickelectric:
> Another thing is that the configuration step is happening on the "Done" step, but it's just another detail.

How so ? The logic behind it did not change, done is only applied once everything is accomplished.

core/frontend/src/components/wizard/ActionStepper.vue
joaoantoniocardoso: maybe gray for a neutral feel?

patrickelectric: > maybe gray for a neutral feel?

I did try gray, but the contrast was terrible on white mode. black was the one that worked fine for white and dark mode.

joaoantoniocardoso:
> > Another thing is that the configuration step is happening on the "Done" step, but it's just another detail.
> 
> How so ? The logic behind it did not change, done is only applied once everything is accomplished.

I see, maybe the name of the step should be "Apply" instead of "Done".

patrickelectric:
> > > > Another thing is that the configuration step is happening on the "Done" step, but it's just another detail.
> > > 
> > > 
> > > How so ? The logic behind it did not change, done is only applied once everything is accomplished.
> > 
> > 
> > I see, maybe the name of the step should be "Apply" instead of "Done".
> 
> There is no Done step, where are you seeing it ?
> 
> There is "Done" state, and the "apply" operation.
> 
> But this is not related to this PR at all, this is the logic that is already on the code.
> 
> Since the feature only implies on the skip operation, we can discuss that later if valid.

I see what you are implying now, you are talking about the stepper called Done, will update to Apply.

patrickelectric:
image

PR #1976, opened 2023-08-11
core: frontend: components: DnsConfigurationMenu: Update timeouts for host_dns

patrickelectric:
It takes around 8 seconds in a pi 3


joaoantoniocardoso (approved):
TY

PR #1988, opened 2023-08-14
Configure timeouts on aiohttp.ClientSession

Williangalvani:
The defaults go up to 5 minutes... that means 5 minutes with no user feedback.
not yet tested


patrickelectric:

PR #2011, opened 2023-08-21
core: Use always in git describe

patrickelectric:
From the helper:
```
       --always
           Show uniquely abbreviated commit object as fallback.
```

With that, people that fork the project will not be obligated to push a tag or something for the build to work, the hash will be used as fallback.


Williangalvani (approved):

RaulTrombin:
I was able to run action after create a simple tag on master branch.
Just checked, apparently you only need a first tag to make it works.
image
image

RaulTrombin (approved):

PR #2020, opened 2023-08-22
core: Improve startup time by starting high priority services first

joaoantoniocardoso:
This represents a 12% time reduction.
> note: Reduces QGC connection from 102 seconds to 90 seconds total boot time.

### DIFF
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -55,18 +55,27 @@ else
     echo "Waring: Using dockers's resolv.conf, this may cause DNS issues if the host's network configuration changes."
 fi
 
+# These services have priority because they do the fundamental for the vehicle to work,
+# and by initializing them first we reduce the time users have to wait to control the vehicle.
+# From tests with QGC and Pi3, the reboot time was ~1min42s when not using this strategy,
+# and ~1min30s using this strategy.
+# From that 1min30s, the startup time is about ~25s, and originally, ~37s, meaning that the
+# remaining (~65 seconds) is the docker shutting down, and the Linux booting up.
+PRIORITY_SERVICES=(
+    'autopilot',"$SERVICES_PATH/ardupilot_manager/main.py"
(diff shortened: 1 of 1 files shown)


core/start-blueos-core
```diff
@@ -55,18 +55,24 @@ else
     echo "Waring: Using dockers's resolv.conf, this may cause DNS issues if the host's network configuration changes."
 fi
 
+# These services have priority because they do the fundamental for the vehicle to work, and by initializing them first
```
rafaellehmkuhl, line 58: "Those"

rafaellehmkuhl (approved):

joaoantoniocardoso:
I'll test a bit more, and see if that sleep is really needed.

joaoantoniocardoso:
Okay, without the 5 seconds sleep the time increases by ~8 seconds, ending with a total net time decrement of 4 seconds only (~4%). So I believe it is worth keeping. 

I also measured the time it takes from starting the priority services to have a connection on QGC and it is about 5 seconds, so 5 seconds of sleep is probably not far from the optimal sleep time.

With that said, I'm opening it again for review.

joaoantoniocardoso:
Oh, just to mention, from that 1min30s, the startup time is actually about 25s, as can be seen in the video below. The CPU profile can be compared with the one from [here](https://github.com/bluerobotics/BlueOS/issues/1808):

https://github.com/bluerobotics/BlueOS/assets/5920286/6816abce-c524-4b98-b63e-21a7eff74efb

patrickelectric:
So.. what's the real gain with this patch ? 4% over 12% ?

joaoantoniocardoso:
> So.. what's the real gain with this patch ? 4% over 12% ?

If we keep the 5 seconds sleep, it's 12%, otherwise, 4%

patrickelectric (approved):

patrickelectric:
Helps: https://github.com/bluerobotics/BlueOS/issues/1489

PR #2034, opened 2023-08-24
play nicely

Williangalvani:
just for testing purposes


core/start-blueos-core
```diff
@@ -62,29 +62,29 @@ fi
...
+    'video',"nice --19 mavlink-camera-manager --default-settings BlueROVUDP --mavlink tcpout:127.0.0.1:5777 --gst-feature-rank omxh264enc=0,v4l2h264enc=250,x264enc=260 --log-path /var/logs/blueos/services/mavlink-camera-manager --verbose"
     'mavlink2rest',"mavlink2rest --connect=udpin:127.0.0.1:14000 --server 0.0.0.0:6040 --system-id $MAV_SYSTEM_ID --component-id $MAV_COMPONENT_ID_ONBOARD_COMPUTER4"
 )
 
 SERVICES=(
     'beacon',"$SERVICES_PATH/beacon/main.py"
-    'kraken',"$SERVICES_PATH/kraken/main.py"
-    'wifi',"$SERVICES_PATH/wifi/main.py --socket wlan0"
-    'bridget',"$SERVICES_PATH/bridget/main.py"
+    'kraken',"nice -19 $SERVICES_PATH/kraken/main.py"
```
patrickelectric, line 73: You probably want 19 and not -19, -19 is almost highest priority

core/start-blueos-core
```diff
@@ -62,29 +62,29 @@ fi
 # From that 1min30s, the startup time is about ~25s, and originally, ~37s, meaning that the
 # remaining (~65 seconds) is the docker shutting down, and the Linux booting up.
 PRIORITY_SERVICES=(
-    'autopilot',"$SERVICES_PATH/ardupilot_manager/main.py"
+    'autopilot',"nice --19 $SERVICES_PATH/ardupilot_manager/main.py"
```
patrickelectric, line 65: this should be -19 over --19

patrickelectric (changes requested):

core/start-blueos-core
joaoantoniocardoso: `-19` sets to 19, `--19` sets to `-19`

Williangalvani: yep, apparently I messed up nginx too

Williangalvani: wait, no, joao is right =P. but I still messed up nginx

joaoantoniocardoso:
I've been testing, everything seems to be working fine.

joaoantoniocardoso (approved):

patrickelectric:
Are there performance gains on that ?
Will this play nicely with heavy operations, like remove log or any other that can trigger an endpoint timeout on the frontend ?

joaoantoniocardoso:
> Are there performance gains on that ? Will this play nicely with heavy operations, like remove log or any other that can trigger an endpoint timeout on the frontend ?

Yes, there are gains, but the methodology to measure is too hard. Better to have faith in knowledge: it essentially gives more CPU time to the video encoders and the vehicle control service, making the others wait when there's no CPU time for everyone.

When a video jitter occurs, the perception for the user is obvious, it is very easy for them to feel, the amount of time we are exchanging here is milliseconds -- enough for jitter, but it is too small to impact an endpoint response, especially now that we have 5 minutes.

patrickelectric:
> > Are there performance gains on that ? Will this play nicely with heavy operations, like remove log or any other that can trigger an endpoint timeout on the frontend ?
> 
> Yes, there are gains, but the methodology to measure is too difficult. Better to have faith in knowledge: it essentially gives more CPU time to the video encoders and the vehicle control service, making the others wait when there's no CPU time for everyone.
> 
> When a video jitter occurs, the perception for the user is obvious, it is very easy for them to feel, the amount of time we are exchanging here is milliseconds -- enough for jitter, but it is too small to impact an endpoint response, especially now that we have 5 minutes.

I'm worried about timeouts in the backend for services running in low priority when communicating with the frontend.

Williangalvani:
> I'm worried about timeouts in the backend for services running in low priority when communicating with the frontend.

Any particular one?
I only made nice the ones where I thought it couldn't be harmful

joaoantoniocardoso:
> > > Are there performance gains on that ? Will this play nicely with heavy operations, like remove log or any other that can trigger an endpoint timeout on the frontend ?
> > 
> > 
> > Yes, there are gains, but the methodology to measure is too difficult. Better to have faith in knowledge: it essentially gives more CPU time to the video encoders and the vehicle control service, making the others wait when there's no CPU time for everyone.
> > When a video jitter occurs, the perception for the user is obvious, it is very easy for them to feel, the amount of time we are exchanging here is milliseconds -- enough for jitter, but it is too small to impact an endpoint response, especially now that we have 5 minutes.
> 
> I'm worried about timeouts in the backend for services running in low priority when communicating with the frontend.

I believe the chance of that happening is the same as it already is today.

Williangalvani:
so.. let's get this in?

voorloopnul:
Hey guys, let me hijack this conversation, what are your thoughts on having the services as containers?

We could leverage the docker API for resource management (https://docs.docker.com/config/containers/resource_constraints/), and even update the values while running to meet the requirements of specific tasks/tools.

Like, lower CPU for container X, Y, Z, and increase for C if C is recording/streaming video.

It could also potentially benefit from other things like Log management and monitoring...

It feels like the abstraction on top of process management is all there.

@patrickelectric @joaoantoniocardoso @Williangalvani @rafaellehmkuhl @RaulTrombin

joaoantoniocardoso:
Another alternative is to use systemd, and if needed, systemd also has its own containers...

PR #2036, opened 2023-08-24
propagate errors from docker pull to the frontend

Williangalvani:
Please try to break this.

### DIFF
--- a/core/services/versionchooser/utils/chooser.py
+++ b/core/services/versionchooser/utils/chooser.py
@@ -129,8 +129,12 @@ class VersionChooser:
         await response.prepare(request)
 
         # Stream every line of the output back to the client
-        async for line in self.client.images.pull(f"{repository}:{tag}", repo=repository, tag=tag, stream=True):
-            await response.write(json.dumps(line).encode("utf-8"))
+        try:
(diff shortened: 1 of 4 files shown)


core/services/versionchooser/utils/chooser.py
```diff
@@ -144,8 +148,11 @@ async def get_bootstrap_version(self) -> str:
         Returns:
             str: The image version string for the bootstrap container.
         """
-        bootstrap = await self.client.containers.get(self.bootstrap_name)  # type: ignore
-        return str(bootstrap["Config"]["Image"])
+        try:
+            bootstrap = await self.client.containers.get(self.bootstrap_name)  # type: ignore
+            return str(bootstrap["Config"]["Image"])
+        except Exception as e:
+            return "Unknown"
```
patrickelectric, line 155: Can't we just f-string the error here ?
Besides that, we should log the error as well.

patrickelectric (changes requested):

core/services/versionchooser/utils/chooser.py
Williangalvani: f-string-ing the error could cause a huge text to show up in the sidebar, that's why I opted not to...
I'll add logging

patrickelectric (approved):
image

Issue #1824, opened 2023-06-27
[extension manager] Updating an extension is not stopping the older one


joaoantoniocardoso:
I've just updated Cockpit from 0.1.0 to 0.2.0 and now both are running:

https://github.com/bluerobotics/BlueOS-docker/assets/5920286/cb08cf5f-ea98-4977-86d9-a44a0ae3196e

rafaellehmkuhl:
Reopening the issue as it's still happening in the newest version.

rafaellehmkuhl:
Log from the extension manager that shows the bug happening (installation of new version of Cockpit).

[logfile_08-22-2023_16_40_26.log](https://github.com/bluerobotics/BlueOS/files/12439449/logfile_08-22-2023_16_40_26.log)

PR #2040, opened 2023-08-25
core: services: kraken: Wait for delete operation to finish before running

patrickelectric:
Should help with #1824

### DIFF
--- a/core/services/kraken/kraken.py
+++ b/core/services/kraken/kraken.py
@@ -23,6 +23,7 @@ class Kraken:
         self.load_settings()
         self.running_containers: List[DockerContainer] = []
         self.should_run = True
+        self.deleting_in_progress = False
         self._client: Optional[aiodocker.Docker] = None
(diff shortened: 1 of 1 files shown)


core/services/kraken/kraken.py
```diff
@@ -35,6 +36,9 @@ def client(self) -> aiodocker.Docker:
     async def run(self) -> None:
         while self.should_run:
             await asyncio.sleep(5)
+            if self.deleting_in_progress:
+                continue
+
```
rafaellehmkuhl, line 41: Could this lead to a loophole where kraken gets stuck in?

patrickelectric: logically, no, the `remove` operation ends setting `deleting_in_progress` as `False`.

rafaellehmkuhl: As long as it doesn't raise, yup.

Williangalvani (approved):

PR #2047, opened 2023-08-25
core: Add alias to prevent exit

### LINKED ISSUE #2035: bug: [Terminal] : User can't use terminal anymore if execute "exit"
### Bug description

User cannot regain control in the terminal if misuse the exit command.

### Steps to reproduce

CPT2308241522-1166x980

### Primary pain point(s)

_No response_

### Prerequisites

joaoantoniocardoso:
Fixes #2035

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -74,5 +74,8 @@ RUN ln -s /usr/blueos/userdata /shortcuts/userdata
 RUN ln -s /root/.config/ardupilot-manager /shortcuts/ardupilot_logs
 RUN ln -s / /shortcuts/system_root
 
+# Add alias to prevent user to exit from the terminal (they can still exit by using "\exit")
+RUN echo "alias exit='reset'" > /root/.bashrc
+
 # Start
 CMD /bin/bash -i /usr/bin/start-blueos-core && sleep infinity


patrickelectric (approved):

joaoantoniocardoso:
> I'm not 100% ok with this solution because it not covers others cases as `ctrl+d`, also.. if an user does want to exist from a tmux splitted, how he should do it ? The solution may be better to be only applied on the `user_session` tmux.

Sorry, I'm not familiar with Tmux at that level, I think I didn't understand the cases you are describing.

The user can still use the original `exit` by typing `\exit`, as usual.

This method, however, will make scripts using `exit 0` or `exit 1` no longer work, possibly affecting our update script.

The available environment for us are:
```bash
SHELL=/bin/bash
TMUX=/tmp/tmux-0/default,35,16
HOSTNAME=blueos
PYTHON_VERSION=3.9.17
MAV_SYSTEM_ID=1
PWD=/root
PYTHON_SETUPTOOLS_VERSION=58.1.0
HOME=/root
LANG=C.UTF-8
GIT_DESCRIBE_TAGS=1.1.0-beta.28-9-g3fc16d15
GPG_KEY=E3FF2839C048B25C084DEBE9B26995E310250568
TERM=screen
TMUX_PANE=%16
SHLVL=2
PYTHON_PIP_VERSION=23.0.1
PYTHON_GET_PIP_SHA256=45a2bb8bf2bb5eff16fdd00faef6f29731831c7c59bd9fc2bf1f3bed511ff1fe
PYTHON_GET_PIP_URL=https://github.com/pypa/get-pip/raw/9af82b715db434abb94a0a6f3569f43e72157346/public/get-pip.py
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
MAV_COMPONENT_ID_ONBOARD_COMPUTER4=194
OLDPWD=/root
_=/usr/bin/env
```

So, maybe we could `reset` if `TMUX`, and `exit` otherwise?

Issue #2048, opened 2023-08-25
bug: file browser: after "Download files" popup closes, the main browser pane is still unselectable


clydemcqueen:
### Bug description

If I select 2 files and download them, I get a popup asking me to choose a format (zip, etc.) This popup grays out the pane behind it, as expected.

After the popup disappears, the pane is still grayed out / unselectable.

![image](https://github.com/bluerobotics/BlueOS/assets/4961326/bafd54fe-47a6-4674-ba1d-f33c106b81cd)

BlueOS Version: [tags/1.1.0-beta.28-0-g2df6ae03](https://github.com/bluerobotics/blueos-docker/releases/tag/1.1.0-beta.28)
Bootstrap Version: 1.0.1
Build: 8/12/2023, 8:51:38 PM
By [Blue Robotics](https://bluerobotics.com/)

Filebrowser 2.15.0

Client:
Ubuntu 22.04
Version 116.0.5845.110 (Official Build) (64-bit)

### Steps to reproduce

1. select "File browser"
2. select 2 or more files or folders
3. select a format in the popup, e.g., "zip"
4. the file browser pane is grayed out / unselectable

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

joaoantoniocardoso:
I can reproduce it on Chrome-based browsers. It seems to not be a problem on Firefox, so you can try it as a workaround until we fix it.

Issue #1951, opened 2023-08-08
bug: [File-Manager] : User cannot undo menu actions


RaulTrombin:
### Bug description

User cannot undo the upload action when clicked out of it.
![CPT2308081120-1174x724](https://github.com/bluerobotics/BlueOS/assets/80598030/707badc4-7fdb-4ea7-acb4-b10ff9bee97e)

### Steps to reproduce

Files Manager > Upload > Just click out to undo.
OR
cancel the file selection.

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

RaulTrombin:
could it be related to this issue? 
https://github.com/filebrowser/filebrowser/issues/550

PR #2049, opened 2023-08-25
core: tools: Update filebrowser to v2.24.2

joaoantoniocardoso:
It might help #1951 and #2048

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -67,7 +67,7 @@ RUN RCFILE_PATH="/etc/blueosrc" \
     && echo "export GIT_DESCRIBE_TAGS=$GIT_DESCRIBE_TAGS" >> $RCFILE_PATH
 
 # create some Symlinks to make life easier
-RUN mkdir /shortcuts
+RUN mkdir -p /shortcuts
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
Confirmed: 
- closes #1951 
- closes #2048

core/Dockerfile
```diff
@@ -67,7 +67,7 @@ RUN RCFILE_PATH="/etc/blueosrc" \
     && echo "export GIT_DESCRIBE_TAGS=$GIT_DESCRIBE_TAGS" >> $RCFILE_PATH
 
 # create some Symlinks to make life easier
-RUN mkdir /shortcuts
+RUN mkdir -p /shortcuts
```
Williangalvani, line 70: has this ever happened? that is weird.
not a bad patch, I'm just wondering how it could happen.

Williangalvani (approved):
tested, NICE

core/Dockerfile
joaoantoniocardoso: Yeah, I wonder, initially I thought it was some weird cache from GitHub, but it was also happening on my computer without cache :eyes:

PR #2056, opened 2023-08-26
core: tools: linux2rest: Use latest version 0.4.8 <- 0.4.6

patrickelectric:


joaoantoniocardoso:
From sysinfo crate [changelog](https://github.com/GuillaumeGomez/sysinfo/blob/master/CHANGELOG.md):

> # 0.29.8
>  *  Linux: Fix overflow bug.
>
> # 0.28.2
>  *  Linux: Improve CPU usage computation.

joaoantoniocardoso (approved):
Seems to be better, but it still measures peaks instead of averages.

PR #2066, opened 2023-08-28
core: frontend: wizard: make sure a default param set is chosen

Williangalvani:
[Screencast from 2023-08-28 14-51-27.webm](https://github.com/bluerobotics/BlueOS/assets/4013804/3923c207-296e-4831-8c57-daa217fa835a)

### DIFF
--- a/core/frontend/src/components/wizard/DefaultParamLoader.vue
+++ b/core/frontend/src/components/wizard/DefaultParamLoader.vue
@@ -1,16 +1,19 @@
 <template>
   <div class="d-flex flex-column align-center ma-5" style="width: 100%; height: 100%;">
-    <v-select
-      v-model="selected_param_set_name"
-      :items="filtered_param_sets_names"
-      item-text="sanitized"
-      item-value="full"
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/wizard/DefaultParamLoader.vue
```diff
@@ -135,6 +139,15 @@ export default Vue.extend({
...
   methods: {
+    // this is used by Wizard.vue, but eslint doesn't dectect it
+    // eslint-disable-next-line
+    validateParams(): boolean {
+      const element = this.$refs.select as VForm
+      return element.validate()
+    },
+    isNotEmpty(value : string): boolean {
+      return value !== ''
+    },
```
patrickelectric, line 150:
```suggestion
    // this is used by Wizard.vue, but eslint doesn't detect it
    // eslint-disable-next-line
    validateParams(): boolean {
      const element = this.$refs.select as VForm
      return element.validate()
    },
    isNotEmpty(value: string): boolean {
      return value !== ''
    },
```

patrickelectric (changes requested):

patrickelectric (approved):

PR #2073, opened 2023-08-29
Revert "core: tools: linux2rest: Use latest version 0.4.8 <- 0.4.6"

patrickelectric:
Reverts bluerobotics/BlueOS#2056

0.4.8 appears to have problems with cpu frequency and temperature sensor


joaoantoniocardoso:
BTW, the wrong readings still happen on beta 30

image

joaoantoniocardoso (approved):
image

CPU/temperature problem solved.

PR #2074, opened 2023-08-29
Use requirements.txt to install requirements

Williangalvani:
this saves 2 minutes in the full docker build:

master: 
```
docker buildx build --no-cache --build-arg  --build-arg  --platform  . -t   1,79s user 1,42s system 0% cpu 9:56,63 total
```
this patch:
```
docker buildx build --no-cache --progress=plain --build-arg  --build-arg    .  0,68s user 0,44s system 0% cpu 7:59,90 total
```
I also ran into some weird conflicts, which required me to update uvicorn, click, and requests


core/services/requirements.txt
```diff
@@ -0,0 +1,76 @@
...
+typing_extensions==4.7.1
+urllib3==1.26.16
+uvicorn==0.23.2
+validators==0.18.2
+websocket-client==1.6.2
+Werkzeug==2.2.3
+winregistry==1.1.1
+yarl==1.9.2
+zeroconf==0.38.4
+zipp==3.16.2
```
patrickelectric, line 76: I personally think that is not a good decision, we are going to lose track where each package is necessary and decrease organization. This doesn't sound valid for a 2min tradeoff.

patrickelectric: We could parse and generate this file from the content of others setup.py

patrickelectric:
Something that would help even more is the docker cache system to work reliable.

Williangalvani:
> Something that would help even more is the docker cache system to work reliable.

this has nothing to do with docker cache. it all happens in the same layer.

core/services/requirements.txt
Williangalvani: > We could parse and generate this file from the content of others setup.py, maintaining organization and getting the 2min.

that sounds reasonable. ideally we'd use something like virtual envs for each one, but I think all implementations will result in duplicated files.
too bad this is still unresolved: https://github.com/bluerobotics/BlueOS/issues/75#issuecomment-771726745

Williangalvani:
I'll work on a better solution soon.

Williangalvani:
> I'll work on a better solution soon.

What was the better Solution, @Williangalvani ? master broke again...

joaoantoniocardoso:
> > I'll work on a better solution soon.
> 
> What was the better Solution, @Williangalvani ? master broke again...

image

PR #2081, opened 2023-09-04
add /cloud/join for integration with BlueOS Cloud

Williangalvani:
still WIP, we intend to keep the page hidden while we work on it.

### DIFF
--- a/core/frontend/src/views/CloudView.vue
+++ b/core/frontend/src/views/CloudView.vue
@@ -1,0 +1,81 @@
+<template>
+  <v-card
+    elevation="0"
+    class="mx-auto my-12 px-1 py-6 text-center"
+    min-width="370"
+    max-width="720"
+  >
+    <v-card>
+      <v-card-title>Cloud Registration</v-card-title>
+      <v-card-text>
+        <p>
+          In order to register your vehicle, access <a href="https://app.blueos.cloud/vehicle/register/" taget="_blank">
+            BlueOS Cloud
+          </a>
+          And create a new vehicle, then paste the token below.
+        </p>
+      </v-card-text>
+
+      <v-card-text>
+        <strong>Cloud Token:</strong>
+        <v-text-field v-model="cloud_token" outlined />
+      </v-card-text>
+      <v-btn @click="join">
+        Save
+      </v-btn>
+      <v-card-text />
+    </v-card>
+  </v-card>
+</template>
+<script>
+import axios from 'axios'
+import Vue from 'vue'
+
+import Notifier from '@/libs/notifier'
+import bag from '@/store/bag'
+import { cloud_service } from '@/types/frontend_services'
+
+const notifier = new Notifier(cloud_service)
+
+export default Vue.extend({
+  name: 'CloudView',
+  data() {
+    return {
+      cloud_token: '',
+    }
+  },
+  methods: {
+    async join() {
+      axios.put(
+        'https://app.blueos.cloud/api/agent/join/',
+        {},
+        {
+          headers:
+          {
+            Authorization: `Token ${this.cloud_token}`,
+          },
+        },
+      ).catch((error) => {
+        if (error.response.status === StatusCodes.UNAUTHORIZED) {
+          notifier.pushError(
+            'CLOUD_REGISTRATION_ERROR',
+            'Invalid token. Failed to link your vehicle to BlueOS Cloud ',
+            true,
+          )
+        }
+        if (error.response.status === StatusCodes.BAD_REQUEST) {
+          notifier.pushSuccess(
+            'CLOUD_REGISTRATION_SUCCESS',
+            'Token is valid, your vehicle is now linked to BlueOS Cloud',
+            true,
+          )
+          // TODO: save this securely on a backend or file instead of bag
+          bag.setData('vehicle.cloud', { token: this.cloud_token })
+        } else {
+          notifier.pushError('CLOUD_REGISTRATION_ERROR', 'Failed to link your vehicle to BlueOS Cloud ', true)
+        }
+      })
+    },
+  },
+})
+</script>
--- a/core/frontend/src/router/index.ts
+++ b/core/frontend/src/router/index.ts
@@ -5,6 +5,7 @@ import Autopilot from '../views/Autopilot.vue'
 import AvailableServicesView from '../views/AvailableServicesView.vue'
 import BagEditorView from '../views/BagEditorView.vue'
 import BridgesView from '../views/BridgesView.vue'
+import CloudView from '../views/CloudView.vue'
 import Endpoint from '../views/EndpointView.vue'
 import ExtensionManagerView from '../views/ExtensionManagerView.vue'
 import ExtensionView from '../views/ExtensionView.vue'
@@ -135,6 +136,11 @@ const routes: Array<RouteConfig> = [
     path: '/vehicle/parameters',
     name: 'Parameter Editor',
     component: ParameterEditorView,
+  },
+  {
+    path: '/cloud/join',
+    name: 'Cloud',
+    component: CloudView,
   },
   {
     path: '*',
--- a/core/frontend/src/types/frontend_services.ts
+++ b/core/frontend/src/types/frontend_services.ts
@@ -146,3 +146,10 @@ export const version_chooser_service: Service = {
   company: 'Blue Robotics',
   version: '0.1.0',
 }
+
+export const cloud_service: Service = {
+  name: 'Cloud Service',
+  description: 'Manage talking to BlueOS Cloud',
+  company: 'Blue Robotics',
+  version: '0.1.0',
+}


core/frontend/src/views/CloudView.vue
```diff
@@ -0,0 +1,57 @@
...
+export default Vue.extend({
+  name: 'CloudView',
+  data() {
+    return {
+      machineid: undefined,
+      cloud_token: '',
+    }
+  },
+  async mounted() {
+    const result = await commander.command("python -c 'import machineid; print(machineid.hashed_id())'")
```
patrickelectric, line 47: Isn't it better to have helper to provide such information on the frontend over doing python over ts ?

Williangalvani: it definitely is. actually I was thinking it could be done by linux2rest, what do you think?

joaoantoniocardoso: > This ID uniquely identifies the host. It should be considered
       "confidential", and must not be exposed in untrusted
       environments, in particular on the network. If a stable unique
       identifier that is tied to the machine is needed for some
       application, the machine ID or any part of it must not be used
       directly. Instead the machine ID should be hashed with a
       cryptographic, keyed hash function, using a fixed,
       application-specific key. That way the ID will be properly
       unique, and derived in a constant way from the machine ID but
       there will be no way to retrieve the original machine ID from the
       application-specific one. The
       [sd_id128_get_machine_app_specific(3)](https://man7.org/linux/man-pages/man3/sd_id128_get_machine_app_specific.3.html) API provides an
       implementation of such an algorithm.


https://man7.org/linux/man-pages/man5/machine-id.5.html

So I guess we should pass an application key to it, like `machineid.hashed_id({blueos})`

To avoid using commander, we could bind the /etc/machine-id as read-only to the BlueOs container.

Williangalvani: except we dont want to use machine-id, right? since it does a prettye bad job at "uniquely identifying the host"...
I vote that at first we just hash the cpu serial. I have no strong feelings about that being confidential.

For it do actually be confidention, we'd have to implement the "join" logic mostly in the backend, though

patrickelectric: I agree with pascal about the security concerns.
For now, what I can point is:

For a rasp we can use the serial number, and for everything else we could have custom support or use directly https://docs.rs/machineid-rs/1.2.4/machineid_r, that should support everything necessary for hardware identification generation and add it in linux2rest as pointed by @Williangalvani for easy access API. For more information about crates capabilities, check https://docs.rs/machineid-rs/1.2.4/machineid_rs/enum.HWIDComponent.html

But, that may not fix the issue, since we are going to have only a local source that may not work reliable.

I vote for us to have an automatic request by blueos to blueos.cloud requesting an unique identifier based on the HARDWARE_IDENTIFIER.

```
HARDWARE_IDENTIFIER = `${HARDWARE_IDENTIFIER}-${LINUX_MACHINE_ID}`
BLUEOS_CLOUD_IDENTIFIER = request_blueos_cloud_identifier(HARDWARE_IDENTIFIER)
BLUEOS_IDENTIFIER = `${BLUEOS_CLOUD_IDENTIFIER}-${HARDWARE_IDENTIFIER}`
```

With that we ensure a global unique identifier.

patrickelectric: `HARDWARE_IDENTIFIER` is generated during execution time, so with that we're going to check if the two computers share the same CLOUD unique identifier with different hardware identification. That'll happen on the case where the sdcard is cloned. 

How `HARDWARE_IDENTIFIER` will be created depend on us. Right now, I would use the pi serial number + `machineid-rs (SystemID, CPUCores, MacAddress, CPUID)`, where the `machineid-rs` would be the standard way of doing it if there are no custom support for the identified system.

joaoantoniocardoso: > `HARDWARE_IDENTIFIER` is generated during execution time, so with that we're going to check if the two computers share the same CLOUD unique identifier with different hardware identification. That'll happen on the case where the sdcard is cloned.
> 
> How `HARDWARE_IDENTIFIER` will be created depend on us. Right now, I would use the pi serial number + `machineid-rs (SystemID, CPUCores, MacAddress, CPUID)`, where the `machineid-rs` would be the standard way of doing it if there are no custom support for the identified system.

Okay, so you want to do only (B)?

patrickelectric: > > `HARDWARE_IDENTIFIER` is generated during execution time, so with that we're going to check if the two computers share the same CLOUD unique identifier with different hardware identification. That'll happen on the case where the sdcard is cloned.
> > How `HARDWARE_IDENTIFIER` will be created depend on us. Right now, I would use the pi serial number + `machineid-rs (SystemID, CPUCores, MacAddress, CPUID)`, where the `machineid-rs` would be the standard way of doing it if there are no custom support for the identified system.
> 
> Okay, so you want to do only (B)?

It's `C`, where we keep track of both, we ask `BLUEOS_CLOUD_IDENTIFIER` that is always unique. With this, we are going to be sure that is always a single unique identifier per vehicle.

# SD Card cloned

The `HARDWARE_IDENTIFIER` will not be the same when comparing to `BLUEOS_CLOUD_IDENTIFIER` on the cloud database, we are going to detect this case.

# New SD Card, old registered hardware

The `HARDWARE_IDENTIFIER` will match but `BLUEOS_CLOUD_IDENTIFIER` not once connects to `blueos.cloud`. We are going to detect this case as well.

voorloopnul (approved):

core/frontend/src/store/commander.ts
```diff
@@ -38,6 +38,28 @@ class CommanderStore {
       })
   }
 
+  async command(command: string): Promise<undefined | ReturnStruct> {
```
patrickelectric, line 41: Can we break this in another commit ?

patrickelectric, line 41: Can we break this in another commit ?

core/frontend/src/store/commander.ts
```diff
@@ -38,6 +38,28 @@ class CommanderStore {
...
+        i_know_what_i_am_doing: true,
+      },
+    })
+      .then((response) => response.data)
+      .catch((error) => {
+        if (error === backend_offline_error) {
+          return undefined
+        }
+        const message = `Could not send command to docker: ${error.response?.data ?? error.message}.`
+        console.log(message)
```
patrickelectric, line 57: please remove development messages

core/frontend/src/store/commander.ts
```diff
@@ -38,6 +38,28 @@ class CommanderStore {
...
+        notifier.pushError('COMMANDER_COMMAND_HOST_FAIL', message, true)
```
patrickelectric, line 58: This is not command_host but command_blouos

core/frontend/src/views/CloudView.vue
```diff
@@ -0,0 +1,80 @@
...
+          },
+        },
+      ).catch((error) => {
+        if (error.response.status === 401) {
+          notifier.pushError(
+            'CLOUD_REGISTRATION_ERROR',
+            'Invalid token. Failed to link your vehicle to BlueOS Cloud ',
+            true,
+          )
+        } else if (error.response.status === 400) {
```
patrickelectric, line 65: else is not necessary, an early return can be performed.

patrickelectric, line 65: StatusDode.UNAUTHORIZED

core/frontend/src/views/CloudView.vue
```diff
@@ -0,0 +1,80 @@
...
+          notifier.pushSuccess(
+            'CLOUD_REGISTRATION_SUCCESS',
+            'Token is valid, your vehicle is now linked to BlueOS Cloud',
+            true,
+          )
+          // TODO: save this securely on a backend or file instead of bag
+          bag.setData('vehicle.cloud', { token: this.cloud_token })
```
patrickelectric, line 72: How this information is going to be used ? Does this PR work ?

patrickelectric (changes requested):

core/frontend/src/views/CloudView.vue
Williangalvani: For now it just validates the token with the backend and stores it locally. it will only be consumed in subsequent PRs

Williangalvani:
I think we decided to have this be in an extension together with major tom, is that right?

patrickelectric:
> I think we decided to have this be in an extension together with major tom, is that right?

yes

voorloopnul:
I'm _hoping_ to not have a web service running in the major_tom extension, just a daemon, so if we can have the token/auth process on blueos itself, I'll not complain :D 

Also, I think we discussed before that the cloud joining would eventually happen in the wizard, not sure how easy it is to mix things being hosted as an extension with the overall UI of BlueOS.

patrickelectric:
> I'm _hoping_ to not have a web service running in the major_tom extension, just a daemon, so if we can have the token/auth process on blueos itself, I'll not complain :D
> 
> Also, I think we discussed before that the cloud joining would soon happen in the wizard, not sure how easy it is to mix things being hosted as an extension with the overall UI of BlueOS.
> 
> @Williangalvani @patrickelectric

Could we have only the joining / token processes in BlueOS and share that somehow with extensions ?

voorloopnul:
Yes, this PR does precisely that. It validates and saves the token in the bag.vehicle.cloud.
Also, this PR creates a blueOS URL "/cloud/join/" but doesn't advertise it anywhere; it is hidden, so only selected users can know and use it in the early stages, but later should be integrated in the wizard.

PR #2088, opened 2023-09-08
bootstrap: startup.json.default: Add /etc/blueos

patrickelectric:


voorloopnul:
Just to ensure I'm following, when the /etc/machine-id will be populated?

Williangalvani:
> Just to ensure I'm following, when the /etc/machine-id will be populated?

IIRC we decided not to use it. we'll eventually create a new field for a "hardware fingerprint"

patrickelectric:
This already helps with the id conflict while we don't have the hardware fingerprint.

patrickelectric:
Can we have this in ?

voorloopnul (approved):
Looks all right to me.

ES-Alexander:
Should document that this telemetry exists (as part of internet access checking?, and "What's new" section), and explain about the map on [blueos.cloud](https://blueos.cloud), GeoIP (including precision), and the ability to obscure it with a VPN, or prevent it by blocking `telemetry.blueos.cloud` in router settings or avoiding connecting the BlueOS device to the internet.

PR #2089, opened 2023-09-08
install: install: Add uninitialized machine-id

patrickelectric:
We should have the file otherwise docker binds will fail

Follows #2086

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -215,6 +215,12 @@ curl -fsSL $ROOT/install/network/avahi.sh | bash
 # Following https://systemd.io/BUILDING_IMAGES/
 echo "Restarting machine-id."
 rm -rf /etc/machine-id /var/lib/dbus/machine-id
+# The file needs to exist for docker to bind
+## if /etc/machine-id does not exist, this is a first boot. During early boot, systemd will write "uninitialized\n"
+## to this file and overmount a temporary file which contains the actual machine ID.
+## Later (after first-boot-complete.target has been reached), the real machine ID will be written to disk.
+## If /etc/machine-id contains the string "uninitialized", a boot is also considered the first boot. The same mechanism as above applies.
+echo "uninitialized" > /etc/machine-id
 echo "Restarting random-seeds."
 rm -rf /var/lib/systemd/random-seed /loader/random-seed
 


install/install.sh
```diff
@@ -215,6 +215,12 @@ curl -fsSL $ROOT/install/network/avahi.sh | bash
 # Following https://systemd.io/BUILDING_IMAGES/
 echo "Restarting machine-id."
 rm -rf /etc/machine-id /var/lib/dbus/machine-id
+# The file needs to exist for docker to bind
+## if /etc/machine-id does not exist, this is a first boot. During early boot, systemd will write "uninitialized\n"
+## to this file and overmount a temporary file which contains the actual machine ID.
+## Later (after first-boot-complete.target has been reached), the real machine ID will be written to disk.
+## If /etc/machine-id contains the string "uninitialized", a boot is also considered the first boot. The same mechanism as above applies.
+echo "uninitialized" > /etc/machine-id
 echo "Restarting random-seeds."
```
joaoantoniocardoso, line 224: can you make the /lib/var/dbus/machine-id a symlink to it?

patrickelectric: Why ? Is that necessary ? They are unrelated.

patrickelectric: When a machine is booted with [systemd(1)](https://www.freedesktop.org/software/systemd/man/systemd.html#) the ID of the machine will be established. If systemd.machine_id= or --machine-id= options (see first section) are specified, this value will be used. Otherwise, the value in /etc/machine-id will be used. If this file is empty or missing, systemd will attempt to use the D-Bus machine ID from /var/lib/dbus/machine-id.

It looks safer to not do it, and I don't see a reason to do.

joaoantoniocardoso: Tested here, we actually just need to remove `/lib/var/dbus/machine-id`, then the symlink will be handled by systemd itself.

In the case `/lib/var/dbus/machine-id` is not removed (like it is on this PR), we are not going to create a new machine-id, but only restore the already existing one.

joaoantoniocardoso:
```suggestion
rm -r /lib/var/dbus/machine-id
echo "Restarting random-seeds."
```

joaoantoniocardoso: nevermind, it is being deleted, I missed the DIFF >.<

joaoantoniocardoso (approved):

PR #2093, opened 2023-09-11
core: ArdupilotManager: allow setting the SITL frame

### LINKED ISSUE #2094: all SITL instances are using "vectored" models
This means rover, copter, and place code will try to fly (the simplified physics of) an ROV, which doesn't work great (or at all)

Williangalvani:
without this, all SITL instances try to use the "vectored" sub model, which doesn't work great with rover(and others) code.

fix #2094, we'd still need a ui update though

### DIFF
--- a/core/services/ardupilot_manager/ArduPilotManager.py
+++ b/core/services/ardupilot_manager/ArduPilotManager.py
@@ -45,7 +45,6 @@ class ArduPilotManager(metaclass=Singleton):
         self.mavlink_manager = MavlinkManager()
         self.mavlink_manager.set_logdir(self.settings.log_path)
         self._current_board: Optional[FlightController] = None
-        self._current_sitl_frame: SITLFrame = SITLFrame.UNDEFINED
 
         # Load settings and do the initial configuration
         if self.settings.load():
@@ -64,6 +63,7 @@ class ArduPilotManager(metaclass=Singleton):
 
         self.should_be_running = False
         self.remove_old_logs()
+        self.current_sitl_frame = self.load_sitl_frame()
 
     def remove_old_logs(self) -> None:
         def need_to_remove_file(file: pathlib.Path) -> bool:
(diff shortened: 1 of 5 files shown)


rafaellehmkuhl (approved):
Haven't tested, but the code looks good to me.

core/services/install-services.sh
```diff
@@ -35,6 +34,7 @@ SERVICES=(
     ping
     versionchooser
     wifi
+    ardupilot_manager # ardupilot_manager's setup.py requires the dependencies it is supposed to install...
```
patrickelectric, line 37: how so ? ardupilot_manager does not have direct dependency to the others services.

core/services/ardupilot_manager/ArduPilotManager.py
```diff
@@ -257,10 +257,25 @@ def start_serial(self, board: FlightController) -> None:
...
+        self.configuration["sitl_frame"] = frame
+        self.settings.save(self.configuration)
+
+    def load_sitl_frame(self) -> SITLFrame:
+        if self.settings.sitl_frame != SITLFrame.UNDEFINED:
+            return self.settings.sitl_frame
+        frame = SITLFrame.VECTORED
+        logger.warning(f"SITL frame is undefined. Setting {frame} as current frame.")
+        self.configuration["sitl_frame"] = frame
+        self.settings.save(self.configuration)
```
patrickelectric, line 271:
```suggestion
        self.set_sitl_frame(frame)
```

patrickelectric (changes requested):

core/services/install-services.sh
Williangalvani: setup.py tries to use our settings service, which requires pydantic

patrickelectric: And that'll be installed by setup.py on the following loop. My questions is, why the order of ardupilot_manager was changed since there is no dependency to others services.

Williangalvani: this is the reason:
https://github.com/bluerobotics/BlueOS/actions/runs/6176088878/job/16764395632?pr=2093#step:7:12948

I'm open to alternative solutions

patrickelectric: Check here: https://github.com/bluerobotics/BlueOS/blob/master/core/services/install-services.sh#L41

Williangalvani: updated

patrickelectric (approved):

PR #2098, opened 2023-09-12
Use Pipenv for CI/pre-push-hooks

Williangalvani:
This should make pre-push-hook usable on multiple systems, with the same package version used in CI

it warn the user if pipenv is not installed, and install the required files from core/tools 

I also sneaked in some parallelization for pytest and mypy


core/tools/ardupilot_tools/bootstrap.sh
```diff
@@ -7,32 +7,50 @@ set -e
...
 SUDO="$(command -v sudo || echo '')"
 readonly SUDO
-$SUDO wget "$REMOTE_URL_UPLOADER" -O "$LOCAL_PATH_UPLOADER"
-$SUDO chmod +x "$LOCAL_PATH_UPLOADER"
-
+if [ ! -f "$LOCAL_PATH_UPLOADER" ]; then
+  $SUDO wget "$REMOTE_URL_UPLOADER" -O "$LOCAL_PATH_UPLOADER"
+  $SUDO chmod +x "$LOCAL_PATH_UPLOADER"
+else
+  echo "File already exists. Skipping download."
```
joaoantoniocardoso, line 22:
```suggestion
  echo "File $LOCAL_PATH_UPLOADER already exists. Skipping download."
```

core/tools/mavlink_router/bootstrap.sh
```diff
@@ -13,5 +19,5 @@ if [[ "$(uname -m)" == "x86_64"* ]]; then
     REMOTE_BINARY_URL="${GITHUB_REMOTE}/releases/download/${VERSION}/mavlink-routerd-libc-x86-64"
 fi
 
-wget "$REMOTE_BINARY_URL" -O "$LOCAL_BINARY_PATH"
-chmod +x "$LOCAL_BINARY_PATH"
\ No newline at end of file
+$SUDO wget "$REMOTE_BINARY_URL" -O "$LOCAL_BINARY_PATH"
+$SUDO chmod +x "$LOCAL_BINARY_PATH"
```
joaoantoniocardoso, line 23: missing the if to not download if already exists

joaoantoniocardoso (approved):
Great job, 

beyond the benefits of allowing anyone to run **the same** tools as the CI, it improved from ~11 to ~3 minutes on the Python tests/lint CI, and it's taking ~1min42s the first pass locally, and then ~1min20s after.

PR #2104, opened 2023-09-14
core: services: Add limit_ram_usage

patrickelectric:
Limit ram usage by python programs to 100MB

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/general.py
+++ b/core/libs/commonwealth/commonwealth/utils/general.py
@@ -1,4 +1,5 @@
 import os
+import resource
 import subprocess
 import uuid
 from functools import cache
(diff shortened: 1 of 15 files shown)


Williangalvani:
what happens is the service hits the memory limit?

patrickelectric:
> what happens is the service hits the memory limit?

The code raise a `MemoryLimit` exception

Williangalvani:
I have mixed feelings here. I like the idea, but I think if should be further handled. should we have all services check for this exception?
or perhaps have it kill and restart them?

Williangalvani:
some of them, like Kraken, are find, but I wonder about the basic funcionalities one, mostly the ones related to piloting/networking

patrickelectric:
We are on master branch, the main point is for us to find where the ram problems are happening and logging it. 
For a stable release we can just remove or increase the limit.
Right now having it at least allow the system to recover. Otherwise the user looses the entire vehicle and a power reboot is necessary.

joaoantoniocardoso (approved):
I agree to have this. We might need to figure out the limits, but this could prevent the situation where a service triggers Linux's OOM killer, and the main process is killed, being unable to recover.

Williangalvani:
fair enough, let's get it in. needs a rebase though

PR #2123, opened 2023-09-21
core: ping: detect all serial ports

Williangalvani:
depends on https://github.com/pyserial/pyserial/pull/709


Williangalvani:
I vote we switch to this fork:
https://github.com/pyserial/pyserial/pull/709#issuecomment-2387049649
I don't think we should keep waiting any longer.

PR #2124, opened 2023-09-21
core: services: kraken: Add size of extension

patrickelectric:
image
ps: Yes, it's a 128GB sdcard

### DIFF
--- a/core/frontend/src/components/kraken/InstalledExtensionCard.vue
+++ b/core/frontend/src/components/kraken/InstalledExtensionCard.vue
@@ -71,6 +71,30 @@
                 </v-progress-linear>
               </td>
             </tr>
+            <tr v-if="extension.enabled">
+              <td>Disk usage</td>
+              <td>
+                <v-progress-linear
+                  :value="getDiskUsage()"
+                  color="green"
+                  height="25"
+                >
+                  <template #default>
+                    <SpinningLogo
+                      v-if="loading || extension.enabled && !container"
+                      size="20px"
+                    />
+                    <strong v-else-if="getDiskUsage()?.toFixed">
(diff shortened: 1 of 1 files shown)


patrickelectric:
ping @joaoantoniocardoso @Williangalvani

core/services/kraken/kraken.py
```diff
@@ -310,11 +315,18 @@ async def load_stats(self) -> Dict[str, Any]:
             except KeyError:
                 memory_usage = "N/A"
 
+            try:
+                print(show)
```
Williangalvani, line 319:
```suggestion
```
debugging code

Williangalvani (changes requested):
it misbehaved a couple of times here, where it only loaded stats for some of the containers, but I don't think it is related to this pr.

Other than that print, it looks great

joaoantoniocardoso (approved):

core/frontend/src/components/kraken/InstalledExtensionCard.vue
```diff
@@ -199,6 +223,14 @@ export default Vue.extend({
...
       return total_kb ?? undefined
     },
+    main_disk(): number | undefined {
+      const disks = system_information.system?.disk
+      return disks?.find((sensor) => sensor.mount_point === '/')
+    },
+    main_disk_size(): number {
+      const value = this.main_disk?.total_space_B ?? 0
+      return value / 1024 // Move to kB
+    },
```
Williangalvani, line 233: My brain is not working well today. but the typing here looks wrong.

if `main_disk()` is a number, how can we access `total_space_B` of it in the second method?

And why is CI not catching this?

Williangalvani (changes requested):
sorry, minor typing issue

patrickelectric:
ping @Williangalvani

PR #2128, opened 2023-09-21
Create authorized_keys with correct owner

rotu:
Previously `~pi/.ssh/authorized_keys` was created with the current user (root) as owner. This causes `ssh-copy-id` to fail when the user tries to upload their ssh key to the host. Now it should have the correct permissions.

### DIFF
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -1,7 +1,6 @@
 #! /usr/bin/env python3
 import logging
 import os
-import shlex
 import shutil
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


patrickelectric:
Hi @rotu thanks for you contribution! 
There are some things necessary for you PR to be merged, the CI need to be pass, in this case you just need to replace `'` with `"`.

Williangalvani:
@rotu you can automatically fix the styling by running `.hooks/pre-push --fix`

rotu:
I think this is ready for review now!

patrickelectric (changes requested):
Hi @rotu,

Thanks for updating your PR.
Can you organize the commits in a way that makes more clear what's being changed ? Looking at the code it just appears to be a syntactic change plus the `chown` and `chmod`.
Organizing better your changes in atomic commits will explain the idea behind this PR and what is fixing exactly.

rotu:
> Hi @rotu,
> 
> Thanks for updating your PR. Can you organize the commits in a way that makes more clear what's being changed ? Looking at the code it just appears to be a syntactic change plus the `chown` and `chmod`. Organizing better your changes in atomic commits will explain the idea behind this PR and what is fixing exactly.

The commit history represents what happened and when - I try not to rebase unless necessary but if you'd still like to me to I can. There was a lot of floundering in trying to make this PR, so I'll explain the changes below better than I would in any commit message:

1. (fix) Ensure the owner and permissions are correct. This was my original intent in opening the PR. I'm sure there's a reason, but it *is* weird to me that `chown` lives in `shutil` whereas `chmod` lives in `pathlib`. 
2. (style) use `pathlib.Path` where possible for clarity, extract the file names to variables for clarity
3. (minor fix) use is_file to check that the public key is a file instead of a folder or a broken symlink
4. (style) don't use shlex to parse interpolated strings. While the previous style *looks* more like what you'd type in a command line, it doesn't actually quote the path nor escape spaces, backslashes, etc.
5. (minor fix) don't `touch authorized_keys` to ensure it exists. This always changes the modification time, which is not what we want. Instead, only write the file if we need to. (as an aside, I also prefer using library functions instead of subprocess calls, where convenient)
6. (style) Use `read_text` and `write_text` instead of `open( ... ,'r+')` to read the file.
  - `r+` would have to get changed to `w+` anyway (since I removed the `touch` call)
  - I think it's more clear to open the file for reading and opening for writing only if it must be written (but I don't feel *too* strongly about this)

patrickelectric:
Hi @rotu, Thanks for the clarification, now everything makes more sense.
Can you divide your commits in a why that reflects the item in your comment ?
You can also squash everything in a single commit with a commit body with all raised points. 
The two alternatives work out for us to merge.

rotu:
@patrickelectric go ahead and [squash merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges)!

patrickelectric (approved):

patrickelectric:
Thanks @rotu!

PR #2138, opened 2023-09-26
core: frontend: fix tag for vue tour

Williangalvani:
This fails weirdly on my end. I get just a white screen and this in the console:
image

this is what I am trying to fix:
image


Williangalvani:
doesnt matter what I do, it seems to be trying to load vtour from vuetify:
```
[vite-plugin-pwa:build] 'VTour' is not exported by node_modules/vuetify/lib/index.js, imported by src/App.vue
```

PR #2142, opened 2023-09-27
Add autopilot type banner

patrickelectric:
image
image


joaoantoniocardoso (approved):
fancy

ES-Alexander:
Advanced usage docs should have an updated screenshot.

Also need to look into whether PX4 and OpenPilot firmwares are supported for installing/updating, or just detection.

PR #2171, opened 2023-10-16
core: autopilot manager: bump memory limit to 200mb

### LINKED ISSUE #2170: Autopilot Manager runs out of memory when decompressing the firmware manifest
Screenshot 2023-10-16 at 15 57 14

Williangalvani:
fix #2170

### DIFF
--- a/core/services/ardupilot_manager/main.py
+++ b/core/services/ardupilot_manager/main.py
@@ -30,7 +30,8 @@ from typedefs import Firmware, FlightController, Parameters, Serial, SITLFrame, 
 
 FRONTEND_FOLDER = Path.joinpath(Path(__file__).parent.absolute(), "frontend")
 
-limit_ram_usage()
+# We need additional RAM in order to decompress the manifest file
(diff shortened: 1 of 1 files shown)


Williangalvani:
autopilotmanager idles at ~ 25mb, the unzipped manifest file is 36mb. loading the whole thing as json in memory seems to be what is causing the large memory usage.
We can either merge this or try to find a better json parser.

patrickelectric (approved):
The issue in the end is having both in memory at the same time. We can let it go since it happens only once with the manifest decompression.

patrickelectric:
For more information:
[memray-flamegraph-output.html.zip](https://github.com/bluerobotics/BlueOS/files/13035094/memray-flamegraph-output.html.zip)

Williangalvani:
possibly relevant further on
https://pythonspeed.com/articles/json-memory-streaming/

patrickelectric:
In the end the problem was with gzip and not the json, @Williangalvani you can check the flamegraph plot.

PR #2188, opened 2023-11-09
Don't modify /etc/dhcpd.conf if it does not exist

rotu:
If /etc/dhcpd.conf didn't exist, this would crash with "sed: can't read /etc/dhcpcd.conf: No such file or directory".

Prevents a crash of the install script when installing on top of a clean install of Raspberry Pi OS.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -162,11 +162,16 @@ echo "Going to install blueos-docker version ${VERSION}."
 echo "Downloading and installing udev rules."
 curl -fsSL $ROOT/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
 
-echo "Disabling automatic Link-local configuration in dhcpd.conf."
-# delete line if it already exists
-sed -i '/noipv4ll/d' /etc/dhcpcd.conf
-# add noipv4ll
-sed -i '$ a noipv4ll' /etc/dhcpcd.conf
+if [ -f /etc/dhcpcd.conf ]
+then
+    echo "Disabling automatic Link-local configuration in dhcpd.conf."
+    # delete line if it already exists
+    sed -i '/noipv4ll/d' /etc/dhcpcd.conf
+    # add noipv4ll
+    sed -i '$ a noipv4ll' /etc/dhcpcd.conf
+else
+    echo "Not modifying /etc/dhcpcd.conf - file does not exist"
+fi
 
 # Do necessary changes if running in a Raspiberry
 command -v raspi-config && (


install/install.sh
```diff
@@ -162,11 +162,16 @@ echo "Going to install blueos-docker version ${VERSION}."
 echo "Downloading and installing udev rules."
 curl -fsSL $ROOT/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
 
-echo "Disabling automatic Link-local configuration in dhcpd.conf."
-# delete line if it already exists
-sed -i '/noipv4ll/d' /etc/dhcpcd.conf
-# add noipv4ll
-sed -i '$ a noipv4ll' /etc/dhcpcd.conf
+if [ -e /etc/dhcpcd.conf ]
```
rafaellehmkuhl, line 165: Would use `-f` here to check if it is truly a file and not a directory (of course very unlikely).

install/install.sh
```diff
@@ -229,4 +234,4 @@ echo "You can access after the reboot:"
 echo "- The computer webpage: http://blueos.local"
 echo "- The ssh client: pi@blueos.local"
 echo "System will reboot in 10 seconds."
-sleep 10 && reboot
```
rafaellehmkuhl, line 232: No need to commit this change.

PS: our linter should be getting those cases, but it isn't.

rafaellehmkuhl (changes requested):

rafaellehmkuhl:
@rotu can you run the install script and see if everything is running fine once we approve?

install/install.sh
rotu: I didn't mean to make that change... I don't know what you mean about the linter.

rafaellehmkuhl: This was probably an auto-formatting from your IDE. Are you on Windows? If I remember correctly it does this because the line ending character is different. You can just unstage this before commiting.

About the linter, I'm just noticing that the linter we have in the repository should be getting those cases automatically.

install/install.sh
rotu: Done. Sorry I rebased instead of a simple commit - my setup is being weird

rafaellehmkuhl: No problem.

patrickelectric (approved):

rotu:
What is this? https://github.com/bluerobotics/BlueOS/actions/runs/6815507685/job/18534767388?pr=2188#step:4:170
> Invalid author string. Must be in the format: John Smith <john@example.com>

patrickelectric:
> What is this? [bluerobotics/BlueOS/actions/runs/6815507685/job/18534767388?pr=2188#step:4:170](https://github.com/bluerobotics/BlueOS/actions/runs/6815507685/job/18534767388?pr=2188#step:4:170)
> 
> > Invalid author string. Must be in the format: John Smith [john@example.com](mailto:john@example.com)

Hi @rotu, It's an issue related to pyproject, it does not affect your changes. There is a fix already in #2189. 
To avoid breaking CI for now, we are just going to merge that before yours.

install/install.sh
rotu: Yep. I'm in VSCode on Windows.

No, it's not the different line ending - I just have it set up to add a newline at the end of the file. Most of the scripts in the repo *do* have a newline at the end, and I'm surprised this one doesn't and that your linter doesn't enforce it!

PR #2212, opened 2023-12-05
Stream logs

Williangalvani:


Williangalvani:
current downside is that it only streams for 30 seconds. we can't tell if the connection is closed, so we need a timeout.

patrickelectric (approved):

PR #2223, opened 2023-12-06
core: ArdupilotManager: fix mavlink-router tcp endpoints

### LINKED ISSUE #2222: MAvlinkRouter TCPServer and TCPClient endpoints are switched
out default master endpoint is a tcpserver for the router, yet we have ardupilot as a the server
image

Williangalvani:
fix #2222 

This will only affect SITL, since Navigator uses UDP.

It also doesn't really affect it, since the behavior is the same, we were just calling things via the wrong names


core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
```diff
@@ -26,9 +26,9 @@ def assemble_command(self, master_endpoint: Endpoint) -> str:
         # Convert endpoint format to mavlink-router format
         def convert_endpoint(endpoint: Endpoint) -> str:
             # TCP uses a special argument and only works with localhost
-            if endpoint.connection_type == EndpointType.TCPServer:
-                return f"--tcp-port {endpoint.argument}"
             if endpoint.connection_type == EndpointType.TCPClient:
+                return f"--tcp-port {endpoint.argument}"
```
patrickelectric, line 30: Wait, here you are using tcp-port as a client and in line 68 as a server.

IIRC the tcp-port receives only the port because it binds on the host computer, being a server by nature.

patrickelectric (changes requested):

core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
Williangalvani: good catch, it turned out that only the master endpoint was switched

patrickelectric (approved):

PR #2243, opened 2023-12-08
core: frontend: do not update motors when blueos is not in focus

### LINKED ISSUE #2206: Blueos Motors Setup page conflicts with QGC
Maybe we can disable it if there's another GCS connected?

https://github.com/bluerobotics/BlueOS/assets/4013804/5a2a3ed5-9d03-4680-afbb-da13d9a7fb58

Williangalvani:
should fix #2206 

needs testing

### DIFF
--- a/core/frontend/src/components/vehiclesetup/PwmSetup.vue
+++ b/core/frontend/src/components/vehiclesetup/PwmSetup.vue
@@ -25,7 +25,7 @@
                   </th>
                   <th>
                     <v-switch
-                      v-model="is_armed"
+                      v-model="desired_armed_state"
                       :loading="desired_armed_state !== (is_armed) ? 'warning' : null"
                       :disabled="!is_manual"
                       class="mx-1 flex-grow-0"
@@ -415,10 +415,9 @@ export default Vue.extend({
         console.warn('Disarming failed!')
       }, 5000)
     },
-    arm_disarm_switch_change(): void {
-      this.desired_armed_state = !this.is_armed
+    arm_disarm_switch_change(should_arm: boolean): void {
       // eslint-disable-next-line no-unused-expressions
-      this.is_armed ? this.disarm() : this.arm()
+      should_arm ? this.arm() : this.disarm()
     },
     armDisarm(arm: boolean, force: boolean): void {
       mavlink2rest.sendMessage(
--- a/core/frontend/src/components/vehiclesetup/PwmSetup.vue
+++ b/core/frontend/src/components/vehiclesetup/PwmSetup.vue
@@ -9,6 +9,11 @@
           <vehicle-viewer :highlight="highlight" :transparent="true" :autorotate="false" />
         </v-card>
         <v-card class="mt-3">
+          <v-overlay :value="!has_focus">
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso:
I didn't test it, but just a reminder: it should show feedback that it was canceled because the user changed the window.

patrickelectric (approved):

core/frontend/src/components/vehiclesetup/PwmSetup.vue
```diff
@@ -384,10 +412,9 @@ export default Vue.extend({
         console.warn('Disarming failed!')
       }, 5000)
     },
-    arm_disarm_switch_change(): void {
-      this.desired_armed_state = !this.is_armed
+    arm_disarm_switch_change(new_state: boolean): void {
```
patrickelectric, line 415: Can we have new_state with a better name that reflects the arm state ?
Like.. is_armed ?

core/frontend/src/components/vehiclesetup/PwmSetup.vue
```diff
@@ -350,11 +372,17 @@ export default Vue.extend({
     },
     printParam,
     zero_motors() {
+      if (!this.has_focus) {
+        return
```
patrickelectric, line 376: This is a case where we should be safe to run if out of focus no ?

patrickelectric (changes requested):

core/frontend/src/components/vehiclesetup/PwmSetup.vue
Williangalvani: I'd prefer to stop completely in case the user is trying to do something else.
We need to check if rover has a decent motor test timeout, actually.

core/frontend/src/components/vehiclesetup/PwmSetup.vue
Williangalvani: this is not what it means. this is the state of the switch itself, which we use to arm/disarm.
maybe `wants_to_be_armed` would be better?

patrickelectric: `should_arm` ?

core/frontend/src/components/vehiclesetup/PwmSetup.vue
patrickelectric: But this is zero_motors.. isn't this supposed to be the safest way than letting a undefined state ?

Williangalvani: we don't want to interfere if the user is using another software. we should probably send one zero_motors and then stop completely

patrickelectric: Can we have a comment explaining that ? This is not clear from the code.

Williangalvani: > Can we have a comment explaining that ? This is not clear from the code.

the code wasn't doing it before. I added a boolean to keep track of it.

patrickelectric (approved):

PR #2254, opened 2023-12-12
Use python:3.11.7-slim-bullseye as base image

patrickelectric:
By default buster-slim comes with python 3.7 that is EOL already


Williangalvani:
CI is definteily not happy, in a very weird way

patrickelectric:
done in #2263

PR #2256, opened 2023-12-13
core: Dockerfile: Update to use new BlueOS base image

### LINKED ISSUE #1175: suggestion: transition to a more recent Python version
[Python 3.10](https://docs.python.org/3/whatsnew/3.10.html) has been the latest stable release for almost a year now, and is about to be superseded by [Python 3.11](https://docs.python.org/3.11/whatsnew/3.11.html) later this month.

For major relevant improvements, 3.10 offers
- more precise error tracebacks
- better type hints
- structural pattern matching (`match`-`case`)
- various performance optimisations

3.11 offers
- even more precise error tracebacks
- exception groups (handle multiple unrelated exceptions simultaneously)
- more type hint improvements
- significant performance optimisations

At this point I expect the libraries we use in BlueOS core (and our extensions) will most likely all have been tested with 3.10, and many have likely been tested with 3.11 as well. Stability wise upgrading to 3.10 would be a reasonably conservative move. As I understand it both 3.10 and 3.11 should be more stable at release than previous releases, because since [Python switched to a fixed 12-month release cycle](https://peps.python.org/pep-0602/) there's a fixed 5 month period of "no new features" before each release, and library maintainers have a known date to aim for to test their packages by.

Philosophically I think it's important to avoid the trap of Companion where the initial version of everything is kept until it's well and truly dead and buried. Doing so results in a developer experience that degrades over time, unnecessarily misses out on efficiency improvements, and may lose us contributions from external developers who want to work with recent features.

patrickelectric:
tested
[x] Move to use https://www.piwheels.org/project/aiohttp/
Fix #1175 
https://github.com/bluerobotics/blueos-docker-base/releases/tag/v0.1.0

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -12,12 +12,12 @@ RUN bun install --cwd /home/pi/frontend
 RUN bun run --cwd /home/pi/frontend build
 
 # Download binaries
-FROM bluerobotics/blueos-base:v0.0.10 as downloadBinaries
+FROM bluerobotics/blueos-base:v0.1.0 as downloadBinaries
 COPY tools /home/pi/tools
 RUN /home/pi/tools/install-static-binaries.sh
 
 # BlueOS-docker base image
-FROM bluerobotics/blueos-base:v0.0.10
+FROM bluerobotics/blueos-base:v0.1.0
 
 # Install necessary tools
 COPY tools /home/pi/tools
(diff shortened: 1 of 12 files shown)


joaoantoniocardoso (approved):
Tested, and everything seems to be working correctly :)

PR #2268, opened 2023-12-20
core: do not compile unnecessarily

Williangalvani:
For a faster build.

This replaces "setup.py install" with "pip install", which will use the additional index-url.
it also stops installing pre-requisites when we don't need to build anything.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -19,6 +19,15 @@ RUN /home/pi/tools/install-static-binaries.sh
 # BlueOS-docker base image
 FROM bluerobotics/blueos-base:v0.1.0
 
+# Define arguments for target platform
+# These arguments are defined automatically by buildx when using `--platform`
+ARG TARGETARCH
+ARG TARGETVARIANT
+
+# Set these arguments as environment variables
+ENV TARGETARCH=${TARGETARCH}
+ENV TARGETVARIANT=${TARGETVARIANT}
+
 # Install necessary tools
 COPY tools /home/pi/tools
 RUN /home/pi/tools/install-system-tools.sh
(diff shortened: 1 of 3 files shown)


core/Dockerfile
```diff
@@ -19,6 +19,14 @@ RUN /home/pi/tools/install-static-binaries.sh
 # BlueOS-docker base image
 FROM bluerobotics/blueos-base:v0.1.0
 
+# Define arguments for target platform
+ARG TARGETARCH
```
patrickelectric, line 23: Can we add a comment about how this arguments are generated by --platform ?

```
# These arguments are defined automatically by buildx when using `--platform`
```

core/libs/install-libs.sh
```diff
@@ -7,23 +7,30 @@ BUILD_PACKAGES=(
...
 )
 
-apt update
-apt install -y --no-install-recommends ${BUILD_PACKAGES[*]}
+echo "Target architecture: $TARGETARCH"
+echo "Target variant: $TARGETVARIANT"
+# Install build packages if not on armv7 or amd64, which are the only platforms that have pre-built wheels
+if ! { [ "$TARGETARCH" == "arm" ] && [ "$TARGETVARIANT" == "v7" ]; } && [ "$TARGETARCH" != "amd64" ]; then
+    apt update
+    apt install -y --no-install-recommends ${BUILD_PACKAGES[*]}
```
patrickelectric, line 15: Should we remove this at all ? Everything else should have pre-built binaries, no ? piwhell will mostly ensure that.

Williangalvani: oddly enough, there were binaries missing for x86_64. IIRC it was aiohttp

core/Dockerfile
Williangalvani: done

patrickelectric (approved):

Issue #2267, opened 2023-12-20
Kraken: Container logs are not live


Williangalvani:
Happens when looking at container logs in the UI
Side-effect from migration to 3.11, apparently.

Williangalvani:
probably due to the 3.11 change in here:
https://docs.python.org/3/library/asyncio-task.html#id12
![image](https://github.com/bluerobotics/BlueOS/assets/4013804/c699ab56-025c-4e62-bc3c-146d475db3f6)

joaoantoniocardoso:
Yep, Kraken is broken on master, making it impossible to install extensions:

```
2023-12-22 15:19:49.532 | ERROR    | logging:callHandlers:1706 - Exception in ASGI application

Traceback (most recent call last):

  File "/home/pi/services/kraken/main.py", line 175, in <module>
    loop.run_until_complete(server.serve())
    │    │                  │      └ <function Server.serve at 0xb5aaf7a8>
    │    │                  └ <uvicorn.server.Server object at 0xb551a5f0>
    │    └ <function BaseEventLoop.run_until_complete at 0xb62d4758>
    └ <_UnixSelectorEventLoop running=True closed=False debug=False>

  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 640, in run_until_complete
    self.run_forever()
    │    └ <function BaseEventLoop.run_forever at 0xb62d4708>
    └ <_UnixSelectorEventLoop running=True closed=False debug=False>
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 607, in run_forever
    self._run_once()
    │    └ <function BaseEventLoop._run_once at 0xb62d6668>
    └ <_UnixSelectorEventLoop running=True closed=False debug=False>
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 1922, in _run_once
    handle._run()
    │      └ <function Handle._run at 0xb6382f28>
    └ <Handle <TaskStepMethWrapper object at 0xb6128f70>()>
  File "/usr/local/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
    │    │            │    │           │    └ <member '_args' of 'Handle' objects>
    │    │            │    │           └ <Handle <TaskStepMethWrapper object at 0xb6128f70>()>
    │    │            │    └ <member '_callback' of 'Handle' objects>
    │    │            └ <Handle <TaskStepMethWrapper object at 0xb6128f70>()>
[...]

PR #2269, opened 2023-12-21
Update kraken

Williangalvani:
helps #2267 

it still looks like some data gets stuck, buffered somewhere.


joaoantoniocardoso (approved):
Nice, Kraken is working again :)

PR #2271, opened 2023-12-22
install: install: Use current user as default user

patrickelectric:
The current user also works for sudo commands


joaoantoniocardoso (approved):

ES-Alexander:
Relevant to #1073

PR #2273, opened 2023-12-22
core: tools: scripts: red-pill: Add option to select user

patrickelectric:


core/tools/scripts/red-pill
```diff
@@ -1,5 +1,38 @@
...
+      usage
+      ;;
+    : )
+      echo "Invalid Option: -$OPTARG requires an argument" 1>&2
+      usage
+      ;;
+  esac
+done
+shift $((OPTIND -1))
+
```
joaoantoniocardoso, line 33:
```suggestion

# Don't simply ignore additional arguments
if [ $# -gt 0 ]; then
  echo "Invalid Argument: $1" 1>&2
  usage
fi

```

patrickelectric: I'm not sure if I follow, what is the use case for this ?

joaoantoniocardoso: `redpill myusername` would print the usage, instead of ignoring the argument. It's pretty common for people to miss the argument identifier (say, `-u`) before the argument.

patrickelectric: Oh, nice!

ES-Alexander:
Relevant to #1073

PR #2265, opened 2023-12-19
bootstrap: Fix delta time calculation using monotonic clock

patrickelectric:
`time.time()` use system clock to return current time. If the system time changes while in operation, with the timezone or connecting to the internet NTP server, our delta calculation will be wrong.

This pr uses `time.monotonic()` that:
> Return the value (in fractional seconds) of a monotonic clock, i.e. a clock that cannot go backwards. The clock is not affected by system clock updates. The reference point of the returned value is undefined, so that only the difference between the results of two calls is valid.

Fix #2260 #2248 #2105 

Detected from Tony log:
```
2023-12-07 08:16:14.876 | INFO     | bootstrap.bootstrap:start:180 - Starting bluerobotics/blueos-core
2023-12-07 08:16:16.935 | INFO     | bootstrap.bootstrap:start:197 - core (bluerobotics/blueos-core:1.1.1) started
2023-12-07 08:16:16.936 | WARNING  | bootstrap.bootstrap:run:245 - core is not running, starting..
2023-12-07 08:16:16.963 | WARNING  | bootstrap.bootstrap:is_version_chooser_online:222 - Could not talk to version chooser for 64728.144294023514: HTTPConnectionPool(host='localhost', port=80): Max retries exceeded with url: /version-chooser/v1.0/version/current (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0xb5857470>: Failed to establish a new connection: [Errno 111] Connection refused'))
2023-12-07 08:16:16.964 | WARNING  | bootstrap.bootstrap:run:262 - Core has not responded in 5 minutes, resetting to factory...
```

PR #2277, opened 2024-01-02
1.1: bootstrap: Fix delta time calculation using monotonic clock

patrickelectric:
Cherry-pick #2265 to be available on next beta 

I'm waiting @vshie further comments about it before moving


patrickelectric:
From tony:
> yes, go ahead an add it! I continue to use vehicles without revert!

PR #2286, opened 2024-01-04
1.1: core: vehicle-setup: fix ArduSub lights mappings

patrickelectric:
From #2285


ES-Alexander (approved):
Thanks for the backport :-)

PR #2288, opened 2024-01-04
github: workflows: test-and-deploy: Add timeout for deploy-raspberry-image

patrickelectric:


.github/workflows/test-and-deploy.yml
```diff
@@ -249,6 +249,7 @@ jobs:
   deploy-raspberry-image:
     needs: deploy-docker-images
     if: github.event_name != 'pull_request'
+    timeout-minutes: 120 # Detect if we get in a infinity loop or in a weird state
```
RaulTrombin, line 252: Detect if it gets into an infinite loop or some unexpected state.

RaulTrombin (commented):
2h seems fine checking previous jobs, 1~1:30

PR #2299, opened 2024-01-14
core: services: ardupilot_manager: Increase memory limit

patrickelectric:


Williangalvani (approved):
we should consider changing how we handle the json file. maybe theres a more memory-efficient way? this solves the issue at hand, though, so merge away

patrickelectric:
> we should consider changing how we handle the json file. maybe theres a more memory-efficient way? this solves the issue at hand, though, so merge away

I did try to find something, but it appears that something bigger is wrong, the total amount of the memory used on this manifest is around 30MB from my tests. There are others parts of the ardupilot manager that is using more RAM. I'm not sure if the 200MB includes mavlink-router and ardupilot itself, since both are initialized by ardupilot manager.

PR #2300, opened 2024-01-14
ardupilot_manager: mavlink_proxy: Increase wait time for process to start and finish

patrickelectric:
1 second appears to be not enough, with 3 seconds it improves the check for the router.

### DIFF
--- a/core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
@@ -102,7 +102,7 @@ class AbstractRouter(metaclass=abc.ABCMeta):
             *shlex.split(command), stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
         )
 
-        await asyncio.sleep(1)  # Non-blocking sleep
+        await asyncio.sleep(3)  # Non-blocking sleep
         if not await self.is_running():
             raise MavlinkRouterStartFail("Failed to initialize Mavlink router")
         await self.start_house_keepers()
@@ -112,11 +112,11 @@ class AbstractRouter(metaclass=abc.ABCMeta):
             if self._subprocess is not None:
                 logger.error("terminate")
                 self._subprocess.terminate()
-                await asyncio.sleep(1)  # Non-blocking sleep
+                await asyncio.sleep(3)  # Non-blocking sleep
                 if await self.is_running():
                     logger.error("kill")
                     self._subprocess.kill()
-                    await asyncio.sleep(1)
+                    await asyncio.sleep(3)
                 await self._subprocess.wait()  # Wait for the subprocess to terminate
         else:
             logger.debug("Tried to stop router, but it was already not running.")
--- a/core/services/ardupilot_manager/mavlink_proxy/test_all.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/test_all.py
@@ -238,7 +238,7 @@ async def serial_test_mavp2p(valid_output_endpoints: Set[Endpoint], valid_master
     )
 
 
-@pytest.mark.timeout(60)
+@pytest.mark.timeout(120)
 @pytest.mark.asyncio
 async def test_router(valid_output_endpoints: Set[Endpoint], valid_master_endpoints: Set[Endpoint]) -> None:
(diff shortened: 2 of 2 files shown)


joaoantoniocardoso (changes requested):
This patch (probably) creates this bug: https://github.com/bluerobotics/BlueOS/issues/2306

update: false positive. The bug is also in master.

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
```diff
@@ -112,11 +112,11 @@ async def exit(self) -> None:
             if self._subprocess is not None:
                 logger.error("terminate")
                 self._subprocess.terminate()
-                await asyncio.sleep(1)  # Non-blocking sleep
+                await asyncio.sleep(3)  # Non-blocking sleep
                 if await self.is_running():
                     logger.error("kill")
                     self._subprocess.kill()
-                    await asyncio.sleep(1)
+                    await asyncio.sleep(3)
```
joaoantoniocardoso, line 119: can't we await for the kill to return instead of waiting for a fixed time?

joaoantoniocardoso: perhaps:
```python
self._subprocess.kill()
await self._subprecess.wait()
```

patrickelectric: for that we could move to asyncio subprocess, we could work that on a future PR and open an issue.

patrickelectric: ouch, you are correct, will update

patrickelectric: @joaoantoniocardoso btw, we already such thing on the following line

joaoantoniocardoso: uh, I wonder why that isn't enough?

patrickelectric: I'll do more tests to be sure

joaoantoniocardoso: 1. if it is running, terminate the process
2. wait for 3 seconds
3. if it is still running, kill the process
4. wait for 3 seconds (why, what for?)
5. wait for the process

joaoantoniocardoso: maybe something using a timeout like:

```python
if await self.is_running():
  if self._subprocess is not None:

    self._subprocess.terminate()
    try:
      await asyncio.wait_for(self._subprocess, timeout=3)
    except asyncio.TimeoutError:
      if await self.is_running(): 
        self._subprocess.kill()
        await self._subprecess.wait() # Wait for the subprocess to terminate
        # or maybe instead we need to: await self._subprocess.communicate()
```

patrickelectric:
> This patch (probably) creates this bug: #2306

how is is this connected to the PR ?

core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py
patrickelectric: I did try that and now the issue is back again. Since we have limited time, is there a problem with the original approach ? Besides the verbosity/timed logic.

joaoantoniocardoso: > I did try that and now the issue is back again. Since we have limited time, is there a problem with the original approach ? Besides the verbosity/timed logic.

The original approach seems fragile. It might not work on different binary versions or other scenarios like different CPU loads...

joaoantoniocardoso:
> > This patch (probably) creates this bug: #2306
> 
> how is is this connected to the PR ?

Desconsider that. I found the same bug in the master as well.

patrickelectric:
@joaoantoniocardoso I just tested the original code of this PR again and it works without problems. I vote to merge this now and deal with the syntax or helper functions later.

joaoantoniocardoso (approved):
Let's move. Let's just suspect this subprocess thing if similar errors happen.

PR #2302, opened 2024-01-14
core: frontend: vite.config: Use blueos-avahi over blueos

patrickelectric:
-avahi is available over all network interfaces


joaoantoniocardoso (approved):

joaoantoniocardoso:
Coudln't we turn 'blueos-avahi' into 'blueos', and 'blueos' into 'blueos-eth'?

patrickelectric:
> Couldn't we turn 'blueos' into 'blueos-eth', and 'blueos-avahi' into 'blueos'?

this will affect directly people that use the tether, by default we use blueos.local to work only over ethernet to avoid connection issues when using poor network interfaces. But that's up to discussion.

PR #2317, opened 2024-01-17
Add support for Raspberry pi OS Bookworm (and, to some extent, Pi5)

Williangalvani:


RaulTrombin:
@Williangalvani also we need to add `arm_64bit=0` to the config.txt file. If not it will load the 64 bit.

RaulTrombin:
> I expect this to break AT LEAST the pi4 serial port mapping
> 
> known issues:
> 
> * [ ]  ssh is not getting enabled automatically, even with the ssh file at /boot
> * [ ]  pi5 i2c ports are not working, probably need some update to the overlays

@Williangalvani Have you tryed adding the pi user?

generate the encrypted 'raspberry' password and copy
echo 'raspberry' | openssl passwd -6 -stdin

nano /boot/userconfig
`pi:openssl_thing_passwd`

nano /boot/ssh

Williangalvani:
> @Williangalvani Have you tryed adding the pi user?

yes, that is the second commit

PR #2318, opened 2024-01-17
core: frontend: warn developers if they use a bad BLUEOS_ADDRESS environment variable

Williangalvani:


joaoantoniocardoso (approved):

joaoantoniocardoso:
TY

PR #2322, opened 2024-01-19
Bridges update

### LINKED ISSUE #2304: Serial ports are being affected by ping service
An user reported that they were getting no data. the issue seems to be that the serial port baudrate is incorrect when bridges tries to receive data. doing `screen /dev/ttyAMA1 38400` seems to solve the issue and trigger the data into flowing.

### LINKED ISSUE #2305: Bridges should allow individual configuration of uplink and downlink, on the same port
Some software do not use the outgoing port for receiving data, and we should deal with it

Williangalvani:
This splits udp port handling into udp target port and udp lister port.

client will always send data to `udp_target_port`.
server will always listen at `udp_listen_port`.
additionally, this now allows the client to pick the port it will bind to, using `udp_listen_port`.
this is useful for implementations where whatever is talking to us is sending/receiving packets naively, without an actual udp "connection". 

~depends on https://github.com/patrickelectric/bridges/pull/11~

fix #2305 
fix #2304

### DIFF
--- a/core/services/bridget/bridget.py
+++ b/core/services/bridget/bridget.py
@@ -11,7 +11,7 @@ from serial.tools.list_ports_linux import SysFS
 from settings import BridgeSettingsSpecV1, SettingsV1
 
 
-class BridgeSpec(BaseModel):
+class BridgeFrontendSpec(BaseModel):
     """Basic interface for 'bridges' links."""
 
     serial_path: str
@@ -26,8 +26,8 @@ class BridgeSpec(BaseModel):
         return hash(str(self))
 
     @staticmethod
-    def from_settings_spec(settings_spec: BridgeSettingsSpecV1) -> "BridgeSpec":
-        return BridgeSpec(
+    def from_settings_spec(settings_spec: BridgeSettingsSpecV1) -> "BridgeFrontendSpec":
+        return BridgeFrontendSpec(
             serial_path=settings_spec.serial_path,
             baud=settings_spec.baudrate,
             ip=settings_spec.ip,
@@ -39,13 +39,13 @@ class Bridget:
     """Manager for 'bridges' links."""
 
     def __init__(self) -> None:
-        self._bridges: Dict[BridgeSpec, Bridge] = {}
+        self._bridges: Dict[BridgeFrontendSpec, Bridge] = {}
         self._settings_manager = Manager("bridget", SettingsV1)
         self._settings_manager.load()
         for bridge_settings_spec in self._settings_manager.settings.specs:
             try:
                 logging.debug(f"Adding following bridge from persistency '{bridge_settings_spec}'.")
-                self.add_bridge(BridgeSpec.from_settings_spec(bridge_settings_spec))
+                self.add_bridge(BridgeFrontendSpec.from_settings_spec(bridge_settings_spec))
             except Exception as error:
                 logging.exception(f"Could not add bridge '{bridge_settings_spec}'. {error}")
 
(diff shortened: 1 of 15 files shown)


core/start-blueos-core
```diff
@@ -15,6 +15,7 @@ MAV_COMPONENT_ID_ONBOARD_COMPUTER4=194
 # Enable Rust backtrace for all programs
 RUST_BACKTRACE=1
 
+RUN_AS_BLUEOS_USER="sudo -u blueos"
```
joaoantoniocardoso, line 18: in case `sudo` is not available, you can `su blueos -c "commandline with arguments"`

Williangalvani: the quotes in there make commands insanely hard to assemble. Do you want to take a look at it? I tried for a while but ended up giving up

Williangalvani:
ping @patrickelectric @joaoantoniocardoso

core/services/bridget/settings.py
```diff
@@ -42,3 +42,58 @@ def migrate(self, data: Dict[str, Any]) -> None:
...
+                new_specs.append(
+                    {
+                        "serial_path": spec["serial_path"],
+                        "baudrate": spec["baudrate"],
+                        "ip": spec["ip"],
+                        "udp_target_port": 0 if server else spec["udp_port"],
+                        "udp_listen_port": spec["udp_port"] if server else 0,
+                    }
+                )
+            data["specsv2"] = new_specs
```
patrickelectric, line 99: why not just replacing specs directly ?

core/start-blueos-core
```diff
@@ -85,6 +86,12 @@ else
     echo "Waring: Using dockers's resolv.conf, this may cause DNS issues if the host's network configuration changes."
 fi
 
+# fix permissions for logging folders:
+find /var/logs/blueos -type d -exec chmod 777 {} \;
```
patrickelectric, line 90: The default folder permission should be "755", no ?

core/services/bridget/bridget.py
```diff
@@ -8,44 +9,50 @@
...
 
 
 class Bridget:
     """Manager for 'bridges' links."""
 
     def __init__(self) -> None:
-        self._bridges: Dict[BridgeSpec, Bridge] = {}
-        self._settings_manager = Manager("bridget", SettingsV1)
+        self._bridges: Dict[BridgeFrontendSpec, Bridge] = {}
+        self._settings_manager = Manager("bridget", SettingsV2, USERDATA / "settings" / "bridget")
```
patrickelectric, line 50: Why are we moving this to userdata over our standard settings folder  ?

patrickelectric (changes requested):

core/start-blueos-core
joaoantoniocardoso: it should be something like:
```Bash

BLUEOS_USER='blueos'

run_as_blueos_user () {
  su "$BLUEOS_USER" -c "$1"
}

run_as_blueos_user 'ls -lah /dev/video*'
```

core/services/bridget/bridget.py
Williangalvani: non-root users have no access to our regular folder because it is under /root/

patrickelectric: It would be nice to have a comment explaining that in the code

core/start-blueos-core
Williangalvani: yep, updated

core/services/bridget/settings.py
Williangalvani: sounds good, that is just how my brain works. updated.

patrickelectric (approved):
Be free to merge it

Williangalvani:
cool, I'm trying @joaoantoniocardoso 's suggestion and I'll merge it after I'm done

core/start-blueos-core
Williangalvani: didn't quite work, I think this need to be on its own file so it can be used inside the new tmux sessions.
I'll try it in a subsequent pr

PR #2325, opened 2024-01-20
Try to recover the system if available size in critical

patrickelectric:
Once the system gets in a critical state where the disk is full, it gets in a condition that recover is almost impossible. The system is unstable, the backend services and frontend may not work or be available for the user to recover at all. As a last resource, we delete BlueOS logs if the filesystem. 

Fix https://github.com/bluerobotics/BlueOS/issues/2323

### DIFF
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -20,7 +20,7 @@ from fastapi_versioning import VersionedFastAPI, version
 from loguru import logger
 
 SERVICE_NAME = "commander"
-LOG_FOLDER_PATH = "/var/logs/blueos"
+LOG_FOLDER_PATH = os.environ.get("BLUEOS_LOG_FOLDER_PATH", "/var/logs/blueos")
 
 limit_ram_usage()
 
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -15,6 +15,22 @@ MAV_COMPONENT_ID_ONBOARD_COMPUTER4=194
 # Enable Rust backtrace for all programs
 RUST_BACKTRACE=1
 
+# Set BlueOS log folder
+BLUEOS_LOG_FOLDER_PATH="/var/logs/blueos"
+
+# The system may start with full disk, in this case resulting in an unstable behavior.
+# As an attempt to recover the system, lets delete the log files before current boot
+AVAILABLE_SPACE_MB=$(($(stat -f / --format="%a*%S/1024**2")))
+CRITICAL_SPACE_LIMIT_MB=100
+echo "Available disk space: ${AVAILABLE_SPACE_MB}MB"
+(( AVAILABLE_SPACE_MB < CRITICAL_SPACE_LIMIT_MB )) && (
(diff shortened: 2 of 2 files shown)


core/services/commander/main.py
```diff
@@ -21,6 +21,7 @@
 
 SERVICE_NAME = "commander"
 LOG_FOLDER_PATH = "/var/logs/blueos"
+LOG_FOLDER_PATH = os.environ.get("BLUEOS_LOG_FOLDER_PATH", "/var/logs/blueos")
```
joaoantoniocardoso, line 24:
```suggestion
LOG_FOLDER_PATH = os.environ.get("BLUEOS_LOG_FOLDER_PATH", "/var/logs/blueos")
```

patrickelectric: Thanks!

joaoantoniocardoso (approved):
The code seems correct, and 100MB of threshold feels okay since this is before the system starts up. 

How I imagine this patch acting:
1. some subsystem spams a log, depleting the disk space
2. the user feels the system is unusable
3. the user reboots the system
4. assuming there is some space left for the kernel and services, the system boots
5. critical space is detected
6. blueos logs are removed

I can see this patch not working if the kernel or docker fails to start because of insufficient disk space.

It would be nice to:
- inform the user of what happened
- inform us via telemetry system

Now, to _really_ protect the system, we should put the logs into a separate partition from the root (`/`), and the log manager service be responsible for cleaning it when necessary.

patrickelectric:
> The code seems correct, and 100MB of threshold feels okay since this is before the system starts up.
> 
> How I imagine this patch acting:
> 
> 1. some subsystem spams a log, depleting the disk space
> 2. the user feels the system is unusable
> 3. the user reboots the system
> 4. assuming there is some space left for the kernel and services, the system boots
> 5. critical space is detected
> 6. blueos logs are removed
> 
> I can see this patch not working if the kernel or docker fails to start because of insufficient disk space.
> 
> It would be nice to:
> 
> * inform the user of what happened
> * inform us via telemetry system
> 
> Now, to _really_ protect the system, we should put the logs into a separate partition from the root (`/`), and the log manager service be responsible for cleaning it when necessary.

Check: https://github.com/bluerobotics/BlueOS/issues/2327, https://github.com/bluerobotics/BlueOS/issues/2323, https://github.com/bluerobotics/BlueOS/issues/2326, https://github.com/bluerobotics/BlueOS/issues/1015

The docker is able to start, but everything after that just results in unstable behavior.
Some points that you suggested are already available as issues, others are relevant to recover the system.
- We should delete old logs when doing the rotation is noticing the the disk space is almost full.
- We should stop logging if the disk space is almost full.
- We should clean up old dockers that are not being used.
- We should clean up old docker artifacts that are not being used.
- We should allow user to delete all unused docker images.
- We should warn the user though cockpit that the companion computer is almost full in disk.
- We should erase older tlog or bin files if the disk is almost full.
- We should warn the user though BlueOS header that the disk is almost full and in critical state.
- We **may** not allow the user to arm the vehicle if the disk is almost full.
- We **may** do some of this steps automatically to try to recover the system once it starts.

PR #2332, opened 2024-01-23
core: frontend: components: wizard: Add 'remind me later' when skipping wizard

### LINKED ISSUE #2330: Add a remind me later when asking to update to new version

patrickelectric:
Fix #2330 

image

### DIFF
--- a/core/frontend/src/components/wizard/Wizard.vue
+++ b/core/frontend/src/components/wizard/Wizard.vue
@@ -50,12 +50,15 @@
               If your vehicle is already set up, you can skip this wizard.
             </v-card>
             <v-row class="pa-5 justify-space-between">
-              <v-btn
-                color="warning darken"
-                @click="setWizardVersion(); cancel()"
-              >
-                Skip Wizard
-              </v-btn>
+              <v-row class="pl-3 pt-2">
+                <v-btn
+                  color="warning darken"
+                  class="mr-5"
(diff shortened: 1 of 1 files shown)


Williangalvani:
I think it would be better if it was just an independent button. while this is obvious, it... _feels_ weird.

patrickelectric:
> I think it would be better if it was just an independent button. while this is obvious, it... _feels_ weird

Adding a third button just for that is kind of weird to me, this way we enforce the user to indeed disable the checkbox, make it more clear what he's doing.

joaoantoniocardoso:
btw, _DON'T SHOW AGAIN_ seems conflicting with _Remind me later_

patrickelectric:
> btw, _DON'T SHOW AGAIN_ seems conflicting with _Remind me later_
> 
> Maybe by positioning the checkbox below the _SKIP WIZARD_ button, it would look more like two separate groups (the _skip_ vs the _don't show_)?

That's why it's near the skip button. Putting in the bottom makes the entire thing weird with a space underneath.

patrickelectric:
@Williangalvani @joaoantoniocardoso changed to add a second dialog to be more clear.

ES-Alexander:
I don't think this needs to be documented - it's pretty obvious what the buttons do, so documentation would basically just be repeating the description that's included in the popup.

PR #2333, opened 2024-01-23
core: frontend: split vehicle setup configurations

Williangalvani:
This is the base for implementing all the sensor calibrations and peripherals setup. I'm opening this PR so we could parallelize some of the tasks at hand.
image

### DIFF
--- a/core/frontend/src/components/vehiclesetup/Configure.vue
+++ b/core/frontend/src/components/vehiclesetup/Configure.vue
@@ -1,56 +1,59 @@
 <template>
-  <div>
-    <v-row class="mt-5">
-      <v-col
-        cols="12"
-        sm="7"
+  <v-container fluid>
+    <v-tabs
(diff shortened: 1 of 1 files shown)


patrickelectric (commented):
Can we have at least what current exists on master before merging it ? Like gripper and etc.
We could also develop this in a branch in BlueOS, I'm aiming to make master stable and avoid removing stuff/break.

Williangalvani:
> Can we have at least what current exists on master before merging it ? Like gripper and etc. We could also develop this in a branch in BlueOS, I'm aiming to make master stable and avoid removing stuff/break.

While it looks like it, this PR is not actually removing anything useful, all the things it removes are present in the first "overview" page, and were just duplicated here. (I assume you asked because the diff is removing some stuff)

patrickelectric (approved):

PR #2336, opened 2024-01-24
Fix mavlink updateMessage usage  and barometer data

patrickelectric:
- MAVLink use filter as a regex argument, we should use the messages names that we receive over the regex as key for the structure
- setMessageRefreshRate accepts a regex, not a literal. We need to enforce the string that we want using ^..$

This PR:
Peek 24-01-2024 01-57

Master:
Peek 24-01-2024 01-59

### DIFF
--- a/core/frontend/src/store/mavlink.ts
+++ b/core/frontend/src/store/mavlink.ts
@@ -67,9 +67,9 @@ class MavlinkStore extends VuexModule {
       // We should not use `message.messageName` as dictionary key since it's a regex,
       // the best approach is to use the message name as key
       const messageName = (message.messageData.message as any).type
-      Vue.set(this.available_messages, messageName, message)
-      const header = message.messageData.header
+      const { header } = message.messageData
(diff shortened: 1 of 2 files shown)


core/frontend/src/store/mavlink.ts
```diff
@@ -64,7 +64,9 @@ class MavlinkStore extends VuexModule {
     if (message) {
       // TODO: Check if this is the best possible way to update `available_messages`
       // Reference: https://github.com/bluerobotics/blueos-docker/pull/508#discussion_r718729077
-      Vue.set(this.available_messages, message.messageName, message)
+      // We should not use `message.messageName` as dictionary key since it's a regex,
+      // the best approach is to use the message name as key
+      Vue.set(this.available_messages, (message.messageData.message as any).type, message)
```
Williangalvani, line 69: should we also update line 76? for consistency?

patrickelectric: Good catch!

Williangalvani (approved):

PR #2347, opened 2024-01-25
Added installation support for Ubuntu

agregghai:
Modified the install and bcm_27xx scripts to detect Ubuntu OS and make necessary changes to install successfully. Tested on Rpi4 running aarch64 Ubuntu 22.04.

### DIFF
--- a/install/boards/bcm_27xx.sh
+++ b/install/boards/bcm_27xx.sh
@@ -2,7 +2,7 @@
 
 echo "Configuring BCM27XX board (Raspberry Pi 4).."
 
-VERSION="${VERSION:-ubuntu_aarch64}"
+VERSION="${VERSION:-master}"
 GITHUB_REPOSITORY=${GITHUB_REPOSITORY:-HavocAI/BlueOS}
 REMOTE="${REMOTE:-https://raw.githubusercontent.com/${GITHUB_REPOSITORY}}"
 ROOT="$REMOTE/$VERSION"
--- a/install/boards/bcm_27xx.sh
+++ b/install/boards/bcm_27xx.sh
@@ -3,7 +3,7 @@
 echo "Configuring BCM27XX board (Raspberry Pi 4).."
 
 VERSION="${VERSION:-master}"
-GITHUB_REPOSITORY=${GITHUB_REPOSITORY:-bluerobotics/blueos-docker}
+GITHUB_REPOSITORY=${GITHUB_REPOSITORY:-HavocAI/BlueOS}
 REMOTE="${REMOTE:-https://raw.githubusercontent.com/${GITHUB_REPOSITORY}}"
 ROOT="$REMOTE/$VERSION"
 CMDLINE_FILE=/boot/firmware/cmdline.txt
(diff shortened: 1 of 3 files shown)


install/boards/bcm_27xx.sh
```diff
@@ -6,16 +6,33 @@ VERSION="${VERSION:-master}"
...
 alias curl="curl --retry 6 --max-time 15 --retry-all-errors"
 
+# Determine OS
+OS=$(lsb_release -is)
+
+if [ "$OS" = "Raspbian" ]; then
+    echo "Running on Raspbian."
+    CMDLINE_FILE="/boot/cmdline.txt"
+    CONFIG_FILE="/boot/config.txt"
+    OVERLAYS_DIR="/boot/overlays"
```
patrickelectric, line 18: @Williangalvani, since you was working on the support to the latest raspbian version, are these paths the same or they changed ?
My point is that such paths can be a consequence of the debian version that such distros are based over (Debian), so the checks should be done with the debian version and not with the distro name. @agregghai any idea ?

install/boards/bcm_27xx.sh
```diff
@@ -6,16 +6,33 @@ VERSION="${VERSION:-master}"
...
+elif [ "$OS" = "Ubuntu" ]; then
+    echo "Running on Ubuntu."
+    CMDLINE_FILE="/boot/firmware/cmdline.txt"
+    CONFIG_FILE="/boot/firmware/config.txt"
+    OVERLAYS_DIR="/boot/firmware/overlays"
+else
+    echo "Unknown OS"
+    exit 1
```
patrickelectric, line 26: We should not abort, we could try to check if the files exist and try to select it automatically or provide fault ones.
But in both cases, warning the user that the system is not supported.

install/install.sh
```diff
@@ -217,8 +231,24 @@ docker create \
...
+    Description=Start Docker container blueos-bootstrap at system start
+
+    [Service]
+    ExecStart=/usr/bin/docker start blueos-bootstrap
+
+    [Install]
+    WantedBy=multi-user.target" > /etc/systemd/system/blueos-bootstrap.service
+
+    systemctl daemon-reload
+    systemctl enable blueos-bootstrap.service
```
patrickelectric, line 250: Why docker start does not work with ubuntu like raspbian ? We rely on docker to start with the system, and that as consequence will start bootstrap.

patrickelectric (changes requested):

install/install.sh
agregghai: Ubuntu does not appear to use rc.local, so to get the bootstrap docker to start with the system, you need to use some other method like a systemd service, as I did here.

patrickelectric: Oh, thanks!

install/boards/bcm_27xx.sh
Williangalvani: yes, they are still the same

Williangalvani:
This is a great contribution, thank you for the PR! 
Other than @patrickelectric 's suggestions, it would be nice if you prepended the commit messages with "install: " and squashed the commits where possible. If you need help with anything, let us know =]

patrickelectric:
Hi @agregghai are you interested to continue with this PR ?

PR #2350, opened 2024-01-25
core: frontend: components: version-chooser: Allow browsers to save repository history

patrickelectric:
From a friendly stackoverflow:
> In most browsers, you can make your browser display values previously entered in input fields by giving a name to the input.

image


joaoantoniocardoso (approved):
Works for chrome, not for Firefox.

PR #2351, opened 2024-01-25
core: frontend: components: autopilot: Avoid usage of form

patrickelectric:
Fix NS_BINDING_ABORTED issue


joaoantoniocardoso (approved):
Brilliant. Thanks, @patrickelectric, I can confirm that this solves the issue!

PR #2357, opened 2024-01-31
Fix conflict between kill_autopilot and auto_restart_autopilot

patrickelectric:
The main problem here is that `should_run` was not set as False on the start of the function, resulting on `auto_restart_autopilot` to conflict with the firmware installation procedure.


Williangalvani (approved):
code looks good and it behaved well in my tests.

PR #2368, opened 2024-02-06
Create configuration page for lights

Williangalvani:
image

### DIFF
--- a/core/frontend/src/components/vehiclesetup/configuration/lights.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/lights.vue
@@ -1,0 +1,148 @@
+<template>
+  <div class="d-flex">
+    <v-card>
+      <v-card-title> Lights 1 (RCIN9) </v-card-title>
+      <v-card-text>
+        Here you can configure what pin outputs the signal for the first set of lights.
+        <v-select
+          v-model="lights1_new_param"
+          :items="recommended_params"
+          :item-text="friendlyName"
+          :item-value="'name'"
+          label="Lights 1"
+        />
+      </v-card-text>
+    </v-card>
+    <v-card>
+      <v-card-title> Lights 2 (RCIN10) </v-card-title>
+      <v-card-text>
+        What pin outputs the signal for the second set of lights.
+        <v-select
+          v-model="lights2_new_param"
+          :items="recommended_params"
+          :item-text="friendlyName"
+          :item-value="'name'"
+          label="Lights 2"
+        />
+      </v-card-text>
+    </v-card>
+    <v-card>
+      <v-card-title> Joystick steps </v-card-title>
+      <v-card-text>
+        How many button presses it takes to go from 0% to 100% brightness.
+        <br>
+        {{ light_steps?.value }} steps result in a {{ (100 / light_steps?.value).toFixed(1) }}%
+        increase per button press.
+        <v-text-field
+          ref="steps_input"
+          v-model="steps_new_value"
+          label="Joystick steps"
+          type="number"
+          @input="setSteps"
+        />
+      </v-card-text>
+    </v-card>
+  </div>
+</template>
+
+<script lang="ts">
+import mavlink2rest from '@/libs/MAVLink2Rest'
+import autopilot_data from '@/store/autopilot'
+import Parameter, { printParam } from '@/types/autopilot/parameter'
+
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/vehiclesetup/configuration/lights.vue
```diff
@@ -0,0 +1,156 @@
...
+          if (old_param.name === new_param_name) {
+            continue
+          }
+          if (old_param.value === Lights.Lights2) {
+            // set the old parameter to DISABLED
+            mavlink2rest.setParam(old_param.name, Lights.DISABLED, autopilot_data.system_id)
+          }
+        }
+        mavlink2rest.setParam(new_param_name, Lights.Lights2, autopilot_data.system_id)
+      }
```
patrickelectric, line 134: Can we encapsulate this in a single function for both lights ?

core/frontend/src/components/vehiclesetup/configuration/lights.vue
```diff
@@ -0,0 +1,156 @@
...
+          :item-text="friendlyName"
+          :item-value="'name'"
+          label="Lights 2"
+        />
+      </v-card-text>
+    </v-card>
+    <v-card>
+      <v-card-title> Joystick steps </v-card-title>
+      <v-card-text>
+        How many button presses it takes to go from 0% to 100% brightness.
```
patrickelectric, line 32: I believe that asking the user a percentage per step over the number of buttons press is more friendly. Maybe rephrasing it helps.

patrickelectric (changes requested):

Williangalvani:
updated

core/frontend/src/components/vehiclesetup/configuration/lights.vue
Williangalvani: improved the wording and added a live-updated percentage

core/frontend/src/components/vehiclesetup/configuration/lights.vue
Williangalvani: done

core/frontend/src/components/vehiclesetup/configuration/lights.vue
```diff
@@ -0,0 +1,148 @@
...
+  DISABLED = 0,
+}
+
+export default {
+  name: 'LightsConfigration',
+  data() {
+    return {
+      lights1_new_param: undefined as (string | undefined), // item-value must be a primitive
+      lights2_new_param: undefined as (string | undefined),
+      steps_new_value: 0,
```
patrickelectric, line 65: could we use a default value, like 10 ?
Other than that is just weird to start with 0 steps.

patrickelectric (changes requested):

core/frontend/src/components/vehiclesetup/configuration/lights.vue
Williangalvani: that should never happen, as we set it to the parameter value both when mounting (if the param is available) or when we detect a parameter change in the watcher. but I did update it anyway.

patrickelectric: When you configure it for the first time, if the lights are not configured, it starts with 0.

ES-Alexander:
Can we add a slider for the steps?
And min/max PWM limits?

Could also be nice to have a little graph showing the steps visually (with display options for both linear and Lumen-exponential brightness outputs).

It might make sense to have the channel controls in the top row, and then the steps slider + graph in a second row (instead of the current approach with everything in a single row).

Thinking something like this:
image

patrickelectric:
There is also another thing, the page end up showing for rover as well, should we remove it ?

Williangalvani:
> Can we add a slider for the steps? And min/max PWM limits?
> 
> Could also be nice to have a little graph showing the steps visually (with display options for both linear and Lumen-exponential brightness outputs).
> 
> It might make sense to have the channel controls in the top row, and then the steps slider + graph in a second row (instead of the current approach with everything in a single row).
> 
> Thinking something like this: image

Fancy!
Idk.
I'm 50% "we can do this later as there's more calibration/setup stuff to be done" and 50% "let's do it now or it won't get done".
@patrickelectric thoughts?

patrickelectric (approved):
We can move further required changes to issues.

PR #2379, opened 2024-02-12
core: frontend: fix nmea-injector validations

### LINKED ISSUE #2363: Creating nmea sockets is broken:
image

Williangalvani:
fix #2363


joaoantoniocardoso (approved):
The code looks good, but wait, why the linter didn't catch this?

Williangalvani:
> The code looks good, but wait, why the linter didn't catch this?

The linter is not as smart as we'd like. It often assumes some things wrong or just trusts our signatures are correct...

PR #2401, opened 2024-02-22
core: Improve docker compose structure

patrickelectric:
Needs #2399 indirectly


core/compose/compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      - ./workspace/config:/root/.config
+      - ./workspace/blueos:/etc/blueos
+      # Allow changes on repository reflects directly on the running container
+      - ../start-blueos-core:/usr/bin/start-blueos-core
+      - ../services:/home/pi/services
+    pid: "host"
+    environment:
+      - BLUEOS_DISABLE_SERVICES=cable_guy,wifi,commander
+      - BLUEOS_DISABLE_MEMORY_LIMIT=true
+    ports:
```
joaoantoniocardoso, line 27: Shouldn't the websockets also be here?

Couldn't we just use `network_mode: "host"` instead?

patrickelectric: The websockets use the same ports as the apis, no ? At least the is the case for mavlink2rest.

there is a comment explaining about network_mode host on the top.

joaoantoniocardoso: the MCM uses 6021 for the websocket

patrickelectric: @joaoantoniocardoso added

core/compose/compose.yml
```diff
@@ -0,0 +1,48 @@
...
+      - "80:80"
+      - "81:81"       # Helper
+      - "2748:2748"   # NMEA Injector
+      - "5173:5173"   # Cockpit development port
+      - "6020:6020"   # MAVLink Camera Manager REST API and interface
+      - "6021:6021"   # MAVLink Camera Manager Stun and Signalling server
```
joaoantoniocardoso, line 33:
```suggestion
      - "6021:6021"   # MAVLink Camera Manager Signalling server
```

patrickelectric: Can you explain why ?
Because the code says otherwise..

image

joaoantoniocardoso: https://github.com/mavlink/mavlink-camera-manager/issues/348

core/compose/compose.yml
```diff
@@ -0,0 +1,49 @@
...
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - /run/udev:/run/udev
+      - /etc/resolv.conf:/etc/resolv.conf
+      - /etc/machine-id:/etc/machine-id
+      - ./workspace/userdata:/usr/blueos/userdata
+      - ./workspace/logs:/var/logs/blueos
+      - ./workspace/config:/root/.config
+      - ./workspace/blueos:/etc/blueos
+      # Allow changes on repository reflects directly on the running container
```
Williangalvani, line 20:
```suggestion
      # Allow changes made to the repository to be directly reflected in the running container
```

core/compose/compose.yml
```diff
@@ -0,0 +1,49 @@
+version: '3.7'
+services:
+  blueos-core:
+    image: bluerobotics/blueos-core:1.2.0-beta.6
```
Williangalvani, line 4: should we use master here? this will get outdated very quickly

core/compose/compose.yml
joaoantoniocardoso: Should we add the RTSP server too? (8554)

patrickelectric: added

core/compose/compose.yml
patrickelectric: done

core/compose/compose.yml
patrickelectric: done

core/compose/compose.yml
joaoantoniocardoso: I'm not seeing it :x

patrickelectric: Sorry! I put the code and forget to push it. Should be there now

joaoantoniocardoso (approved):

PR #2403, opened 2024-02-22
core: services: helper: main: Add some header properties

patrickelectric:
This allows it to work with cockpit vite development server

Tested with #2401 

image

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -230,8 +230,13 @@ class Helper:
         conn: Optional[Union[http.client.HTTPConnection, http.client.HTTPSConnection]] = None
         request_response = SimpleHttpResponse(status=None, decoded_data=None, as_json=None, timeout=False, error=None)
 
+        # Based on requests library
+        headers = {
+            "User-Agent": "python",
+            "Accept-Encoding": "gzip, deflate",
+            "Accept": "*/*",
+        }
         # Prepare the header for json request
-        headers = {}
         if try_json:
             headers["Accept"] = "application/json"
 


Williangalvani:
that is wild. do you know _why_ this fixes the issue?

joaoantoniocardoso:
> that is wild. do you know _why_ this fixes the issue?

The _why_ is beyond my understandings, but I believe this is coming from what Patrick got from the Requests library, which defines its default as:

```Python
from urllib3.util import make_headers, parse_url

# Ensure that ', ' is used to preserve previous delimiter behavior.
DEFAULT_ACCEPT_ENCODING = ", ".join(
    re.split(r",\s*", make_headers(accept_encoding=True)["accept-encoding"])
)

def default_user_agent(name="python-requests"):
    """
    Return a string representing the default user agent.

    :rtype: str
    """
    return f"{name}/{__version__}"

def default_headers():
    """
    :rtype: requests.structures.CaseInsensitiveDict
    """
    return CaseInsensitiveDict(
        {
            "User-Agent": default_user_agent(),
            "Accept-Encoding": DEFAULT_ACCEPT_ENCODING,
            "Accept": "*/*",
            "Connection": "keep-alive",
        }
    )
```

The [make_headers from urllib3](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.make_headers) will give `‘gzip,deflate’`.

patrickelectric:
> > that is wild. do you know _why_ this fixes the issue?
> 
> The _why_ is beyond my understandings, but I believe this is coming from what Patrick got from the Requests library, which defines its default as:
> 
> ```python
> from urllib3.util import make_headers, parse_url
> 
> # Ensure that ', ' is used to preserve previous delimiter behavior.
> DEFAULT_ACCEPT_ENCODING = ", ".join(
>     re.split(r",\s*", make_headers(accept_encoding=True)["accept-encoding"])
> )
> 
> def default_user_agent(name="python-requests"):
>     """
>     Return a string representing the default user agent.
> 
>     :rtype: str
>     """
>     return f"{name}/{__version__}"
> 
> def default_headers():
>     """
>     :rtype: requests.structures.CaseInsensitiveDict
>     """
>     return CaseInsensitiveDict(
>         {
>             "User-Agent": default_user_agent(),
>             "Accept-Encoding": DEFAULT_ACCEPT_ENCODING,
>             "Accept": "*/*",
>             "Connection": "keep-alive",
>         }
>     )
> ```
> 
> The [make_headers from urllib3](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.make_headers) will give `‘gzip,deflate’`.

Yep, it was from the requests lib, I used Wireshark and found out that the header was necessary.

joaoantoniocardoso (approved):
Can we add a note that this is using the default headers from Requests as the reference?

PR #2407, opened 2024-02-23
frontend: components: VideoThumbnail: Show last thumbnail over fetching animation

patrickelectric:
A problem on the video page is that mostly of the time is showing the fetching animation over showing the thumbnail itself.
The PR changes the logic to show the last thumbnail, and only show the fetching animation when a thumbnail is not available.

### DIFF
--- a/core/frontend/src/components/video-manager/VideoThumbnail.vue
+++ b/core/frontend/src/components/video-manager/VideoThumbnail.vue
@@ -5,12 +5,8 @@
       :width="width"
       tile
     >
-      <img
-        v-if="thumbnail?.status === 200"
-        :src="thumbnail?.source"
-      >
       <span
-        v-else-if="not_available"
+        v-if="not_available"
         :style="{ border: '2px dashed' }"
         class="text-caption"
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/video-manager/VideoThumbnail.vue
```diff
@@ -110,7 +110,10 @@ export default Vue.extend({
   },
   methods: {
     updateThumbnail() {
-      this.thumbnail = video.thumbnails.get(this.source)
+      const result = video.thumbnails.get(this.source)
+      if (result.status === 200) {
```
joaoantoniocardoso, line 114:
```suggestion
      if (result?.status === 200) {
```

joaoantoniocardoso (approved):
Noice. Just missing the '?'.

core/frontend/src/components/video-manager/VideoThumbnail.vue
patrickelectric: Thanks!

PR #2411, opened 2024-02-26
Bridges updates,  part two

Williangalvani:
requires #2322

This improves management of brigdes, including `communicate` to check the status of the process and to consume data, as well as adding a status flag and textual output available to the frontend


patrickelectric:

PR #2418, opened 2024-02-27
core: tools: blueos_startup_update: Abort if startup file does not exist

patrickelectric:


Williangalvani:
is this mostly for when running in docker compose? can we get a comment in the code mentioning when it is relevant?

patrickelectric:
> is this mostly for when running in docker compose? can we get a comment in the code mentioning when it is relevant?

Not necessary, it's also helpful if someone is running BlueOS without bootstrap

Williangalvani (approved):
I still think a comment saying that this is not expected to happen in a "production" system

PR #2419, opened 2024-02-27
core: libs: commonwealth: general: Use losof -b to avoid kernel blocks

patrickelectric:


Williangalvani:
is it working as expected?
[the manual seems to say it wont accept "names",](https://man7.org/linux/man-pages/man8/lsof.8.html#AVOIDING_KERNEL_BLOCKS) though I'm not quite sure of what that means.
image

patrickelectric:
In the end it did not fix the blocking problem

PR #2420, opened 2024-02-27
core: libs: commonwealth: general: Add timeout for lsof operation

patrickelectric:
This fixes the lsof operation that is blocking and using 100% of cpu in some cases.

```
2024-02-27 19:22:03.150 | DEBUG    | __main__:zip_files:29 - Deleted file: /shortcuts/system_logs/ardupilot-manager/logfile_02-22-2024_10:39:21.log
2024-02-27 19:22:08.175 | ERROR    | commonwealth.utils.general:file_is_open:46 - Failed to check if file /shortcuts/system_logs/ardupilot-manager/logfile_02-27-2024_11:30:22.log is open, Command '['lsof', PosixPath('/var/logs/blueos/services/ardupilot-manager/logfile_02-27-2024_11:30:22.log')]' timed out after 5 seconds
```

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/general.py
+++ b/core/libs/commonwealth/commonwealth/utils/general.py
@@ -37,8 +37,19 @@ def delete_everything(path: Path) -> None:
 
 
 def file_is_open(path: Path) -> bool:
-    result = subprocess.run(["lsof", path.resolve()], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
-    return result.returncode == 0
+    try:
+        kernel_functions_timeout = str(2)
+        result = subprocess.run(
+            ["lsof", "-t", "-n", "-P", "-S", kernel_functions_timeout, path.resolve()],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            check=False,
+            timeout=5,
+        )
+        return result.returncode == 0
+    except Exception as error:
+        logger.error(f"Failed to check if file {path} is open, {error}")
+        return True
 
 
 @cache


core/libs/commonwealth/commonwealth/utils/general.py
```diff
@@ -37,8 +37,14 @@ def delete_everything(path: Path) -> None:
...
-    result = subprocess.run(["lsof", path.resolve()], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
-    return result.returncode == 0
+    try:
+        result = subprocess.run(
+            ["lsof", path.resolve()], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False, timeout=5
+        )
+        return result.returncode == 0
+    except Exception as error:
+        logger.error(f"Failed to check if file {path} is open, {error}")
+        return True
```
joaoantoniocardoso, line 47: If timed out, the file is considered open.

just commenting so we think about the behavior we are choosing

core/libs/commonwealth/commonwealth/utils/general.py
```diff
@@ -37,8 +37,14 @@ def delete_everything(path: Path) -> None:
 
 
 def file_is_open(path: Path) -> bool:
-    result = subprocess.run(["lsof", path.resolve()], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
-    return result.returncode == 0
+    try:
+        result = subprocess.run(
+            ["lsof", path.resolve()], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False, timeout=5
+        )
```
joaoantoniocardoso, line 43:
```suggestion
        kernel_functions_timeout = str(2)
        result = subprocess.run(
            ["lsof", "-t", "-n", "-P", "-S", kernel_functions_timeout, path.resolve()], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False, timeout=5
        )
```

From its [man pages](https://man7.org/linux/man-pages/man8/lsof.8.html):
>       -t     specifies that lsof should produce terse output with
>              process identifiers only and no header - e.g., so that the
>              output may be piped to [kill(1)](https://man7.org/linux/man-pages/man1/kill.1.html).  -t selects the -w option.
>       -n     inhibits the conversion of network numbers to host names
>              for network files.  Inhibiting conversion may make lsof
>              run faster.  It is also useful when host name lookup is
>              not working properly.
>       -P     inhibits the conversion of port numbers to port names for
>              network files.  Inhibiting the conversion may make lsof
>              run a little faster.  It is also useful when port name
>              lookup is not working properly.
>       -S [t] specifies an optional time-out seconds value for kernel
>              functions - [lstat(2)](https://man7.org/linux/man-pages/man2/lstat.2.html), [readlink(2)](https://man7.org/linux/man-pages/man2/readlink.2.html), and [stat(2)](https://man7.org/linux/man-pages/man2/stat.2.html) - that
>              might otherwise deadlock.  The minimum for t is two; the
>              default, fifteen; when no value is specified, the default
>              is used.

patrickelectric: We are not using numbers at all, only paths, I believe that -n and -P is kind of unnecessary

joaoantoniocardoso (approved):

Issue #2362, opened 2024-02-01
Bad report of reaching BlueOS through a non-wired connection


Williangalvani:
![image](https://github.com/bluerobotics/BlueOS/assets/4013804/1d4e5ff7-d2b6-40b3-b824-102b22ab56b0)
![image](https://github.com/bluerobotics/BlueOS/assets/4013804/13726a74-ff7e-487c-9dd2-ce687e014c97)

JoaoMario109:
Closing it per now due to not be able to replicate and that #2423 added new checks that would avoid this issue to happen

PR #2423, opened 2024-02-29
frontend: Beacon: Add double check for wireless network

JoaoMario109:
* Adds a double check for check in `is_connected_to_wifi` potentially avoid unexpected behavior where both wired and wireless interfaces could have the same IP
* Change computed property `wireless_interface_domains` to instead of checking for interfaces with type different than WIRED to consider only WIFI and HOTSPOT as Wireless interfaces, since usb interfaces currently holds type unknown and would be classified as Wireless

Probably linked to #2362


JoaoMario109:
Should we only use a `console.debug` in case of unexpected behavior or would some other approach be better?

patrickelectric:
@JoaoMario109 please try to rebase over later

PR #2428, opened 2024-03-06
Add extra check for empty file input in version chooser manual upload

JoaoMario109:


patrickelectric (commented):
Can we move alway from v-form ?
Reason: https://github.com/bluerobotics/BlueOS/pull/2351

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -596,6 +602,28 @@ export default Vue.extend({
...
       return this.available_versions.local.some((image) => image.sha === sha)
     },
+    get_input_file(): File | null {
+      const { files } = document.getElementById('file') as HTMLInputElement
+
+      if (files !== null && files.length > 0) {
+        return files[0]
+      }
+
+      return null
```
joaoantoniocardoso, line 612:
```suggestion
      if (files == null || files.length <= 0) {
          return null
      }
      
      return files[0]
```

just so the main/intended flow is obvious.

JoaoMario109: Changed to the suggested!

joaoantoniocardoso (approved):

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -596,6 +602,28 @@ export default Vue.extend({
...
       return this.available_versions.local.some((image) => image.sha === sha)
     },
+    get_input_file(): File | null {
+      const { files } = document.getElementById('file') as HTMLInputElement
+
+      if (files == null || files.length <= 0) {
+        return null
+      }
+
+      return files[0]
```
patrickelectric, line 612:
```suggestion
      return files?.[0]
```

patrickelectric (changes requested):

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -596,6 +602,28 @@ export default Vue.extend({
...
+    },
+    is_file_input_not_empty(v: File | null): true | string {
+      this.file_input_error = ''
+      return !!v || this.input_file_required_message
+    },
+    validateInputFileForm(): boolean {
```
patrickelectric, line 618: The style should follow the others functions as well

patrickelectric (changes requested):

core/frontend/src/components/version-chooser/VersionChooser.vue
JoaoMario109: Changed both the methods and the computed property to follow the convention, only error messages remained snake case due the others data be in this way also

patrickelectric (approved):

Issue #2274, opened 2023-12-23
bug: Beacon services list does not contain the IP of the network USB connection


rafaellehmkuhl:

JoaoMario109:
Closing it by now as fixed in #2433, if more related problems appears it can be reopened

PR #2433, opened 2024-03-08
core: services: Fix USB not being listed in Beacon

JoaoMario109:
Add a new interface type in beacon to avoid USB interfaces to get listed as UNKNOWN interface type and change to use guess_from_name in set_hostname instead of replicating the same logic.

Probably linked to #2274


patrickelectric (approved):

patrickelectric:
@JoaoMario109 pls check CI

PR #2445, opened 2024-03-13
Add check for free space in kraken prior to install new extension

### LINKED ISSUE #2307: kraken, check if there is enough space before installing a new version

JoaoMario109:
Adds a check for free space in kraken, right now only check a hard coded value o 1GB.

Closes #2307


core/services/kraken/kraken.py
```diff
@@ -346,5 +346,13 @@ async def load_stats(self) -> Dict[str, Any]:
             }
         return result
 
+    async def has_enough_disk_space(self, path: str = "/", required: int = 1073741824) -> bool:
```
patrickelectric, line 349:
```suggestion
    async def has_enough_disk_space(self, path: str = "/", required_mb: int = 2 ** 10) -> bool:
```

core/services/kraken/kraken.py
```diff
@@ -346,5 +346,13 @@ async def load_stats(self) -> Dict[str, Any]:
...
+        try:
+            free_space = psutil.disk_usage(path).free
+            # Right now only check if there is 1GB of free space as a hardcoded value
+            return free_space > required
```
patrickelectric, line 353:
```suggestion
            return free_space > required_md * 2 ** 20
```

patrickelectric (changes requested):

core/services/kraken/kraken.py
JoaoMario109: Added!

core/services/kraken/kraken.py
JoaoMario109: Added!

PR #2446, opened 2024-03-13
core: tools: linux2rest: Update to v0.5.4

patrickelectric:
This new version follows the same log structure as mavlink-camera-manager and also allows the user to log custom system information

### DIFF
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -109,7 +109,7 @@ SERVICES=(
     'nmea_injector',250,"nice -19 $SERVICES_PATH/nmea_injector/nmea_injector/main.py"
     'helper',250,"$SERVICES_PATH/helper/main.py"
     'iperf3',250," iperf3 --server --port 5201"
-    'linux2rest',250,"linux2rest"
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso (changes requested):
1. We need to update the command-line to add the log settings in the PR, without it, it'll just not run.

I was testing with `--log-system netstat=10,platform=10,serial-ports=10,system-cpu=10,system-disk=30,system-info=10,system-memory=10,system-network=10,system-process=10,system-temperature=10,system-unix-time-seconds=10`

2. The BlueOS system-information frontend seems to be failing to get the temperature:
image

3. Keep in mind that CPU peaks might be a little high for the max frequencies:

https://github.com/bluerobotics/BlueOS/assets/5920286/aa352f05-801a-4fff-93ff-d3b5733f9c50

4. The .log (wrongly named) contains no logged data, just warn/info messages.

patrickelectric:
@joaoantoniocardoso the filename should be fixed now.
But, I want to do the tunning of the logging system in following PRs.

joaoantoniocardoso:
> @joaoantoniocardoso the filename should be fixed now. But, I want to do the tunning of the logging system in following PRs.

I meant.. if we merge this as it was, there would be no linux2rest running. It was panicking

patrickelectric:
@joaoantoniocardoso thanks for pointing, indeed it was crashing. This final version should have everything fixed.

joaoantoniocardoso (approved):

PR #2458, opened 2024-03-14
Correct parameters name in ParameterLoader

### LINKED ISSUE #2447: Load default parameters should show a enum names over values
Autopilot Parameters:
image
Load default parameters:
image

Load default parameters should show enum name and not value, it should also provide a way for the user to see the parameter description.
The description may show when the mouse is hovering. UI up to discussion.

JoaoMario109:
Change to use correct parameters values and units instead of just showing the raw enumeration value

Closes #2447

### DIFF
--- a/core/frontend/src/components/parameter-editor/ParameterLoader.vue
+++ b/core/frontend/src/components/parameter-editor/ParameterLoader.vue
@@ -34,7 +34,7 @@
         </v-row>
         <!-- display all parameters in a concise table using virtual scroller -->
         <v-virtual-scroll
-          :items="printable(different_param_set)"
+          :items="parametersFromSet(different_param_set)"
           height="300"
           item-height="30"
           class="virtual-table"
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/parameter-editor/ParameterLoader.vue
```diff
@@ -270,12 +288,22 @@ export default Vue.extend({
...
+          current: currentParameter,
+          new: { ...currentParameter, value },
+        }
+      })
+    },
+    prettyNameFromParameter(parameter: Parameter) {
+      const paramValueText = printParam(parameter)
+      const paramUnitsText = parameter?.units ? `[${parameter.units}]` : ''
+
+      return paramValueText + paramUnitsText
```
joaoantoniocardoso, line 306:
```suggestion
      return paramValueText + ' ' + paramUnitsText
```

I don't remember the correct syntax, but the idea is to have something like "10 [A]" instead of "10A".

joaoantoniocardoso (commented):
I'm not against the proposal, but I'd prefer to have a separate column with the units instead of having it appended to each value.

JoaoMario109:
> I'm not against the proposal, but I'd prefer to have a separate column with the units instead of having it appended to each value.

I used this way to keep concise with the already existing implementation in

https://github.com/bluerobotics/BlueOS/blob/d65793d19769a06a8aa586c7b2458c794d02e54f/core/frontend/src/components/parameter-editor/ParameterEditor.vue#L65

But is actually easy to change if you guys think its better

patrickelectric (approved):

patrickelectric:
@JoaoMario109 can you rebase over master ?

PR #2459, opened 2024-03-14
Add force tag when deleting images in version chooser

### LINKED ISSUE #2431: bug: Add force remove when deleting local images
### Bug description

Add a force tag when deleting local images in version chooser since some of images are referenced and cannot be deleted

### Steps to reproduce

Try to remove the local version and got this error

### Primary pain point(s)

_No response_

### Additional context

_No response_

### Prerequisites

JoaoMario109:
Closes #2431

### DIFF
--- a/core/services/versionchooser/utils/chooser.py
+++ b/core/services/versionchooser/utils/chooser.py
@@ -287,7 +287,7 @@ class VersionChooser:
         # actually attempt to delete it
         logger.info(f"Deleting image {image}:{tag}...")
         try:
-            await self.client.images.delete(full_name, force=False, noprune=False)
+            await self.client.images.delete(full_name, force=True, noprune=False)
             logger.info("Image deleted successfully")
             return web.Response(status=200)
         except Exception as e:


joaoantoniocardoso (approved):

Williangalvani:
will this actually alllow the api to delete the currently running image? would this kill the container?

patrickelectric (approved):

JoaoMario109:
> will this actually alllow the api to delete the currently running image? would this kill the container? can you try it at http://blueos.local:8081/v1.0/ui/#/default/main.delete_version ?

I added the force option because it to enables the removal of the image even in cases where a container was stopped but not terminated. Within our existing model, the BlueOS frontend initiates a delete operation only for images that are not currently running. Therefore, introducing the force option will not alter the existing behavior. However, it will provide a safeguard by ensuring that images can still be removed under certain conditions. For instance, if a container is manually stopped on the Raspberry Pi but not terminated, and a subsequent attempt is made to delete the image, without the force option, the image would remain as a dangling image, preventing its deletion.

Williangalvani:
> > will this actually alllow the api to delete the currently running image? would this kill the container? can you try it at http://blueos.local:8081/v1.0/ui/#/default/main.delete_version ?
> 
> I added the force option because it enables the removal of the image even in cases where a container was stopped but not terminated. Within our existing model, the BlueOS frontend initiates a delete operation only for images that are not currently running. Therefore, introducing the force option will not alter the existing behavior. However, it will provide a safeguard by ensuring that images can still be removed under certain conditions. For instance, if a container is manually stopped on the Raspberry Pi but not terminated, and a subsequent attempt is made to delete the image, without the force option, the image would remain as a dangling image, preventing its deletion.

I understand your point, my concern is that someone could use the API to delete the image from the currently running container, in that case I'm not sure what would happen.

BUT, after seeing the code, I see we already have a safeguard for that, so we're good =]

Williangalvani (approved):

Issue #2109, opened 2023-09-14
bug: store should not show extensions that are not compatible with current architecture


patrickelectric:
### Bug description

It's possible to install an aarch64 extension in a raspberry pi 3

### Steps to reproduce

Install ROS extension

### Primary pain point(s)

It just crashes

### Additional context

We should not install invalid architectures 

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

PR #2466, opened 2024-03-19
Fix some problems related to order of operations in kraken update extesion

### LINKED ISSUE #2326: We should delete older extensions once update is done
The system is not being cleaned, when updating a extension to a newer version we should delete the previous one

JoaoMario109:
Closes #2326 and #2018


JoaoMario109:
Keep this holding until finishing the solution PR for #2109 after that I'll push the refactor with fixes related to both install/uninstall ordering and sizes checking

PR #2488, opened 2024-03-26
frontend: cosmos: Add last

patrickelectric:


core/frontend/src/cosmos.ts
```diff
@@ -15,7 +16,12 @@ declare global {
 
 // eslint-disable-next-line
 Array.prototype.first = function<T> (this: T[]): T | undefined {
-  return this.isEmpty() ? undefined : this[0]
+  return this[0]
+}
+
+// eslint-disable-next-line
+Array.prototype.last = function<T> (this: T[]): T | undefined {
+  return this.slice(-1)[0]
```
JoaoMario109, line 24: Maybe we could use `at`

```suggestion
  return this.at(-1)
```

patrickelectric: Thanks!

JoaoMario109 (approved):

Williangalvani:
nice

PR #2500, opened 2024-03-28
Add manifest typing and lib for caching in kraken

JoaoMario109:
Needed in #2486

### DIFF
--- a/core/services/kraken/manifest/manifest.py
+++ b/core/services/kraken/manifest/manifest.py
@@ -1,0 +1,152 @@
+import dataclasses
+import datetime
+from typing import List, Optional
+
+# Extra
+import aiohttp
+from dataclass_wizard import fromdict
+
+# Models
+from manifest.models import ExtensionVersion, RepositoryEntry
+
+# Current Extension Manifest host
+REPO_URL = "https://bluerobotics.github.io/BlueOS-Extensions-Repository/manifest.json"
+
+
+@dataclasses.dataclass
+class ManifestContent:
+    """
+    Represents the manifest content with cache control.
+
+    Attributes:
+        extensions (List[RepositoryEntry]): List of extensions in the manifest.
+        issued_at (Optional[int]): Timestamp when the manifest was issued.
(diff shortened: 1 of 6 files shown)


core/services/kraken/manifest/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+    def is_expired(self) -> bool:
+        """
+        Checks if the manifest is expired.
+
+        Returns:
+            bool: True if the manifest is expired, False otherwise.
+        """
+
+        return (int(datetime.datetime.now(datetime.timezone.utc).timestamp()) - self.issued_at) > self.expires_in
+
```
joaoantoniocardoso, line 41: Consider unifying the timestamp into something like:

```Python
@classmethod
def _current_timestamp(cls) -> int:
  return int(datetime.datetime.now(datetime.timezone.utc).timestamp())
```

core/services/kraken/manifest/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        Returns:
+            ManifestContent: Manifest content.
+
+        Raises:
+            RuntimeError: If the manifest could not be fetched.
+        """
+
+        async with aiohttp.ClientSession() as session:
+            async with session.get(REPO_URL) as resp:
```
joaoantoniocardoso, line 79: I know it was already implemented, but maybe it's a good idea to add a `, headers={'Accept': 'application/json'}`?

joaoantoniocardoso (commented):
Haven't tested yet

core/services/kraken/manifest/manifest.py
JoaoMario109: Added!

core/services/kraken/manifest/manifest.py
JoaoMario109: Added!

patrickelectric (changes requested):
@JoaoMario109 can you check what CI is detecting ?

patrickelectric:
merged in https://github.com/bluerobotics/BlueOS/pull/2486

PR #2501, opened 2024-03-28
Fix shellcheck warning listed in deploy

JoaoMario109:
Fixes this warning making clear the append statement

```
DEV_PART2+=2
^-------^ SC2324 (warning): var+=1 will append, not increment. Use (( var += 1 )), declare -i var, or quote number to silence.
```

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -58,7 +58,7 @@ MOUNT_LOCATION=/tmp/blueos_deploy
 mkdir -p $MOUNT_LOCATION
 
 DEV_PART2=$DEV_DISK
-DEV_PART2+=2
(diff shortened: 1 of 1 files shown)


deploy/deploy.sh
```diff
@@ -58,7 +58,7 @@ MOUNT_LOCATION=/tmp/blueos_deploy
 mkdir -p $MOUNT_LOCATION
 
 DEV_PART2=$DEV_DISK
-DEV_PART2+=2
+DEV_PART2="${DEV_PART2}2"
```
joaoantoniocardoso, line 61: I didn't even know that `+=` was a valid shell syntax

patrickelectric (approved):

PR #2503, opened 2024-03-28
core: frontend: fix prop type for compassDisplay

Williangalvani:
see https://github.com/bluerobotics/BlueOS/blob/master/core/frontend/src/components/vehiclesetup/configuration/compass/ArdupilotMavlinkCompassSetup.vue#L237

### DIFF
--- a/core/frontend/src/components/vehiclesetup/configuration/compass/CompassDisplay.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/compass/CompassDisplay.vue
@@ -49,8 +49,8 @@ export default Vue.extend({
       required: true,
     },
     colors: {
-      type: Array as PropType<string[]>,
-      default: () => ['green', 'blue', 'purple'],
+      type: Object as PropType<Dictionary<string>>,
+      default: () => ({} as Dictionary<string>),
     },
   },
   data() {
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/vehiclesetup/configuration/compass/CompassDisplay.vue
```diff
@@ -49,8 +49,8 @@ export default Vue.extend({
       required: true,
     },
     colors: {
-      type: Array as PropType<string[]>,
-      default: () => ['green', 'blue', 'purple'],
+      type: Object as PropType<Dictionary<string>>,
+      default: () => ({}),
```
joaoantoniocardoso, line 53: Is this empty default okay?

joaoantoniocardoso (approved):

core/frontend/src/components/vehiclesetup/configuration/compass/CompassDisplay.vue
patrickelectric: @Williangalvani any comment on that ?

Williangalvani: I found a different issue while looking at it, I'll come back to it

patrickelectric (approved):

PR #2504, opened 2024-03-28
Refactor COMMAND_LONG code

Williangalvani:
This creates `mavlink2rest.sendCommandLong()` and `mavlink2rest.waitForAck()` and updates the current calibration code to use them

### DIFF
--- a/core/frontend/src/libs/MAVLink2Rest/index.ts
+++ b/core/frontend/src/libs/MAVLink2Rest/index.ts
@@ -8,6 +8,8 @@ import { Dictionary } from '@/types/common'
 import Endpoint from './Endpoint'
 import Listener from './Listener'
 import messageId from './MessageID'
+import autopilot_data from '@/store/autopilot'
+import { MAVLinkType, MavCmd } from './mavlink2rest-ts/messages/mavlink2rest-enum'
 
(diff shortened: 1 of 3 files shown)


core/frontend/src/libs/MAVLink2Rest/index.ts
```diff
@@ -95,6 +97,67 @@ class Mavlink2RestManager {
...
+        },
+        target_system: autopilot_data.system_id,
+        target_component: 1,
+        confirmation: 0,
+      },
+    })
+  }
+
+
+  waitForAck(command: MavCmd, timeout: number = 3): Promise<any> {
```
patrickelectric, line 141: It would be nice to have a documentário or something explaining what is the unit of timeout

patrickelectric (changes requested):

core/frontend/src/libs/MAVLink2Rest/index.ts
patrickelectric: ping @Williangalvani

Williangalvani: I renamed timeout to timeout_seconds

PR #2527, opened 2024-04-11
Lock set operations in version chooser

patrickelectric:
Needs testing

Locks all set operations in version chooser to wait for operations that are already in progress.


patrickelectric:

PR #2530, opened 2024-04-11
Add plain shader loading overlay in BrIframe

JoaoMario109:
Based on discussions with @patrickelectric , this PR introduces a loading component in BrIFrame using shader and canvas for faster loading. It also implements a simple bubble fragment shader for the loading screen, laying the groundwork for the application's splash screen.

image

Currently, the loading logo geometries are rendered by the fragment shader directly. For better efficiency, it might be preferable in the future to transition to a vertex buffer and render on a second pass, rather than utilizing rect functions in fShaders.

### DIFF
--- a/core/frontend/src/components/utils/BrIframe.vue
+++ b/core/frontend/src/components/utils/BrIframe.vue
@@ -1,8 +1,8 @@
 <template>
   <v-sheet
-    id="brIframe"
     width="100%"
     height="100%"
+    style="overflow: hidden;"
   >
     <spinning-logo
       v-if="!gl_compatible && !iframe_loaded"
@@ -10,8 +10,7 @@
       subtitle="Loading external application..."
     />
     <canvas
-      v-if="!iframe_loaded"
-      v-show="gl_compatible"
+      v-if="gl_compatible && !iframe_loaded"
       ref="webglCanvas"
     />
     <iframe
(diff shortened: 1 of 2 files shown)


rafaellehmkuhl:
By curiosity, what was the performance issue and what was causing it?

core/frontend/src/components/utils/BrIframe.vue
```diff
@@ -24,8 +29,21 @@
 <script lang="ts">
 import Vue from 'vue'
 
+import { basic2dVertexShaderSource, bubbles2dFragmentShaderSource } from '../../utils/shaders'
```
Williangalvani, line 32:
```suggestion
import { basic2dVertexShaderSource, bubbles2dFragmentShaderSource } from '@/utils/shaders'
```

Williangalvani (commented):
I'm not sure how this makes loading faster, but it definitely looks great, nice job!

I just have one request. could we have full names for things?
like "uniform" instead of "u" and "location" instead of "loc" ?
Or comments as you did in the shader itself, that works great for being explicit but also succinct when there is math involved.

I think this is the first shader in this repo, it would be nice to have things be as explicit as possible.

core/frontend/src/components/utils/BrIframe.vue
JoaoMario109: Added both changes, the qualifiers names to complete attribute, location and uniforms and also the import path!

About "how it make faster" right now its not making that much difference, but this will serve as groundwork for future splash screen, and in that scenario it will probably shipped together with index making  the splash screen appears as soon as possible.

patrickelectric (approved):

PR #2535, opened 2024-04-15
install: fix configuration of config.txt

Williangalvani:
this should make blueos_startup_update stop running in the first boot.

I have a following patch that adds tests of blueos_startup_update and bcm_27xx to our pre push hook tests


joaoantoniocardoso:
This improvement seems a bit easy to break: every time we update the startup update script, we'd have actually to check if the boot happens, somehow.

Couldn't we import and use the startup update functions directly?

Williangalvani:
> This improvement seems a bit easy to break: every time we update the startup update script, we'd have actually to check if the reboot happens, somehow.
> 
> Couldn't we import and use the startup update functions directly?

it wont be breakable after my tests get in, but yes, that is an interesting idea, I'll play with it

Williangalvani:
I think we should probably go with this minimal patch for 1.2.x and refactor the whole thing to use blueos_startup_update for master

PR #2537, opened 2024-04-15
Boot fix stage 2

Williangalvani:
This tests our install.sh scripts against blueos_startup_update to ensure the current code state will never require a subsequent reboot after blueos_startup_update

### DIFF
--- a/.hooks/pre-push
+++ b/.hooks/pre-push
@@ -48,6 +48,10 @@ for arg in "$@"; do
   esac
 done
 
+cd install
+TESTING=true pipenv run ./test.sh
+cd -
+
 echo "Running pre push hook!"
 repository_path=$(git rev-parse --show-toplevel)
 tag_name=$(git tag --points-at=HEAD | head -n 1)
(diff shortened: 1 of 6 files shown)


Williangalvani:
@patrickelectric @joaoantoniocardoso this is what I had in mind for making sure our install scripts match our blueos_startup_update. I am open for suggestions and/or better alternatives.

One thing I'm worried about is that our startup.json.default should always match what is actually in blueos. so I don't think it should be manipulated by blueos_startup_update during the first boot.

Williangalvani:
[this](https://github.com/bluerobotics/BlueOS/pull/2537/files#diff-1b6f61dd35ba0a33383f81ee7443065b1d88729fba34996c986f41235dc8c70b) is the actual test part

joaoantoniocardoso:
I think this idea works, as it contemplates the requirements for the config file:
- allow users to customize
- protect our configuration from users' change
- keep up-to-date

I think the next thing we should do is to generate the file in the tree as a git pre-commit client-side hook that runs whenever the script changes.

Williangalvani:
@joaoantoniocardoso  @JoaoMario109 @patrickelectric how do you guys feel about this?
I personally don't like it but I think it is helpful...

patrickelectric:

PR #2311, opened 2024-01-16
Fix assets location

joaoantoniocardoso:
This fixes the following errors in the development env:
```
 cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/SIMPLEROV_3.json, put the file in the src directory, and use /src/assets/vehicles/models/sub/SIMPLEROV_3.json instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/SIMPLEROV_4.json, put the file in the src directory, and use /src/assets/vehicles/models/sub/SIMPLEROV_4.json instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/VECTORED_6DOF.json, put the file in the src directory, and use /src/assets/vehicles/models/sub/VECTORED_6DOF.json instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/VECTORED.json, put the file in the src directory, and use /src/assets/vehicles/models/sub/VECTORED.json instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/boat/UNDEFINED.glb, put the file in the src directory, and use /src/assets/vehicles/models/boat/UNDEFINED.glb instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/rover/unknown.glb, put the file in the src directory, and use /src/assets/vehicles/models/rover/unknown.glb instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/SIMPLEROV_3.glb, put the file in the src directory, and use /src/assets/vehicles/models/sub/SIMPLEROV_3.glb instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/SIMPLEROV_4.glb, put the file in the src directory, and use /src/assets/vehicles/models/sub/SIMPLEROV_4.glb instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/SIMPLEROV_5.glb, put the file in the src directory, and use /src/assets/vehicles/models/sub/SIMPLEROV_5.glb instead.
Assets in public cannot be imported from JavaScript.
Instead of /public/assets/vehicles/models/sub/BLUEROV1.glb, put the file in the src directory, and use /src/assets/vehicles/models/sub/BLUEROV1.glb instead.
Assets in public cannot be imported from JavaScript.
[...]

PR #2539, opened 2024-04-15
Fix missing assets

### LINKED ISSUE #2489: 3D models are not working
It appears that the problem was introduced in https://github.com/bluerobotics/BlueOS/pull/2311

patrickelectric:
Reverts: #2311
Fix #2489


patrickelectric:
@joaoantoniocardoso it would be nice if we can put this in before next beta

PR #2541, opened 2024-04-15
core: tools: mavlink-camera-manager: Update to t3.12.11

joaoantoniocardoso:


rafaellehmkuhl:
Just to point out that I've already tested the image, and it's working perfectly.
All the locking problems I was facing appear to be gone.

rafaellehmkuhl (approved):

PR #2546, opened 2024-04-16
Implements is_valid_version in version chooser to avoid setting a version that does not exists locally

JoaoMario109:


joaoantoniocardoso (approved):

core/services/versionchooser/test_versionchooser.py
```diff
@@ -121,11 +123,18 @@ async def test_set_version(write_mock: AsyncMock) -> None:
 async def test_set_version_invalid_settings(json_mock: mock.MagicMock) -> None:
     client = mock.MagicMock()
     chooser = VersionChooser(client)
+
+    # Image will not exists locally, bt for test lets suppose it exists
```
joaoantoniocardoso, line 127:
```suggestion
    # Image does not exist locally, but for the test let's fake it exists
```

core/services/versionchooser/test_versionchooser.py
```diff
@@ -216,6 +225,13 @@ async def test_set_version_json_exception(json_mock: mock.MagicMock) -> None:
     client = mock.MagicMock()
     json_mock.side_effect = Exception()
     chooser = VersionChooser(client)
+
+    # Image will not exists locally, bt for test lets suppose it exists
```
joaoantoniocardoso, line 229:
```suggestion
    # Image does not exist locally, but for the test let's fake it exists
```

core/services/versionchooser/test_versionchooser.py
JoaoMario109: Added!

core/services/versionchooser/test_versionchooser.py
JoaoMario109: Added!

patrickelectric (approved):

PR #2557, opened 2024-04-23
Fix install from outside manifest

JoaoMario109:
Now only throws arch compatibility error if extensions is found in manifest and its not compatible, if extension is not found or if digest is null tries to proceed with the installation


core/services/kraken/main.py
```diff
@@ -84,7 +84,8 @@ async def install_extension(extension: Extension) -> Any:
             detail="Not enough disk space to install the extension",
         )
     compatible_digest = await kraken.is_compatible_extension(extension.identifier, extension.tag)
-    if not compatible_digest:
+    # Only throw error if a explicit false is given, if none is given, the extension is not found in manifest
```
patrickelectric, line 87: Please add in the commend what is the case and usage of having a `None`

patrickelectric (changes requested):

patrickelectric (approved):

PR #2559, opened 2024-04-23
Hide camera and gripper tabs from vehicle setup

Williangalvani:
they are still empty


patrickelectric (approved):

patrickelectric (commented):
Can't we just check the page component ?

Williangalvani:
> Can't we just check the page component ?

we could, but I don't think if is worth it or changes anything...

patrickelectric:
> > Can't we just check the page component ?
> 
> we could, but I don't think if is worth it or changes anything...

The change is the original code desire to add such configurations on the code. But ok.

PR #2564, opened 2024-04-24
eslint: add no-else-return

Williangalvani:
This catches unnecessary else blocks. sadly doesn't seems to work after `throw`

also, this works with --fix, which is neat


patrickelectric (commented):
Wait.. don't we have any occurrance of that in the code ?

Williangalvani:
> Wait.. don't we have any occurrance of that in the code ?

apparently not. good job. I had to manually do it in order to to test

patrickelectric (approved):

Issue #2567, opened 2024-04-26
bug: vehicle-setup: barometer representation issues


ES-Alexander:
### Bug description

1. Aerial barometers have their type displayed as "Freshwater Pressure", but should be something like "Internal Pressure" / "Aerial Pressure"
    - This seems to only be a Sub issue, although in Rover it comes up as "Unknown Pressure", which isn't great either
3. Bar30 (`MS5837-30BA`) shows up as `MS5611` (I2C)

<img width="484" alt="Screenshot 2024-04-26 at 1 03 04 PM" src="https://github.com/bluerobotics/BlueOS/assets/25898329/bc8c6365-8a99-4316-81ec-c40473c3dddb">

### Steps to reproduce

1. Connect flight controller to BlueOS computer, and view the "Autopilot Sensors" section of the Vehicle Setup page
2. Connect a Bar30 and restart the autopilot

### Primary pain point(s)

1. This is incorrect, and misleadingly may cause people to believe they have one or more connected water pressure sensors that are actually internal / aerial
2. Showing the wrong sensor type is confusing, especially in the Pixhawk which already has an internal MS5611 barometer connected via SPI

### Additional context

[A somewhat relevant forum thread](https://discuss.bluerobotics.com/t/bar30-not-detected-by-pixhawk/12848), from a while ago.

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

ES-Alexander:
1. From some digging after a discussion with @Williangalvani, I think [this](https://github.com/bluerobotics/BlueOS/blob/master/core/frontend/src/components/vehiclesetup/overview/OnboardSensors.vue#L250-L255) should just be an `if-else`, so the "not water sensor" doesn't get overwritten with the water sensor value
2. This is likely an ArduPilot mis-reporting issue, but until that's resolved it may still be worth hedging our bets that an external `MS5611` sensor connected via I<sup>2</sup>C and with a water density is very likely actually a Bar30 (`MS5837-30BA`)

ES-Alexander:
#2560 fixes most of this, and ArduSub-4.5 (currently beta) fixes the sensor name issue (correctly displays MS5837 for the Bar30, and MS5611 for the internal/aerial one).

PR #2570, opened 2024-04-26
core: frontend: onboardSensors: fix internal baro showing as water pressure sensor

Williangalvani:
needs testing (I can't find my spare bar30)


ES-Alexander:
Nice!

Per [my comment](https://github.com/bluerobotics/BlueOS/issues/2567#issuecomment-2079248064) in #2567, this fixes a bug but is only a partial fix for the observed behaviour, because we’re saying MS5611 is a water sensor (to handle the current misrepresentation of the MS5837 on the Bar30), and the Pixhawk has a genuine MS5611 as its onboard barometer (so that one will still show up as water unless we handle it explicitly). Alternatively we could stop treating MS5611 as water except in the special case where we expect it’s actually an MS5837, e.g. when it’s used on a sub vehicle and connected via I2C instead of SPI.

Williangalvani:
> Nice!
> 
> Per [my comment](https://github.com/bluerobotics/BlueOS/issues/2567#issuecomment-2079248064) in #2567, this fixes a bug but is only a partial fix for the observed behaviour, because we’re saying MS5611 is a water sensor (to handle the current misrepresentation of the MS5837 on the Bar30), and the Pixhawk has a genuine MS5611 as its onboard barometer (so that one will still show up as water unless we handle it explicitly). Alternatively we could stop treating MS5611 as water except in the special case where we expect it’s actually an MS5837, e.g. when it’s used on a sub vehicle and connected via I2C instead of SPI.

hmmm makes sense! we can check if it is on i2c and connected to bus BARO_EXT_BUS

patrickelectric (approved):

patrickelectric:
@ES-Alexander what is the current status of this PR ? Should we merge it ?

ES-Alexander:
> what is the current status of this PR ? Should we merge it ?

@patrickelectric it wouldn't _hurt_ to merge it as-is, but I think it makes more sense for the `is_water_baro` function to get updated as part of this PR, with the additional check that's described in the discussion here, since otherwise this PR isn't particularly useful (and I believe it should be a pretty small change).

I don't know how to actually do those checks, but I think an appropriate pseudocode update would be something like:

```js
        if (['MS5837', 'MS5611', 'KELLERLD'].includes(compass.deviceName ?? '--')
        && "baro is on I2C, on BARO_EXT_BUS"
        && autopilot.vehicle_type === 'Submarine') {
          results[compass.param] = true
        }
```

Also @Williangalvani 
1. I'm not sure why that's currently checking compasses instead of baros - is that check actually against all connected sensors (and the variable is just poorly named), or is that another bug?
2. Would it make sense to also check for boats, or do they not yet have support for water sensors?

ES-Alexander (approved):
Confirmed working as expected! :-D

Nice work - this one ended up being more of a slog than expected...

ES-Alexander:
Vehicle setup overview docs may need an updated screenshot.

PR #2579, opened 2024-04-30
Create orientation picker, Add initial frame setup

Williangalvani:
I think this is mostly ready. I'll re-check all the code and make the arrows reorient with the model

[Screencast from 2024-04-30 15-33-41.webm](https://github.com/bluerobotics/BlueOS/assets/4013804/f339e6de-0a3a-4a6d-9654-6123e2b25ecf)

## Summary by Sourcery

Add orientation picker and vehicle body configuration for board setup

New Features:
- Implement interactive 3D orientation picker for board configuration
- Add vehicle body setup page in configuration
- Create model helper utility for vehicle and board models

Enhancements:
- Refactor model loading and vehicle type detection
- Improve vehicle setup configuration interface

Chores:
- Extract model-related logic into a separate helper module
- Simplify component structure for vehicle setup


rafaellehmkuhl:
This is muuuuuuuuuch better!!

patrickelectric (changes requested):
We should do something about the navigator model, 10MB looks insane. 
The advantage is that we can now replace the svgs with hardcoded red squares with component highlights, like we do in the vehicle setup.

Williangalvani:
> We should do something about the navigator model, 10MB looks insane. The advantage is that we can now replace the svgs with hardcoded red squares with component highlights, like we do in the vehicle setup.

huh, I had it down to 1.3mb. I think I forgot to enable compression on the latest export. will check.

Williangalvani:
> We should do something about the navigator model, 10MB looks insane. The advantage is that we can now replace the svgs with hardcoded red squares with component highlights, like we do in the vehicle setup.

updated, good catch

rafaellehmkuhl:
Back from the dead.

ES-Alexander (changes requested):
Niiiice! This will make a huge difference to the people who actually need to change the board orientation! :D

Some requests/ideas:
1. There's a weirdly large amount of padding at the top of each card:
    Screenshot 2025-05-23 at 10 53 46 pm
1. Would it be possible to make the `FRAME_CONFIG` tooltip have bolded parts for the headings? (e.g. "**Description:** Set this parameter ...") - they're a bit hard to differentiate at the moment:
    Screenshot 2025-05-23 at 10 53 24 pm
1. Board Orientation has no parameter name/tooltip for `AHRS_ORIENTATION`
1. It would be nice if board orientation allowed selecting angles visually (instead of just options from a dropdown), including support for the custom orientation options:
    image
    - Fair enough if this is turned into an Issue and left for another PR
1. Frame selection could also be improved with 3D model views and the like, but I know that's already on the way, just not sure whether it'll be in this PR or a later one

Williangalvani:
> 3. Board Orientation has no parameter name/tooltip for `AHRS_ORIENTATION`

will fix!


> 4. It would be nice if board orientation allowed selecting angles visually (instead of just options from a dropdown), 

That is planned for a next iteration with proper gizmos for rotation and moving (to set offsets too)

PR #2590, opened 2024-05-06
Add better parameters description table in Default Parameters Loader in Wizard

### LINKED ISSUE #2588: Improve parameter table on wizard
It should show human friendly names over hardcoded values, it also should show a pretty table.
image

joaomariolago:
Closes #2588


JoaoMario109:
This PR moves the "beautified" print function to the parameters file, though the table structure is still being created in each file. Since the parameters tables across the software right now have different columns but are similar ion styles, maybe in future we should create some kind of more generic parameter table that allow to easily follow the same design across the software.

patrickelectric (approved):

PR #2622, opened 2024-05-23
blueos_startup_update: find config.txt and cmdline.txt dynamically

Williangalvani:
Required for Bookworm support.

Tested on current builds and on pi5

Ideally we'd make it a class instead of using globals, but I'd rather keep the change small.

~This will error if neither of the files exist, how should we handle that?~ I made it so we never halt start-blueos-core due to errors in blueos_startup_update

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update
+++ b/core/tools/blueos_startup_update/blueos_startup_update
@@ -6,7 +6,7 @@ import json
 import re
 import sys
 import time
-from typing import Tuple, List
+from typing import Tuple, List, Optional
 from loguru import logger
 from commonwealth.utils.logs import InterceptHandler, init_logger
 import logging
@@ -65,6 +65,9 @@ DELTA_JSON = {
 # However, it is important to note that conflicting configurations can happen, potentially impacting the kernel's loading process or causing harm to BlueOS.
 CONFIG_USER_PROTECTION_WORD = 'custom'
 
+config_file = None
+cmdline_file = None
+
 import collections
 from commonwealth.utils.commands import run_command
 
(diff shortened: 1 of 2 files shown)


core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -253,7 +270,6 @@ def update_cgroups() -> bool:
     logger.info("Running cgroup update..")
 
     # Retrieve the cmdline file
-    cmdline_file = '/boot/cmdline.txt'
     cmdline_content = load_file(cmdline_file).replace('\n','').split(' ')
```
patrickelectric, line 273: We should skip this commands if cmdline_field is none

patrickelectric (changes requested):

core/tools/blueos_startup_update/blueos_startup_update
```diff
@@ -421,6 +434,12 @@ def main() -> int:
     match = re.match(r'(?P<tag>.*)-(?P<commit_number>\d+)-(?P<commit_hash>[a-z0-9]+)', current_git_version)
     tag, commit_number, commit_hash = match['tag'], match['commit_number'], match['commit_hash']
     logger.info(f"Running BlueOS: {tag=}, {commit_number=}, {commit_hash=}")
+    global config_file
+    global cmdline_file
+    config_file = locate_config_file()
+    logger.info(f"config.txt found at {config_file}")
+    cmdline_file = locate_cmdline_file()
+    logger.info(f"cmdline.txt found at {cmdline_file}")
```
joaoantoniocardoso, line 442: Shouldn't we abort if the config files were not found?

core/tools/blueos_startup_update/blueos_startup_update
Williangalvani: I updated the tests using cmdline/config so they emit a warning and skip

core/tools/blueos_startup_update/blueos_startup_update
Williangalvani: not entirely? we have some other patches that could be useful, I think...

joaoantoniocardoso: yeah, I mean, skip these that depend on the config files

joaoantoniocardoso (approved):

PR #2623, opened 2024-05-23
Cgroups updates

Williangalvani:
Our startup and blueos_startup_update files are inconsistent. also it _should_ require rw access for cgroups to work. it seems to be working regardless...

needs testing


Williangalvani:
tested

patrickelectric (approved):

PR #2625, opened 2024-05-24
Installed Extensions Page: Fix single extension card min width

ArturoManzoli:
When there is a single item on the installed extensions page, the card min-width is too small:
image

Fixed to:
image


patrickelectric (approved):

JoaoMario109 (changes requested):
Maybe we could add in the `.installed-extension-card` some `min-width: 400px;` or some other value just to avoid the flex from not wrapping the elements like this.

image

JoaoMario109 (approved):

PR #2626, opened 2024-05-24
Ardupilot udp serial port fix

### LINKED ISSUE #2627: "udpclient:ip:pport" serial settings do not work in ArdupilotManager
The correct setting is "udp" and not "udpclient"

Williangalvani:
Screenshot 2024-05-24 at 11 36 55

fix #2627

### DIFF
--- a/core/frontend/src/components/autopilot/AutopilotSerialConfiguration.vue
+++ b/core/frontend/src/components/autopilot/AutopilotSerialConfiguration.vue
@@ -200,7 +200,7 @@ export default Vue.extend({
       if (input.startsWith('/dev/')) {
         return true
       }
-      return 'unknown format'
+      return 'unknown format. Supported formats are "tcpclient", "udpclient", "tcpin", "udpin", and system device paths'
     },
   },
 })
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
nothing like a consistent API :)

joaoantoniocardoso:
so, Ardupilot Manager here doesn't need to use the same description names as MAVLink endpoint types (see the class EndpointType above)?

patrickelectric (approved):

patrickelectric (changes requested):
It appears that there is indeed an inconsistent as pointed by @joaoantoniocardoso

Williangalvani:
> so, Ardupilot Manager here doesn't need to use the same description names as MAVLink endpoint types (see the class EndpointType above)?

that is correct, these are not "Mavlink Endpoints". They are often not mavlink at all.

These are the serial port commandline arguments taken by ardupilot. which are [udp, udpin, tcp].
The ardupilot code code be updated to also take `udpout`, but that would only take effect in a future release.

PR #2628, opened 2024-05-24
core: blueos_startup_update: one link is enough for resolv.conf.host

Williangalvani:


Williangalvani:
something else is off. I'll close this for now. we do actually need two links, the first points to the original file, the second to the in-docker file. I'll close this while I try to find a better approach

PR #2637, opened 2024-05-29
versionchooser: differentiate arm64 from arm

Williangalvani:


patrickelectric (approved):

patrickelectric:
@JoaoMario109 can you take a look as well ?

Williangalvani:
I tested it here, btw. my dockers were not showing in aarch64 because we were filtering for armv7, and I was only building for  arm64. This generally is not an issue if the image has both builds, as the "pull" step of docker __tries__ to pull the right one.

JoaoMario109 (approved):

PR #2640, opened 2024-05-30
Kraken: Add docker wrapper module with async context to manage containers and related operations

JoaoMario109:


core/services/kraken/harbor/container.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class ContainerManager:
+    @staticmethod
+    async def get_raw_container_by_name(client: Docker, container_name: str) -> DockerContainer:
+        containers = await client.containers.list(filters={"name": {container_name: True}})  # type: ignore
+        if not containers:
+            raise ContainerNotFound(f"Container {container_name} not found in running containers")
+
+        return cast(DockerContainer, containers[0])
```
joaoantoniocardoso, line 23: Shouldn't this return a List[DockerContainers] instead of the first, as `name` is not a unique identifier, and let the caller decide if it wants the first or all?

patrickelectric (approved):
Nice!

joaoantoniocardoso (approved):
cool!

PR #2641, opened 2024-05-30
Kraken: Add extension Module V2

### LINKED ISSUE #2326: We should delete older extensions once update is done

JoaoMario109:
Depends on #2640

Closes #2326


core/services/kraken/api/v1/routers/extension.py
```diff
@@ -17,48 +18,42 @@
...
+
+    await asyncio.gather(*[ext.uninstall() for ext in extensions])
 
 
 @extension_router_v1.post("/update_to_version", status_code=status.HTTP_201_CREATED)
-async def update_extension(extension_identifier: str, new_version: str) -> Any:
-    return StreamingResponse(streamer(kraken_instance.update_extension_to_version(extension_identifier, new_version)))
+async def update_extension(extension_identifier: str, new_version: str) -> StreamingResponse:
+    extension = cast(Extension, await Extension.from_manifest(extension_identifier, new_version))
+
```
patrickelectric, line 37: just a minor thing, is this spaces necessary ? in all functions here ?

patrickelectric (commented):
Minor things

core/services/kraken/extension/extension.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+class Extension:
+    """
+    Extension class to manage extensions.
+    """
+
+    # If an extension is being installed the key will be the extension identifier if is being removed the key is the
+    # container name.
+    locked_entries: Dict[str, Literal[True]] = {}
```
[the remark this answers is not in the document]
JoaoMario109: Usually I prefer in this kind of lock operation to use dict just to avoid duplicated keys and keep the lock, we could just change to a list with filter also in the unlock, but its just a preference, if you think is better I can change to list.

core/services/kraken/api/v1/routers/extension.py
JoaoMario109: They aren't necessary, already removed!

patrickelectric (approved):

PR #2642, opened 2024-05-30
Commonwealth: create upload_file()

Williangalvani:
the current "echo" method can break files with `"`

Can be tested on my remote williangalvani/blueos-core:upload_file

just take something out of cmdline.txt or config.txt and restart the core container

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/commands.py
+++ b/core/libs/commonwealth/commonwealth/utils/commands.py
@@ -76,3 +76,68 @@ def run_command(command: str, check: bool = True, log_output: bool = True) -> "s
     if ret.stderr:
         logger.error(f"stderr: {ret.stderr}")
     return ret
+
+
+def upload_file_with_password(
+    source: str, destination: str, check: bool = True
+) -> "subprocess.CompletedProcess['str']":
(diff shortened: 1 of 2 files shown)


core/libs/commonwealth/commonwealth/utils/commands.py
```diff
@@ -76,3 +76,65 @@ def run_command(command: str, check: bool = True, log_output: bool = True) -> "s
...
+    logger.debug(f"uploading to {destination}")
+    with open(temp_file_in_container, "w", encoding="utf-8") as f:
+        f.write(file_content)
+    try:
+        ret = upload_file_with_ssh_key(temp_file_in_container, temp_file_in_host, check)
+    except KeyNotFound:
+        logger.warning("SSH key not found, falling back to password authentication")
+        ret = upload_file_with_password(temp_file_in_container, temp_file_in_host, check)
+    logger.debug(ret)
+    run_command(f"sudo mv {temp_file_in_host} {destination}")
```
joaoantoniocardoso, line 139: I think we should only move if the command succeeds

joaoantoniocardoso (approved):
:1st_place_medal: for using scp

PR #2646, opened 2024-05-30
core: blueos_startup_update: load pi5 overlays in runtime

Williangalvani:


core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -438,6 +438,75 @@ def ensure_user_data_structure_is_in_place() -> bool:
...
+        # SPI0: LED
+        "spi0-led",
+    ]
+
+    for dtparam in dtparams_to_load:
+        if dtparam in loaded_dt_params:
+            logger.info(f"dtparam {dtparam} is already loaded")
+        else:
+            logger.info(f"dtparam {dtparam} is not loaded, loading")
+            logger.info(run_command(f"sudo dtparam {dtparam}", False))
```
joaoantoniocardoso, line 483: consider the following for better readability, update the other calls if accepted

```suggestion
            command_result = run_command(f"sudo dtparam {dtparam}", False)
            logger.info(command_result)
```

Williangalvani: hmmm mixed feelings
we are already using the `logger.info(run_command())` structure at least in 4 other insteances in this file.

We could switch all of them to having an intermediate variable, but I wouldn't do it in this pr. I can fix this  up and do a follow up to update the others later, though. thoughts @patrickelectric  @JoaoMario109 ?

Williangalvani: actually, I do think readability wins. I'll update the patch

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -438,6 +438,79 @@ def ensure_user_data_structure_is_in_place() -> bool:
     return False
 
 
+def load_pi5_overlays_in_runtime() -> bool:
+    loaded_overlays = run_command("dtoverlay -l", check=False).stdout.split("\n")
+    loaded_overlays = [overlay.split(": ")[1].strip() for overlay in loaded_overlays if ": " in overlay]
+    loaded_overlay_names = [overlay.split(" ")[0] for overlay in loaded_overlays if "dtparam" not in overlay]
+    loaded_dt_params = [overlay.split(" ")[-1] for overlay in loaded_overlays if "dtparam" in overlay]
```
patrickelectric, line 445: I would recommend using a simple regex over doing this logic, it'll be simpler and more readable.

`^(?<index>\d):\s+(?<name>\S+)($|\s+(?<args>.*))`

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -438,6 +438,79 @@ def ensure_user_data_structure_is_in_place() -> bool:
...
+
+    for dtparam in dtparams_to_load:
+        if dtparam in loaded_dt_params:
+            logger.info(f"dtparam {dtparam} is already loaded")
+        else:
+            logger.info(f"dtparam {dtparam} is not loaded, loading")
+            command_result = run_command(f"sudo dtparam {dtparam}", False)
+            logger.info(command_result)
+
+    command_result = run_command("sudo modprobe i2c-dev", False)
```
patrickelectric, line 486: we are missing `lsmod` validation

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -438,6 +438,79 @@ def ensure_user_data_structure_is_in_place() -> bool:
...
+        elif overlay_name in loaded_overlay_names:
+            logger.info(f"Overlay {overlay} is already loaded with different parameters")
+            command_result = run_command(f"sudo dtoverlay -r {overlay_name} && sudo dtoverlay {overlay}", False)
+            logger.info(command_result)
+        else:
+            logger.info(f"Overlay not present, loading overlay {overlay}")
+            command_result = run_command(f"sudo dtoverlay {overlay}", False)
+            logger.info(command_result)
+        src = "/dev/i2c-3"
+        dst = "/dev/i2c-4"
```
patrickelectric, line 502: Isn't there a way for us to avoid that via ardupilot ?

patrickelectric (changes requested):

core/tools/blueos_startup_update/blueos_startup_update.py
Williangalvani: only if we define a new board type, which we may have to in order to build 64bit binaries...

core/tools/blueos_startup_update/blueos_startup_update.py
joaoantoniocardoso: I agree with Patrick here, it's easier to follow the parsing from the regex line.

core/tools/blueos_startup_update/blueos_startup_update.py
Williangalvani: I removed it meanwhile

Williangalvani:
I re-wrote most of it to do it right. it is a lot more code, but it should also be a lot more robust, and easier to follow.

core/libs/commonwealth/commonwealth/utils/boot/pi5overlays.py
```diff
@@ -0,0 +1,168 @@
...
+            loaded_overlays.append(Overlay(overlay_name, overlay_args.split(" ") if overlay_args else []))
+
+    logger.info("Loaded dt params: ")
+    for dt_param in loaded_dt_params:
+        logger.info(dt_param)
+
+    for dtparam in dtparams_to_load:
+        if dtparam in loaded_dt_params:
+            logger.info(f"dtparam {dtparam} is already loaded")
+            print(dtparam)
```
patrickelectric, line 145: lost print

patrickelectric (changes requested):

Williangalvani:
> this will move to ardupilot manager right ?

yes, that is the current plan

core/tools/blueos_startup_update/boot/pi5overlays.py
```diff
@@ -0,0 +1,167 @@
...
+        command_result = run_command(f"sudo dtoverlay -r {self.name} && sudo dtoverlay {self}", False)
+        logger.info(command_result)
+
+    def load(self) -> None:
+        command_result = run_command(f"sudo dtoverlay {self}", False)
+        logger.info(command_result)
+
+    @staticmethod
+    def from_string(overlay: str) -> "Overlay":
+        match = re.match(r"^((?P<index>\d+):\s+)?(?P<name>(\S)+)(?:$|\s+(?P<args>.*))", overlay)
```
joaoantoniocardoso, line 40: Can we extract this line into a class method so we can call it on line 127 instead of having the regex duplicated?

Williangalvani:
closing this as we should only do these things if we detect a navigator (which only requires i2c overlays)

see #2670 and #2678

PR #2647, opened 2024-05-30
blueos_startup_update: build spi0-led overlay if it doesn't exist

Williangalvani:
What should we do with the overlay source file? we can't get stuff from parent directories in the dockerfile...
ideas?

requires #2642


Williangalvani:
should we just include the built overlay instead?

core/Dockerfile
```diff
@@ -64,6 +64,9 @@ ENV TARGETVARIANT=${TARGETVARIANT}
 COPY tools /home/pi/tools
 RUN /home/pi/tools/install-system-tools.sh
 
+# Copy overlay source in case we need to build it in runtime
+COPY install/ /install
```
patrickelectric, line 68: Since etc/blueos is already being binded to the host, I believe that best place would be /usr/share/blueos/install 
@joaoantoniocardoso

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -438,6 +438,20 @@ def ensure_user_data_structure_is_in_place() -> bool:
...
 
 
+def build_led_overlay():
+    overlay_exists = locate_file(["/boot/overlays/spi0-led.dtbo", "/boot/firmware/overlays/spi0-led.dtbo"])
+    if overlay_exists:
+        logger.info(f"spi0-led overlay found at {overlay_exists}")
+        return False
+    with open("/install/spi0-led.dts", "r", encoding="utf-8") as f:
+        dts = f.read()
+        save_file("/tmp/spi0-led.dts", dts, "")
```
patrickelectric, line 448: We should have a variable to hold /tmp/spi0-led.dts path

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -438,6 +438,20 @@ def ensure_user_data_structure_is_in_place() -> bool:
...
+    command = "sudo dtc -@ -Hepapr -I dts -O dtb -o /boot/overlays/spi0-led.dtbo /tmp/spi0-led.dts && sudo cp /boot/overlays/spi0-led.dtbo /boot/firmware/overlays/spi0-led.dtbo"
```
patrickelectric, line 449: Just to be sure, both paths exist in both versions ? (bullseye and bookworm)
Shouldn't this be done based in the existence of the path ?

patrickelectric (changes requested):

core/Dockerfile
joaoantoniocardoso: Since `install` is a script, I think it's more likely for it to find a comfy home as `/opt/blueos/install`

Williangalvani: > Since `install` is a script, I think it's more likely for it to find a comfy home as `/opt/blueos/install`. `/opt/blueos` is a general good home for all sorts of little creatures :)

in this case it is as folder with the .dts file in it...

plus my largerst concern is where it should live in the git repo, as this (/core/install/spi0-led.dts) is a redundant copy of the one at the /install/overlays/spi0-led.dts

core/tools/blueos_startup_update/blueos_startup_update.py
Williangalvani: good catch. it does't exist on bullseye. I added a check

Williangalvani:
closing it as it does not belong in blueos_startup_update. this is a navigator-specific file

PR #2649, opened 2024-05-30
blueos_startup_update: add 'remove_console' step

Williangalvani:
this install step was not in blueos_startup_update
https://github.com/bluerobotics/BlueOS/blob/master/install/boards/bcm_27xx.sh#L83-L84

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -446,6 +446,24 @@ def run_command_is_working():
     return True
 
 
+def remove_console():
+    """
+    removes "console=serial0,115200" from cmdline.txt
+    """
+    if not cmdline_file:
+        logger.error("cmdline unknown, aborting console patch")
+        return False
+    cmdline_content = load_file(cmdline_file).strip().split(" ")
+    unpatched_cmdline_content = cmdline_content.copy()
+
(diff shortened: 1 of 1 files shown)


core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -446,6 +446,24 @@ def run_command_is_working():
...
 
 
+def remove_console():
+    """
+    removes "console=serial0,115200" from cmdline.txt
+    """
+    if not cmdline_file:
+        logger.error("cmdline unknown, aborting console patch")
+        return False
+    cmdline_content = load_file(cmdline_file).strip().split(" ")
```
joaoantoniocardoso, line 456: why are we parsing the file differently here? Can we extract the correct way to parse  this file into a function?

Williangalvani: yes, I intend to do that in another pr, where I also intend to stop opening it multiple times.
plus this is virtually the same structure in update_cgroups()

joaoantoniocardoso: okay

joaoantoniocardoso (approved):

patrickelectric:
Just saying that this step makes not possible to debug kernel problems via serial

PR #2651, opened 2024-05-31
core: add boot-loop detection

### LINKED ISSUE #2639: blueos_startup_update should have a mechanism to prevent boot loops

Williangalvani:
simple fix for #2639

It can be tested with williangalvani/blueos-core:fox_boot_loop_test

which includes this patch and the broken commit from #2649

fix #2639

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -18,6 +18,8 @@ SERVICE_NAME = "blueos_startup_update"
 logging.basicConfig(handlers=[InterceptHandler()], level=0)
 init_logger(SERVICE_NAME)
 
+BOOT_LOOP_DETECTOR = "/root/.config/.boot_loop_detector"
+
 # Any change made in this DELTA_JSON dict should be also made
 # into /bootstrap/startup.json.default too!
 DELTA_JSON = {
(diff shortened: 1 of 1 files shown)


core/start-blueos-core
```diff
@@ -41,6 +43,8 @@ echo "Total RAM size: ${TOTAL_RAM_MB} MB"
 if [ -z "${BLUEOS_DISABLE_STARTUP_UPDATE}" ]; then
     blueos_startup_update.py || true  # We NEVER want to prevent booting
 fi
+# if we made it here, it means the patcher didn't require a reboot
+rm $BOOT_LOOP_DETECTOR
```
Williangalvani, line 47: this actually could be inside blueos_startup_update.py

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -448,6 +450,10 @@ def run_command_is_working():
 
 def main() -> int:
     start = time.time()
+    # check if boot_loop_detector exists
+    if os.path.isfile(BOOT_LOOP_DETECTOR):
+        logger.warning("It seems we are in a boot loop, skipping patches.")
```
patrickelectric, line 455: "It seems that we are starting from a previous startup configuration, skipping patches."

Williangalvani: that sounds... vague
what about "It seems we the startup patches were just applied on the last boot. skipping patches..." ?

patrickelectric: Yep

patrickelectric (approved):

PR #2656, opened 2024-06-03
Kraken: Add default extensions

JoaoMario109:


install/kraken/set_default_extensions.sh
```diff
@@ -0,0 +1,85 @@
...
+mkdir -p "${SETTINGS_BASE_DIR}"
+
+# Check if the directory creation was successful
+if [[ ! -d "${SETTINGS_BASE_DIR}" ]]; then
+  echo "Failed to create directory ${SETTINGS_BASE_DIR}"
+  exit 1
+fi
+
+# Creates kraken V1 settings
+cat > "${SETTINGS_BASE_DIR}/settings-1.json" <<EOF
```
patrickelectric, line 57: What is the reason for us to have v1 here ?

JoaoMario109: Actually is not needed, already removed

patrickelectric (approved):

PR #2675, opened 2024-06-10
Add kraken core/UI improvements

JoaoMario109:
* Add check for required image size when installing from manifest sources


JoaoMario109:
Implemented by:
#2735 
#2777 
#2798

PR #2677, opened 2024-06-11
fix pre-push hooks, and move to Poetry

joaoantoniocardoso:
This migrates the hooks/pre-push to use Poetry directly, and makes all the files contained within the project's virtual environment, instead of loading files on the users' folder, making the development environment more replicable.

To use the pre-push, the user must have a Python 3.11 available. Before, it would use whatever the system's Python was, failing to run when the user had 3.12 or 3.10 for example.

To test:
- checkout to this PR's branch
- clean the project: `\rm -rf .venv poetry.lock`
- run the pre-push hook: `.hooks/pre-push`

To check whether it had any impact on the normal installation:
- Try out the `joaoantoniocardoso/blueos-core`:`fix-pre-push` on BlueOS

I have tested it myself, but more samples are welcome.

After this patch is merged, we can all clean our system:

* `ls "$(python -m site --user-site)"`
* `ls "$(python -m site --user-base)/bin"`

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -29,20 +29,9 @@ jobs:
 
     - name: Install dependencies
       run: |
-        sudo apt install shellcheck
+        sudo apt-get install --no-install-recommends --assume-yes shellcheck parallel
         python -m pip install --upgrade pip
-        pip install pipenv
-        pipenv install poetry && pipenv run poetry install
-        echo "installing mavlink-router"
-        pipenv run ./core/tools/mavlink_router/bootstrap.sh
-        echo "installing mavp2p"
-        pipenv run ./core/tools/mavp2p/bootstrap.sh
-        echo "installing ardupilot_tools"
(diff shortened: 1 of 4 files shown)


Williangalvani:
works great once it works! but I had to manually configure a local python3.11 with pyenv. can we add a check for python version and add hints on how to set it up?

 - install pyenv
 - run `pyenv install 3.11 && pyenv local 3.11`

joaoantoniocardoso:
@Williangalvani is the command `python3.11` available when using pyenv, or should I use some pyenv command to figure it out?

joaoantoniocardoso:
@Williangalvani 

Okay, I rewrote the Python check code to automatically use pyenv if it's available

Williangalvani:
working well locally here

Williangalvani (approved):
lgtm! this is so much better!

joaoantoniocardoso:
@Williangalvani I just rebased, some of the commits were already merged.

PR #2678, opened 2024-06-11
Linux refactor: Episode I

Williangalvani:
splitting it up to make the review process smoother

tested on pi4/bullseye
test on pi5/bookworm
~port mapping on pi4/bullseye has changed =[~
updated  serial mapping for pi4 on bullseye

### DIFF
--- a/core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
+++ b/core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
@@ -1,0 +1,41 @@
+from typing import List, Type
+
+from smbus2 import SMBus
+
+from typedefs import FlightController, PlatformType, Serial
+
+
+class LinuxFlightController(FlightController):
+    """Linux-based Flight-controller board."""
+
+    @property
+    def type(self) -> PlatformType:
+        return PlatformType.Linux
+
+    def detect(self) -> bool:
+        raise NotImplementedError
+
+    def get_serial_cmdlines(self) -> str:
+        cmdlines = [f"-{entry.port} {entry.endpoint}" for entry in self.get_serials()]
+        return " ".join(cmdlines)
+
+    def get_serials(self) -> List[Serial]:
+        raise NotImplementedError
+
+    def check_for_i2c_device(self, bus_number: int, address: int) -> bool:
+        try:
+            bus = SMBus(bus_number)
+            bus.read_byte_data(address, 0)
+            return True
+        except OSError:
+            return False
+
+    @classmethod
+    def get_all_boards(cls) -> List[Type["LinuxFlightController"]]:
+        all_subclasses = []
+
+        for subclass in cls.__subclasses__():
+            all_subclasses.append(subclass)
+            all_subclasses.extend(subclass.get_all_boards())
+
+        return all_subclasses
--- a/core/services/ardupilot_manager/flight_controller_detector/linux/navigator.py
+++ b/core/services/ardupilot_manager/flight_controller_detector/linux/navigator.py
@@ -1,0 +1,81 @@
+from typing import List
+
+from commonwealth.utils.commands import load_file
+
+from flight_controller_detector.linux.linux_boards import LinuxFlightController
+from typedefs import Platform, Serial
+
+
+class Navigator(LinuxFlightController):
+    name = "Navigator"
+    manufacturer = "Blue Robotics"
+    platform = Platform.Navigator
+
+    def is_pi5(self) -> bool:
+        with open("/proc/cpuinfo", "r", encoding="utf-8") as f:
+            return "Raspberry Pi 5" in f.read()
+
+    def detect(self) -> bool:
+        return False
+
+    def get_serials(self) -> List[Serial]:
+        raise NotImplementedError
+
+
+class NavigatorPi5(Navigator):
+    devices = {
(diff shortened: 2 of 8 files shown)


core/services/ardupilot_manager/flight_controller_detector/linux/navigator.py
```diff
@@ -0,0 +1,84 @@
...
+        "AK09915": (0x0C, 1),
+        "BME280": (0x76, 1),
+        "PCA9685": (0x40, 4),
+    }
+
+    def get_serials(self) -> List[Serial]:
+        release = "Bullseye"
+        os_release = load_file("/etc/os-release")
+        if "bookworm" in os_release:
+            release = "Bookworm"
```
patrickelectric, line 62: Can we use an enum here ?

core/services/ardupilot_manager/flight_controller_detector/linux/navigator.py
```diff
@@ -0,0 +1,84 @@
...
+    name = "Navigator"
+    manufacturer = "Blue Robotics"
+    platform = Platform.Navigator
+
+    def is_pi5(self) -> bool:
+        is_pi5 = False
+        with open("/proc/cpuinfo", "r", encoding="utf-8") as f:
+            if "Raspberry Pi 5" in f.read():
+                is_pi5 = True
+        return is_pi5
```
patrickelectric, line 19:
```suggestion
        with open("/proc/cpuinfo", "r", encoding="utf-8") as f:
            return "Raspberry Pi 5" in f.read()
        return False
```

core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
```diff
@@ -0,0 +1,78 @@
...
+            and not filename.endswith("linux_boards.py")
+        ]
+
+        for module in module_files:
+            logger.info(f"Loading module: {module}")
+            __import__(f"flight_controller_detector.linux.{module}")
+
+    @classmethod
+    def detect_boards(cls, ignore_cache: bool = False) -> Optional["LinuxFlightController"]:
+        if not hasattr(cls, "previously_detected"):
```
patrickelectric, line 67: How is this possible ? Since is a derivation of a static property it should exist. 
Why not just calling the static property directly ?

core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
```diff
@@ -0,0 +1,78 @@
...
+            os.path.basename(filename)[:-3]  # Remove the '.py' extension
+            for filename in os.listdir(current_directory)
+            if filename.endswith(".py")
+            and not filename.endswith("__init__.py")
+            and not filename.endswith("linux_boards.py")
+        ]
+
+        for module in module_files:
+            logger.info(f"Loading module: {module}")
+            __import__(f"flight_controller_detector.linux.{module}")
```
patrickelectric, line 63: You should be able to load by just importing the files, check AbstractRouter from ardupilot manager: https://github.com/bluerobotics/BlueOS/blob/bd8069d45e6d816ab83ee5d928dabfb4236e9cef/core/services/ardupilot_manager/mavlink_proxy/Manager.py#L8

patrickelectric (changes requested):
Nice approach, just some minors

joaoantoniocardoso:
Nice one, eager for the next episodes!

core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
Williangalvani: true, this was probably due a mistake on my part during early development

core/services/ardupilot_manager/flight_controller_detector/linux/navigator.py
Williangalvani: can this wait?
I plan to move the checks for os_release and pi version to commonwealth in the next PR

core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
Williangalvani: the idea was that all files dropped here would be loade automagically, but it is a lot of code.
One issue is that importing them here causes a cyclic redundancy, so they (or at least argonot) would need to be imported within the function here

patrickelectric: Can't be imported as the others and keep the subclass for subclass logic ?

Williangalvani: actually I think I have a solution for that

patrickelectric (commented):
Should we merge this before 1.3 ? I'm more inclined to merge it in 1.4, but I'm open to discuss.

joaoantoniocardoso:
> Should we merge this before 1.3 ? I'm more inclined to merge it in 1.4, but I'm open to discuss.

Yeah, might be a smarted move if this is not required for 1.3

Williangalvani:
> Should we merge this before 1.3 ? I'm more inclined to merge it in 1.4, but I'm open to discuss.

Yes, it is probably wiser to keep it for the next sable

patrickelectric (approved):

PR #2690, opened 2024-06-15
Free some space before building docker

patrickelectric:
This removes 20GB inside the docker action


joaoantoniocardoso (approved):
Awesome! And it takes only 15s!

PR #2707, opened 2024-06-19
Disable delete button if two images or less are available

### LINKED ISSUE #2706: Delete button should be deactivated when the version/image is the only one available
image

patrickelectric:
Fix #2706

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionCard.vue
+++ b/core/frontend/src/components/version-chooser/VersionCard.vue
@@ -221,6 +221,10 @@ export default Vue.extend({
       type: Boolean,
       default: false,
     },
+    enableDelete: {
+      type: Boolean,
+      default: true,
+    },
     deleting: {
       type: Boolean,
       default: false,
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso (approved):
I agree w/ the solution, assuming there will always be a 'factory' image

core/frontend/src/components/version-chooser/VersionCard.vue
```diff
@@ -262,7 +266,7 @@ export default Vue.extend({
       return value.replace('sha256:', '').substring(0, 8)
     },
     imageCanBeDeleted() {
-      return (this.image.tag !== 'factory' || this.image.repository !== DEFAULT_REMOTE_IMAGE) && !this.deleting
+      return (this.image.tag !== 'factory' || this.image.repository !== DEFAULT_REMOTE_IMAGE) && !this.deleting && this.enableDelete
```
joaoantoniocardoso, line 269: bun needs this line to be breaked

patrickelectric: Done

joaoantoniocardoso (approved):

PR #2711, opened 2024-06-19
core: frontend: components: version-chooser: Make bootstrap update less dangerous

patrickelectric:
image

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionCard.vue
+++ b/core/frontend/src/components/version-chooser/VersionCard.vue
@@ -96,13 +96,13 @@
 
         <v-card>
           <v-card-title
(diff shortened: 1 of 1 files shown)


JoaoMario109 (approved):

core/frontend/src/components/version-chooser/VersionCard.vue
```diff
@@ -116,10 +116,10 @@
             </v-btn>
             <v-spacer />
             <v-btn
-              color="warning"
+              color="success"
```
joaoantoniocardoso, line 119: I don't agree with success color for this button

patrickelectric: @joaoantoniocardoso check now

PR #2713, opened 2024-06-20
core: components: GpsTrayMenu: First version

patrickelectric:
Fix https://github.com/bluerobotics/BlueOS/issues/1749

image

image


joaoantoniocardoso (commented):
I don't have a GPS to test, but it looks all right

PR #2726, opened 2024-06-22
core: frontend: health: GpsTrayMenu: Update description to contain number of satellites

patrickelectric:


core/frontend/src/components/health/GpsTrayMenu.vue
```diff
@@ -104,6 +94,7 @@ export default Vue.extend({
 
       this.last_message_date = new Date()
       this.gps_raw_int = message?.message as GpsRawInt
+      this.gps_connected = this.gps_raw_int?.fix_type?.type !== GpsFixType.GPS_FIX_TYPE_NO_GPS
```
patrickelectric, line 97: We should still add a time validation for the case where mavlink data is not received

Williangalvani:
I think we should check for the MAV_SYS_STATUS_SENSOR_GPS bit in [MAV_SYS_STATUS_SENSOR_GPS](https://mavlink.io/en/messages/common.html#SYS_STATUS)'s "onboard_control_sensors_present" field

patrickelectric:
> I think we should check for the MAV_SYS_STATUS_SENSOR_GPS bit in [MAV_SYS_STATUS_SENSOR_GPS](https://mavlink.io/en/messages/common.html#SYS_STATUS)'s "onboard_control_sensors_present" field

What should we do with it ? This is just an information that a GPS is connected, but no further information. No status, number of satellites and etc.

core/frontend/src/components/health/GpsTrayMenu.vue
```diff
@@ -1,18 +1,17 @@
...
+    v-if="gps_connected"
+    :content="satellites_number"
+    :value="satellites_number"
+    :title="`${connection_description} (Number of satellites: ${satellites_number})`"
+    :color="number_color"
+    class="mr-2"
+    overlap
+  >
+    <v-icon color="white">
+      {{ satellites_number ? "mdi-crosshairs-gps" : "mdi-crosshairs-question" }}
```
ES-Alexander, line 12: This is still very unintuitive to me, I'm afraid.

It's not the icon itself that's the issue (I can see why crosshairs would indicate position, even if it's not super familiar to me), my concern is with the combination of crosshairs + number of satellites, because there aren't multiple locations being stored, or multiple GPS sensors being averaged between to estimate the location, which is how my brain wants to interpret that display.

patrickelectric: There is a tooltip saying that is the number of satellites, but I'm open to suggestions.

Williangalvani:
> > I think we should check for the MAV_SYS_STATUS_SENSOR_GPS bit in [MAV_SYS_STATUS_SENSOR_GPS](https://mavlink.io/en/messages/common.html#SYS_STATUS)'s "onboard_control_sensors_present" field
> 
> What should we do with it ? This is just an information that a GPS is connected, but no further information. No status, number of satellites and etc. Is that better than GPS_RAW_INT/GLOBAL_POSITION_INT ?

hide the icon entirely if there is no gps connected, such as in rovs

patrickelectric:
> > > I think we should check for the MAV_SYS_STATUS_SENSOR_GPS bit in [MAV_SYS_STATUS_SENSOR_GPS](https://mavlink.io/en/messages/common.html#SYS_STATUS)'s "onboard_control_sensors_present" field
> > 
> > 
> > What should we do with it ? This is just an information that a GPS is connected, but no further information. No status, number of satellites and etc. Is that better than GPS_RAW_INT/GLOBAL_POSITION_INT ?
> 
> hide the icon entirely if there is no gps connected, such as in rovs

But we are doing that already here.

core/frontend/src/components/health/GpsTrayMenu.vue
Williangalvani: maybe the type of fix? like "rtk", "dgps", "3d" and friends?

patrickelectric: This is already in the description.

ES-Alexander: > There is a tooltip saying that is the number of satellites ...

That helps, but feels like the reasoning is backwards - ideally the reaction to an interface should be "it makes sense, and has extra detail if you're interested", rather than "it's confusing so we had to explain it". 

> ... I'm taking suggestions

1. I'd suggest using a satellite icon when there is a valid satellite lock
2. If that's not possible, I think just the word `GPS` would be easier to understand than the target icon (when there's a satellite number connected to it)
3. If the target icon is critical then I think the satellite number should _only_ be included in the tooltip, to avoid confusion, although that does then provide less information "at a glance" to people who know what they're looking at

patrickelectric: 1. A satellite icon implies on satellite communication. (Latest iphone as example)
2. "GPS" does not follow our current UI where we use icons
3. I'm ok to remove the icon at all, but in general that'll result Ina worst UI in my opinion. Based on our current userbase I believe that the current UI is fine.

I also against removing the badge because we have different colors for different stats.

About the icon choice, for Android and google is pretty common.
Screenshot_20240624-233340
Screenshot_20240624-233321

patrickelectric:
Btw, this PR fix a corrente problem where the icon show in a vehicle that does not have a GPS, UI discussions can happen in an issue.

core/frontend/src/components/health/GpsTrayMenu.vue
Williangalvani: I think the issue with the icon is that this is icon is commonsly associated with an action "center on my current position", and is not usually representing a status. but yes, we can follow the discussion on an issue

Williangalvani (approved):
This does solve the main issue at hand, which is displaying the icon on ROVs with no GPS connected, right?
in that case I think we can get it in as is in order to release a new beta

PR #2736, opened 2024-06-24
kraken: Add treatment for ext start when offline

### LINKED ISSUE #2727: Extensions are failing to start in beta.5 with no internet
This is causing issues on remote deployments with no internet.
we got two complaints over the weekend

image

JoaoMario109:
Fix #2727

* Add ManifestBackendOffline error in manifest fetch
* Add treatment to try to start anyway if fails to fetch manifest


Williangalvani (approved):
lgtm!
tested with no internet and working as expected

PR #2741, opened 2024-06-25
frontend: VideStreamCreation: Add more RTSPs variants (UDP, TCP, HTTP) support

### LINKED ISSUE #2737: Add better support to blueos RTSP options (TCP/UDP) and etc

JoaoMario109:
Closes #2737 

When in dev mode:
image

If set some pirate mode variant in dev mode but switching back to normal mode only that one will be added, and if change it will not more be available:
image
image

### DIFF
--- a/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
+++ b/core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
@@ -49,12 +49,32 @@
             class="d-flex justify-space-between align-center"
           >
             <v-select
-              :items="[StreamType.UDP, StreamType.RTSP]"
-              :value="endpoint.startsWith('rtsp://') ? StreamType.RTSP : StreamType.UDP"
+              :items="availableStreamTypes(endpoint)"
+              :value="selectedStreamType(endpoint)"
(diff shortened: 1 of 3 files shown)


core/frontend/src/components/video-manager/VideoStreamCreationDialog.vue
```diff
@@ -356,6 +384,40 @@ export default Vue.extend({
     showDialog(state: boolean) {
       this.$emit('visibilityChange', state)
     },
+    selectedStreamType(endpoint: string): StreamType {
+      if (endpoint.startsWith('rtspu://')) return StreamType.RTSPU
+      if (endpoint.startsWith('rtsp://')) return StreamType.RTSP
+      if (endpoint.startsWith('rtspt://')) return StreamType.RTSPT
+      if (endpoint.startsWith('rtsph://')) return StreamType.RTSPH
```
patrickelectric, line 391: As a future note, we can use an Object.keys(StreamType).find((this_protocol) => new Url(endpoint).protocol.toUpper()) or something like that

patrickelectric (approved):

PR #2748, opened 2024-06-26
frontend: add failsafe configurations

### LINKED ISSUE #2723: core: frontend: vehicle-setup: allow configuring failsafes
- [Supported in QGC](https://docs.qgroundcontrol.com/master/en/qgc-user-guide/setup_view/safety_ardupilot.html#sub)

- Maybe include an editable checklist, with default values of common pre-dive/launch checks
    - Charged batteries
    - Ensure enclosures and vent are sealed
    - Equipment packed
    - etc

---

# Failsafe Parameters
- Control Station Heartbeat Loss
    - [x] `FS_GCS_ENABLE` (optional)
    - [ ] `FS_GCS_ENABL` (Plane -_-)
    - [x] `FS_GCS_TIMEOUT` (optional)
- Pilot Input Loss (Sub)
    - [x] `FS_PILOT_INPUT`
    - [x] `FS_PILOT_TIMEOUT`
- Radio Control Loss
    - [ ] `RC_OVERRIDE_TIME`
    - [ ] `RC_FS_TIMEOUT`
    - [ ] `SERVO_RC_FS_MSK`
- Leak Detection (Sub)
    - [x] `FS_LEAK_ENABLE`
    - [ ]  Configuration details should (in future) be handled by a separate Leak Sensor page
    - [x] `LEAK1_PIN`
    - [x] `LEAK1_LOGIC`
    - [x] `LEAK1_TYPE` (optional)
    - [ ] `LEAK2_PIN`
    - [ ] `LEAK2_LOGIC`
    - [ ] `LEAK2_TYPE` (optional)
    - [ ] `LEAK3_PIN`
    - [ ] `LEAK3_LOGIC`
    - [ ] `LEAK3_TYPE` (optional)
- Excess Internal Pressure (Sub)
    - [x] `FS_PRESS_ENABLE`
    - [x] `FS_PRESS_MAX`
- Excess Internal Temperature (Sub)
    - [x] `FS_TEMP_ENABLE`
    - [x] `FS_TEMP_MAX`
- Terrain Loss During AUTO (Sub)
    - [ ] `FS_TERRAIN_ENAB`
- Low Battery
    - [ ] `BATTn_FS_VOLTSRC`
    - [ ] `BATTn_LOW_VOLT`
    - [ ] `BATTn_LOW_TIMER`
    - [ ] `BATTn_LOW_MAH`
    - [ ] `BATTn_FS_LOW_ACT`
    - [ ] `BATTn_CRT_VOLT`
    - [ ] `BATTn_CRT_MAH`
    - [ ] `BATTn_FS_CRT_ACT`
- Excess Power Draw
    - [ ] `MOT_BAT_VOLT_MAX`
    - [ ] `MOT_BAT_VOLT_MIN`
    - [ ] `MOT_BAT_CURR_MAX`
    - [ ] `MOT_BAT_CURR_TC`
    - [ ] `MOT_BAT_IDX`
- Disarm
    - [ ] `MOT_SAFE_DISARM`
    - [ ] `MOT_SAFE_TIME`
    - [ ] `LOG_DISARMED`
    - [ ] `LOG_REPLAY`
    - [ ] `LOG_FILE_DSRMROT`
    - [ ] `VTX_OPTIONS`
- Plane Disarm (Plane)
    - [ ] `LAND_DISARM_DELAY`
    - [ ] `LAND_THEN_NEUTRL`
    - [ ] `Q_OPTIONS`
    - [ ] `Q_M_SAFE_DISARM`
    - [ ] `Q_M_SAFE_TIME`
- Sensor Fusion Uncertainty
    - [x] `FS_EKF_ACTION` (optional)
    - [x] `FS_EKF_THRESH`
    - [ ] `FS_EKF_FILT` (optional, Copter)
- Crash Detection
    - [x] `FS_CRASH_CHECK`
- X Crash Detection (Plane)
    - [ ] `CRASH_ACC_THRESH`
    - [ ] `CRASH_DETECT`
- Low Throttle
    - [ ] `FS_THR_ENABLE`
    - [ ] `FS_THR_VALUE`
- Excess Vibrations
    - [ ] `FS_VIBE_ENABLE`
- Dead Reckoning Recovery
    - [ ] `FS_DR_ENABLE`
    - [ ] `FS_DR_TIMEOUT`
- Fence Breach
    - [ ] `FENCE_ENABLE`
    - [ ] `FENCE_ACTION`
    - [ ] `FENCE_OPTIONS` (optional)
    - [ ] `FENCE_ALT_MAX` (optional)
    - [ ] `FENCE_MARGIN`
    - [ ] `FENCE_ALT_MIN`
    - [ ] `FENCE_AUTOENABLE` (optional)
    - [ ] `FENCE_RET_RALLY`
    - [ ] `FENCE_TYPE`
- Parachute Required
    - [ ] `CHUTE_ENABLED`
    - [ ] `CHUTE_ALT_MIN`
    - [ ] `CHUTE_DELAY_MS`
    - [ ] `CHUTE_CRT_SINK`
    - [ ] Configuration details should be handled by a separate Parachute page
    - [ ] `CHUTE_TYPE`
    - [ ] `CHUTE_SERVO_ON`
    - [ ] `CHUTE_SERVO_OFF`
- Miscellaneous
    - [ ] `FS_OPTIONS`
    - [ ] `OSD_FS_SCR`
    - [ ] `FS_SHORT_ACTN` (Plane)
    - [ ] `FS_SHORT_TIMEOUT` (Plane)
    - [ ] `FS_LONG_ACTN` (Plane)
    - [ ] `FS_LONG_TIMEOUT` (Plane)
    - [ ] `FS_ACTION` (Rover)
    - [ ] `FS_TIMEOUT` (Rover)
- Advanced Failsafe
    - [ ] `AFS_ENABLE`
    - [ ] `AFS_MAN_PIN`
    - [ ] `AFS_HB_PIN`
    - [ ] `AFS_TERMINATE`
    - [ ] `AFS_TERM_ACTION`
    - [ ] `AFS_TERM_PIN`
    - [ ] `AFS_WP_COMMS`
    - [ ] `AFS_WP_GPS_LOSS`
    - [ ] `AFS_MAX_GPS_LOSS`
    - [ ] `AFS_MAX_COM_LOSS`
    - [ ] `AFS_AMSL_LIMIT`
    - [ ] `AFS_AMSL_ERR_GPS`
    - [ ] `AFS_DUAL_LOSS`
    - [ ] `AFS_GCS_TIMEOUT`
    - [ ] `AFS_QNH_PRESSURE`
    - [ ] `AFS_GEOFENCE`
    - [ ] `AFS_RC`
    - [ ] `AFS_RC_MAN_ONLY`
    - [ ] `AFS_RC_FAIL_TIME`
    - [ ] `AFS_MAX_RANGE`
    - [ ] `AFS_OPTIONS`

# Pre-Arm Parameters
- [ ] `GPS_HDOP_GOOD`
- [ ] `SCR_DEBUG_OPTS`
- [ ] `SCR_LD_CHECKSUM`
- [ ] `SCR_RUN_CHECKSUM`
- [ ] `AUTO_OPTIONS`
- [ ] `GUID_OPTIONS`
- [ ] `EFI_SVF_ARMCHECK`
- [ ] `EFI_SP_ST_DISARM`
- [ ] `ARMING_ACCTHRESH`
- [ ] `ARMING_RUDDER`
- [ ] `ARMING_MIS_ITEMS`
- [ ] `ARMING_CHECK`
- [ ] `ARMING_MAGTHRESH`
- [ ] `ARMING_CRSDP_IGN`
- [ ] `BARO_ALTERR_MAX`
- [ ] `BATTn_ARM_VOLT`
- [ ] `BATTn_ARM_MAH`
- [ ] `BRD_SAFETYOPTION`
- [ ] `BRD_VBUS_MIN`
- [ ] `BRD_VSERVO_MIN`
- [ ] `BRD_OPTIONS`
- [ ] `BRD_HEAT_LOWMGN`
- [ ] `COMPASS_OFFS_MAX`
- [ ] `RC_OPTIONS`
- [ ] `TMODE_FLAGS`
- [ ] `TKOFF_RPM_MIN`
- [ ] `TKOFF_THR_MINACC` (Plane)
- [ ] `TKOFF_ACCEL_CNT` (Plane)
- [ ] `TKOFF_TIMEOUT` (Plane)
- [ ] `RTL_AUTOLAND` (Plane)
- [ ] `FLIGHT_OPTIONS` (Plane)
- [ ] `ARMING_BBOX_SPD` (Plane)
- [ ] `ARMING_REQUIRE` (Plane)
- [ ] `ICE_OPTIONS` (Plane)
- [ ] `Q_ASSIST_SPEED` (Plane)
- [ ] `Q_TAILSIT_ENABLE` (Plane)

Williangalvani:
This implements the (for me) more relevant/important failsafes
image

helps #2723


patrickelectric:
We should ask for marketing to come up with the icons if we want that in the next stable.

patrickelectric (changes requested):

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
```diff
@@ -0,0 +1,231 @@
+<template>
+  <v-form
```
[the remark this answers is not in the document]
Williangalvani: I don't quite recall what the issue was. care to elaborate?
also it looks like we use forms quite a lot? is this a different situation?

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+          params: [
+            {
+              replacementTitle: 'Action',
+              icon: 'mdi-lightning-bolt',
+              name: 'FS_PILOT_INPUT',
+            },
+            {
+              replacementTitle: 'Timeout',
+              icon: 'mdi-timer-outline',
+              name: 'FS_PILOT_TIMEOUT',
```
ES-Alexander, line 79:
```suggestion
              replacementTitle: 'Timeout',
              icon: 'mdi-timer-outline',
              name: 'FS_PILOT_TIMEOUT',
            },
            {
              replacementTitle: 'Action',
              icon: 'mdi-lightning-bolt',
              name: 'FS_PILOT_INPUT',
```
We should be consistent with the ordering.
As a general principle I think the Action should go at the bottom, because everything else has to happen before the action actually triggers.

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+              name: 'LEAK1_TYPE',
+            },
+            {
+              replacementTitle: 'Leak 1 pin',
+              name: 'LEAK1_PIN',
+            },
+            {
+              replacementTitle: 'Leak 1 logic level when dry',
+              name: 'LEAK1_LOGIC',
+            },
```
ES-Alexander, line 104: Is this temporary? If so, might want to add a code comment about it or something. Per an earlier review comment, I think the Action should typically be the last field, since it's conditional on the other fields.

For the record, we've discussed having the pin management and configuration handled in a separate "Leak Sensor" configuration page, in which case it's less problematic to have the parameters for all three supported connections, while also simplifying what needs to get shown in the failsafe card (which is just the failsafe action, and preferably a link to the sensor configuration page).

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+        {
+          name: 'Internal Pressure',
+          generalDescription: 'Triggered when the internal pressure is too high. this could help in detecting a leak'
+          + ' and in avoiding rapid unplanned disassembly.',
+          image: (await import('@/assets/img/configuration/failsafes/pressure.svg')).default as string,
+          params: [
+
+            {
+              replacementTitle: 'Maximum internal pressure',
+              name: 'FS_PRESS_MAX',
```
ES-Alexander, line 116:
```suggestion
              replacementTitle: 'Maximum internal pressure',
              icon: 'mdi-gauge-full',
              name: 'FS_PRESS_MAX',
```

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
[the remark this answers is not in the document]
joaoantoniocardoso: > I don't quite recall what the issue was. care to elaborate? also it looks like we use forms quite a lot? is this a different situation? (first result in google is purple but I don't recall what happened)

https://github.com/bluerobotics/BlueOS/issues?q=is%3Aissue+NS_BINDING_ABORTED+is%3Aclosed

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+            },
+            {
+              replacementTitle: 'Timeout',
+              icon: 'mdi-timer-outline',
+              name: 'FS_PILOT_TIMEOUT',
+            },
+          ],
+        },
+        {
+          name: 'Leak detection',
```
ES-Alexander, line 84:
```suggestion
          name: 'Leak Detection',
```
Consistency with other cards

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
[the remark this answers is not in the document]
joaoantoniocardoso: and: https://github.com/bluerobotics/BlueOS/pull/2351/files

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+      this.failsafes = failsafes
+    })
+  },
+  methods: {
+    async get_failsafes(): Promise<FailsafeDefinition[]> {
+      const failsafes: FailsafeDefinition[] = [
+        {
+          name: 'Ground Station Heartbeat Loss',
+          generalDescription: 'The heartbeat failsafe is triggered when the vehicle does not receive '
+          + 'a heartbeat from the GCS within 3 seconds.',
```
ES-Alexander, line 47:
```suggestion
          generalDescription: 'Triggers when the vehicle does not receive a heartbeat from the GCS '
          + 'within the timeout (default 3 seconds).',
```
The name already says what it's for. Given the active tone in the name we should continue that in the description.

Not sure if we can somehow make the end dynamically determined, where it says "the timeout" if the parameter exists, or "3 seconds" if the parameter doesn't exist. If not then I think this suggestion is an ok compromise.

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+              icon: 'mdi-lightning-bolt',
+              name: 'FS_GCS_ENABLE',
+              replacementDescription: '',
+            },
+          ],
+        },
+        {
+          name: 'Pilot Input Loss',
+          generalDescription: 'The pilot input failsafe is triggered when the vehicle does not receive '
+          + 'any pilot input for a given amount of time.',
```
ES-Alexander, line 68:
```suggestion
          generalDescription: 'Triggers when the vehicle does not receive any pilot input for a given '
          + 'amount of time.',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+            {
+              replacementTitle: 'Timeout',
+              icon: 'mdi-timer-outline',
+              name: 'FS_PILOT_TIMEOUT',
+            },
+          ],
+        },
+        {
+          name: 'Leak detection',
+          generalDescription: 'Triggered when a leak is detected. We recoomend keeping control electronics dry.',
```
ES-Alexander, line 85:
```suggestion
          generalDescription: 'Triggers when a leak is detected. We recommend keeping control electronics dry.',
```
Love the dry humour here :-P

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+              icon: 'mdi-lightning-bolt',
+              name: 'FS_LEAK_ENABLE',
+            },
+            {
+              replacementTitle: 'Leak probe 1 type',
+              name: 'LEAK1_TYPE',
+            },
+            {
+              replacementTitle: 'Leak 1 pin',
+              name: 'LEAK1_PIN',
```
ES-Alexander, line 99:
```suggestion
              replacementTitle: 'Leak 1 pin',
              icon: `mdi-pin',
              name: 'LEAK1_PIN',
```
Not sure if this section is even staying here, but wherever it goes it may as well make use of the icon 🤷‍♂️

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+            {
+              replacementTitle: 'Leak 1 logic level when dry',
+              name: 'LEAK1_LOGIC',
+            },
+          ],
+        },
+        {
+          name: 'Internal Pressure',
+          generalDescription: 'Triggered when the internal pressure is too high. this could help in detecting a leak'
+          + ' and in avoiding rapid unplanned disassembly.',
```
ES-Alexander, line 110:
```suggestion
          name: 'Excess Internal Pressure',
          generalDescription: 'Triggers when the internal pressure is too high. This may help to detect a leak, '
          + 'and to avoid rapid unplanned disassembly.',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+              replacementTitle: 'Action',
+              icon: 'mdi-lightning-bolt',
+              name: 'FS_PRESS_ENABLE',
+            },
+          ],
+        },
+        {
+          name: 'Battery Level',
+          generalDescription: 'Triggered when the battery reaches given thresholds.\n This can help in avoiding '
+          + 'damage to the battery and potentially loss of the vehicle.',
```
ES-Alexander, line 128:
```suggestion
          name: 'Low Battery',
          generalDescription: 'Triggers when the voltage goes below specified thresholds.\n This can help to avoid '
          + 'damage to the battery and potentially loss of the vehicle.',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+          image: (await import('@/assets/img/configuration/failsafes/battery.svg')).default as string,
+          params: [
+            // TODO: add support for the MAH params and coloumb counting
+            {
+              replacementTitle: 'Voltage Measurement Source',
```
ES-Alexander, line 133:
```suggestion
              replacementTitle: 'Voltage measurement source',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+            },
+          ],
+        },
+        {
+          name: 'EKF tolerance',
+          generalDescription: 'Triggered when the EKF variances surpass a given threshold.',
+          image: (await import('@/assets/img/configuration/failsafes/ekf.svg')).default as string,
+          params: [
+            {
+              replacementTitle: 'Variance Threshold',
```
ES-Alexander, line 165:
```suggestion
              replacementTitle: 'Variance threshold',
```

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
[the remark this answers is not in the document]
Williangalvani: ok that is weird. It must be a combination of things, though, right? maybe some request initialized in the form and then aborted because of the submit action?
We do have 15 other v-forms in the code, and I'm actively relying on the form validation in this component.
I'm not looking forward to changing it if the issue isn't really there.
@joaoantoniocardoso can you try the image plz? I'll do a bunch of testing here too.

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+            {
+              replacementTitle: 'Action',
+              icon: 'mdi-lightning-bolt',
+              name: 'FS_EKF_ACTION',
+            },
+          ],
+        },
+        {
+          name: 'Crash Detection',
+          generalDescription: 'Triggered when a crash is detected.',
```
ES-Alexander, line 178:
```suggestion
          generalDescription: 'Triggers when a crash has occurred.',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+            {
+              replacementTitle: 'Action when battery is critical',
+              icon: 'mdi-skull-crossbones-outline',
+              name: 'BATT_FS_CRT_ACT',
+            },
+          ],
+        },
+        {
+          name: 'EKF tolerance',
+          generalDescription: 'Triggered when the EKF variances surpass a given threshold.',
```
ES-Alexander, line 161:
```suggestion
          name: 'Sensor Fusion Uncertainty',
          generalDescription: 'Triggers when the EKF variances surpass a given threshold, '
          + 'so it loses trust in its state estimates.',
```

ES-Alexander (changes requested):
Great start - thanks for all the hard work on this! :-)

A few comments and suggestions, mostly around language consistency and clarity.

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+          params: [
+            // TODO: add support for the MAH params and coloumb counting
+            {
+              replacementTitle: 'Voltage Measurement Source',
+              icon: 'mdi-source-branch',
+              name: 'BATT_FS_VOLTSRC',
+            },
+            {
+              replacementTitle: 'Low voltage threshold',
+              icon: 'mdi-gauge-empty',
```
ES-Alexander, line 139:
```suggestion
              icon: 'mdi-battery-30',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+              name: 'BATT_LOW_VOLT',
+            },
+            {
+              replacementTitle: 'Action when battery is low',
+              icon: 'mdi-lightning-bolt',
+              name: 'BATT_FS_LOW_ACT',
+            },
+            {
+              replacementTitle: 'Critical voltage threshold',
+              icon: 'mdi-gauge-empty',
```
ES-Alexander, line 149:
```suggestion
              icon: 'mdi-battery-alert-variant-outline',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+          ],
+        },
+        {
+          name: 'EKF tolerance',
+          generalDescription: 'Triggered when the EKF variances surpass a given threshold.',
+          image: (await import('@/assets/img/configuration/failsafes/ekf.svg')).default as string,
+          params: [
+            {
+              replacementTitle: 'Variance Threshold',
+              icon: 'mdi-gauge-empty',
```
ES-Alexander, line 166:
```suggestion
              icon: 'mdi-chart-bell-curve',
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
Williangalvani: I think it can be done, but I'd rather keep this PR shorter. we could have a field to override the description text that uses a computed property 🤔

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,200 @@
...
+          name: 'Ground Station Heartbeat Loss',
+          generalDescription: 'The heartbeat failsafe is triggered when the vehicle does not receive '
+          + 'a heartbeat from the GCS within 3 seconds.',
+          image: (await import('@/assets/img/configuration/failsafes/heartbeat.svg')).default as string,
+
+          params: [
+            {
+              replacementTitle: 'Timeout',
+              icon: 'mdi-timer',
+              name: 'FS_GCS_TIMEOUT',
```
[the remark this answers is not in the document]
Williangalvani: yes, when parameters are marked optional, as this one, they are just ommited, otherwise, if a parameter is missing, the whole card is hidden

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
Williangalvani: yeah my goal is to allow links to "advanced" settings wherever applicable, right now I have in mind doing this for battery and leak at least

Williangalvani: and maybe keeping this simpler version here, not 100% sure about that

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,202 @@
...
+              replacementTitle: 'Voltage measurement source',
+              icon: 'mdi-source-branch',
+              name: 'BATT_FS_VOLTSRC',
+            },
+            {
+              replacementTitle: 'Low voltage threshold',
+              icon: 'mdi-battery-30',
+              name: 'BATT_LOW_VOLT',
+            },
+            {
```
ES-Alexander, line 143:
```suggestion
            },
            {
              replacementTitle: 'Low voltage timeout',
              icon: 'mdi-timer-outline',
              name: 'BATT_LOW_TIMER',
            },
            {
```
Realised this one got missed. Not sure if that was intentional 🤷‍♂️

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
ES-Alexander: > and maybe keeping this simpler version here, not 100% sure about that

I think it's cleaner to separate the concerns - the Action is (high level) failsafe configuration, but the pins and logic level stuff are (lower level) hardware configuration, so can be relegated to the page for that.

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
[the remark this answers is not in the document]
ES-Alexander: Neato :-)

Could consider including `FS_EKF_FILT` as optional for Copter then if you want, but 🤷‍♂️

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
Williangalvani: I hadn't seem this one, interesting!

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
Williangalvani: makes sense. but this can wait until we have that page, right? I'll add a TODO

core/frontend/src/assets/img/configuration/failsafes/pressure.svg
ES-Alexander, line 1: Would be better if the exclamation didn't change with the light/dark theme. Here's a fixed one:
 
pressure

patrickelectric:
> We should ask for marketing to come up with the icons if we want that in the next stable.

SMS will work on that

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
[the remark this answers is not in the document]
patrickelectric: I would be okay if @joaoantoniocardoso fails to replicate the issue, he's pretty good on that :laughing:

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
```diff
@@ -0,0 +1,231 @@
...
+    ref="form"
+    v-model="is_form_valid"
+    @submit.prevent="saveEditedParam()"
+  >
+    <template v-if="!custom_input && param?.bitmask">
+      <v-checkbox
+        v-for="(key, keyvalue) in param?.bitmask"
```
patrickelectric, line 9: the optional does not make sense since it was validated by previous if

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
```diff
@@ -0,0 +1,231 @@
...
+        v-model="selected_bitflags"
+        dense
+        hide-details
+        :loading="waiting_for_param_update"
+        :label="key"
+        :value="2 ** keyvalue"
+      />
+    </template>
+    <v-autocomplete
+      v-else-if="!custom_input && Object.entries(param?.options ?? []).length > 10"
```
patrickelectric, line 20: is there an reason for why enabling it if bigger than 10 ? Why not always ?

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
```diff
@@ -0,0 +1,231 @@
...
+      v-model.number="new_value"
+      variant="solo"
+      :items="as_select_items"
+    />
+    <v-select
+      v-else-if="!custom_input && param?.options"
+      v-model.number="new_value"
+      dense
+      :items="as_select_items"
+      :rules="[() => true]"
```
patrickelectric, line 30: what ?

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
```diff
@@ -0,0 +1,231 @@
...
+      :indeterminate="waiting_for_param_update"
+      @change="updateVariables"
+    />
+    <v-text-field
+      v-if="
+        custom_input
+          || (param
+            && !param.options
+            && !param.bitmask
+          )
```
patrickelectric, line 40:
```suggestion
        custom_input || (!param?.options && !param?.bitmask)
```

ES-Alexander (changes requested):
Looking great! Just a few more small changes / suggestions, and these few extra things that I couldn't tie to specific lines of code:

1. It seems a little weird to be able to set negative values for the battery voltages, but that might not be something we have control over within this PR
2. ~When a parameter is set to a custom value (as is the case for the default value of the EKF variance threshold) the dropdown just shows up empty~
    - ~This is also a problem in the Parameters page, but it at least shows as the custom value in the main table there, and has the option to set other custom values (but I think it's ok for the failsafes page to not support setting custom values, would just be nice if it displayed them somehow)~
    
    https://github.com/bluerobotics/BlueOS/assets/25898329/2883fb5b-48b8-4eb9-9505-66fdf4a22d99
    - ~If this isn't fixable in this PR, I can open it as a separate issue~
    - **Resolved**
3. I think it would be more intuitive if fields that are linked to an Action get greyed out/disabled when the action is set to disabled
    - I'm not sure how hard that is to implement, and I can add it as a future TODO in the Issue if you don't think it's worth doing in this PR
4. I believe this box is incorrectly being set to have a background colour, when it should be transparent so the cards are surrounded by blue (like in the other configuration pages)
    image

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
```diff
@@ -0,0 +1,231 @@
...
+      this.updateSelectedFlags()
+    },
+  },
+  mounted() {
+    this.new_value = this.param?.value ?? 0
+    this.updateVariables()
+    this.updateSelectedFlags()
+  },
+  methods: {
+    isInRange(input: number | string): boolean | string {
```
patrickelectric, line 137: Can't we use the parameter editor dialog ? Or break this in a common ground ?

patrickelectric (changes requested):

core/frontend/src/assets/img/configuration/failsafes/pressure.svg
patrickelectric: image

I see only a yellow thing here

ES-Alexander:
> SMS will work on that, we can merge after stable

@patrickelectric feedback so far has been positive on the existing hand-drawn designs. I've mentioned that I can neaten them up or change them to not drawn ones, but that hasn't been requested or desired by the five people who've responded.

Daniel did mention that it might be a bit weird to have a different style only in this page, but we can have a think about whether that's ok since we don't have such images anywhere else, or if we perhaps want to add those kinds of images to some of our other pages as well :-)

core/frontend/src/assets/img/configuration/failsafes/pressure.svg
Williangalvani: we have custom css applied, so the way it shows in github is not the same as in blueos

patrickelectric:
> > SMS will work on that, we can merge after stable
> 
> @patrickelectric feedback so far has been positive on the existing hand-drawn designs. I've mentioned that I can neaten them up or change them to not drawn ones, but that hasn't been requested or desired by the five people who've responded.
> 
> Daniel did mention that it might be a bit weird to have a different style only in this page, but we can have a think about whether that's ok since we don't have such images anywhere else, or if we perhaps want to add those kinds of images to some of our other pages as well :-)

I agree, the feedback from SMS was: "I think we could probably keep the essence of these icons, but maybe tweak them just a little!"

patrickelectric:
@ES-Alexander @Williangalvani after talking with @rjehangir we agreed that the current icons can be used but sms will be working to make them sharp. Meanwhile, if the PR gets ready, we can merge it before 1.3 stable.

ES-Alexander:
@Williangalvani there are cleaned up SVGs downloadable [in the Issue](https://github.com/bluerobotics/BlueOS/issues/2723#issuecomment-2198254328), which should probably be switched to for now.

core/frontend/src/assets/img/configuration/failsafes/battery.svg
ES-Alexander, line 1: battery

core/frontend/src/assets/img/configuration/failsafes/crash.svg
ES-Alexander, line 1: crash

core/frontend/src/assets/img/configuration/failsafes/ekf.svg
ES-Alexander, line 1: ekf

core/frontend/src/assets/img/configuration/failsafes/heartbeat.svg
ES-Alexander, line 1: heartbeat

core/frontend/src/assets/img/configuration/failsafes/leak.svg
ES-Alexander, line 1: leak

core/frontend/src/assets/img/configuration/failsafes/pilot-input.svg
ES-Alexander, line 1: pilot-input

core/frontend/src/assets/img/configuration/failsafes/pressure.svg
ES-Alexander, line 1: pressure

core/frontend/src/assets/img/configuration/failsafes/temperature.svg
ES-Alexander, line 1: temperature

core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
```diff
@@ -0,0 +1,122 @@
...
+
+<style>
+
+i.svg-icon svg {
+  height: 100% !important;
+  min-width: 180px;
+}
+
+i.svg-outline-dark path {
+  fill: white;
```
ES-Alexander, line 105:
```suggestion
  fill: #d1eaf1;
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
```diff
@@ -0,0 +1,122 @@
...
+}
+
+i.svg-outline-light path {
+  fill: black;
```
ES-Alexander, line 109:
```suggestion
  fill: #002f45;
```

ES-Alexander (changes requested):
SMS requested colour changes

core/frontend/src/components/vehiclesetup/configuration/failsafes/Failsafes.vue
```diff
@@ -0,0 +1,229 @@
...
+              replacementTitle: 'Action',
+              icon: 'mdi-lightning-bolt',
+              name: 'FS_PRESS_ENABLE',
+            },
+          ],
+        },
+        {
+          name: 'Excess Internal Temperature',
+          generalDescription: 'Triggers when the internal temperature is too high. This may help preventing '
+          + 'damage to electronics due to overheating.',
```
ES-Alexander, line 132:
```suggestion
          generalDescription: 'Triggers when the internal temperature is too high. This may help to prevent '
          + 'damage to electronics due to overheating.',
```

ES-Alexander (changes requested):
I've confirmed that both aspects of the no-value parameter thing has been fixed - thanks! 😀 
I did notice that now (?[^1]) when you click save on a parameter in the Autopilot Parameters page the dialog goes into the "unchanged" state (with the save button greyed out). I think it would make more sense to just close the dialog, since the user is done with it, and otherwise the 'correct' interaction sequence is "change param, click save, click cancel", which is unintuitive.

[^1]: This might just be an existing interface issue with the Parameter page, rather than being new, but I've just noticed it now.

Would be good if you can comment on the numbered items 1, 3, and 4 of my earlier review, as to whether they'll be included in this PR or I should open issues for them :-)

There are a couple of colour issues in my drawn70-set that I didn't notice before, but they should be fixed in the new SMS colours set anyway 👍

Williangalvani:
@ES-Alexander the negative voltage is somewhat hard to control, the limits should be defined in the parameter metadata.

As for blocking the settings if we don't have a threshold set, I think it should be a subsequent PR, and it would probably mean we should move the action up, no? 👀

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
Williangalvani: well, that's your code :eyes: , I just copied it over, but the logic is that we don't need autocomplete if we have few choices

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
Williangalvani: also old code, doesnt's seem like it is required. removed

core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
Williangalvani: refactored so editordialog uses this

ES-Alexander:
> the negative voltage is somewhat hard to control, the limits should be defined in the parameter metadata.

Yeah, fair enough.

> As for blocking the settings if we don't have a threshold set, I think it should be a subsequent PR, and it would probably mean we should move the action up, no? 👀

Nah, I think the actions should stay at the bottom, but the rest of the card can be greyed out / disabled, so it's clearer that the other fields aren't effective until the (normal-coloured + interactive) Action field has been configured to do something :-)

ES-Alexander (approved):
Looking great! :D

Only remaining comments are:
1. It'd be nice to have a non-zero minimum horizontal margin around the cards, to prevent this from being possible:
    Screenshot 2024-07-05 at 10 07 04 AM
2. I think low battery is more important than the pressure/temp ones, and it would also look better if it was moved up to directly after the Leak Sensor (so when they're paired up the pressure/temp ones go together and can shrink the card height)
    Screenshot 2024-07-05 at 10 15 05 AM
3. I'm not sure how the flat blue background is applied, but it might be worth showing the page background gradient, or at least making sure the flat background colour is defined in terms of the theme variables

I do think 1 and 2 are worth fixing, but none of those points are critical, so approving now. 

Again, excellent work with this - it'll hopefully make the failsafes more approachable and widely used :-)

patrickelectric (approved):

PR #2754, opened 2024-06-26
frontend: App: Add new way to enter in dev mode

JoaoMario109:
* Add new conditionals for dev mode
* Add new way to enter dev mode in App
* Remove cloud integration unless in dev mode

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -61,7 +61,7 @@
         <internet-tray-menu />
         <wifi-tray-menu />
         <ethernet-tray-menu />
-        <cloud-tray-menu />
+        <cloud-tray-menu v-if="settings.is_dev_mode" />
         <notification-tray-button />
       </v-app-bar>
     </v-card>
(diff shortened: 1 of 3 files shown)


core/frontend/src/App.vue
```diff
@@ -719,6 +734,12 @@ export default Vue.extend({
         this.$router.push('/')
       }
     },
+    buildDateClick(): void {
+      this.build_clicks = (this.build_clicks + 1) % 10
+      if (this.build_clicks === 0) {
+        settings.is_dev_mode = true
+      }
```
patrickelectric, line 741: Is there a reason for not doing something simpler ?
```suggestion
      this.build_clicks += 1
      settings.is_dev_mode = this.build_clicks > 10
```

patrickelectric (changes requested):

core/frontend/src/App.vue
JoaoMario109: Just to keep consistent that if the user activate and deactivate the dev mode without refreshing it still will require 10 clicks and not only one in the second activation.

patrickelectric (approved):

PR #2759, opened 2024-06-27
frontend: CompassDisplay: fix ekf arrow being black

### LINKED ISSUE #2757: bug: EKF needle color in new compass widget does not match color key
### Bug description

Testing 1.3.0-beta.6

The new compass widget is nifty! However, the color key seems to be wrong: the EKF needle is black, not cyan. 
image

### Steps to reproduce

1. blueos.local/vehicle/setup
2. click on configure
3. click on setup

### Primary pain point(s)

User confusion

### Additional context

_No response_

### Prerequisites

Williangalvani:
fix #2757

### DIFF
--- a/core/frontend/src/components/vehiclesetup/configuration/compass/CompassDisplay.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/compass/CompassDisplay.vue
@@ -235,7 +235,7 @@ export default Vue.extend({
       for (const [index, angleDegrees] of this.headings.entries()) {
         if (angleDegrees === null) continue
         const paramValue = this.compasses?.[index]?.paramValue
-        const color = paramValue ? this.colors[paramValue] : 'black'
+        const color = paramValue ? this.colors[paramValue] : this.primaryBaseColor
         ctx.save()
         this.renderVariables.yawAngleDegrees[index] = angleDegrees
         ctx.rotate(glMatrix.toRadian(angleDegrees))


ES-Alexander:
Confirmed the arrow colour is fixed 👍

That said, there are some other small compass page issues that should probably get fixed in the same PR:

- [ ] The Pixhawk only has one listed compass, but still shows an arrow for compass 2, just set pointing north
    - it should be hidden if there is no compass for it
- [ ] The "Compass Learn" dropdown description has a typo where it says "...then _driver_ the vehicle around" (should be _drive_), and is missing the final period
- [ ] The "Compass Learn" popup window has the large vehicle calibration title
- [ ] The second word in the "Compass Learn" popup description is unnecessarily capitalised ("A Valid position" should be "A valid position")
- [ ] The "Large Vehicle Calibration" popup is using the "Compass Learn" popup's description
- [ ] There's no "back" / "undo" option when you select to enter custom coordinates in the "Large Vehicle Calibration" popup, so to use the GeoIP coordinates from that point you need to either refresh the whole page or copy them from the section above
    - It might make more sense to not have the two buttons at all, and just start with the custom coordinates view but pre-filled with the GeoIP coordinates, so the user can either click "use these coordinates" straight away, or change them first if they want custom ones
    - To make it extra obvious we could change the confirmation button text to be "Use GeoIP coordinates" when the input coords match the GeoIP ones, and then change it to "Confirm custom coordinates" when they don't match
- [ ] The background of all the compass calibration popup window title bars is hardcoded to the light-mode grey
    Screenshot 2024-06-28 at 2 54 01 PM

patrickelectric:
@ES-Alexander your comment was moved to an issue to track the other issues

patrickelectric (approved):

PR #2769, opened 2024-06-29
Create component for "gating" pages and wait until parameters finish loading

Williangalvani:
Also apply it to all of the vehicle-setup section, which makes everything smoother as the data doesn't get populated slowly, and removes duplicated code

we have to check how loading default parameters deal with it


Williangalvani:
> we have to check how loading default parameters deal with it

actually this will only be triggered when the parameters table gets reset, so I think we're good

patrickelectric (approved):

Issue #2328, opened 2024-01-22
Deal with issues when the disk is full


patrickelectric:
Check: https://github.com/bluerobotics/BlueOS/issues/2327, https://github.com/bluerobotics/BlueOS/issues/2323, https://github.com/bluerobotics/BlueOS/issues/2326, https://github.com/bluerobotics/BlueOS/issues/1015

The docker is able to start, but everything after that just results in unstable behavior.
Some points that you suggested are already available as issues, others are relevant to recover the system.
- [x] We should delete old logs when doing the rotation and noticing the the disk space is almost full.
- [ ] We should stop logging if the disk space is almost full.
    - This conflicts with rotation configuration in loguru 
- [ ] We should clean up old dockers that are not being used.
- [ ] We should clean up old docker artifacts that are not being used.
- [ ] We should allow user to delete all unused docker images.
- [ ] We should warn the user though cockpit that the companion computer is almost full in disk.
- [ ] We should erase older tlog or bin files if the disk is almost full.
    - #1257
- [ ] We should warn the user though BlueOS header that the disk is almost full and in critical state.
- [ ] We **may** not allow the user to arm the vehicle if the disk is almost full.
- [ ] We **may** do some of this steps automatically to try to recover the system once it starts.  
- [ ] We **may** need a page like [filelight](https://apps.kde.org/pt-br/filelight/) on BlueOS to help identify the root of such problems.
- [x] We should limit journald max size

_Originally posted by @patrickelectric in https://github.com/bluerobotics/BlueOS/issues/2325#issuecomment-1902117968_

joaoantoniocardoso:
I know it might be trickier to manage the installation, but another valid strategy is to put `/var` in another partition.

patrickelectric:
#2359

voorloopnul:
I installed one extension [Nortek Nucleus], that grew the docker log to 18GB in about a week, maybe kraken could add a limit in the size of the docker logs:

https://docs.docker.com/config/containers/logging/configure/#configure-the-default-logging-driver

`--log-opt max-size=100m`

joaoantoniocardoso:
I've just freed ~5 gb here by doing:

```bash
sudo docker system prune  # ~3 gb
sudo journalctl --vacuum-time=2d  # ~1 gb
sudo apt-get clean  # ~1 gb
```

patrickelectric:
@joaoantoniocardoso how we end up with 1GB of unnecessary stuff in our apt ?

[...]

PR #2790, opened 2024-07-05
Add cron worker with journalctl cleaner

patrickelectric:
Helps #2328

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -79,6 +79,8 @@ COPY services /home/pi/services
 RUN /home/pi/services/install-services.sh
 COPY start-blueos-core /usr/bin/start-blueos-core
 COPY run-service.sh /usr/bin/run-service
+RUN rm -f /etc/cron.d/*
+COPY cron /etc/cron.d
 
 # Copy binaries and necessary folders from downloadBinaries to this stage
 COPY --from=downloadBinaries \
--- a/core/cron/waste-collector
+++ b/core/cron/waste-collector
@@ -1,0 +1,1 @@
+0 */4 * * * root journalctl-vacuum # Run every 4 hours
--- a/core/services/install-services.sh
+++ b/core/services/install-services.sh
@@ -6,12 +6,18 @@ set -e
 BUILD_PACKAGES=(
     g++
 )
+EXTERNAL_PROGRAMS=(
+    cron
+)
+
+apt update
 
 # Install build packages if not on armv7, which we have all pre-built wheels for
 if ! { [ "$TARGETARCH" == "arm" ] && [ "$TARGETVARIANT" == "v7" ]; }; then
-    apt update
     apt install -y --no-install-recommends ${BUILD_PACKAGES[*]}
 fi
+
+apt install -y --no-install-recommends ${EXTERNAL_PROGRAMS[*]}
 
 # Wifi service:
 ## Bind path for wpa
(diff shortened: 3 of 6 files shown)


Williangalvani:
should we use a script instead? we need something to cleanup the autopilot and mavlink logs from time to time, too :thinking:

patrickelectric:
> should we use a script instead? we need something to cleanup the autopilot and mavlink logs from time to time, too 🤔

what you mean by script ?

Williangalvani:
> what you mean by script ?

never mind, if we have cron working, we can just had other entries for cleaning different things

patrickelectric:
> > what you mean by script ?
> 
> never mind, if we have cron working, we can just had other entries for cleaning different things

yes, that the plan

joaoantoniocardoso (changes requested):
The idea of adding a cron is interesting for our system, however, journalctl is a log manager that doesn't require user intervention: it has configurations like (1) log size, (2) free disk size, (3) rate limit, (4) compression, and it will vacuum automatically.

https://www.freedesktop.org/software/systemd/man/latest/journald.conf.html

patrickelectric:
What is better, calling journalctl configuration command or working on journalctl configuration file/parameters, dealing with user custom configuration and identifying (based in the OS), which of the following files to change ?
/etc/systemd/journald.conf
/run/systemd/journald.conf
/usr/local/lib/systemd/journald.conf
/usr/lib/systemd/journald.conf
/etc/systemd/journald.conf.d/*.conf
/run/systemd/journald.conf.d/*.conf
/usr/local/lib/systemd/journald.conf.d/*.conf

patrickelectric:
To me it adds extra complexity at the moment, what is the problem of the current approach ?

joaoantoniocardoso:
> To me it adds extra complexity at the moment, what is the problem of the current approach ?

(1) we are taking power from the administrator, and (2) we are reinventing the wheel instead of configuring a basic system tool.

What I recommend is that when installing the system, we should overwrite (with a backup) `/etc/systemd/journald.conf`, print a message about it, and if the administrator wants a different configuration from our default, he can just change the configs there after installing it.

patrickelectric:
> > To me it adds extra complexity at the moment, what is the problem of the current approach ?
> 
> (1) we are taking power from the administrator, and (2) we are reinventing the wheel instead of configuring a basic system tool.
> 
> What I recommend is that when installing the system, we should overwrite (with a backup) `/etc/systemd/journald.conf`, print a message about it, and if the administrator wants a different configuration from our default, he can just change the configs there after installing it.

Installation does not fix the problem for current BlueOS systems.

joaoantoniocardoso:
> > > To me it adds extra complexity at the moment, what is the problem of the current approach ?
> > 
> > 
> > (1) we are taking power from the administrator, and (2) we are reinventing the wheel instead of configuring a basic system tool.
> > What I recommend is that when installing the system, we should overwrite (with a backup) `/etc/systemd/journald.conf`, print a message about it, and if the administrator wants a different configuration from our default, he can just change the configs there after installing it.
> 
> Installation does not fix the problem for current BlueOS systems.

Can't we install it if the file doesn't exist?

patrickelectric:
> > > > To me it adds extra complexity at the moment, what is the problem of the current approach ?
> > > 
> > > 
> > > (1) we are taking power from the administrator, and (2) we are reinventing the wheel instead of configuring a basic system tool.
> > > What I recommend is that when installing the system, we should overwrite (with a backup) `/etc/systemd/journald.conf`, print a message about it, and if the administrator wants a different configuration from our default, he can just change the configs there after installing it.
> > 
> > 
> > Installation does not fix the problem for current BlueOS systems.
> 
> Can't we install it if the file doesn't exist?

By default this file comes from the OS.

patrickelectric:
I'm working in an alternative solution.

patrickelectric:

PR #2802, opened 2024-07-09
frontend: fix off-by-8 issue with pixhawk channels

Williangalvani:


ES-Alexander (approved):
Nice :-)

A bit undecided on whether we should have AUX or Aux, but 🤷‍♂️

patrickelectric (approved):

PR #2806, opened 2024-07-09
add interface for Gyro calibration

Williangalvani:
image


Williangalvani:
opening as I think it is useful enough as is. I would like to make a better ui to show the current readings, but I cant think of anything good right now.

patrickelectric (commented):
There are two things that appears a bit weird, using 0 over NaN or something like it when we failed to fetch board values. And the word Gyro over gyroscope, it makes the text and reading a bit weird.. as "gyros" for the plural of gyroscopes.

PR #2819, opened 2024-07-17
tools: bump mavp2p to 1.2.1

Williangalvani:
needs testing


Williangalvani:
closing as I think we should figure out why it stopped working before we update it

PR #2829, opened 2024-07-19
Frontend: ExtManagerView: Fix passing a reference of installed extension to edit extension card

JoaoMario109:
This fix cases where user edits the tag and the installed extension card changes instantaneously without create being pressed


joaoantoniocardoso (approved):
Tested, thanks for the fix!

PR #2847, opened 2024-07-23
helper: raise timeout to 2 seconds, try probing up to 3 times by default

### LINKED ISSUE #2846: extension probing sometimes fail, showing a 404 page
image
image

Williangalvani:
as the issue is hard to replicate, I'll be running this image for a while. 
should fix #2846

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -303,16 +303,21 @@ class Helper:
 
         return request_response
 
+    # pylint: disable=too-many-arguments,too-many-branches
     @staticmethod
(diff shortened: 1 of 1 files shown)


core/services/helper/main.py
```diff
@@ -305,14 +305,18 @@ def simple_http_request(
 
     @staticmethod
     @temporary_cache(timeout_seconds=1)  # a temporary cache helps us deal with changes in metadata
```
patrickelectric, line 307: We should update the timeout as well

joaoantoniocardoso:
I think the retries should be implemented inside the Helper.simple_http_request instead

PR #2850, opened 2024-07-29
frontend: InstalledExtCard: Add Major Tom update

JoaoMario109:
* Add manual option for users to update Major Tom


JoaoMario109:
Decided to not allow user to manually update

PR #2852, opened 2024-07-30
Frontend/Kraken: Add continuous logging for better Extensions debugging

JoaoMario109:
* Add continuous logging in backend
* Add heartbeats in streamer to force data to frontend
* Add auto reconnect in frontend

### DIFF
--- a/core/frontend/src/views/ExtensionManagerView.vue
+++ b/core/frontend/src/views/ExtensionManagerView.vue
@@ -26,11 +26,25 @@
       width="80%"
     >
       <v-card>
-        <v-card-text>
-          <!-- eslint-disable -->
-          <pre class="logs" v-html="log_output" />
-          <!-- eslint-enable -->
-        </v-card-text>
+        <v-app-bar dense>
+          <v-spacer />
+          <v-toolbar-title>
+            {{ log_info_output }}
+          </v-toolbar-title>
+          <v-spacer />
+          <v-checkbox
+            v-model="follow_logs"
+            label="Follow Logs"
+            hide-details
+          />
+        </v-app-bar>
+        <v-sheet>
+          <v-card-text ref="logContainer" class="scrollable-content">
+            <!-- eslint-disable -->
+            <pre class="logs" v-html="log_output" />
+            <!-- eslint-enable -->
+          </v-card-text>
+        </v-sheet>
       </v-card>
     </v-dialog>
     <v-toolbar>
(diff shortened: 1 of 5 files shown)


core/frontend/src/views/ExtensionManagerView.vue
```diff
@@ -410,61 +431,58 @@ export default Vue.extend({
...
-            if (!output) {
+            if (result) {
+              lastDecode = ansi.ansi_to_html(result)
+              this.log_info_output = `Logs for ${extension.name}`
+              this.$set(this, 'log_output', outputBuffer + lastDecode)
+            }
+            this.$nextTick(() => {
+              const logContainer = this.$refs.logContainer as HTMLElement
+              if (logContainer) {
+                logContainer.scrollTop = logContainer.scrollHeight
```
patrickelectric, line 465: Can we avoid scrolling to the bottom every time that a new message is received when the scroll is not in the bottom ?
The use case is that if someone is trying to look at the logs, it should be able to look to previous log messages when new ones are being received.

patrickelectric (changes requested):

core/services/kraken/api/v1/routers/index.py
```diff
@@ -37,10 +37,19 @@ async def list_containers() -> Any:
...
-async def log_containers(container_name: str) -> StreamingResponse:
-    return StreamingResponse(
-        timeout_streamer(ContainerManager.get_container_log_by_name(container_name)), media_type="text/plain"
-    )
+async def log_containers(container_name: str, timeout: Optional[int] = None) -> StreamingResponse:
+    """
+    Fetch logs of a given container.
+    If timeout is provided, the stream will be closed after no log line is received for the given timeout.
+    If continuous is set to True, the stream will be kept open until the request be aborted by the client, even if the
+    container is stopped, it will keep trying to reconnect to the container log.
```
joaoantoniocardoso, line 45: There is no `continuous` parameter here.. is this docstring up to date??

JoaoMario109: Already removed, it was an old parameter, thanks!

core/frontend/src/views/ExtensionManagerView.vue
JoaoMario109: Added a Follow Logs checkbox to allow users to not auto scroll

patrickelectric (approved):

PR #2857, opened 2024-08-09
Add camera configuration UI

Williangalvani:


patrickelectric (commented):
@ES-Alexander (Mr. Elioto), can you put some of you magic touch in this range image ?

ES-Alexander:
@patrickelectric these were my original design suggestions: 

image

I expect @Williangalvani is starting out with a simpler interface, although given this PR is still a draft it's quite possibly still a WIP.

We did have some discussions about the diagram for this being much more dynamic than the others (e.g. adjusting the diagram based on the specified values as well as the autopilot telemetry for the current orientation and mount tilt angles), which does also make it trickier to have hand-drawn aspects for things like the arrows, if we want them to have variable lengths depending on the angles being covered...

There was also some discussion around a 3D interface, especially if we want to visualise the field of view:

Williangalvani:
@ES-Alexander that is correct. I decided to start smaller as this will cover 90% of what we need. Just to avoid sinking too much time into it. I also had misunderstood how some of those parameters work, which is why they are gone.

patrickelectric (changes requested):
Is 0 degrees expected here ?

image

patrickelectric:
ping @Williangalvani

eldinmiller:
I second that it would be useful to have a "center" or "home", it may not be the center of the two values.  For example, you may want to have less range of motion up than down, but still center straight ahead.

Williangalvani:
> I second that it would be useful to have a "center" or "home", it may not be the center of the two values. For example, you may want to have less range of motion up than down, but still center straight ahead.

hmmm interesting. that would require a small change on ardusub, but I think it is worth it

Williangalvani:
Updated to include most of the functionality I originally intended to have.
The downside is that this now requires Sub >= 4.5
Screenshot 2024-09-17 at 15 08 51

patrickelectric (approved):

PR #2873, opened 2024-08-27
Frontend Add sentry feedback option in Report Menu

JoaoMario109:
image
image
image
image


core/frontend/src/components/app/ReportMenu.vue
```diff
@@ -80,6 +95,40 @@ export default Vue.extend({
     openDiscuss(): void {
       window.open(this.discussUrl(), '_blank')
     },
+    async openSimpleReport(): Promise<void> {
+      // @ts-expect-error - Theme is not defined in the type
+      const { isDark, currentTheme } = this.$vuetify.theme
+
+      const light = '#FFF'
+      const dark = '#1E1E1E'
```
patrickelectric, line 103: Just to be sure, did you check our style folder ?

https://github.com/bluerobotics/BlueOS/tree/b0a2c9f230b56ff82ba4bc1aa7310fd7126c1814/core/frontend/src/style

If such colors are not here, it would be nice to add it as well

JoaoMario109: We don't have these colors there since they are v-sheet colors used by vuetify itself, added them in default colors since some other places will probably use it in future also

patrickelectric (approved):

PR #2875, opened 2024-08-27
Fix Endpoint comparison, always overwrite mavlink2rest one

### LINKED ISSUE #2874: Mavlink2rest breaks if updating to 1.3.0 with a pre-existing settings file
The main symptom is a lack of communication in blueos (no heartbeat) and in cockpit.
This happens ONLY if the user manually changed endpoints, with trigger a settings file to be created.

Williangalvani:
fix #2874


JoaoMario109 (approved):

patrickelectric (approved):
After talking with @JoaoMario109 we agreed that will be safer if we cherry-pick this to 1.3.

PR #2881, opened 2024-08-30
Handle (for now) unsupported hotspot functionality in bookworm

Williangalvani:
This greys out the hotspot buttons and prevents the backend from trying to enable it if the underlying OS is not supported

~needs testing~ tested on pi4/bookworm

### DIFF
--- a/core/services/wifi/Hotspot.py
+++ b/core/services/wifi/Hotspot.py
@@ -44,7 +44,7 @@ class HotspotManager:
         self.ipr = IPRoute()
 
         self._ap_interface_name = ap_interface_name
-        self._supports_hotspot = self.check_hotspot_support()
+        self.supports_hotspot = self.check_hotspot_support()
         try:
             dev_id = device_id()
         except Exception:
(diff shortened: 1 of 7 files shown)


core/services/wifi/main.py
```diff
@@ -167,8 +172,8 @@ async def disconnect() -> Any:
 
 @app.get("/hotspot", summary="Get hotspot state.")
 @version(1, 0)
-def hotspot_state() -> Any:
-    return wifi_manager.hotspot.is_running()
+def hotspot_state() -> HotspotStatus:
```
patrickelectric, line 175: This is an API break, a v2 field should be used or a new endpoint.

patrickelectric (changes requested):

core/services/wifi/main.py
Williangalvani: good catch. I updated it to use a new endpoint

patrickelectric (changes requested):

core/services/wifi/Hotspot.py
```diff
@@ -80,6 +80,10 @@ def binary_name() -> str:
     def binary(self) -> pathlib.Path:
         return self._binary
 
+    def check_hotspot_support(self) -> bool:
+        # Support for Bookworm should arrive with NetworkManager support
+        return bool(get_host_os() == HostOs.Bullseye)
```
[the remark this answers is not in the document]
Williangalvani: ask mypy....

patrickelectric (approved):

Issue #2752, opened 2024-06-26
Safety considerations and suggested changes


Williangalvani:
BlueOS
 - Implement an automatic “safe mode”, that by default is on when armed
   - Basically for BlueOS-level pre-arm checks, since BlueOS can’t actually prevent arming
   - Should show indications of what is blocked when engaged

 - Things that should be blocked in safe mode
     - Firmware flashing
     - BlueOS updates
     - Extensions installations
     - Vehicle Setup (configuration, failsafes, etc)
     - Loading parameter files
     - Setting reboot-required parameters

 - Add warnings for
     - network interface changes
     - Mavlink Endpoints
     - speed tests
     - Rebooting (core and host os)
     - Accessing the File Browser
     - Downloading logs

 - Warning dialog should have options for “disarm and do the thing”, “ignore and do the thing”, “don’t do the thing”, with a  -  -  - checkbox for “don’t warn me again” (with a configurable duration?)
 - Have the safe/unsafe state accessible from anywhere (BlueOS itself and extensions). Maybe an endpoint?
 - Allow dedicating some number of cores to the autopilot functionality
 - BlueOS could send status messages on the MAVLink stream when things get critical (throttling, temperature, cpu, disk, etc)
 - BlueOS could also send messages about temperature, storage, etc, to the MAVLink stream so Cockpit can show it to the pilot

PR #2888, opened 2024-09-09
creating "safe mode"

Williangalvani:
Screenshot 2024-09-09 at 12 18 49

helps #2752

### DIFF
--- a/core/frontend/src/store/autopilot.ts
+++ b/core/frontend/src/store/autopilot.ts
@@ -36,6 +36,8 @@ class AutopilotStore extends VuexModule {
 
   system_id = 1
 
+  verhicle_armed = false
+
   get parameter() {
     return (name: string): Parameter | undefined => this.parameters.find((parameter) => parameter.name === name)
   }
@@ -51,6 +53,11 @@ class AutopilotStore extends VuexModule {
     return (user_filter: (param: Parameter) => boolean) => this.parameters.filter(
       (param: Parameter) => user_filter(param),
     )
+  }
+
+  get is_safe() {
+    // We can potentially check for external things here
+    return !this.verhicle_armed
   }
 
   @Mutation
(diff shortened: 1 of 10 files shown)


core/frontend/src/store/autopilot.ts
```diff
@@ -53,6 +55,11 @@ class AutopilotStore extends VuexModule {
     )
   }
 
+  get is_safe() {
+    // We can potentially check for external things here
+    return !this.verhicle_armed
+  }
+
```
joaoantoniocardoso, line 62: why is this not calling the backend instead?

core/services/ardupilot_manager/main.py
```diff
@@ -290,6 +290,12 @@ async def available_boards() -> Any:
     return await autopilot.available_boards(True)
 
 
+@app.get("/safe")
```
patrickelectric, line 293: Doesn't make more sense to move to "armed" ? Safe is to abstracted, and armed is a common term.

joaoantoniocardoso: @patrickelectric I think `safe` can abstract more than just `arm`, in case we want to check for other things..?

patrickelectric: any idea of other things that the backend of ardupilot_manager may know ?
It would be nice to have a comment about that.

core/frontend/src/store/autopilot.ts
Williangalvani: This will give us feedback at 1hz, I don't want yet another thing polling the backends at 1hz for no reason...
I can add to the comment "such as ardupilotManager's /safe endpoint"

patrickelectric: So why we have the safe endpoint in the backend ?

joaoantoniocardoso: hmm, so why was the backend endpoint added?

Williangalvani: for subsequent backend-based validations, such in ardupilot_manager.py and whatever else

core/services/ardupilot_manager/main.py
Williangalvani: if there's a flashing operation in progress, for example, I'll add comments and/or more checking

Williangalvani: comment added, I'll leave the actual blocking checks to after @JoaoMario109 's refactor is done

patrickelectric (approved):

PR #2890, opened 2024-09-10
frontend: create widget to show upload/download speeds

Williangalvani:
Screenshot 2024-09-10 at 10 13 14

ideally the math should be done by linux2rest itself...
a current bug is that if you go to the system page, you get doubled the requests, and as they seem to be cached in linux2rest's side, the value jumps to zero and back

The values are also not super consistent with other tools, but that could be due to the super low sampling rate


patrickelectric:
I'll add the speed calculation in linux2rest

patrickelectric (approved):

PR #2891, opened 2024-09-10
versionchooser: bump dockerhub page_size to 200

Williangalvani:


ES-Alexander (approved):
I confirmed this functionality works.

@Williangalvani mentioned this is preferable to dynamically fetching more pages with fewer per page because limits are around API calls, not their contents.

Only potential concerns are around 
1. whether too large a number would cause the fetch to fail if there aren't enough images
    - this seems highly unlikely
1. whether it's too much data to fetch for people on slow connections
    - if they can't fetch this much metadata then they almost certainly can't fetch an actual image, so this seems like a non-issue

patrickelectric (approved):

PR #2894, opened 2024-09-11
Networking - listen for IPv6 connections

rotu:
Previously, attempting to connect to the web interface via IPv6 address would fail. Now it succeeds.

Note this does *not* affect DNS.
- e.g. `http://blueos` will still resolve to the IPv4 address.

This *will* prevent problems with services that use host networking and connect to `localhost`. As mentioned [here](https://github.com/bluerobotics/BlueOS/issues/2467#issuecomment-2339354034), `getent hosts localhost` (resolve using `hosts` file instead of DNS) resolves to the IPv6 address.

Caveats:
- When testing this, use `service nginx reload` NOT `service nginx restart`. In the `blueos-core` image, the config file `nginx.conf` is loaded from `/home/pi/tools/nginx/` instead of `/etc/nginx/`. Using `restart` will cause the server to use the default config


core/start-blueos-core
```diff
@@ -102,7 +102,7 @@ PRIORITY_SERVICES=(
     'autopilot',0,"nice --19 $SERVICES_PATH/ardupilot_manager/main.py"
     'cable_guy',0,"$SERVICES_PATH/cable_guy/main.py"
     'video',0,"nice --19 mavlink-camera-manager --default-settings BlueROVUDP --mavlink tcpout:127.0.0.1:5777 --mavlink-system-id $MAV_SYSTEM_ID --gst-feature-rank omxh264enc=0,v4l2h264enc=250,x264enc=260 --log-path /var/logs/blueos/services/mavlink-camera-manager --verbose"
-    'mavlink2rest',0,"mavlink2rest --connect=udpout:127.0.0.1:14001 --server 0.0.0.0:6040 --system-id $MAV_SYSTEM_ID --component-id $MAV_COMPONENT_ID_ONBOARD_COMPUTER4"
+    'mavlink2rest',0,"mavlink2rest --connect=udpin:127.0.0.1:14000 --server '[::]:6040' --system-id $MAV_SYSTEM_ID --component-id $MAV_COMPONENT_ID_ONBOARD_COMPUTER4"
```
patrickelectric, line 105: This udpin change appears to be something that showed up from your merge.
Please, do a `git rebase origin/master` to fix this.

rotu: Fixed and rebased.

patrickelectric (approved):

PR #2895, opened 2024-09-11
ArdupilotManager: fix loading sitl frame

Williangalvani:
I don't know why, but `self.settings.sitl_frame` and `self.configuration["sitl_frame"]` are not matching.

I don't know why that is happening, but I have already lost 2 hours on this and I got more important stuff to work on.

This fixes the issue where the SITL frame saved in the settings is ignored and overwritten.


rafaellehmkuhl (approved):
Fixed the problem here!

PR #2905, opened 2024-09-18
Frontend: add more checks

Williangalvani:
needs testing (mostly worried about the calibration section)
we also need to test the build


Williangalvani:
closing in favor of #2907

PR #2911, opened 2024-09-20
Fix service links when using IPv6 addresses

rotu:
Previously, the hostname would get truncated at the first colon.


patrickelectric (approved):

patrickelectric:
I'll wait @Williangalvani to check this as well since he wrote this part of code.

Williangalvani (approved):

PR #2815, opened 2024-07-16
Improve remote access support

Williangalvani:
reviving #1842 as a fallback for remote control

This introduces a new optional `works_in_relative_paths`  to the services metadata.
These will use a relative path (under `/extensionv2/`) to load the extension in the browser.
This allows them to be used seamlessly behind a webserver with ssl.

PR #2913, opened 2024-09-22
Nginx: use prefixes instead of regexes

rotu:
Nginx supports full regex-based rules, but we don't actually need them

1. Switch to prefix-based locations, which are easier to understand and gets rid of unnecessary rewrite rules.
2. This has the effect of also allowing the server to rewrite "Location" and "Refresh" headers, which would otherwise require manually using `proxy_redirect`.
3. This avoids hardcoding the protocol and IP address (which make troubleshooting network issues very confusing!)

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -485,9 +485,8 @@ class Helper:
     def setup_nginx_route(metadata: ServiceMetadata, port: int) -> bool:
         name = metadata.sanitized_name
         text = f"""
-        location /extensionv2/{name} {{
-        rewrite ^/extensionv2/{name}(/|$)(.*)$ /$2 break;
-        proxy_pass http://localhost:{port};
+        location /extensionv2/{name}/ {{
+        proxy_pass :{port}/;
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
--- a/core/services/helper/nginx_parser.py
+++ b/core/services/helper/nginx_parser.py
@@ -8,7 +8,7 @@ def parse_nginx_file(filepath: str) -> dict[int, str]:
         location_block_pattern = re.compile(r"location\s+(?P<location>/[\w-]+)\s*{(?P<block_content>[^}]+)}", re.DOTALL)
 
         # Pattern for extracting proxy_pass directive
-        proxy_pass_pattern = re.compile(r"proxy_pass\s+http://[^:]+:(?P<port>\d+);")
+        proxy_pass_pattern = re.compile(r"proxy_pass\s.*:(?P<port>\d+).*;")
         location_blocks = location_block_pattern.finditer(content)
         result = {}
         for match in location_blocks:
--- a/core/tools/nginx/nginx.conf
+++ b/core/tools/nginx/nginx.conf
@@ -49,161 +49,141 @@ http {
 
         # Endpoint used for backend status checks.
         # It will always return an empty 204 response when online.
-        location /status {
+        location = /status {
             return 204;
         }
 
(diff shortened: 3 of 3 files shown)


rotu:
Note this also causes paths to the root to get normalized to `/` by a 301 redirect as described in the [nginx docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#location), effectively reversing 3b6cd4c.

I think this is a good thing because it prevents accidentally changed behavior between using an API with or without the proxy.

In `http://host:777`, for instance, the relative URL `bar` refers to `http://host:777/bar`. If we remap that to a URL without a trailing slash `http://host/p777`, then the relative URL `bar` will refer to `http://host/bar`. Adding the trailing slash (`http://host/p777/`) makes the relative URL `bar` refer to `http://host/p777/bar`.

Root-relative URLs (e.g. `/bar`) are still a problem, however.

Williangalvani:
The reasoning for this being done in #2815 is that nginx's redirects, when used behind another server (apache on a web server) was redirecting the browser to localhost. perhaps this can be fixed by setting the server with a variable passed from the proxying server...

rotu:
> The reasoning for this being done in #2815 is that nginx's redirects, when used behind another server (apache on a web server) was redirecting the browser to localhost. perhaps this can be fixed by setting the server with a variable passed from the proxying server...

I think the actual issue is *not* the redirect but the fact that the `Location` header is not being rewritten, e.g. by the `proxy_redirect` rule. I think that's done for you automagically when using `proxy_pass` *with* a base path.

Per the [docs](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect) (Confusingly (to me, at least), URI here means the path name):
> In some cases, the part of a request URI to be replaced cannot be determined:
>  ...
> - When the URI is changed inside a proxied location using the [rewrite](https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite) directive, and this same configuration will be used to process a request (break):
> location /name/ {
>     rewrite    /name/([^/]+) /users?name=$1 break;
>     proxy_pass http://127.0.0.1;
> }
> In this case, the URI specified in the directive is ignored and the full changed request URI is passed to the server.

If you disagree with my assessment, what are the URLs and response headers you see in the problematic case (especially between Apache and nginx?)

Williangalvani:
> If you disagree with my assessment, what are the URLs and response headers you see in the problematic case (especially between Apache and nginx?)

I do not disagree, but I need input from @voorloopnul who's on vacation right now.
Is the current situation a blocker for you?

If so, we could revert for a while until he's available again.

patrickelectric (approved):
I think that is safe for us to merge and see how it goes. 
We can revert before stable 1.4 if necessary.

patrickelectric:
@rotu it appears that the PR broke master, can you take a look ?

rotu:
@patrickelectric Taking a look now. What are you looking at in particular?

patrickelectric:
> @patrickelectric Taking a look now. What are you looking at in particular?

Frontend appears to be broken, api is not accessible and bootstrap is putting in factory mode.

rotu:
Turns out it did *not* like the leading colon in URLs.
```
2024/09/24 23:19:17 [emerg] 423#423: invalid URL prefix in /home/pi/tools/nginx/nginx.conf:58
```

I was confused because `service nginx reload` didn't cause the server to barf.

rotu:
Turns out it did *not* like the leading colon in URLs.
```
2024/09/24 23:19:17 [emerg] 423#423: invalid URL prefix in /home/pi/tools/nginx/nginx.conf:58
```

I was confused because `service nginx reload` didn't cause the server to fall over in the same way that an invalid config would with `service nginx restart`.

patrickelectric:
Thanks for fixing it!

PR #2930, opened 2024-09-27
Add MAVLinkServer

patrickelectric:
Working fine with SITL, the problem when connecting with an autopilot is probably a missing stream-request in the router. That should be configured with command line as mavp2p.
https://github.com/bluerobotics/mavlink-server/issues/30

### DIFF
--- a/core/tools/mavlink_server/bootstrap.sh
+++ b/core/tools/mavlink_server/bootstrap.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Immediately exit on errors
+set -e
+
+VERSION="master"
+PROJECT_NAME="mavlink-server"
+REPOSITORY_ORG="bluerobotics"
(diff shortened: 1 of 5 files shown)


Williangalvani:
@patrickelectric can you rebase as master was broken?

patrickelectric:
> @patrickelectric can you rebase as master was broken?

Rebased

core/tools/mavlink_server/bootstrap.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Immediately exit on errors
+set -e
+
+VERSION="master"
```
joaoantoniocardoso, line 6: just a reminder to fixate a version once we validate it

Williangalvani (commented):
looks good and it is working as expected here, but I agree with @joaoantoniocardoso that we should change the version from master

PR #2939, opened 2024-10-02
Fix docker name validation logic

rotu:
1. Previously, valid tags such as 'myRegistry:5000/path/to/image' would be incorrectly rejected as containing a tag. Now private registries are tolerated.
2. Previously, the validation would not work as intended, since it was missing boundary assertions to match the whole input.
3. Previously, the path would permit capital letters.


rotu:
Converting to draft since I have not yet tested this locally.

rotu:
I was running into the same issues as docker has trying to build this locally:

> 135.6   WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLEOFError(8, '[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1006)'))': /simple/python-dateutil/python_dateutil-2.9.0.post0-py2.py3-none-any.whl
> 141.1 ERROR: Could not install packages due to an OSError: HTTPSConnectionPool(host='www.piwheels.org', port=443): Max retries exceeded with url: /simple/python-dateutil/python_dateutil-2.9.0.post0-py2.py3-none-any.whl (Caused by SSLError(SSLEOFError(8, '[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1006)')))

patrickelectric:
Hi @rotu any update on that ? Just wandering if this should be in draft still.

rotu:
@patrickelectric In truth, I haven't been able to successfully build and deploy the blueos-core image on my pi 3B. (does uploading via the version chooser actually work for you?)

patrickelectric:
> @patrickelectric In truth, I haven't been able to successfully build and deploy the blueos-core image on my pi 3B. (does uploading via the version chooser actually work for you?)

It works, but depends of swap and ram usage. I would recommend increasing you swapfile. Docker put everything in memory, the file and the uncompress version before installing. You can also install through the internet using your own GitHub fork as well. Our CI configuration work on our forks, you just need to setup the GitHub env variables.

rotu:
No dice after increasing the swapfile to 2GB. AFAICT, it doesn't load the image to docker (`docker image ls --all` doesn't show the image on my pi) and then it restarts to "factory" version.
[system_logs.zip](https://github.com/user-attachments/files/17440474/system_logs.zip)

rotu:
Okay, I *was* able to load the image by running the docker client locally against the remote docker host on the pi. So I think the "Manual upload" is just borked.

rotu:
Alright. Finally tested this locally and caught a few bugs. I *still* could not get the manual uploader to work.

Anyway, you can test this with a fully qualified docker image name, like: `index.docker.io:443/library/debian-slim`. Or a completely unqualified name like `debian-slim` (though of course neither will work as BlueOS extensions as-is!).

JoaoMario109 (approved):

PR #2941, opened 2024-10-03
Core: AutoPilot Manager: Move from Ardu to Auto names only (Serial PX4 support part 4) 

JoaoMario109:
Depends on #2940

Moves from ArduPilot occurrences to AutoPilot in cases where does not affect like paths


patrickelectric (approved):

patrickelectric:
@JoaoMario109 please rebase

patrickelectric (approved):

PR #2956, opened 2024-10-10
wifi: add stdout to errors

Williangalvani:
Yes this could be somewhat verbose, but verbosity is better than no information

needs testing (letting CI build the image)


Williangalvani:
looks like I missed something
image

PR #2974, opened 2024-10-24
core: tools: mavlink-server: Update to 0.2.0

patrickelectric:


JoaoMario109 (changes requested):
As mentioned to @joaoantoniocardoso, using this version generates around 16 rejected connection logs per second on the old Mav2Rest compatibility endpoint. We believe it’s better to wait for the new version that converts this error to a trace or silences it entirely.

image

joaoantoniocardoso:
Replaced by #2976

PR #2983, opened 2024-11-05
frontend: parameter-table: allow using metadata override

Williangalvani:
This allows integrators to override the parameters metadata with their own. It is specially useful if they changed the firmware version.


JackSkellet:
Can Confirm it working as intended, Parameters that had no Descriptions before now have full descriptions

patrickelectric (approved):

PR #2985, opened 2024-11-07
backend: helper: Allow set source IP on speedtest

JoaoMario109:
* Allow to set a source IP address when getting best server to allow testing internet speed using other interfaces


patrickelectric (approved):

patrickelectric:
We can configure the timeout in future PRs

PR #2986, opened 2024-11-08
frontend: defaultParameterLoader: add trailing slash to filters

Williangalvani:
This fixes the issue of double parameter entries as Navigator64 `includes` "Navigator"


Williangalvani:
we might want to backport this to 1.3

patrickelectric (approved):

PR #2988, opened 2024-11-11
Change DHCP range api and set server as default mode

patrickelectric:


core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -16,7 +16,7 @@ def __init__(
         interface: str,
         ipv4_gateway: IPv4Address,
         subnet_mask: Optional[IPv4Address] = None,
-        ipv4_lease_range: Optional[tuple[IPv4Address, IPv4Address]] = None,
+        lease_range: Union[int, int] = (101, 200),
```
Williangalvani, line 19:
```suggestion
        lease_range: Tuple[int, int] = (101, 200),
```

patrickelectric: nice

patrickelectric:
ping @Williangalvani @JoaoMario109

JoaoMario109 (approved):

goasChris:
Does this mean that blueOS would act as a DHCP server on the default eth0 network? If that is the case, then that worries me a little bit, since we usually have an existing device there that does that already, and having two DHCP servers is usually not a great experience.

Or does it only kick in if it did not already get a DHCP lease from an existing DHCP server?

PR #3001, opened 2024-11-18
Fix commonwealh packages structure and json dumps for modern Pydantic versions.

JoaoMario109:


core/libs/commonwealth/commonwealth/settings/bases/pydantic_base.py
```diff
@@ -99,7 +100,11 @@ def save(self, file_path: pathlib.Path) -> None:
 
         with open(file_path, "w", encoding="utf-8") as settings_file:
             logger.debug(f"Saving settings on: {file_path}")
-            settings_file.write(self.json(indent=4))
+            if pydantic_version > "1.8.0":
```
patrickelectric, line 103: Just to be sure, this type do eq with semver ?

patrickelectric (approved):

PR #3004, opened 2024-11-19
Core: Tools: Zenoh: Add basic Zenoh support

JoaoMario109:


core/tools/zenoh/bootstrap.sh
```diff
@@ -0,0 +1,66 @@
...
+done
+echo "Downloaded all binaries, now installing to $BINARY_PATH"
+
+cd "$DOWNLOAD_FOLDER"
+unzip "*.zip"
+rm -- ./*.zip
+mv -- ./* "$BINARY_PATH"
+cd - && rm -rf "$DOWNLOAD_FOLDER"
+
+echo "Install complete!"
```
joaoantoniocardoso, line 66:
```suggestion

for BINARY in "${BINARIES[@]}"; do
  echo "Installed binary type: $(file "$(which "$BINARY_PATH/$BINARY")")"
done

echo "Finished installing $PROJECT_NAME"
```

JoaoMario109: Here when building with the suggested code I'm getting this

```
#16 24.88 Installed binary type: cannot open `' (No such file or directory)
#16 24.88 Installed binary type: cannot open `' (No such file or directory)
#16 24.88 Installed binary type: cannot open `' (No such file or directory)
#16 24.88 Finished installing zenoh
```

joaoantoniocardoso: I see, just make it static, then:
```bash
cd "$DOWNLOAD_FOLDER"
unzip "*.zip"
rm -- ./*.zip
INSTALLED_FILES=(*)
mv -- ./* "$BINARY_PATH"
cd - && rm -rf "$DOWNLOAD_FOLDER"

for BINARY in "${INSTALLED_FILES[@]}"; do
  echo "Installed binary type: $(file "$(which "$BINARY_PATH/$BINARY")")"
done

echo "Finished installing $PROJECT_NAME"
```

core/tools/zenoh/bootstrap.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Step 2: Prepare the installation and tools paths
+
+if [ -n "$VIRTUAL_ENV" ]; then
+    BIN_DIR="$VIRTUAL_ENV/bin"
+else
+    BIN_DIR="/usr/bin"
+fi
+
+BINARY_PATH="$BIN_DIR/$PROJECT_NAME"
```
joaoantoniocardoso, line 44: I think we actually want this instead:

```suggestion
BINARY_PATH="$BIN_DIR/"
```

otherwise we'd have to add that to the PATH

joaoantoniocardoso (approved):

PR #3015, opened 2024-11-28
blueos_startup_update: only create dns link on Bullseye

Williangalvani:
This is not working on Bookworm, and it is causing unnecessary reboots.
This change makes it happen on Bullseye only. On bookworm we should probaby use NetworkManager to configure dns isntead of doing the resolv.conf thing

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -460,7 +460,6 @@ def main() -> int:
         update_startup,
         ensure_user_data_structure_is_in_place,
         ensure_nginx_permissions,
-        create_dns_conf_host_link,
         fix_ssh_ownership,
     ]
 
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso:
I don't know. I feel we should configure NetworkWanager to leave `resolv.conf` managed by the user. I had so many problems with services trying to manage `resolv.conf` and creating all sorts of issues.

Williangalvani:
The issue with the current implementation is that it (understandbly) binds the file instead of all of /etc. But on Bookworm, dns changes are applied by (I think) NetworkManager erasing the file and creating a new one, isntead of editing it.
This breaks the links we are relying on

joaoantoniocardoso:
> The issue with the current implementation is that it (understandbly) binds the file instead of all of /etc. But on Bookworm, dns changes are applied by (I think) NetworkManager erasing the file and creating a new one, isntead of editing it. This breaks the links we are relying on

ok, but why is the NetworkManager trying to change `resolv.conf`?

Williangalvani:
> ok, but why is the NetworkManager trying to change `resolv.conf`?

good question. probably to use the router's advertised dns? I'll try configuring to not do it

Williangalvani:
if we do disable networkmanager's handling dns, we shouild probably set one by ourselves, manually...

joaoantoniocardoso:
> if we do disable networkmanager's handling of dns, we shouild probably set one by ourselves, manually...

Yeah. It should be initialized with a copy of the host dns, but blueos could give options in the wizard.

PR #3020, opened 2024-12-04
core: helper: Only abort if server returns bad request

patrickelectric:

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -317,7 +317,7 @@ class Helper:
             "127.0.0.1", port=port, path="/", timeout=1.0, method="GET", follow_redirects=10
         )
         log_msg = f"Detecting service at port {port}"
-        if response.status != http.client.OK:
+        if response.status == http.client.BAD_REQUEST:
             # If not valid web server, documentation will not be available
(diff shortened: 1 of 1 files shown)


core/services/helper/main.py
```diff
@@ -317,7 +317,7 @@ def detect_service(port: int) -> ServiceInfo:
             "127.0.0.1", port=port, path="/", timeout=1.0, method="GET", follow_redirects=10
         )
         log_msg = f"Detecting service at port {port}"
-        if response.status != http.client.OK:
+        if response.status == http.client.BAD_REQUEST:
```
joaoantoniocardoso, line 320: could you explain the reason behind this?

joaoantoniocardoso: shouldn't the servers give a 200 when valid?

patrickelectric: In the commit body:
> Sometimes it can return a redirect or any other, register service will take care for us

patrickelectric: My case was moving openvscode to vscode where the / was not an OK but something

joaoantoniocardoso: okay, it looks prone to false positives, but it might be better than false negatives.

PR #3025, opened 2024-12-09
Add support to more (all?) ardupilot serial boards

Williangalvani:
This should add support to all ardupilot boards

TODO:
  - [ ] generate boards.json on CI


core/services/ardupilot_manager/firmware/FirmwareInstall.py
```diff
@@ -16,13 +16,355 @@
...
+        platform = board.name
+    else:
+        platform = board
     ardupilot_board_ids = {
         Platform.Pixhawk1: 9,
         Platform.Pixhawk4: 50,
         Platform.Pixhawk6X: 53,
         Platform.Pixhawk6C: 56,
         Platform.CubeOrange: 140,
+        "PX4_FMU_V1": 5,
```
patrickelectric, line 31: We should refactor typedefs.Platform no ? It would be nice to unify both implementation

Williangalvani: We can, yes. I was trying to not touch stuff that currently works.
We could leave this PR for after 1.4 and then I can rework it to unify everything.

core/services/ardupilot_manager/flight_controller_detector/metadata_preprocessor.py
```diff
@@ -0,0 +1,197 @@
...
+        if self.load_existing_data(output_file):
+            print(f"Using existing data from {output_file}")
+            # Print summary of existing data
+            device_count = len(self._usb_devices)
+            platform_count = sum(len(platforms) for platforms in self._usb_devices.values())
+            print(f"\nFound {device_count} unique USB IDs")
+            print(f"Total platform mappings: {platform_count}")
+            return
+
+        # If file is invalid or too old, download and process new data
```
Williangalvani, line 166: it doesnt look like we are checking the age of the file anywhere

Williangalvani: it is at `is_file_valid()`

PR #3028, opened 2024-12-10
Helper: add support for cockpit widgets

Williangalvani:
related to https://github.com/bluerobotics/ping-viewer-next/pull/46
and https://github.com/bluerobotics/cockpit/pull/1499

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -107,6 +107,7 @@ class ServiceMetadata(BaseModel):
     api: str
     sanitized_name: Optional[str]
     works_in_relative_paths: Optional[bool]
+    extras: Optional[Dict[str, str]]
 
 
 class ServiceInfo(BaseModel):


core/services/helper/main.py
```diff
@@ -93,6 +93,11 @@ class WebsiteStatus(BaseModel):
     error: Optional[str] = None
 
 
+class CockpitWidget(BaseModel):
+    name: str
+    url: str
+
```
joaoantoniocardoso, line 99: Thinking on compatibility, it should include a widget version, and a cockpit target version as well

Williangalvani: yeah, this is the absolutely minimum. I'm thinking of how to expand things without messing up things too much.
would it be an issue if we we had a complex json in here?
should we instead just use this as an aggregator and have it link to another json on the extension?
then we could also manually input that json into cockpit

Williangalvani:
I added more info about it at https://github.com/bluerobotics/cockpit/pull/1499

joaoantoniocardoso (approved):

core/services/helper/main.py
```diff
@@ -107,6 +107,7 @@ class ServiceMetadata(BaseModel):
     api: str
     sanitized_name: Optional[str]
     works_in_relative_paths: Optional[bool]
+    extras: Optional[Dict[str, str]]
```
joaoantoniocardoso, line 110: maybe we should name it `integrations`

Williangalvani: I think that limits the usage. I don't want to tell people what they can or can't add in these jsons. not sure if that makes sense.

joaoantoniocardoso: I see!

PR #3046, opened 2025-01-10
frontend: disable wifi-updater and let user know if wifi service is disabled

Williangalvani:
makes it clear to the user when the service got disabled. Also points into the direction to enable it.
More than that, it stops trying to make a bunch of requests (wifi-updater) when the service is not even running.


core/frontend/src/components/wifi/WifiTrayMenu.vue
```diff
@@ -48,6 +83,11 @@ export default Vue.extend({
       return wifi_strenght_icon(wifi.current_network.signal)
     },
   },
+  mounted() {
+    commander.getEnvironmentVariables().then((environment_variables) => {
+      this.disabled_services = (environment_variables?.BLUEOS_DISABLE_SERVICES as string).split(',') as string[]
```
JoaoMario109, line 88:
```suggestion
      this.disabled_services = (environment_variables?.BLUEOS_DISABLE_SERVICES as string ?? '').split(',') as string[]
```

core/start-blueos-core
```diff
@@ -21,6 +21,9 @@ RUN_AS_REGULAR_USER="sudo -u blueos"
 # Set BlueOS log folder
 BLUEOS_LOG_FOLDER_PATH="/var/logs/blueos"
 
+# Replace spaces with commas in the disabled services list
+BLUEOS_DISABLE_SERVICES=${BLUEOS_DISABLE_SERVICES// /,}
```
patrickelectric, line 25: why the replace is necessary here ? the list should be comma separated already

patrickelectric (approved):

PR #3048, opened 2025-01-10
Unified settings

### LINKED ISSUE #3047: We should unify our settings sources (bag/localstorate)
We currently have some older settings that get saved to the browser 's localStorage. We should probably migrate everything to the bag instead.

There will be a slight change of behavior, as the settings will now carry over between different machines, but I believe that is for the best.

Williangalvani:
fix #3047

### DIFF
--- a/core/frontend/src/store/settings.ts
+++ b/core/frontend/src/store/settings.ts
@@ -6,6 +6,7 @@ import {
 import store from '@/store'
 import beacon from '@/store/beacon'
 import { castString } from '@/utils/helper_functions'
+import bag from '@/store/bag'
 
 @Module({
   dynamic: true,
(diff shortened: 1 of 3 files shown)


core/frontend/src/store/settings.ts
```diff
@@ -74,70 +75,48 @@ class SettingsStore extends VuexModule {
...
-   * Save a variable on the settings system
-   * @param string name
-   * @param T value
-   */
-  static saveVariable<T>(name: string, value: T): void {
-    // eslint-disable-next-line
-    window.localStorage.setItem(SettingsStore.settingsName(name), JSON.stringify(value))
+  static async load(): Promise<void> {
+    const data = await bag.getData('settings')
+    if (data) {
```
patrickelectric, line 93: just do an early return

PR #3049, opened 2025-01-10
bluoes_startup_update: add fine-grained patch suppression

### LINKED ISSUE #3035: core: bootstrap: reduce aggressiveness of blueos_startup_update.py in overwriting custom pin assignments in Pi overlay
### Current behaviour

after editing config.txt (aka Pi overlay) with new i2c-3 assignments and reduced uart assignments, including the #custom tag,  a reboot shows that the serial assignment was added.   This causes a dmesg reported GPIO error.

It would be nice to be able to aggressively change GPIO assignments in the overlay file.  I.e. to throttle the list of desired items in the variable navigator_configs_with_match_patterns    
A flag file in the system to modify the aggressiveness of this update script would be helpful.   Looking at the calling script, it seems that there are many activities that are still needed, but the environment variable BLUEOS_DISABLE_STARTUP_UPDATE  will prevent the entire script from running.   Therefore, a more granular option is desired. 

### Prerequisites

Williangalvani:
tested by running `BLUEOS_DISABLE_PATCHES=userdata,cgroups,nginx,dwc2,ssh,i2c4,dns blueos_startup_update.py`

fix #3035

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -48,6 +48,8 @@ CONFIG_USER_PROTECTION_WORD = "custom"
 config_file = None
 cmdline_file = None
 
+disabled_patches = [entry.strip() for entry in os.getenv("BLUEOS_DISABLE_PATCHES", "").split(",")]
+
 # Copyright 2016-2022 Paul Durivage
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

drzwey:
Great, thanks!   I'll check it out.

ES-Alexander:
Would be good to mention this functionality in the dev docs :-)

PR #3050, opened 2025-01-10
core: move setup_ssh() over to startup_blueos_update

### LINKED ISSUE #3038: bookworm: first boot can fail to detect the Navigator board
I suspect something is wrong with the overlays, that gets fixed by blueos_startup_update
[_blueos_startup_update.zip](https://github.com/user-attachments/files/18215063/_blueos_startup_update.zip)

Williangalvani:
~not~tested.

The issue here is that the ssh setup was being done _after_ blueos_startup_update, when commander started up.
This meant it would always fail on the first boot, to succeed later on after commander ran at least once.

fix #3038

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -5,6 +5,8 @@ import logging
 import os
 import re
 import time
+import subprocess
+from pathlib import Path
 from typing import List, Tuple
 import configparser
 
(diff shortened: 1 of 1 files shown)


dougsland:
looks sane to me. LGTM

Williangalvani:
just re-tested by deleting the content of `/home/pi/.ssh/authorized_keys` on the host, and the keys on the container at 
`/root/.config/.ssh` and rebooting. everything seems in order.

Williangalvani:
rebased

joaoantoniocardoso (approved):

PR #3074, opened 2025-01-17
Add isBackendOffline

### LINKED ISSUE #3073: 25% of our errors in sentry are Network Error
Network Error is related to the network and not BlueOS, being a point that the surface computer does not have access to the vehicle (maybe because is changing network) or the vehicle is not online at all.
We should deal with this case as the vehicle is offline.

patrickelectric:
Fix #3073

### DIFF
--- a/core/frontend/src/utils/api.ts
+++ b/core/frontend/src/utils/api.ts
@@ -6,8 +6,13 @@ const backend_offline_error = new Error('Backend is offline')
 backend_offline_error.name = 'BackendOffline'
 export { backend_offline_error }
 
-const axios_backend_instance: AxiosInstance = axios.create()
+export const isBackendOffline = (error: any): boolean => {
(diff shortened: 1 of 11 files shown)


core/frontend/src/utils/api.ts
```diff
@@ -6,8 +6,13 @@ const backend_offline_error = new Error('Backend is offline')
 backend_offline_error.name = 'BackendOffline'
 export { backend_offline_error }
 
-const axios_backend_instance: AxiosInstance = axios.create()
+export const isBackendOffline = (error: any): boolean => {
```
Williangalvani, line 9: I just woke up... but is this not causing an infinite recursion?

patrickelectric: wrong global replace!

Williangalvani (approved):

PR #3077, opened 2025-01-17
frontend: Add wifi animation when doing actions

patrickelectric:
https://github.com/user-attachments/assets/4d3e1336-3ea8-4176-8a4d-3308b31f776a


JoaoMario109 (approved):

JoaoMario109:
Just to leave a note to migrate the /connect and /remove to WiFi store in future so we can keep centralized the control of states like is_loading.

PR #3082, opened 2025-01-20
ardupilot_manager: use new numeric --serialN ports instead of confusing (and deprecated) --uartX

### LINKED ISSUE #2522: ArdupilotManager: we should try using --serialN instead of --uartX for assigning serial ports
see https://github.com/ArduPilot/ardupilot/pull/26156

--uartX has been deprecated for a while, but ArduSub stable still requires it. I propose we try to use --serialN, and if that is an unknown argument, we switch over to --serialN.

Williangalvani:
fix #2522

### DIFF
--- a/core/frontend/src/components/autopilot/AutopilotSerialConfiguration.vue
+++ b/core/frontend/src/components/autopilot/AutopilotSerialConfiguration.vue
@@ -3,11 +3,11 @@
     class="v-card v-sheet theme--light mt-5"
   >
     <v-combobox
-      v-for="port in Object.keys(port_map)"
+      v-for="port in range(1, 9)"
       :key="port"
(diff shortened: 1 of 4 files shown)


joaoantoniocardoso (commented):
@Williangalvani the code looks okay to me. Could you rebase this so I can try it on the top of the current master?

patrickelectric:
ping @joaoantoniocardoso

joaoantoniocardoso (commented):
This patch does work; however, it breaks both the API and settings backward compatibility.

@patrickelectric and @Williangalvani what do you think?

patrickelectric:
if it's just a "set"  we can handle that in the endpoint and accept both

Issue #3090, opened 2025-01-24
bug: No network configuration on first boot


patrickelectric:
### Bug description

There is no network configuration for eth0 and usb0 on the first boot

### Steps to reproduce

Flash 1.4.0-beta.10
Wait for led to start blinking
It's not possible to connect to the vehicle

### Primary pain point(s)

First use is painful 

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

PR #3091, opened 2025-01-24
core: commonwealth: DHCPServerManager: Try to start DHCP server 3 times

patrickelectric:
No idea how to test it in a fresh image

Helps #3090

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
+++ b/core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
@@ -101,16 +101,23 @@ class Dnsmasq:
         ]
 
     def start(self) -> None:
-        try:
-            # pylint: disable=consider-using-with
-            self._subprocess = subprocess.Popen(self.command_list(), shell=False, encoding="utf-8", errors="ignore")
-            time.sleep(3)
-            if not self.is_running():
+        attempts = 3
+        for attempt in range(attempts):
+            try:
+                # pylint: disable=consider-using-with
+                self._subprocess = subprocess.Popen(self.command_list(), shell=False, encoding="utf-8", errors="ignore")
+                time.sleep(3)
+                if self.is_running():
+                    logger.info("DHCP Server started.")
+                    return
+
                 exit_code = self._subprocess.returncode
-                raise RuntimeError(f"Failed to initialize Dnsmasq ({exit_code}).")
-            logger.info("DHCP Server started.")
-        except Exception as error:
-            raise RuntimeError("Unable to start DHCP Server.") from error
+                logger.error(f"Attempt {attempt} failed to initialize Dnsmasq ({exit_code}).")
+            except Exception as error:
+                logger.error(f"Attempt {attempt} encountered an error: {error}")
+                time.sleep(1)
+
+        raise RuntimeError("Unable to start DHCP Server")
 
     def stop(self) -> None:
         if self.is_running():


core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -101,16 +101,23 @@ def command_list(self) -> List[Union[str, pathlib.Path]]:
         ]
 
     def start(self) -> None:
```
dougsland, line 103: Just starting follow the project, congrats guys first of all. I am wondering if the attempts and others params could be loaded from config file and if not setup the config file or exists use the default value.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -101,16 +101,23 @@ def command_list(self) -> List[Union[str, pathlib.Path]]:
...
-            # pylint: disable=consider-using-with
-            self._subprocess = subprocess.Popen(self.command_list(), shell=False, encoding="utf-8", errors="ignore")
-            time.sleep(3)
-            if not self.is_running():
+        attempts = 3
+        for attempt in range(attempts):
+            try:
+                # pylint: disable=consider-using-with
+                self._subprocess = subprocess.Popen(self.command_list(), shell=False, encoding="utf-8", errors="ignore")
+                time.sleep(3)
```
dougsland, line 109: how about create a method for this code ? Also both args can be from config file so users can define ?  
Some code here as example for the idea.

```
import configparser

class DHCPServer:
     CONFIG_PATH = "/etc/blueos/dhcp-server-manager.conf" 

    def __init__(self, command_list, config=None):
        <snip>
        self.config = self.load_config()  # Load configuration from the system file
        self.timeout = float(self.config.get("timeout", 5))  # Default to 5 seconds
        self.interval = float(self.config.get("interval", 0.5))  # Default to 0.5 seconds
        self.attempts = int(self.config.get("attempts", 3))  # Default to 3 attempts
 
       <snip>

   @classmethod
    def load_config(cls):
        """
        Load configuration from the system-wide config file.
        Returns a dictionary of configuration values.
        """
        config = configparser.ConfigParser()
        if os.path.exists(cls.CONFIG_PATH):
            try:
                config.read(cls.CONFIG_PATH)
                logger.info(f"Loaded configuration from {cls.CONFIG_PATH}")
                return config["DHCPServer"]  # Return the specific section for DHCPServer
            except Exception as e:
                logger.error(f"Error reading configuration file: {e}")
        else:
            logger.warning(f"Configuration file not found: {cls.CONFIG_PATH}")
        return {}


      def is_running(self):
        return self._subprocess and self._subprocess.poll() is None

    def wait_for_startup(self):
        """Wait for the server to start up within the configured timeout."""
        max_checks = int(self.timeout / self.interval)
        for _ in range(max_checks):
            if self.is_running():
                logger.info("DHCP Server started successfully.")
                return True
            time.sleep(self.interval)
        logger.warning("DHCP Server did not start within the expected timeout.")
        return False

    def start(self):
        for attempt in range(self.attempts):
            try:
                logger.info(f"Starting DHCP Server, attempt {attempt + 1}/{self.attempts}...")
                with subprocess.Popen(
                    self.command_list(),
                    shell=False,
                    encoding="utf-8",
                    errors="ignore"
                ) as process:
                    self._subprocess = process
                    if self.wait_for_startup():
                        return
            except Exception as e:
                logger.error(f"Attempt {attempt + 1} failed with error: {e}")
        
        logger.error("Failed to start DHCP Server after multiple attempts.")
        raise RuntimeError("Unable to start DHCP Server.")
  
```


Format: /etc/blueos/dhcp-server-manager.conf
```
[DHCPServer]
timeout = 10
interval = 1
attempts = 5
```

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
patrickelectric: Thanks, do you have any project running BlueOS or any idea of how to use it ? You can reach us via discord or bluerobotics forum if you wish.

patrickelectric: The default configuration are loaded via the main file, each project has its own settings.

core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
patrickelectric: Thanks for the suggestion, but the code is simpler as it is.

dougsland: I wouldn't put static `attempts` and `sleep` in the code because it's simpler.  I am just robotic and open source engineer that likes to explore new projects (like this one), probably acquire a bluerov or vessel to execute real tests.  Whats the discord you guys use?

patrickelectric: I just replied you on the raised issue: https://github.com/bluerobotics/BlueOS/issues/3104

patrickelectric:
Closing since it appears unnecessary anymore

PR #3098, opened 2025-01-24
github: workflows: test-and-deploy: Add a suffix to pi4 and pi5

patrickelectric:


.github/workflows/test-and-deploy.yml
```diff
@@ -331,12 +331,21 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+      - name: Set asset name
+        run: |
+          if [[ ${{ matrix.runner }} == 'pi5-builder' ]]; then
+            echo "ASSET_NAME_SUFFIX=-pi5" >> $GITHUB_ENV
+          elif [[ ${{ matrix.os }} == 'bullseye' ]]; then
```
dougsland, line 338: same here: `elif [[ ${{ matrix.os }} == *'bullseye'* ]]; then` ?

.github/workflows/test-and-deploy.yml
```diff
@@ -331,12 +331,21 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+      - name: Set asset name
+        run: |
+          if [[ ${{ matrix.runner }} == 'pi5-builder' ]]; then
```
dougsland, line 336: should use contain for safety this bash operation? `elif [[ ${{ matrix.os }} == *'pi5-builder'* ]];` then ?

patrickelectric: It's ok to use the machine name as it's.

.github/workflows/test-and-deploy.yml
patrickelectric: The OS matrix name os bullseye, there is no need to regex.

dougsland: Sure

PR #3103, opened 2025-01-27
Overwrite /etc/resolv.conf.host if it exists

rotu:
This prevents the install script from crashing if run on an existing blueos machine.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -253,7 +253,7 @@ echo "Restarting random-seeds."
 rm -rf /var/lib/systemd/random-seed /loader/random-seed
 
 echo "creating dns link"
-sudo ln /etc/resolv.conf /etc/resolv.conf.host
+sudo ln --force /etc/resolv.conf /etc/resolv.conf.host
 
 echo "Installation finished successfully."
 echo "You can access after the reboot:"


install/install.sh
```diff
@@ -253,7 +253,7 @@ echo "Restarting random-seeds."
 rm -rf /var/lib/systemd/random-seed /loader/random-seed
 
 echo "creating dns link"
-sudo ln /etc/resolv.conf /etc/resolv.conf.host
+sudo ln --force /etc/resolv.conf /etc/resolv.conf.host
```
dougsland, line 256: How about create a function to execute such operation and also check if the file exists or not and use --force? 

# Function to create a symbolic link with optional --force
create_symlink() {
    local src=$1
    local dst=$2
    local options=""

    # Check if the destination file exists
    if [ -e "$dst" ]; then
        echo "Destination $dst exists. Adding --force option."
        options="--force"
    else
        echo "Destination $dst does not exist. Creating link without --force."
    fi

    # Create the symbolic link
    ln $options "$src" "$dst"

    # Check if the operation was successful
    if [ $? -eq 0 ]; then
        echo "Successfully linked $src to $dst"
    else
        echo "Failed to link $src to $dst"
        return 1
    fi
}

create_symlink "/etc/resolv.conf" "/etc/resolv.conf.host"

rotu: I don't understand. What benefit does that have over my suggestion?

dougsland: logging and only use force if required but it's just a suggestion.

patrickelectric: @dougsland, thanks for the suggestion, but it appears to add unnecessary logic/code for such simple change.

patrickelectric (approved):

install/install.sh
rotu: Thanks. The `--force` flag here only serves to overwrite the file if it exists, so it’s not like using it only if the file exists is making it at all safer.

some other thoughts:

1. I’m not totally sure why we create this file on the host system; it seems in the docker image we only need to *read* the file, and there’s no need to pollute the *host* filesystem with an extra file. So it’s possible that we shouldn’t be doing this step at all.
2. If we wanted to change how we move/link files, I would look to the standard `install` utility which also provides options to set permissions, rename files instead of clobber them, and not modify files if they match. https://linux.die.net/man/1/install

PR #3108, opened 2025-01-28
frontend: BrIframe: Allow extensions to perform operations with the clipboard

rafaellehmkuhl:
Today, extensions do not have allowance to read from or write to the clipboard.


rafaellehmkuhl:
@ES-Alexander could you test that Cockpit PR with this version of BlueOS?

rafaellehmkuhl:
As noticed by @JoaoMario109, this PR will not work since the clipboard API requires secure context. Closing.

PR #3109, opened 2025-01-28
VersionChooser: Allow login for docker daemon by version chooser interface

JoaoMario109:
Allows users to perform a Docker login to the default Docker registry or custom registries using their Docker username and password or a Personal Access Token (PAT).

Effectively allows for the following:
* Mirrors the functionality of the `docker login` command.
* Saves credentials to `.docker/config.json` for both the normal user and the root user (if specified).
* Affects the entire Docker daemon context, enabling all BlueOS services to fetch images or other resources from the authenticated registry.

To access users must be in `pirate mode` on the version chooser page:
image

---

By default, the dialog logs in using only the normal Raspberry Pi user (not root) and the default Docker registry (`https://index.docker.io/v1/`). 

If needed, users can:
- Authenticate to a custom registry by enabling the **Custom Registry** switch.
- Use the authentication for `sudo docker` commands by enabling the **Login as Root User** switch. 

image

Connected accounts can be seen at the accounts tab. Ones with the `shield` icon means that the credentials are valid for `sudo docker` commands.

image

Validation:

* Trying to pull private repository without login:
image_2025-01-28_15-53-12
* Trying to pull private repository after performing login:
image

---

NOTE: This PR does not includes the *on login* authentication check, and does not allows users to test their credentials to check if they still valid. A following PR will add the check, probably matching the [auth](https://docs.docker.com/reference/api/engine/version/v1.39/#tag/System/operation/SystemAuth) endpoint docs.

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -31,6 +31,8 @@ DELTA_JSON = {
             "/etc/machine-id": {"bind": "/etc/machine-id", "mode": "ro"},
             "/etc/resolv.conf.host": {"bind": "/etc/resolv.conf.host", "mode": "ro"},
             "/home/pi/.ssh": {"bind": "/home/pi/.ssh", "mode": "rw"},
+            "/home/pi/.docker": {"bind": "/home/pi/.docker", "mode": "rw"},
+            "/root/.docker": {"bind": "/root/.docker", "mode": "rw"},
             "/run/udev": {"bind": "/run/udev", "mode": "ro"},
             "/sys/": {"bind": "/sys/", "mode": "rw"},
             "/usr/blueos/bin": {"bind": "/usr/blueos/bin", "mode": "rw"},
--- a/bootstrap/startup.json.default
+++ b/bootstrap/startup.json.default
@@ -38,6 +38,14 @@
                 "bind": "/home/pi/.ssh",
                 "mode": "rw"
             },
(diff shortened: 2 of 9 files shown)


core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -31,6 +31,8 @@
             "/etc/machine-id": {"bind": "/etc/machine-id", "mode": "ro"},
             "/etc/resolv.conf.host": {"bind": "/etc/resolv.conf.host", "mode": "ro"},
             "/home/pi/.ssh": {"bind": "/home/pi/.ssh", "mode": "rw"},
+            "/home/pi/.docker/config.json": {"bind": "/home/pi/.docker/config.json", "mode": "rw"},
+            "/root/.docker/config.json": {"bind": "/root/.docker/config.json", "mode": "rw"},
```
patrickelectric, line 35: Since there is no guarantee of such files to exist, this should bind the folder not the file. If the file does not exist, docker will create a folder with the name config.json.

patrickelectric (changes requested):

core/tools/blueos_startup_update/blueos_startup_update.py
JoaoMario109: Changed to bind the folder!

patrickelectric (approved):
really nice!

goasChris:
amazing!! This saves my butt when you have a whole swarm of robots on the same outgoing IP ;) thank you very much!

PR #3115, opened 2025-01-30
use lower() when comparing lower chars

dougsland:
Just checking some code, better use good practices.


dougsland:
@patrickelectric is there a ./build-locally to test such changes locally? Still learning the process to contribute.

patrickelectric:
> @patrickelectric is there a ./build-locally to test such changes locally? Still learning the process to contribute.

We usually run the docker build command or let CI build for us, if you configure the docker CI username and password secrets. 

We have a small wiki for development here: https://github.com/bluerobotics/BlueOS/wiki/Development

PR #3118, opened 2025-01-31
core:tools:blueos_startup_update: No IPV6 allowed

### LINKED ISSUE #3117: Default endpoint for fake stream is invalid
Seems like it's trying to use an IPv6.

Image

JoaoMario109:
Make sure IPV6 disabled entries are present and are not commented from `/etc/sysctl.conf`

Closes #3117


rafaellehmkuhl:
Worked here on a Pi3B+, without increasing the boot time for reboots other than the first.

dougsland:
howdy @JoaoMario109 

Question:

No plans to support IPV6 in the future? 
Why not just ignore in the UI if the IPV6 information is available (even if it's not used yet in the system yet)? Isn't simple approach and make sure IPV6 is supported in the (near) future? 

Example for UI: A toggle button in the UI "show IPV6 interfaces information" off by default ("work in progress").

patrickelectric (approved):

Issue #1818, opened 2023-06-21
vehicle-setup: allow configuring PWM/motor output parameters


Williangalvani:
- [x] motors test should allow reversing
- [x] changing trim/min/max

Williangalvani:
Also, the motor test sliders do not take the reverse parameter into consideration. Should it?

ES-Alexander:
I think these things should actually be added to the table, since they're also relevant to the other PWM peripherals (e.g. I just had to change the min/max for my camera tilt servo manually via the parameters, to get some extra rotation range).

PR #3122, opened 2025-02-04
ServoFunctionEditorDialog: add sliders for min/trim/max

Williangalvani:
image

## Summary by Sourcery

New Features:
- Added interactive sliders for controlling servo PWM ranges (min, trim, max).


patrickelectric (changes requested):
image

It's possible to set values outside the limit

Williangalvani:
> It's possible to set values outside the limit

that was on purpose so users could stull use this UI if they wanted to use a wider range. But I've personally never needed to go beyond 800-2200 anyway. I updated the range to 800-2200.

ES-Alexander:
Completes #1818

PR #3132, opened 2025-02-07
wifi: networkmanager: disable ipv6

Williangalvani:
## Summary by Sourcery

New Features:
- Disabled IPv6 for new NetworkManager connections.

### DIFF
--- a/core/services/wifi/wifi_handlers/networkmanager/networkmanager.py
+++ b/core/services/wifi/wifi_handlers/networkmanager/networkmanager.py
@@ -4,7 +4,7 @@ import select
 import signal
 import subprocess
 from concurrent.futures import CancelledError
-from typing import Any, List, Optional
+from typing import Any, List, MutableMapping, Optional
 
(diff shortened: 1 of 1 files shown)


JoaoMario109 (approved):

patrickelectric:
It's failing on the python tests, is that expected ?

JoaoMario109:
> It's failing on the python tests, is that expected ?

Probably will need to suppress it. This modification was tested before so we know it works, but seems that mypy is not liking it.

JoaoMario109:
Maybe adding a better type hint can make mypy ok with it.

```suggestion
connection: MutableMapping[str, Any] = {
```

Williangalvani:
> Maybe adding a better type hint can make mypy ok with it.
> 
> ```
> connection: MutableMapping[str, Any] = {
> ```

that seemed to do it, thanks!

patrickelectric (approved):

PR #3134, opened 2025-02-08
wip: Service manager

Williangalvani:
needs some cleanup
image


Williangalvani:
@ES-Alexander mentioned this should probably be on "available services" instead. that does mean we need to make sure the "tmux" names of the services can be properly mapped to the available services entries

Williangalvani:
we should probably just switch to the "multiple-containers using the same image" approach isntead of these hacks

PR #3154, opened 2025-02-19
Add "Backup DHCP Server" option to only enable the server if another one isnt found at startup

### LINKED ISSUE #3143: Enable DHCP server only when there is no other in the network
> Does this mean that blueOS would act as a DHCP server on the default eth0 network? If that is the case, then that worries me a little bit, since we usually have an existing device there that does that already, and having two DHCP servers is usually not a great experience.
Or does it only kick in if it did not already get a DHCP lease from an existing DHCP server?

_Originally posted by @goasChris in https://github.com/bluerobotics/BlueOS/issues/2988#issuecomment-2653363256_

Williangalvani:
fix #3143 

needs testing

## Summary by Sourcery

Adds a "Backup DHCP Server" option that only enables the DHCP server if another one isn't found at startup. This prevents conflicts when multiple DHCP servers are running on the same network.

New Features:
- Adds a "Backup DHCP Server" option to the DHCP server configuration, which only enables the server if no other DHCP server is detected on the network at startup.
- Adds a checkbox in the frontend to enable the "Backup DHCP Server" option.
- Adds a new AddressMode for "backupServer" to distinguish backup DHCP servers from regular ones.

Tests:
- Adds nmap to the list of system tools to install, which is used to detect other DHCP servers on the network.

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
+++ b/core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+import asyncio
+import re
+import sys
+from typing import List
+
+from loguru import logger
+
+
+class DHCPDiscoveryError(Exception):
+    """Base exception for DHCP discovery errors"""
+
+
+async def discover_dhcp_servers(iface: str, timeout: int = 15) -> List[str]:
+    """
+    Discover DHCP servers on the network using nmap's DHCP discovery script
+
+    Args:
+        interface: Network interface to use
+        timeout: Time to wait for responses in seconds
+
+    Returns:
+        List of DHCP server IP addresses found
+
+    Raises:
+        DHCPDiscoveryError: If discovery fails
+    """
+    try:
+        # Run nmap with DHCP discovery script
+        cmd = [
+            "sudo",  # Need root privileges
+            "nmap",
+            "--script",
+            "broadcast-dhcp-discover",
+            "-e",
+            iface,  # Specify interface
+            "--script-timeout",
+            f"{timeout}s",
+        ]
+
+        # Run nmap asynchronously
+        logger.info(f"Running nmap command: {' '.join(cmd)}")
+        process = await asyncio.create_subprocess_exec(
+            *cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+        )
+
+        stdout, stderr = await process.communicate()
+        output = stdout.decode()
+        error = stderr.decode()
+
+        if process.returncode != 0:
+            logger.info(f"nmap output: {output}")
+            logger.info(f"nmap error: {error}")
+            raise DHCPDiscoveryError(f"nmap failed: {error}")
+
+        # Parse nmap output for DHCP servers
+        servers = []
+
+        # Look for Server Identifier or DHCP Server lines
+        server_pattern = r"Server(?:\s+Identifier|\s*:)\s*(\d+\.\d+\.\d+\.\d+)"
+        for match in re.finditer(server_pattern, output):
+            server_ip = match.group(1)
+            if server_ip not in servers:
+                servers.append(server_ip)
+
+        return servers
+
+    except FileNotFoundError as e:
+        raise DHCPDiscoveryError("nmap is not installed. Please install nmap package.") from e
+    except Exception as e:
+        raise DHCPDiscoveryError(f"Failed to discover DHCP servers: {str(e)}") from e
+
+
+async def main() -> None:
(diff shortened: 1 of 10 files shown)


core/libs/commonwealth/commonwealth/utils/DHCPServerManager.py
```diff
@@ -18,12 +22,14 @@ def __init__(
         subnet_mask: Optional[IPv4Address] = None,
         lease_range: Tuple[int, int] = (101, 200),
         lease_time: str = "24h",
+        backup: bool = False,
```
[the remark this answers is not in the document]
Williangalvani: you both told me 

> The introduction of async/await could be done incrementally to reduce the scope of the changes.

and to add this change, which would require even more changes to async

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+        stdout, stderr = await process.communicate()
+        output = stdout.decode()
+        error = stderr.decode()
+
+        if process.returncode != 0:
+            print(f"nmap output: {output}")
+            print(f"nmap error: {error}")
+            raise DHCPDiscoveryError(f"nmap failed: {stderr.decode()}")
```
joaoantoniocardoso, line 54: I think we can extract this into a function and share this code with [`core/services/autopilot_manager/mavlink_proxy/AbstractRouter.py:AbstractRouter::start`](https://github.com/bluerobotics/BlueOS/blob/1d6851042c65aea0fc893a7741140358d36ee6a9/core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py#L96) and [`core/services/helper/main.py:ping`](https://github.com/bluerobotics/BlueOS/blob/1d6851042c65aea0fc893a7741140358d36ee6a9/core/services/helper/main.py#L634).

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
```diff
@@ -0,0 +1,117 @@
...
+
+        return servers
+
+    except FileNotFoundError as e:
+        raise DHCPDiscoveryError("nmap is not installed. Please install nmap package.") from e
+    except Exception as e:
+        raise DHCPDiscoveryError(f"Failed to discover DHCP servers: {str(e)}") from e
+
+
+async def has_dhcp_server(interface: str, timeout: int = 15) -> bool:
```
joaoantoniocardoso, line 74: This function seems to be adding an unnecessary abstraction, the code here seems unlikely to be reused, so we can be inline and simplify it in the only place it is used.

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
```diff
@@ -0,0 +1,117 @@
...
+        process = await asyncio.create_subprocess_exec(
+            *cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+        )
+
+        stdout, stderr = await process.communicate()
+        output = stdout.decode()
+        error = stderr.decode()
+
+        if process.returncode != 0:
+            print(f"nmap output: {output}")
```
patrickelectric, line 52: shouldn't we be using logger here ?

Williangalvani: yep. fixed!

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
Williangalvani: I'm trying not to touch other services as we are almost stable...

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
Williangalvani: removed!

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
joaoantoniocardoso: Right, then we should prepare a PR with that change to be merged after stable

patrickelectric (approved):

core/libs/commonwealth/commonwealth/utils/DHCPDiscovery.py
joaoantoniocardoso: I've just moved this to [this issue](https://github.com/bluerobotics/BlueOS/issues/3194).

goasChris:
Thank you Willian, you are the best <3

PR #3159, opened 2025-02-19
vehicle-setup: disable calibrate button if vehicle has no position available

Williangalvani:
The reasoning here is that without a position, the WMM cannot be used, resulting in a worse calibration. This does NOT require a gps or dvl, there's a UI for setting a rough position estimate we can rely on.

## Summary by Sourcery

Enhancements:
- Disable the compass calibration button in the vehicle setup UI when the vehicle has no position available.

### DIFF
--- a/core/frontend/src/components/vehiclesetup/configuration/compass/FullCompassCalibrator.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/compass/FullCompassCalibrator.vue
@@ -81,7 +81,7 @@
         <v-btn
           v-if="state !== states.CALIBRATING && state !== states.FAILED"
           color="primary"
-          :disabled="!compass_mask"
+          :disabled="!compass_mask || !coordinates"
           @click="calibrate()"
         >
           Calibrate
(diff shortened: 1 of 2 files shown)


core/frontend/src/components/vehiclesetup/configuration/compass/FullCompassCalibrator.vue
```diff
@@ -78,7 +78,12 @@
       </v-card-text>
       <v-card-actions>
         <v-spacer />
-        <v-btn v-if="state !== states.CALIBRATING" color="primary" :disabled="!compass_mask" @click="calibrate()">
+        <v-btn
+          v-if="state !== states.CALIBRATING"
+          color="primary"
+          :disabled="!compass_mask || !coordinates"
```
patrickelectric, line 84: shouldn't this be a "and" ?

ES-Alexander:
```
disabled if:
- there are no compasses detected OR
- there is no position estimate
```

seems correct to me?

ES-Alexander (approved):
This seems to work as expected - nice! :D

Only suggestion would be to maybe add some warning text to the left or on top of the button (or a tooltip, but that's not mobile friendly), to explain why it's disabled - e.g.

- "No operating region provided" /
- "No compasses selected"

maybe even small, and with a newline between them if both are the case.

PR #3184, opened 2025-03-04
Revert "core:tools:blueos_startup_update: No IPV6"

rotu:
This reverts commit 715af4e8b0d3805a9b4b685417641410f0587a6a.

### DIFF
--- a/core/services/wifi/wifi_handlers/networkmanager/networkmanager.py
+++ b/core/services/wifi/wifi_handlers/networkmanager/networkmanager.py
@@ -4,7 +4,7 @@ import select
 import signal
 import subprocess
 from concurrent.futures import CancelledError
-from typing import Any, List, MutableMapping, Optional
+from typing import Any, List, Optional
 
 import sdbus
 from commonwealth.utils.general import device_id
@@ -219,7 +219,7 @@ class NetworkManagerWifi(AbstractWifiManager):
             return
 
         # If no existing connection, create a new one
-        connection: MutableMapping[str, Any] = {
+        connection = {
             "connection": {
                 "type": ("s", "802-11-wireless"),
                 "id": ("s", credentials.ssid),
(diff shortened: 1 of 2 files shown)


Williangalvani:
I don''t disagree that this is "wrong" in principle. The issue is that historically, ipv6 has given us more issues that it has helped, and we don't current have enough workforce to address the issue fully.

This patch will likely case more issues as it will result in a mix of ipv4+ipv6 and ipv4-only systems. the correct fix would be to actually have a patch reverting the systems which are already ipv4-only, too.

Meanwhile, I recommend relying on https://github.com/bluerobotics/BlueOS/pull/3049 to disable the "noIPV6" patch

rotu:
> The issue is that historically, ipv6 has given us more issues that it has helped, and we don't current have enough workforce to address the issue fully.

Even so, I'm not sure that this code correctly addresses issues caused by IPv6.

Loopback is still using IPv6 so I think that line `net.ipv6.conf.lo.disable_ipv6=1` is not working at all by default. `localhost` will still have any problems that you associate with IPv6.

Disabling IPv6 on Ethernet causes Windows to prefer literally any other interface that has IPv6, even if it has a higher metric (like wifi)! I'm now realizing that trying to disable IPv4 is responsible for `blueos-avahi.local` preferring the wifi interface and zerotier over the ethernet adapter. AFAICT, the *reason* you [need to edit the wifi connection](https://github.com/bluerobotics/BlueOS/commit/62fbcb4279cd32e3c4bad3e25d5596986ab37aa8) is to fix the problems caused by disabling IPv6 on ethernet.

I think the right fix, if you want to not *accidentally* use IPv6 is to address the host using a static IPv4 address. If you want to use automatic routing with multiple network interfaces, then the solution is to to configure the MDNS daemon appropriately.

> the correct fix would be to actually have a patch reverting the systems which are already ipv4-only, too.

Which systems are ipv4-only?

PR #3186, opened 2025-03-05
Fix unresolved imports

joaoantoniocardoso:
Solves this:
image

I saw that part of our config came from moving to vite, then I searched for vite projects and saw that they were using `eslint-import-resolver-vite`. Then I configured it following their [docs](https://www.npmjs.com/package/eslint-import-resolver-vite), and it worked.

As a side quest, I made unresolved imports throw error in build time, and the build failed on the autopilot import part, so I moved it to a runtime fetch (@Williangalvani is this correct?).

## Summary by Sourcery

Fixes unresolved imports by configuring eslint to work with vite aliases. Additionally, moves the autopilot import to a runtime fetch to prevent build errors.

### DIFF
--- a/core/frontend/src/types/autopilot/px4/metadata-fetcher.ts
+++ b/core/frontend/src/types/autopilot/px4/metadata-fetcher.ts
@@ -41,7 +41,11 @@ async function fetchPX4Metadata(): Promise<PX4ParametersMetadata[]> {
   try {
     metadata = await fetchPX4MetadataFromBoard()
   } catch (e) {
-    metadata = (await import('@/PX4-parameters/master/parameters.json')).parameters
+    const response = await fetch('/PX4-parameters/master/parameters.json');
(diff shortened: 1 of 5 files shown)


core/frontend/src/types/autopilot/px4/metadata-fetcher.ts
```diff
@@ -41,7 +41,11 @@ async function fetchPX4Metadata(): Promise<PX4ParametersMetadata[]> {
   try {
     metadata = await fetchPX4MetadataFromBoard()
   } catch (e) {
-    metadata = (await import('@/PX4-parameters/master/parameters.json')).parameters
+    const response = await fetch('/PX4-parameters/master/parameters.json');
```
Williangalvani, line 44: It doesn't look like this path exists on the built code. In order for this to work I think you'd need to move it to /public

PR #3192, opened 2025-03-07
core: frontend: components: vehiclesetup: PwmSetup: Move from SERVO\d_FUNCTION to Output \d

patrickelectric:
As requested


ES-Alexander:
We should update the screenshot, and potentially mention that these are based on the `SERVOn_FUNCTION` params.

PR #3203, opened 2025-03-14
[cable-guy] Add Routes API

joaoantoniocardoso:
This PR introduces a new feature: the Routes API, which allow us to add/delete arbitrary routes, like a default, or a multicast.

---

### Example 1: Add an (unmanaged) default route to `eth0` interface with priority of 1500:
```bash
curl -X 'POST' \
  'http://192.168.2.4:9090/v1.0/route?interface_name=eth0' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
        "destination": "0.0.0.0/0",
        "gateway": "192.168.2.1",
        "priority": 1500,
        "managed": false
      }'
```

### Example 2: Add a multicast route to `eth0` interface and track it:

```bash
curl -X 'POST' \
  'http://192.168.2.4:9090/v1.0/route?interface_name=eth0' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
        "destination": "224.0.0.0/4",
        "gateway": null,
        "priority": null,
        "managed": true
      }'
```

---

Notes:

1. The `managed` property is a contract to allow compatibility with other managers, meaning that:
  (I) we won't track changes made by other software;
  (II) we won't try to recreate once it's deleted by other software;

2. While this Route represents kernel routes with minimal abstraction, it has a conflict with the way we are dealing with "interface priorities". The latter is an abstract and simplified way of changing the priority of each route, and it will not respect the priority set by the introduced Routes API. This has little effect now, but the right thing is to upgrade all the interface priority API to use the Routes API internally, or conform to its nuances -- each interface has associated routes, and each route has a defined priority (`0` being the default). An interface might not have any route, though, which should be respected. To keep this PR simple (because we need the multicast route for radcam), I'm _not_ addressing this conflict here.

## Summary by Sourcery

Introduce a REST API for managing network routes associated with network interfaces.

New Features:
- Add API endpoints (`POST`, `DELETE`, `GET /route`) to manage network routes.
- Allow distinguishing between managed and unmanaged routes.
- Include route information when retrieving network interface details.
- Persist route configurations in settings.
- Add a default multicast route to the default `eth0` configuration.

### DIFF
--- a/core/services/cable_guy/api/manager.py
+++ b/core/services/cable_guy/api/manager.py
@@ -690,7 +690,7 @@ class EthernetManager:
         """
 
         mismatches: Set[NetworkInterface] = set()
-        current = self.get_ethernet_interfaces()
+        current_interfaces = self.get_ethernet_interfaces()
         if len(self._settings.content) == 0:
             logger.debug("No saved configuration found")
             logger.debug(f"Current configuration: {self._settings}")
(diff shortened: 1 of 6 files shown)


core/services/cable_guy/api/manager.py
```diff
@@ -59,6 +60,14 @@ def _settings(self) -> settings.SettingsV1:
     def __init__(self) -> None:
         self._dhcp_servers: List[DHCPServerManager] = []
 
+        # Make sure we have a multicast for eth0, for migration from BlueOS < 1.4
+        for interface in self.settings.root["content"]:
+            if interface["name"] == "eth0":
+                interface["multicast"] = next(
+                    (interface.multicast for interface in default_configs if interface.name == "eth0")
+                )
+        self.save()
```
joaoantoniocardoso, line 69: note: Possibly no longer needed

joaoantoniocardoso:
Hey @Williangalvani @patrickelectric @joaomariolago, this PR is ready for review.

core/services/cable_guy/typedefs.py
```diff
@@ -31,11 +35,62 @@ class InterfaceInfo(BaseModel):
...
+    def is_multicast(self) -> bool:
+        return ip_network(self.destination).is_multicast
+
+
 class NetworkInterface(BaseModel):
     name: str
     addresses: List[InterfaceAddress]
     info: Optional[InterfaceInfo] = None
     priority: Optional[int] = None
+    routes: List[Route]
```
joaomariolago, line 93: We need to think better if maybe this change shouldn't also be migrated to `NetworkInterfaceV2` since this will result in a total settings loss due to [this](https://github.com/bluerobotics/BlueOS/blob/17be52ee6328ca3705fe1aa2ce20f58b4f9d1fa2/core/services/cable_guy/api/settings.py#L46) in cases where users migrate from `1.3.1` to this PR, probably either next `1.4.1` or `1.5.0`. Since current stable is relatively new maybe this should also be migrated.

core/services/cable_guy/api/manager.py
```diff
@@ -563,6 +596,113 @@ def get_interface_info(self, interface_name: str) -> InterfaceInfo:
...
+
+            self.ipr.route(
+                action,
+                oif=interface_index,
+                dst=str(route.destination_parsed),
+                gateway=str(gateway) if gateway else None,
+                metrics={"metric": route.priority} if route.priority else None,
+            )
+
+            logger.info(f"Removed route to {route.destination_parsed} via {gateway} on {interface_name}")
```
joaomariolago, line 618: Not sure, but I think that the `Removed` should be the `action` instead.

core/services/cable_guy/api/manager.py
```diff
@@ -563,6 +596,113 @@ def get_interface_info(self, interface_name: str) -> InterfaceInfo:
...
+        except Exception as e:
+            logger.error(f"Failed to remove route: {e}")
```
joaomariolago, line 620: Same here. Not sure if its only `remove`.

joaomariolago (changes requested):
Just the small logging changes if needed. Apart from that seems good.

joaomariolago (approved):

PR #3208, opened 2025-03-18
core:frontend:FirmwareManager: Remove timeout

joaomariolago:
Uploader script by itself uses 3 minutes to timeout and kill the script, given that the full install process usually needs a download and cleanup if timeout, we shouldn't impose the timeout on the frontend since once axios release the request the process kill wont be successful on the backend.

## Summary by Sourcery

Migrate firmware upload and installation processes to use asynchronous programming, removing explicit timeout handling in the frontend

Enhancements:
- Refactor firmware upload and installation methods to use asyncio for better concurrency
- Remove explicit timeout configuration in frontend axios request

Chores:
- Update multiple service methods to be asynchronous
- Modify test cases to support async execution

### DIFF
--- a/core/frontend/src/components/autopilot/FirmwareManager.vue
+++ b/core/frontend/src/components/autopilot/FirmwareManager.vue
@@ -323,7 +323,6 @@ export default Vue.extend({
       this.install_status = InstallStatus.Installing
       const axios_request_config: AxiosRequestConfig = {
         method: 'post',
-        timeout: 180000,
       }
       if (this.upload_type === UploadType.Cloud) {
         // Populate request with data for cloud install
--- a/core/services/ardupilot_manager/api/v1/routers/index.py
+++ b/core/services/ardupilot_manager/api/v1/routers/index.py
@@ -176,11 +176,11 @@ async def install_firmware_from_url(
     # In some cases user might install a firmware that implies in a board change but this is not reflected,
     # so if the board is different from the current one, we change it.
     if (
-        auto_switch_board and
-        board and
-        autopilot.current_board and
-        autopilot.current_board.name != board.name and
-        FlightControllerFlags.is_bootloader not in board.flags
+        auto_switch_board
+        and board
+        and autopilot.current_board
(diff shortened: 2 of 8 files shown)


core/frontend/src/components/autopilot/FirmwareManager.vue
```diff
@@ -323,7 +323,6 @@ export default Vue.extend({
       this.install_status = InstallStatus.Installing
       const axios_request_config: AxiosRequestConfig = {
         method: 'post',
-        timeout: 180000,
```
joaoantoniocardoso, line 326: Doesn't a request always have a timeout? Then if we don't choose it, other piece of software will choose it, possibly from the browser. I image the underlying effect here is going to be to grow the timeout from 3 to (probably) 5 minutes.

joaomariolago: From the axios [docs](https://github.com/axios/axios/blob/e367be54dc0e8e3f5dfcba134c69d4a8f1e40324/README.md?plain=1#L350) they state that default is 0 so no timeout. This docs is specific for our version `^0.21.1`. But the modern version also states the same. So not sure about this but I think that it should keep the request waiting indefinitely.

joaomariolago: **Note**: The correct approach, beyond removing the timeout, would be to migrate the upload pipeline in the backend to an asynchronous and implement a cleanup task, as the user may also close the browser tab. However, since this change is more extensive than this fix, it will be implemented in future PRs. This fix now specifically addresses the issue on the Pixhawk1 series because we need some feedback on the frontend and the autopilot manager can not be stuck forever in this upload, since it will required a restart.

joaomariolago: No problems! The feedback was good, after some testing to see if the timeout was working correctly found that the asyncio was starving and the timer wasn't working. So this PR will already need to include the change of the pipeline to async. I'll mark this conversation as resolved and keep updating the async part.

Issue #3225, opened 2025-03-25
Pi3 BlueOS master freezes right after connecting pixhawk


joaoantoniocardoso:
Last words:

![image](https://github.com/user-attachments/assets/113b3c37-33c6-4b34-af4d-26eccb43e641)

note: it also freezes during boot if the pixhawk is connected

joaoantoniocardoso:
Increasing swap (from 200MB to 1024MB) doesn't help

joaoantoniocardoso:
After today's investigation, it was related to power: by powering the Raspberry Pi 3 with an external supply from two pairs of vcc+gnd on the header connector, it worked until we restarted the autopilot service.

The power supply was set to 5.25V, and the current consumption was around 600mA, with peaks up to 780mA.

PR #3228, opened 2025-03-26
core: tools: blueos_startup_update: Add swapfile update

patrickelectric:
## Summary by Sourcery

Add functionality to dynamically update swap file size based on available disk space during system startup

New Features:
- Implement dynamic swap file size adjustment mechanism that checks available disk space and updates swap configuration accordingly

Enhancements:
- Add a new function to check available disk space
- Create a mechanism to modify swap file size based on system resources

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -545,6 +545,57 @@ unmanaged-devices=interface:eth0;interface:usb0
     return True
 
 
+def check_available_space(required_mb: int) -> bool:
+    """Check if there is enough space in disk for the required megabytes"""
+    try:
+        stats = os.statvfs("/")
+        available_mb = stats.f_bavail * stats.f_frsize / (1024 * 1024)
+        return available_mb >= required_mb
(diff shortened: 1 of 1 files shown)


core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -545,6 +545,57 @@ def configure_network_manager() -> bool:
...
+def update_swap_size() -> bool:
+    """Updates the swap size if there is enough space available"""
+    logger.info("Checking and updating swap size...")
+
+    config_file = "/etc/dphys-swapfile"
+
+    if check_available_space(1024):
+        desired_size = 1024
+    elif check_available_space(512):
+        desired_size = 512
```
joaoantoniocardoso, line 568: maybe we should require 30% more space than what we aim for the swap file

patrickelectric: good idea

joaoantoniocardoso (approved):
tested on a Pi3 and Pi4, all good

patrickelectric:
Should fix #3225

PR #3237, opened 2025-03-28
Fix blueos startup update pi3

### LINKED ISSUE #3236: [Pi3] Cameras not working after updating from last stable (1.3.1)
The support team reported that video streams were freezing after updating to 1.4.0-beta.16 on Pi3+Pixhawk.

After a week of work to reproduce the scenario, the suspects were:
- Power supply problems
- Camera problems
- Camera cable problems
- Problems from other devices (Pixhawk, for instance)
- Gstreamer version
- MCM version
- Kernel version

The three first ones (hardware-related problems) were discarded after trying different cameras, cables, and ways to power the Pi3, mostly led from [this](https://github.com/bluerobotics/BlueOS/issues/3225) other problem, which was showing undervoltage kernel messages.

It took a while after the team could gather the pieces and get a stable setup that could reproduce the problem. The materials are:
- Pi3
- A good power supply, preferentially via two pairs of 5V-GND pins.
- The default BR camera w/ good cable

To reproduce the problem:
1. Install BlueOS-1.3.1 on a Pi3
2. Connect the camera
3. Check that you can create and delete video streams via MCM multiple times and always receive thumbnails. Alternatively, kill MCM and run the test stream (shown below).
4. Update to a current beta (we used BlueOS-1.4.0-beta.17 because it was easier to test because [this](https://github.com/bluerobotics/BlueOS/issues/3223) was fixed).
5. Rerun step 3, it will fail at some point. It might require another reboot.

Test stream: 
```
GST_DEBUG='*:3,*v4l*:4' gst-launch-1.0 v4l2src device=/dev/video2 do-timestamp=true num-buffers=10 io-mode=mmap ! video/x-h264,stream-format=byte-stream,alignment=au,width=1920,height=1080,pixel-aspect-ratio=1/1,framerate=30/1 ! fakesink sync=false
```

The same problem happens if a BlueOS-1.4.0-beta.17 is installed (without the update, so BlueOS-1.4.0-beta.17 would be the factory version).

To investigate it methodically, a test matrix was built with the main suspects, and a couple of tests were run, observing `dmesg` and `journalctl` messages:

image

## Conclusion

After analyzing the diff between journals and `lsmod`, we understood that the difference was that when it was working, the USB was using the dwc_otg module, and when it wasn't working, it was using the dwc2 module.

We then diffed both the `/boot/config.txt` and `/boot/cmdline.txt` files and found that:
- `dwc2` parameters were added inside a [pi5] tag in `/boot/config.txt`
- `cgroups` and `dwg2` parameters were added to the `/boot/cmdline.txt`

We first removed the added lines to the `/boot/cmdline.txt`, but the modules were still being loaded, because the boot line is configured from parameters from both files.

By reading the [raspberry config.txt docs](https://www.raspberrypi.com/documentation/computers/config_txt.html), one would expect each device to only react to the parameters at [all], or inside a tag with its name, so a Pi3 shouldn't be reacting to parameters inside a `[pi5]` tag -- _but it was!_

With a clean `/boot/cmdline.txt`, by manually adding and removing those lines in `/boot/config.txt`, we can control the video freezing behavior, which also validates the solution.

## Solution

A PR is being prepared for the fix:
1. Avoid adding those lines for Pi3 (by skipping the `dwc2` patches)
2. Remove those lines (and any other line from other devices) when present

:rocket:

joaoantoniocardoso:
Closes #3236

## Summary by Sourcery

Improve BlueOS startup update script for Raspberry Pi 3 by adding specific configuration and update handling

New Features:
- Add support for Raspberry Pi 3 in CPU type detection
- Implement specific update and cleanup functions for Raspberry Pi 3

Bug Fixes:
- Fix incorrect configuration application for Raspberry Pi 3
- Remove incorrectly applied dwc2 module configuration on Pi3

Enhancements:
- Refactor boot configuration handling functions to use more generic 'section' terminology
- Add support for detecting and handling Raspberry Pi 3 specific configurations

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -128,6 +128,25 @@ def boot_config_add_configuration_at_section(config_content: List[str], config: 
         config_content.insert(section_start + 1, config)
 
 
+def boot_config_remove_section(config_content: List[str], section_name: str) -> None:
+    (section_start, section_end) = boot_config_get_or_append_section(config_content, section_name)
+    del config_content[section_start:section_end]
+
+
+def boot_config_get_available_section(config_content: List[str]) -> List[str]:
+    regex_flags = re.IGNORECASE | re.DOTALL | re.MULTILINE
+
+    section_match_pattern = r"^\[(?P<section>\w+)\].*$"
+
+    section = []
+    for line in config_content:
+        match = re.match(section_match_pattern, line, regex_flags)
+        if match:
+            section.append(match.group("section"))
+
+    return section
+
+
 def boot_config_filter_conflicting_configuration_at_section(
     config_content: List[str], config_pattern_match: str, config: str, section_name: str
 ) -> List[str]:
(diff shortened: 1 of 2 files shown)


core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -263,6 +282,55 @@ def update_i2c4_symlink() -> bool:
...
+        if "dwc2" not in item and "g_ether" not in item:
+            cmdline_content.append(item)
+
+    # Save if needed, with backup
+    if unpatched_cmdline_content == cmdline_content:
+        return False
+    save_file(cmdline_file, " ".join(cmdline_content), "before_revert_update_dwc2")
+
+    # Patch applied and system needs to be restarted for it to take effect
+    return True
```
Williangalvani, line 304: I know I'm the one who wrote this, but for the record, this will drop all of `modules_load` if either `dwc2` or `g_ether` is in there.
I think it is not worth the extra effort/code, as if someone manually add/remove a module, we can assume the offending modules are not there.

Williangalvani (approved):
code looks sane. haven't tested on a pi

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -263,6 +263,29 @@ def update_i2c4_symlink() -> bool:
...
+    return True
```
Williangalvani: oh it seems a rebase may have gone wrong and there are two `return True` on this commit.
not a blocker for me though 🤷‍♂️

joaoantoniocardoso: hahaha yeah, that was there, ty!

core/tools/blueos_startup_update/blueos_startup_update.py
joaoantoniocardoso: yeah, the most correct would be to implement boot_cmdline_remove_module

patrickelectric (approved):

PR #3255, opened 2025-04-04
core: tools: mavlink-server: Update to 0.4.0

patrickelectric:
## Summary by Sourcery

Chores:
- Bump mavlink-server version from 0.3.1 to 0.4.0 in bootstrap script


joaoantoniocardoso (approved):
It does work, however, I guess MS's CI build is missing git submodules during the checkout:

image

Note: the CPU use is around ~28% here (Pi4), versus ~10% before.

PR #3258, opened 2025-04-04
set BlueOS timezone to the same as the user's browsers

Williangalvani:


joaoantoniocardoso (commented):
I'm not sure if this is a good idea. If we want to set that, I believe this should be done in wizard and/or vehicle setup.

At least let's be sure not to get into the following known chaotic scenario:

1. vehicle is on a mission, gathering research data
2. system time is updated
3. services that use system time glitch the data continuity, integrals and derivatives go crazy, and data and logs are possibly going back in time
4. dataloggers using timestamps as keys overwrite old data entries, or end up with duplicate key entries
5. researcher doesn't know what happened and only discovers the data is compromised months later, when analyzing it because the data doesn't make sense
6. the user has no way to prevent that, as they don't realize their topside is also automatic updating the time

Williangalvani:
> I'm not sure if this is a good idea. If we want to set that, I believe this should be done in wizard and/or vehicle setup.
> 
> At least let's be sure not to get into the following known chaotic scenario:
> 
> 1. vehicle is on a mission, gathering research data
> 2. system time is updated
> 3. services that use system time glitch the data continuity, integrals and derivatives go crazy, and data and logs are possibly going back in time
> 4. dataloggers using timestamps as keys overwrite old data entries, or end up with duplicate key entries
> 5. researcher doesn't know what happened and only discovers the data is compromised months later, when analyzing it because the data don't make sense
> 6. the user has no way to prevent that, as they don't realize their topside is also automatic updating the time as they change timezone when connecting to different cellular network zones.

Your points are all valid, but out of scope I believe.
set_time is already used. this will only change the timezone, which afaik will only affect how the time is "rendered"

joaoantoniocardoso:
Yes, `everything should also use clock_monotonic`, but the pain is real :eyes:

Williangalvani:
> Yes, `everything should also use clock_monotonic`, but the pain is real 👀

the pain is real but out of scope :p

joaoantoniocardoso:
> the pain is real but out of scope :p

Changing the timezone will break the continuity of our Python logs that use `loguru`, for example.

Williangalvani:
> > the pain is real but out of scope :p
> 
> Changing the timezone will break the continuity of our Python logs that use `loguru`, for example. The rust services use UTC, so they are ok.

that is exactly what I'm looking for? right now our logs do not match the local time.
plus, as I mentioned, I expect this to be set once.

I do see an issue if someone access the vehicle remotely, though...
perhaps we could do this on the wizard?

joaoantoniocardoso:
> perhaps we could do this on the wizard?

agree, plus a pirate-mode button somewhere just in case

PR #3301, opened 2025-05-13
cable-guy: only skipping adding dhcp server if the ip is on the interface, too

Williangalvani:
This fixes an issue where after booting , the vehicle loses the eth0 ip and blueos doesnt try to restore it.

## Summary by Sourcery

Ensure DHCP server is only considered already present when its gateway IP is configured on the interface, allowing BlueOS to restore lost eth0 IP after boot.

Bug Fixes:
- Only skip adding a DHCP server when the gateway IP is actually present on the interface to prevent losing the eth0 IP after boot.

### DIFF
--- a/core/services/cable_guy/api/manager.py
+++ b/core/services/cable_guy/api/manager.py
@@ -771,7 +771,11 @@ class EthernetManager:
         """
         if self._is_dhcp_server_running_on_interface(interface_name):
             dhcp_on_interface = self._dhcp_server_on_interface(interface_name)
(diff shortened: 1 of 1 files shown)


Williangalvani:
you can reproduce the issue on master by running `ip addr del 192.168.2.2/24 dev eth0` while connected through another ip. cable-guy wont restore that ip.

joaoantoniocardoso (approved):

patrickelectric (approved):

PR #3323, opened 2025-05-22
install.sh: disable NetworkManager-wait-online.service

Williangalvani:


Williangalvani:
> * Run `sudo systemctl disable NetworkManager-wait-online.service` to ensure the command has the needed privileges.
> * Add a check to verify the service exists before disabling to avoid unnecessary error output on systems without 
systemctl is working fine on line 236, so I assume it should befine here, too

PR #3357, opened 2025-06-05
Add sentry SDK on BlueOS backend

joaomariolago:
## Summary by Sourcery

Initialize Sentry in each backend service using new commonwealth utils and standardize service startup logic with async main() functions

New Features:
- Integrate Sentry SDK for error monitoring across BlueOS backend services
- Add async and sync Sentry initialization hooks in commonwealth utilities

Enhancements:
- Refactor all service entrypoints to use async main() with asyncio.run and centralized Server/Config or AppRunner startup patterns
- Replace manual event loop management and uvicorn.run invocations with unified async startup flows

Build:
- Add sentry-sdk as a project dependency

### DIFF
--- a/core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
@@ -1,0 +1,51 @@
+import os
+import re
+from typing import Any, Dict, Optional
+
+import sentry_sdk
+
+
+def _get_sentry_config() -> Optional[Dict[str, Any]]:
+    """
+    Retrieves the Sentry configuration for the BlueOS backend project
+
+    Returns:
+        Optional[ClientConstructor]: The Sentry configuration if valid, otherwise None.
+    """
+
+    SENTRY_PROJECT = "BlueOS"
+    SENTRY_DSN = "https://d93d1be8ddb7d5e1f45fb1eeca287eac@o4507696465707008.ingest.us.sentry.io/4509446521683968"
+    VALID_TAG_PATTERN = r"^\d+\.\d+\.\d+-\d+-g[0-9a-f]{7,}$"
+
+    git_describe_tag = os.environ.get("GIT_DESCRIBE_TAGS")
+
+    if not git_describe_tag or not re.match(VALID_TAG_PATTERN, git_describe_tag):
+        return
+
+    release = f"{SENTRY_PROJECT}@{git_describe_tag}".replace("tags/", "").replace("/", ":")
+
+    return {
+        "dsn": SENTRY_DSN,
+        "release": release,
+        "traces_sample_rate": 1.0,
+        "trace_propagation_targets": [],
+        "send_default_pii": True,
+    }
+
+
+def init_sentry(name: Optional[str] = None) -> None:
+    sentry_config = _get_sentry_config()
+    if sentry_config:
+        sentry_sdk.init(server_name=name, **sentry_config)
+
+
+async def init_sentry_async(name: Optional[str] = None) -> None:
+    """
+    Initializes Sentry when used in an async context.
+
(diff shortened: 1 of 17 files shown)


core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+
+def init_sentry(name: Optional[str] = None) -> None:
+    sentry_config = _get_sentry_config()
+    if sentry_config:
+        sentry_sdk.init(server_name=name, **sentry_config)
+
+
+async def init_sentry_async(name: Optional[str] = None) -> None:
```
patrickelectric, line 42: can you explain why do we need an async function if no method is async ?

core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
```diff
@@ -0,0 +1,45 @@
+import os
+import re
+from typing import Any, Dict, Optional
+
+from sentry_sdk.consts import ClientConstructor
+
+
+def get_sentry_config() -> Optional[ClientConstructor]:
```
patrickelectric: this function should have an underscore

core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
joaomariolago: Per sentry [docs](https://docs.sentry.io/platforms/python/#configure) it is stated that is needed to call from within an async function the init.

core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
joaomariolago: The code is using underscore, I think that maybe it is lost on the rebase. I will see if I can pass to the seconds commit.

core/services/pardal/main.py
```diff
@@ -75,11 +77,26 @@ async def root(request: web.Request) -> web.Response:
...
-app.add_routes([web.get("/ws", websocket_echo)])
-app.router.add_get("/", root, name="root")
-app.router.add_get("/get_file", get_file, name="get_file")
-app.router.add_post("/post_file", post_file, name="post_file")
-web.run_app(app, path="0.0.0.0", port=args.port)
+async def main() -> None:
+    await init_sentry_async(SERVICE_NAME)
+
+    app = web.Application()
+    app._client_max_size = 2 * (2**30)  # 2 GBs
```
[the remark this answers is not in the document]
joaoantoniocardoso: @joaomariolago this looks like a good idea. [Ref](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.client_max_size).

core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
patrickelectric: Can you add a comment in block since this is not clear at all looking at the code ?

core/services/pardal/main.py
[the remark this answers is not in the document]
patrickelectric: I would say that is out of the scope of this PR and should be handled by another PR or via an issue

core/libs/commonwealth/src/commonwealth/utils/sentry_config.py
joaomariolago: Added!

core/services/pardal/main.py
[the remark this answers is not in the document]
joaomariolago: Agree that is out of the scope, I added the suggestion since it is a simple thing and I already tested here and it works ok, but I tried to copy the start code to the async mains as similar as possible. But added this since it was easy to test and validate.

patrickelectric (approved):

PR #3362, opened 2025-06-06
core: services: helper: main: Control number of worker for scan_ports based on hardware

patrickelectric:
## Summary by Sourcery

Adjust scan_ports to determine the thread pool size based on CPU type and the number of ports to scan.

Enhancements:
- Set ThreadPoolExecutor max_workers to 2 on Raspberry Pi 3 or when there are no ports to scan.
- Use the number of ports as max_workers for other CPU types to improve concurrency.


joaoantoniocardoso (approved):

patrickelectric:
@joaoantoniocardoso it would be nice

PR #3383, opened 2025-06-19
backend: Update zenoh to latest version

patrickelectric:
Update everything besides frontend to 1.5.0
The frontend requires a major change since the api changed a lot, will work around it in the next PR


patrickelectric:
> ~latest~?

I didn't follow

joaoantoniocardoso:
> > ~latest~?
> 
> I didn't follow

Oh, it's just because it's using 1.3.4 for the frontend side, so the PR name isn't strictly accurate.

patrickelectric:

PR #3393, opened 2025-06-21
core: frontend: components: zenoh-inspector: ZenohNetwork: Minor typescript fixes

patrickelectric:
## Summary by Sourcery

Refine TypeScript typing in the ZenohNetwork component to ensure proper layout option and error handling types

Enhancements:
- Import and use FcoseLayoutOptions for the cytoscape-fcose layout options
- Assert the layout options object as FcoseLayoutOptions before running the layout
- Type the catch block error as unknown and cast it to Error to safely access its properties in logging


joaoantoniocardoso (approved):
nice!

PR #3395, opened 2025-06-21
core: frontend: kraken: StoreExtensionCard: Fix cross problem

patrickelectric:
## Summary by Sourcery

Bug Fixes:
- Prefix extension logo URLs with corsproxy.io to bypass cross-origin restrictions.

### DIFF
--- a/core/frontend/src/components/kraken/cards/StoreExtensionCard.vue
+++ b/core/frontend/src/components/kraken/cards/StoreExtensionCard.vue
@@ -50,7 +50,7 @@
             v-once
             ref="extension_logo"
             contain
-            :src="extension.extension_logo"
+            :src="'https://corsproxy.io/?url=' + extension.extension_logo"
             height="150px"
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (commented):
Be aware that if we are using it on the free tier (which is meant for development), they say:
- it has a rate limit of 60 per minute (per IP), which means this will be a temporary fix.
- it is limited to localhost, which means it might not work on our BlueOS live preview.

patrickelectric:
> Be aware that if we are using it on the free tier (which is meant for development), they say:
> 
> * it has a rate limit of 60 per minute (per IP), which means this will be a temporary fix.
> * it is limited to localhost, which means it might not work on our BlueOS live preview.

Oh sad..

Williangalvani:
was this fixed by https://github.com/bluerobotics/BlueOS-Extensions-Repository/pull/147?

PR #3398, opened 2025-06-26
core:(frontend/startup): Modify Cloud token system

joaomariolago:
* Change to save cloud token system in independent file instead of the bag
* Add docker binds for file

## Summary by Sourcery

Modify the cloud token system to persist tokens in a dedicated file with fallback and cleanup logic, integrate filebrowser operations, and update Docker binds for the new storage path.

Enhancements:
- Persist Major Tom cloud token to a dedicated file instead of the Vuex bag
- Implement filebrowser-based read, write, and cleanup methods for token management
- Retain backward compatibility by falling back to bag storage when no token file exists

Build:
- Add Docker bind mount for "/root/.majortom" to support file-based token storage

### DIFF
--- a/bootstrap/startup.json.default
+++ b/bootstrap/startup.json.default
@@ -46,6 +46,10 @@
                 "bind": "/root/.docker",
                 "mode": "rw"
             },
+            "/root/.majortom": {
+                "bind": "/root/.majortom",
+                "mode": "rw"
+            },
             "/etc/blueos": {
                 "bind": "/etc/blueos",
                 "mode": "rw"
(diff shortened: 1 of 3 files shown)


voorloopnul (approved):
Looks good to me
One question tho,  why the file is called **token.key** instead of just **token** ?

joaomariolago:
> Looks good to me One question tho, why the file is called **token.key** instead of just **token** ?

Update on this, I tested with no extension and the filebrowser complains to create the file, so I think that keeping with `.key` extension will be better.

patrickelectric (approved):

PR #3408, opened 2025-07-03
core:frontend: Integrates frontend console messages to zenoh

joaomariolago:


core/frontend/src/libs/console-logger.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      if (level === 'error' && clonedArgs[0] instanceof Error) {
+        message.stack = clonedArgs[0].stack
+      }
+
+      const topic = `frontend/${frontend.frontend_id}`
+      const payload = JSON.stringify(message)
+
+      this.session.put(topic, payload)
```
joaoantoniocardoso: we are also trying to add the `encoding` to the messages to help the parser on the other side.

core/frontend/src/store/frontend.ts
```diff
@@ -18,6 +19,8 @@ class FrontendStore extends VuexModule {
 
   backend_offline = false
 
+  frontend_id = nanoid(9)
```
joaoantoniocardoso: Can we log who the browser/client is whenever it gets the ID?

joaoantoniocardoso (commented):
Nice feature! I'm adding some suggestions here, but I haven't tested it yet.

core/frontend/src/libs/console-logger.ts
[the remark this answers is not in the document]
joaomariolago: On version `1.3.4` we are running [`session.put`](https://github.com/eclipse-zenoh/zenoh-ts/blob/54684e084599695af0862d8053cea0ef7a0673ac/zenoh-ts/src/session.ts#L225) is syncronous.

core/frontend/src/store/frontend.ts
joaomariolago: Added!

core/frontend/src/libs/console-logger.ts
joaomariolago: Added `Encoding.APPLICATION_JSON` if you think other encoding would be more appropriated just let me know.

core/frontend/src/libs/console-logger.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      if (level === 'error' && clonedArgs[0] instanceof Error) {
+        message.stack = clonedArgs[0].stack
+      }
+
+      const topic = `frontend/${frontend.frontend_id}`
+      const payload = JSON.stringify(message)
+
+      this.session.put(topic, payload, { encoding: Encoding.APPLICATION_JSON })
```
patrickelectric: I would recommend using https://docs.foxglove.dev/docs/visualization/message-schemas/log
and using as schema for APPLICATION_JSON `Foxglove.Log`

patrickelectric (changes requested):

joaomariolago:
> Just to be sure, did you tested with [bevry/get-current-line](https://github.com/bevry/get-current-line?rgh-link-date=2025-07-07T12%3A56%3A22Z) ?

Wouldn't this give us the location of the `console-logger.ts` instead of the original error? Because the extraction tries to get where the error originates from so is not exactly the line where we process it. I think that maybe the [`stacktrace-parser`](https://www.npmjs.com/package/stacktrace-parser) would be a better approach for the generic errors, but not sure if totally needed, because most of the cases where we can get the stack are unhanded and usually they will already provide the filename and line.

PR #3422, opened 2025-07-15
Add BlueOS recorder

patrickelectric:

## Summary by Sourcery

Add BlueOS recorder support by incorporating its binary into the Docker image, updating the static binaries installer, and supplying a versioned bootstrap installer for different CPU architectures.

New Features:
- Include blueos-recorder binary in the core Docker image
- Add recorder to the static binaries installation list
- Provide a bootstrap.sh installer script to download and install blueos-recorder for multiple architectures

### DIFF
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -133,6 +133,7 @@ SERVICES=(
     'nginx',250,"nice -18 nginx -g \"daemon off;\" -c $TOOLS_PATH/nginx/nginx.conf"
     'log_zipper',250,"nice -20 $SERVICES_PATH/log_zipper/main.py '/shortcuts/system_logs/\\\\*\\\\*/\\\\*.log' --max-age-minutes 60"
     'bag_of_holding',250,"$SERVICES_PATH/bag_of_holding/main.py"
+    'recorder',250,"blueos-recorder --recorder-path /usr/blueos/userdata/recorder"
 )
 
 tmux -f /etc/tmux.conf start-server
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -109,6 +109,7 @@ COPY run-service.sh /usr/bin/run-service
 # Copy binaries and necessary folders from download-binaries to this stage
 COPY --from=download-binaries \
     /usr/bin/blueos_startup_update.py \
+    /usr/bin/blueos-recorder \
     /usr/bin/bridges \
     /usr/bin/linux2rest \
     /usr/bin/machineid-cli \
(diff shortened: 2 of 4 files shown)


core/start-blueos-core
```diff
@@ -133,6 +133,7 @@ SERVICES=(
     'nginx',250,"nice -18 nginx -g \"daemon off;\" -c $TOOLS_PATH/nginx/nginx.conf"
     'log_zipper',250,"nice -20 $SERVICES_PATH/log_zipper/main.py '/shortcuts/system_logs/\\\\*\\\\*/\\\\*.log' --max-age-minutes 60"
     'bag_of_holding',250,"$SERVICES_PATH/bag_of_holding/main.py"
+    'recorder',250,"blueos-recorder --recorder-path /var/logs/blueos/services/recorder"
```
joaoantoniocardoso: I don't think `/var/logs` is the place for recorded data -- the semantics are very different, and we should treat them differently. `/var/` is, in many cases, temporary. One quick example is that when the user deletes the system logs because they are running out of space, it shouldn't delete their mission data.

1. The service should store its data in a distinct location accessible to the user.
2. The service should (asap) add their runtime/debug logs to `/var/logs/blueos/services/recorder`

Now, what location, we may ask, and my vote is for `/usr/blueos/userdata/recorder/`.

patrickelectric: Changed

core/start-blueos-core
```diff
@@ -133,6 +133,7 @@ SERVICES=(
     'nginx',250,"nice -18 nginx -g \"daemon off;\" -c $TOOLS_PATH/nginx/nginx.conf"
     'log_zipper',250,"nice -20 $SERVICES_PATH/log_zipper/main.py '/shortcuts/system_logs/\\\\*\\\\*/\\\\*.log' --max-age-minutes 60"
     'bag_of_holding',250,"$SERVICES_PATH/bag_of_holding/main.py"
+    'recorder',250,"blueos-recorder --recorder-path /usr/blueos/userdata/recorder"
```
joaoantoniocardoso, line 136: Shouldn't we add `--verbose` in here?

joaoantoniocardoso (approved):
This seems to be working well. Testedo n Pi5 w/ 1xUSBCam (1080p) + 1xRadcam (4k + 1080p);

- ~5% CPU use

- 341MB in 3.4 seconds, which gives us ~100MB/s for this case

- A small limitation, that we must consider: FoxGlove format expects `baseline` profile, and radcam is by default using the `main` profile. In the end, Foxglove can't display it:
image

- The USB camera video was fine

- I started and ended many files by arming/disarming, and it seems solid!

---

With all that said, someone should test it on a Pi4 and check the CPU usage.

PR #3434, opened 2025-07-23
Ping: handle timeouts as failures (backport)

Williangalvani:
backporting #3430 to stable.
for extra-safety, It would be nice if @vshie tested it before we merge

## Summary by Sourcery

Enhancements:
- Wrap ping.request calls in detect_highest_baud with exception handling to log failures and count timeouts as failures


joaoantoniocardoso (approved):
let's go

Issue #3292, opened 2025-05-01
core: frontend: Notification for uncalibrated sensors


dgudiel:
### Current behaviour

When setting up a new BlueROV2 with QGC, QGC will tell you that calibration is required and guide you to the relevant calibration screens. 

### Expected or desired behaviour

It would be nice if BlueOS would give some indication/notification/warning that sensors require calibration.

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

PR #3443, opened 2025-07-29
core: frontend: add notification for uncalibrated sensors

nicoschmdt:
fix: #3292

Calibrated:
image

Uncalibrated:
image

Uncalibrated tool-tip:
image

If some sensors are not calibrated:
image

Docker image for testing: nicoschmdt/blueos-core/notification-for-uncalibrated-sensors


core/frontend/src/App.vue
```diff
@@ -50,6 +50,19 @@
             Armed
           </v-card-title>
         </v-card>
+        <v-card
+          v-if="uncalibrated_sensors"
+          v-tooltip="'Some sensors are not calibrated'"
```
joaoantoniocardoso: (optional) perhaps it would be more precise to word it as _"At least one sensor is not calibrated"_ or _"One or more sensors are not calibrated"_.

joaoantoniocardoso:
hmm... I'm not sure whether we want to put the notification in the widgets area, as the other notifications-like things are happening on the opposite side of the top bar.

joaoantoniocardoso:
Hmm, right, although they all share semantics, as you noted, they may differ in priority. The notification system is a good idea too (and perhaps we should do both). I'll let other people input on this. 

Since this is UX, I hope @ES-Alexander can give his view on it?

ES-Alexander:
As I understand it:
- The "Armed" indicator widget is to indicate when the interface is in "safe" mode, which requires additional confirmation before certain actions can occur (to prevent unsafe behaviour)
    - The calibration state does not have such an interface mode, and while action is required it does not meaningfully influence the use of BlueOS (just of the autopilot / vehicle control)
- The notifications list from the bell icon at the far right of the header bar is for events
    - The calibration state is reported regularly, but is not an event so much as a continuous state, so it does not belong in the list (to avoid being continual spam, or getting lost in subsequent notifications)
- The other icons at the right are for a mix of functionalities and state indicators
    - This seems like the best fit for the calibration details

I can think of a couple of potential interface options:
1. Augment the heartbeat indicator
    - Make it orange when uncalibrated
    - Include calibration details in the clicked expansion (at least showing which sensors are uncalibrated, but ideally also including calibration health/quality so it's useful even when calibrated)
    - Add links when uncalibrated to jump to the relevant calibration page in BlueOS
2. Add one or more temporary indicators
    - This would be more like the storage full warning, which only shows up when the problem is happening
    - There could be a single indicator with a tooltip that "one or more sensors are uncalibrated", which clicking on could take to the vehicle setup overview page, OR
    - There could be separate indicators per sensor type, so when several sensors are uncalibrated there are multiple indicators, with click-through links going to the relevant calibration page (in the vehicle setup section)

I personally prefer the heartbeat augmentation approach, because calibration seems like part of the autopilot health, and I'm reluctant to have an ever-growing list of icons people need to know about and be able to interpret. That said, the temporary indicators approach still seems viable, and would at least be better than the current lack of indication. I'm unsure which would be easier to implement - they seem pretty similar in complexity 🤷‍♂️

patrickelectric (approved):

Issue #2372, opened 2024-02-07
core: frontend: vehicle-setup: add page for Lua scripts


ES-Alexander:
### Current behaviour

Lua scripts exist and influence the autopilot, but they're not nicely accessible, other than via the file manager for vehicles using Navigator flight controllers, but finding that requires some digging.

### Expected or desired behaviour

Lua scripts are a pretty integral part of ArduPilot vehicle management, and that seems to only be increasing. It would be useful to have a part of the "Vehicle Setup" page dedicated to managing (and modifying) Lua scripts.

Lua is already used on the BlueBoat for controlling the NavLight, and [has been used for sub too](https://github.com/ArduPilot/ardupilot/issues/25528).

For Navigators this could be implemented as a variant of the existing File Browser, just started from the relevant directory, but for external flight controllers (e.g. Pixhawk, etc) we might need to add a MAVLink-based file-system mapping or abstraction layer.

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

PR #3445, opened 2025-07-29
ardupilot_manager: create shortcut for script directory when using linux

nicoschmdt:
fix: #2372 

image

Since paths may differ across different platforms, the shortcut is only created if the flight controller platform type is Linux.

## Summary by Sourcery

New Features:
- Add setup() in LinuxFlightController to create the firmware scripts directory and symlink at /shortcuts/lua_scripts for Linux platforms and call it during autopilot startup

### DIFF
--- a/core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
+++ b/core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
@@ -1,3 +1,5 @@
+import os
+
 from typing import List, Type
 
 from smbus2 import SMBus
@@ -7,6 +9,9 @@ from typedefs import FlightController, PlatformType, Serial
 
 class LinuxFlightController(FlightController):
     """Linux-based Flight-controller board."""
+
+    STANDARD_SCRIPT_DIRECTORY_PATH = "/root/.config/ardupilot-manager/firmware/scripts"
+    LUA_SCRIPT_DIRECTORY_PATH = "/shortcuts/lua_scripts"
 
     @property
     def type(self) -> PlatformType:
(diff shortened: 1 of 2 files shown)


core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
```diff
@@ -26,6 +28,10 @@ def check_for_i2c_device(self, bus_number: int, address: int) -> bool:
         except OSError:
             return False
 
+    def setup(self) -> None:
+        os.makedirs("/root/.config/ardupilot-manager/firmware/scripts", exist_ok=True)
+        os.symlink("/root/.config/ardupilot-manager/firmware/scripts", "/shortcuts/lua_scripts")
```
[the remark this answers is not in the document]
Williangalvani: I think we do need to address this one

Image

core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
```diff
@@ -26,6 +28,13 @@ def check_for_i2c_device(self, bus_number: int, address: int) -> bool:
         except OSError:
             return False
 
+    def setup(self) -> None:
+        os.makedirs("/root/.config/ardupilot-manager/firmware/scripts", exist_ok=True)
```
Williangalvani: actually... can we not hardcode it? I think this should probably be a LinuxFlightController attribute, so that subclasses could override it.

core/services/ardupilot_manager/flight_controller_detector/linux/linux_boards.py
```diff
@@ -26,6 +31,15 @@ def check_for_i2c_device(self, bus_number: int, address: int) -> bool:
         except OSError:
             return False
 
+    def setup(self) -> None:
+        os.makedirs(LinuxFlightController.STANDARD_SCRIPT_DIRECTORY_PATH, exist_ok=True)
+        try:
+            os.symlink(
+                LinuxFlightController.STANDARD_SCRIPT_DIRECTORY_PATH, LinuxFlightController.LUA_SCRIPT_DIRECTORY_PATH
```
Williangalvani:
```suggestion
                self.STANDARD_SCRIPT_DIRECTORY_PATH, self.LUA_SCRIPT_DIRECTORY_PATH
```


Image

PR #3460, opened 2025-08-05
frontend: add missing route for extensionv2

Williangalvani:


core/frontend/src/router/index.ts
```diff
@@ -108,6 +108,11 @@ const routes: Array<RouteConfig> = [
     name: 'Named Extensions',
     component: ExtensionView,
   },
+  {
+    path: '/extensionv2/:name',
+    name: 'Named Extensions',
```
joaoantoniocardoso:
```suggestion
    name: 'Named Extensions (v2)',
```

PR #3479, opened 2025-08-09
Core: Services: Cable Guy: Fix DHCP Client marker being removed

joaomariolago:
Make sure that when `_execute_route` is called it keeps the DHCP requisition IP marker `0.0.0.0` `Mode.Client` alive if the link changed it's state somehow.

Related to #3466

## Summary by Sourcery

Preserve the DHCP client IP marker (0.0.0.0 Mode.Client) when link state changes by propagating a new include_dhcp_markers flag through interface retrieval methods and using it in route execution.

Bug Fixes:
- Prevent the DHCP client marker from being removed when executing route changes.

Enhancements:
- Add include_dhcp_markers parameter to get_interfaces, get_interface_by_name, and get_ethernet_interfaces to optionally retain DHCP markers in interface listings.

### DIFF
--- a/core/services/cable_guy/api/manager.py
+++ b/core/services/cable_guy/api/manager.py
@@ -383,8 +383,17 @@ class EthernetManager:
 
         self._update_interface_settings(interface_name, saved_interface)
 
-    def get_interface_by_name(self, name: str) -> NetworkInterface:
-        for interface in self.get_ethernet_interfaces():
+    def get_interface_by_name(self, name: str, include_dhcp_markers: bool = False) -> NetworkInterface:
+        """Get interface by name.
+
+        Args:
+            name (str): Interface name
+            include_dhcp_markers (bool, optional): Include DHCP markers. Defaults to False
+
+        Returns:
+            NetworkInterface: Interface object
+        """
+        for interface in self.get_ethernet_interfaces(include_dhcp_markers):
             if interface.name == name:
                 return interface
         raise ValueError(f"No interface with name '{name}' is present.")
@@ -393,11 +402,13 @@ class EthernetManager:
         return next((i for i in self._settings.content if i.name == name), None)
 
     # pylint: disable=too-many-locals
-    def get_interfaces(self, filter_wifi: bool = False) -> List[NetworkInterface]:
+    def get_interfaces(self, filter_wifi: bool = False, include_dhcp_markers: bool = False) -> List[NetworkInterface]:
         """Get interfaces information
 
         Args:
             filter_wifi (boolean, optional): Enable wifi interface filtering
+            include_dhcp_markers (boolean, optional): DHCP marker is the IP 0.0.0.0 AddressMode.Client, used to
+            inform cable guy that a dynamic IP should be acquired if available.
 
         Returns:
             List of NetworkInterface instances available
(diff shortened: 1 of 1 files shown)


core/services/cable_guy/api/manager.py
```diff
@@ -431,6 +433,20 @@ def get_interfaces(self, filter_wifi: bool = False) -> List[NetworkInterface]:
                 else:
                     mode = AddressMode.Unmanaged if is_static_ip and valid_ip else AddressMode.Client
                 valid_addresses.append(InterfaceAddress(ip=ip, mode=mode))
+            # Check if there is a 0.0.0.0 AddressMode.Client in current interface on self._settings.content and if not in valid_addresses add it
+            interface_settings = self.get_saved_interface_by_name(interface)
+            if (
+                include_dhcp_markers
+                and interface_settings
+                and any(
+                    address.ip == "0.0.0.0" and address.mode == AddressMode.Client
```
patrickelectric: Just to be sure... can you compare ip with a string ?

Williangalvani:
we might need to backport this to 1.4

core/services/cable_guy/api/manager.py
joaoantoniocardoso: nope, and that was the point of #3463

joaomariolago: Specifically in this case we should be fine with the comparison, but added cast to make sure it is consistent in case in the future the address changes to a IPv4Address

core/services/cable_guy/api/manager.py
```diff
@@ -393,11 +393,13 @@ def get_saved_interface_by_name(self, name: str) -> Optional[NetworkInterface]:
         return next((i for i in self._settings.content if i.name == name), None)
 
     # pylint: disable=too-many-locals
-    def get_interfaces(self, filter_wifi: bool = False) -> List[NetworkInterface]:
+    def get_interfaces(self, filter_wifi: bool = False, include_dhcp_markers: bool = False) -> List[NetworkInterface]:
```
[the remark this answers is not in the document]
joaomariolago: As this code may be backported to a stable the changes in this PR will be kept minimal. #3499 will address the code complexity reduction since `get_interfaces` method is really convoluted and need to be spited.

core/services/cable_guy/api/manager.py
```diff
@@ -453,13 +469,13 @@ def get_interfaces(self, filter_wifi: bool = False) -> List[NetworkInterface]:
 
         return result
 
-    def get_ethernet_interfaces(self) -> List[NetworkInterface]:
+    def get_ethernet_interfaces(self, include_dhcp_markers: bool = False) -> List[NetworkInterface]:
         """Get ethernet interfaces information
```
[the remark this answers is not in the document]
joaomariolago: Missing docs comments are corrected.

joaoantoniocardoso (approved):

joaoantoniocardoso:
The code looks sane, but only @Williangalvani actually has the setup to reproduce the problem

patrickelectric:
waiting for @Williangalvani test to merge

Williangalvani (approved):
so this "joaomariolago/blueos-core:fix-removed-dhcp-marker" is actually the one I tested.
I'll test the backport now.

joaoantoniocardoso:
> Should be move this to 1.4 ?

It's already there: https://github.com/bluerobotics/BlueOS/pull/3505

Issue #3489, opened 2025-08-12
bug: versionchooser: remote versions is not working


nicoschmdt:
### Bug description

When trying to access remote versions the request returns and empty list of images for any remote repository informed (check https://github.com/docker/hub-feedback/issues/2484). This happens due to a necessary check BlueOS does to ensure the architecture is compatible.  However, DockerHub now requires authentication for returning this specific information.

### Steps to reproduce

1. Go to  the BlueOS Versions page
2. Try to fetch any remote repository

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

Williangalvani:
proposal:

<img width="1796" height="494" alt="Image" src="https://github.com/user-attachments/assets/cbd9c624-733b-41aa-bb68-d09c99eaabb1" />

PR #3490, opened 2025-08-13
Version Chooser: deal (badly) with dockerhub api changes

Williangalvani:
"Fix" #3489 
Screenshot 2025-08-13 at 06 46 54

## Summary by Sourcery

Handle DockerHub API change by falling back to placeholder TagMetadata for tags with no image data and add an integration test to validate TagFetcher against a real repository

Bug Fixes:
- Add fallback logic in fetch_remote_tags to include tags with empty "images" by creating placeholder TagMetadata entries

Tests:
- Introduce TestTagFetcher integration test to fetch and verify real tags from DockerHub for bluerobotics/blueos-core, skipping on network issues

### DIFF
--- a/core/services/versionchooser/utils/dockerhub.py
+++ b/core/services/versionchooser/utils/dockerhub.py
@@ -110,6 +110,22 @@ class TagFetcher:
                 my_architecture = get_current_arch()
                 valid_images = []
                 for tag in tags:
+                    images = tag["images"]
+                    if len(images) == 0:
+                        # this is a hack to deal with https://github.com/docker/hub-feedback/issues/2484
+                        # we lost the ability to properly identify the images as we dont have the digest,
+                        # and also the ability to filter for compatible architectures.
+                        # so we just add the tag and hope for the best.
+                        tag = TagMetadata(
+                            repository=repository,
+                            image=repository.split("/")[-1],
+                            tag=tag["name"],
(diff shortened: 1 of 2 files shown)


core/services/versionchooser/utils/dockerhub.py
```diff
@@ -110,6 +110,22 @@ async def fetch_remote_tags(self, repository: str, local_images: List[str]) -> T
                 my_architecture = get_current_arch()
                 valid_images = []
                 for tag in tags:
+                    images = tag["images"]
+                    if len(images) == 0:
+                        # this is a hack to deal with https://github.com/docker/hub-feedback/issues/2484
+                        # we lost the ability to properly identify the images as we dont have the digest,
+                        # and also the ability to filter for compatible architectures.
+                        # so we just add the tag and hope for the best.
+                        tag = TagMetadata(
```
joaoantoniocardoso, line 119: We should log a warning here

joaoantoniocardoso:
> > so we just add the tag and hope for the best.
> 
> I'm not against this quick fallback, but I want to understand what can happen in this case.

Okay, I'm good with the behavior -- versionchooser lists the wrong architecture images, but it fails when trying to apply. It would only be better if it had warnings to tell us (and also the frontend) what's happening.

PR #3507, opened 2025-08-25
Gripper updates 2

Williangalvani:
mixed feelings about the store. thoughts?
the idea is to slowly move logic away from .vue files

### DIFF
--- a/core/frontend/src/components/vehiclesetup/overview/gripper.vue
+++ b/core/frontend/src/components/vehiclesetup/overview/gripper.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-card class="ma-2 pa-2">
+  <v-card class="ma-2 pa-2" max-width="250px">
     <v-card-title class="align-center">
       Gripper
     </v-card-title>
@@ -14,15 +14,19 @@
     <v-card-text v-else>
       {{ toBoardFriendlyChannel(`SERVO${gripper}_FUNCTION`) }}
     </v-card-text>
+    <v-card-text v-if="misconfigured_gripper" class="red--text">
+      {{ misconfigured_gripper }}
+    </v-card-text>
   </v-card>
 </template>
 
 <script lang="ts">
 import Vue from 'vue'
 
+import ardupilot_capabilities from '@/store/ardupilot_capabilities'
 import autopilot_data from '@/store/autopilot'
 import autopilot from '@/store/autopilot_manager'
-import { BTN_FUNCTION } from '@/types/autopilot/parameter-sub-enums'
+import { BTN_FUNCTION, SERVO_FUNCTION } from '@/types/autopilot/parameter-sub-enums'
 
 import toBoardFriendlyChannel from './common'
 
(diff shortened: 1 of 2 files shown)


core/frontend/src/store/ardupilot_capabilities.ts
```diff
@@ -0,0 +1,29 @@
...
+})
+
+class ArdupilotCapabilitiesStore extends VuexModule {
+
+  get firmware_supports_actuators(): boolean {
+    return autopilot_data.parameter('ACTUATOR1_INC') !== undefined
+  }
+
+  get firmware_supports_light_functions(): boolean {
+    // TODO: light functions were introduced at the same time as actuators
```
Williangalvani, line 22:
```suggestion
    // light functions were introduced at the same time as actuators
```

core/frontend/src/components/vehiclesetup/overview/gripper.vue
```diff
@@ -39,25 +43,95 @@ export default Vue.extend({
...
-      const channel1 = [BTN_FUNCTION.SERVO_1_MIN_MOMENTARY, BTN_FUNCTION.SERVO_1_MAX_MOMENTARY]
-      const channel2 = [BTN_FUNCTION.SERVO_2_MIN_MOMENTARY, BTN_FUNCTION.SERVO_2_MAX_MOMENTARY]
-      const channel3 = [BTN_FUNCTION.SERVO_3_MIN_MOMENTARY, BTN_FUNCTION.SERVO_3_MAX_MOMENTARY]
+      const channel1 = [
+        BTN_FUNCTION.SERVO_1_MIN_MOMENTARY,
+        BTN_FUNCTION.SERVO_1_MAX_MOMENTARY,
+        BTN_FUNCTION.SERVO_1_MIN_TOGGLE,
+        BTN_FUNCTION.SERVO_1_MAX_TOGGLE,
+        BTN_FUNCTION.SERVO_1_MIN,
+        BTN_FUNCTION.SERVO_1_MAX,
```
ES-Alexander, line 52: Haven't these been renamed now to `actuator_n_*`? Not sure how that's best handled from a display perspective, given they were/are called `servo_n_*` for older firmware versions 🤷‍♂️

core/frontend/src/components/vehiclesetup/overview/gripper.vue
```diff
@@ -39,25 +43,95 @@ export default Vue.extend({
...
+        BTN_FUNCTION.SERVO_4_MAX,
+      ]
+      const channel5 = [
+        BTN_FUNCTION.SERVO_5_MIN_MOMENTARY,
+        BTN_FUNCTION.SERVO_5_MAX_MOMENTARY,
+        BTN_FUNCTION.SERVO_5_MIN_TOGGLE,
+        BTN_FUNCTION.SERVO_5_MAX_TOGGLE,
+        BTN_FUNCTION.SERVO_5_MIN,
+        BTN_FUNCTION.SERVO_5_MAX,
+      ]
```
ES-Alexander, line 85: What about Actuator 6?

core/frontend/src/components/vehiclesetup/overview/gripper.vue
```diff
@@ -14,15 +14,19 @@
     <v-card-text v-else>
       {{ toBoardFriendlyChannel(`SERVO${gripper}_FUNCTION`) }}
     </v-card-text>
+    <v-card-text v-if="misconfigured_gripper" class="red--text">
+      {{ misconfigured_gripper }}
```
ES-Alexander, line 18: This seems worth expanding to more general checks of missing outputs corresponding to assigned button functions:
- Lights N
- Actuator N (with gripper mentioned as a potential intended output)
- Relay N

but that's perhaps out of scope for this PR 🤷‍♂️

core/frontend/src/store/ardupilot_capabilities.ts
```diff
@@ -0,0 +1,29 @@
...
+    return autopilot_data.parameter('ACTUATOR1_INC') !== undefined
```
ES-Alexander, line 23: Is there pre-filtering somewhere, that makes this apply only to ArduSub?
IIRC Actuator outputs are more general ArduPilot, but Lights is a Sub thing - I could be wrong on both accounts though 🤷‍♂️

ES-Alexander (changes requested):
Not sure if you wanted reviews yet, since it's still marked as draft, but you did bring it up in the meeting so figured I'd at least take an initial look :-)

ES-Alexander:
Needs to include outputs displaying as RCIN9/10 when using a firmware that supports the new lights outputs (if the output parameters are not using Lights1/2 values)

PR #3512, opened 2025-08-30
Add color background per version type in version-chooser

patrickelectric:
image

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionCard.vue
+++ b/core/frontend/src/components/version-chooser/VersionCard.vue
@@ -1,6 +1,7 @@
 <template>
   <v-sheet
-    class="d-flex flex-column align-center my-4 pa-4 flex-sm-row py-sm-0"
+    class="d-flex flex-column align-center my-1 pa-4 flex-sm-row py-sm-0"
+    :style="{ background: `linear-gradient(135deg, #1E1E1E 20%, ${backgroundColor} 100%)` }"
     :elevation="$vuetify.breakpoint.xs ? 1 : 0"
   >
     <div class="d-flex">
@@ -165,6 +166,7 @@ import TimeAgo from 'javascript-time-ago'
 import en from 'javascript-time-ago/locale/en.json'
 import Vue, { PropType } from 'vue'
 
+import { isBeta, isStable } from '@/components/kraken/Utils'
 import settings from '@/libs/settings'
 import helper from '@/store/helper'
 import { Dictionary } from '@/types/common'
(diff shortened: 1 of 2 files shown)


rafaellehmkuhl:
The idea is awesome and will help a lot when picking a version.

It would be nice to try sticking more to the current color palette. Probably @ArturoManzoli or Elisa can help.

core/frontend/src/components/version-chooser/VersionCard.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <v-sheet
-    class="d-flex flex-column align-center my-4 pa-4 flex-sm-row py-sm-0"
+    class="d-flex flex-column align-center my-1 pa-4 flex-sm-row py-sm-0"
+    :style="{ background: `linear-gradient(135deg, #1E1E1E 20%, ${backgroundColor} 100%)` }"
```
ES-Alexander:
```suggestion
    :style="{ background: `linear-gradient(135deg in oklab, var(--v-outline-base) 20%, ${backgroundColor} 100%)` }"
```
`oklab` and `oklch` are in baseline CSS, so we may as well use perceptually linear colour spaces where we can, _at least_ for transitions/gradients.

I'm defaulting to `oklab` here because it's slightly more intuitive as a transitive space, but `oklch` provides more vibrant (/less grey/muddy) gradients at the "cost" of occasionally introducing interim colours as it sweeps around the hue wheel, so is also worth considering 🤷‍♂️ 

---

Per my earlier review comment, the starting colour should account for the light/dark mode status. I'm unsure whether this is the best approach, but it does use a variable that's already used elsewhere in the software (e.g. for failsafe icon outlines), so at least avoids defining something completely new to maintain and document.

patrickelectric: Thanks, the outline-base did not work, I moved to transparent and the result is much better.

patrickelectric: @ES-Alexander check the new images in the PR

ES-Alexander: > the outline-base did not work

~Curious - I wonder why...~ Ahh yeah, it's the opposite side of the lightness spectrum, so the text gets hidden.  

> I moved to transparent and the result is much better.

Fair enough - nice! :-)

> check the new images in the PR

@patrickelectric it looks good, but the "Running" indicator is seemingly hard to read in light mode, and I expect it would be even harder to read on stable versions (where the background is green). Maybe the background of the "Running" rectangle should be made less transparent, and possibly less green?

PR #3529, opened 2025-09-10
core: libs: commonwealth: settings: managers: Deal with invalid settings file

patrickelectric:


core/libs/commonwealth/src/commonwealth/settings/managers/pykson_manager.py
```diff
@@ -120,7 +119,7 @@ def get_settings_version_from_filename(filename: pathlib.Path) -> int:
                 self._settings = PyksonManager.load_from_file(self.settings_type, valid_file)
                 logger.debug(f"Using {valid_file} as settings source")
                 return
-            except SettingsFromTheFuture as exception:
+            except Exception as exception:
                 logger.debug("Invalid settings, going to try another file:", exception)
 
         self._settings = PyksonManager.load_from_file(self.settings_type, self.settings_file_path())
```
joaomariolago: We still need to protect this last `load_from_file` call on the end of function.
Eg. Fails:
 * V1 and V2 are present, if V2 is corrupted and V1 also the last call will be a load on a corrupted V2
 * If only V2 is present, the last call will also happens on a corrupted V2

core/libs/commonwealth/src/commonwealth/settings/managers/pydantic_manager.py
```diff
@@ -124,7 +123,7 @@ def get_settings_version_from_filename(filename: pathlib.Path) -> int:
                 self._settings = PydanticManager.load_from_file(self.settings_type, valid_file)
                 logger.debug(f"Using {valid_file} as settings source")
                 return
-            except SettingsFromTheFuture as exception:
+            except Exception as exception:
                 logger.debug("Invalid settings, going to try another file:", exception)
 
         self._settings = PydanticManager.load_from_file(self.settings_type, self.settings_file_path())
```
joaomariolago: Same here, we still need to protect this `load_from_file` call on the end of function.

joaomariolago (changes requested):

core/libs/commonwealth/src/commonwealth/settings/managers/pydantic_manager.py
patrickelectric: nice, fixed

core/libs/commonwealth/src/commonwealth/settings/managers/pykson_manager.py
patrickelectric: nice, fixed

PR #3535, opened 2025-09-11
Fix mypy

patrickelectric:
## Summary by Sourcery

Master was broken on master and a bunch of things are now broken, this PR fixes mypy and everything necessary in master to get things moving again

### DIFF
--- a/core/services/wifi/main.py
+++ b/core/services/wifi/main.py
@@ -17,7 +17,7 @@ from fastapi import FastAPI, HTTPException, status
 from fastapi.staticfiles import StaticFiles
 from fastapi_versioning import VersionedFastAPI, version
 from loguru import logger
-from tabulate import tabulate
+from tabulate import tabulate  # type: ignore
 from uvicorn import Config, Server
 
 from exceptions import BusyError


core/libs/commonwealth/src/commonwealth/utils/logs.py
```diff
@@ -117,7 +117,7 @@ def sink(message: _handler.Message) -> None:
             "CRITICAL": 5,
         }
 
-        record = message.record
+        record = message.record # type: ignore
```
joaoantoniocardoso: The `record` is a `Record` from `loguru`, do we really need to ignore its type?

core/services/kraken/api/v2/routers/manifest.py
```diff
@@ -120,7 +119,7 @@ async def enable(identifier: str) -> None:
 
 @manifest_router_v2.post("/{identifier}/disable", status_code=status.HTTP_204_NO_CONTENT)
 @manifest_to_http_exception
-async def disable(identifier: str) -> Response:
+async def disable(identifier: str) -> None:
```
joaoantoniocardoso, line 122: `None` seems suspicious -- does it mean this call could never fail (as in, return some internal error)?

patrickelectric: There is nothing to return, the called functions do not return as well

core/libs/commonwealth/src/commonwealth/utils/logs.py
patrickelectric: This was fixed on the Message type

Issue #3525, opened 2025-09-05
Add a copy button to get the mavlink endpoint as a configuration string for mavlink routers


patrickelectric:
the idea is to have a copy button under the mavlink endpoint
that outputs something like: tcpin:0.0.0.0:14550

rafaellehmkuhl:
Allows for easily copying the address and paste it in the GCS.

Should copy the currently address being used (e.g.: blueos-avahi.local, 192.168.2.2, 192.168.3.1, etc).

PR #3555, opened 2025-09-21
frontend: autopilot: add copy mavlink endpoint button

nicoschmdt:
fix: #3525

Adds a button which copies the the mavlink information in the following format: `currentHost:mavlinkPort`.

image

### Results:
- `blueos-avahi.local:14660` (when accesing through blueos-avahi)
- `192.168.100.37:14000` (when accesing through the IP)


joaoantoniocardoso:
> > I believe that it would be nicer to also copy with `udpin` (for servers), `udpout` for clients and the same thing for tcp.
> > What you think @joaoantoniocardoso ?
> 
> Yup. Ideally the user should be able to just paste this in the cockpit MAVLink2Rest URI field and it should work.

agree!

PR #3578, opened 2025-10-01
[backport] kraken: prevent extension removal when Docker fails to pull

joaoantoniocardoso:
This is a backport of #3563 into 1.4


patrickelectric:
@joaoantoniocardoso I believe that this should not be backported after the comments of @joaomariolago, we should deal with it in a different way, backport multiple PRs or we could not deal with this issue, since it's mostly common in development.

joaomariolago:
We can backport this one instead #3568. This is the one that actually fixes the issue, others are related to code cleaning etc...

PR #3598, opened 2025-10-13
helper: send mavlink message when BlueOS is in factory mode

### LINKED ISSUE #2236: Notify user via mavlink when blueos is in factory mode

nicoschmdt:
fix #2236.

side note: the issue suggested to put this functionality in the Helper service and I can see why. But at the same time it seems like a good option to put it in the VersionChooser service.

image: nicoschmdt/blueos-core:notify-factory-mode

## Summary by Sourcery

Notify external systems via MAVLink when BlueOS boots in factory mode by querying the VersionChooser service and sending a STATUSTEXT message from the Helper service

New Features:
- Send a MAVLink STATUSTEXT message if BlueOS is in factory mode at startup

Enhancements:
- Add Helper.statustext_message to format MAVLink STATUSTEXT messages
- Implement check_and_notify_factory_mode with retry logic querying the VersionChooser service

### DIFF
--- a/core/services/helper/main.py
+++ b/core/services/helper/main.py
@@ -19,7 +19,9 @@ from urllib.parse import urlparse
 from uuid import UUID
 
 import psutil
+import requests
 from bs4 import BeautifulSoup
+from commonwealth.mavlink_comm.MavlinkComm import MavlinkMessenger
 from commonwealth.utils.apis import GenericErrorHandlingRoute, PrettyJSONResponse
 from commonwealth.utils.decorators import temporary_cache
 from commonwealth.utils.general import (
@@ -231,6 +233,8 @@ class Helper:
 
     MAX_ATTEMPTS_LEFT = 3
     attempts_left: Dict[int, int] = {}
+
+    mavlink2rest = MavlinkMessenger()
 
     @staticmethod
     # pylint: disable=too-many-arguments,too-many-branches,too-many-locals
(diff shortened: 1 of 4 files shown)


core/services/helper/main.py
```diff
@@ -711,6 +738,8 @@ async def main() -> None:
     config = Config(app=app, host="0.0.0.0", port=Helper.PORT, log_config=None)
     server = Server(config)
 
+    await Helper.check_and_notify_factory_mode()
```
patrickelectric: GCS may not be connected when the service stars, we should have a loop that runs every N minutes to notify the user.

core/services/helper/main.py
```diff
@@ -536,6 +540,27 @@ def setup_nginx_route(metadata: ServiceMetadata, port: int) -> bool:
             logging.info(f"file updated: {filename}")
         return True
 
+    @staticmethod
+    def statustext_message(text: str) -> Dict[str, Any]:
```
patrickelectric: This should be moved to MAVLinkMessenger

core/services/helper/main.py
```diff
@@ -19,7 +19,9 @@
 from uuid import UUID
 
 import psutil
+import requests
```
patrickelectric, line 22: It should update pyproject of helper

patrickelectric (approved):

Issue #3591, opened 2025-10-11
core: MAVLink SYSID should match vehicle


rmackay9:
### Current behaviour

If the user changes the mavlink system id from the default, the BlueOS processes that publish to mavlink should use that same system id.  I suspect the easiest way to do this is to have a user setting somewhere.  Please note that extensions also need access to this system id somehow unless we want to force extension developers to always have a local setting.

Below is a screen shot of BlueOS 1.5.0-beta15 connected to a vehicle with SysID = 2

<img width="1221" height="659" alt="Image" src="https://github.com/user-attachments/assets/72229732-4602-4e68-842c-1359be5931ed" />

### Expected or desired behaviour

BlueOS (and its extensions) should use the same system id as the vehicle

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

rmackay9:
Hi @joaoantoniocardoso, it does indeed work, thanks very much!

PR #3600, opened 2025-10-13
core: start-blueos-core: Allow MAV_SYSTEM_ID configuration via env vars

joaoantoniocardoso:
This should allow us to set `MAV_SYSTEM_ID` via env var in `/root/.config/blueos/bootstrap/startup.json`.

Helps #3591

## How to test

To test it:
1. Change `SYSID_THISMAV` parameter to `2`
2. Add the following to your `/root/.config/blueos/bootstrap/startup.json`:

```json
"environment": [
  "MAV_SYSTEM_ID=2"
]
```

Then, open MAVLink-Server UI (http://blueos-avahi.local:8080) and check the SYS ID:

image

### DIFF
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -10,7 +10,7 @@ SERVICES_PATH=$BLUEOS_PATH/services
 TOOLS_PATH=$BLUEOS_PATH/tools
 
 # MAVLink configuration
-MAV_SYSTEM_ID=1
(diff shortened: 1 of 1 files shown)


Williangalvani (approved):
tested with both unset and set environment variables.

patrickelectric (approved):

PR #3614, opened 2025-10-22
core: dockerfile: allow hashes longer than 8 chars

Williangalvani:


core/Dockerfile
```diff
@@ -73,7 +73,7 @@ ENV GIT_DESCRIBE_TAGS=${GIT_DESCRIBE_TAGS:-0.0.0-0-g00000000}
 RUN <<-EOF
 set -e
 
-    if ! echo "$GIT_DESCRIBE_TAGS" | grep -Eq -- '-[0-9]+-g[a-f0-9]{8}$'; then
+    if ! echo "$GIT_DESCRIBE_TAGS" | grep -Eq -- '-[0-9]+-g[a-f0-9]{8,}$'; then
```
patrickelectric, line 76: We also need to update bootstrap as well, both use the same regex logic

joaoantoniocardoso: It is worth checking the CI, too (I think it has some tag regex there)

Williangalvani: the two other regexes I found are already allowing for longer hashes

Issue #2343, opened 2024-01-24
changing boards and firmware should trigger frontend to re-download parameters


Williangalvani:
I'm not 100% if the firmware change does it, but the board change does not trigger it.

PR #3623, opened 2025-10-28
frontend: autopilot: trigger frontend to re download autopilot parameters

nicoschmdt:
fix: #2343

Had to add a timeout to ignore old mavlink messages and avoid having the store set up dated information regarding the autopilot firmware.

Tested going between the SITL board and the Navigator. The `Change Board` and `Install Firmware` code flow were both considered.

## Summary by Sourcery

Add a timeout to filter out outdated MAVLink parameter messages after resets, refactor parameter storage indexing to use parameter names, and ensure the frontend store is reset when changing boards to trigger fresh parameter downloads.

Bug Fixes:
- Ignore stale MAVLink parameter messages received within 500ms of a fetcher reset to prevent outdated autopilot data

Enhancements:
- Use parameter names as keys in ParametersTable for direct lookups and simplify update logic
- Trigger a reset of the autopilot parameters store in BoardChangeDialog to force re-download of parameters

### DIFF
--- a/core/frontend/src/types/autopilot/parameter-fetcher.ts
+++ b/core/frontend/src/types/autopilot/parameter-fetcher.ts
@@ -23,6 +23,8 @@ export default class ParameterFetcher {
 
   min_update_interval_ms = 300
 
+  reset_timestamp = 0
+
   constructor() {
     this.setupWs()
   }
@@ -36,6 +38,7 @@ export default class ParameterFetcher {
     this.total_params_count = null
     this.watchdog_last_count = 0
     this.parameter_table.reset()
+    this.reset_timestamp = performance.now()
   }
 
   updateStore(): void {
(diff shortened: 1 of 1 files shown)


core/frontend/src/types/autopilot/parameter-table.ts
```diff
@@ -187,18 +187,17 @@ export default class ParametersTable {
...
-    if (!index) {
+    if (!(param_name in this.parametersDict)) {
       // This is benign and will happen if we receive a parameter update before the parameters table
       // is fully populated. We can safely ignore it.
       console.info(`Unable to update param in store: ${param_name}. Parameter not yet loaded into ParametersTable.`)
       return
     }
-    this.parametersDict[parseInt(index[0], 10)].value = param_value
+    this.parametersDict[param_name].value = param_value
   }
```
patrickelectric, line 203: @nicoschmdt can you explain a little more this commit ? There is no entry in the commit body, but the ID should be unique as the param name

nicoschmdt: when I was changing boards through the install firmware button duplicates started to show up in the ardupilot parameters list. To fix this I changed the dictionary to use the parameter name instead

edit: checked if the error occured without the changes and it did not. I probably had some information cached which caused this error to happen.

patrickelectric:
> Should we backport it? The issue was from Jan 2024

yes

Issue #3006, opened 2024-11-20
Factory version should show the actual version on "BlueOS versions"


Williangalvani:
We should make it easier to know what the factory version actually is. It currently does display correctly on the bottom left

PR #3628, opened 2025-10-29
version-chooser: show BlueOS factory version

nicoschmdt:
fix: #3006 

The UI looks like this now: 
image

## Summary by Sourcery

Add end-to-end support for displaying the BlueOS factory version in the version chooser by introducing a new API endpoint, client utility, and UI update.

New Features:
- Add backend API endpoint to retrieve the BlueOS factory image version
- Implement getFactoryVersion utility in frontend to fetch the factory version from the new API
- Extend the version chooser UI to display the factory version alongside the 'factory' tag

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionCard.vue
+++ b/core/frontend/src/components/version-chooser/VersionCard.vue
@@ -7,14 +7,14 @@
       <v-list-item-avatar>
         <v-icon :class="current ? 'green' : 'grey'">
           {{ remote ? 'mdi-earth' : '' }}
-          {{ current ? 'mdi-checkbox-blank-circle' : 'mdi-checkbox-blank-circle-outline' }}
+          {{ displayIcon }}
         </v-icon>
       </v-list-item-avatar>
 
       <div>
         <p
           class="text-body-1 ma-0"
-          v-text="image.tag"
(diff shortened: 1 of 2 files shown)


core/services/versionchooser/utils/chooser.py
```diff
@@ -150,6 +150,35 @@ async def generate() -> AsyncGenerator[bytes, None]:
...
+            factory_details = await self.client.images.inspect("bluerobotics/blueos-core:factory")
+            env_info = factory_details["Config"]["Env"]
+
+            for info in env_info:
+                if "=" not in info:
+                    continue
+
+                key, val = info.split("=", maxsplit=1)
+                if key == "GIT_DESCRIBE_TAGS" and val:
+                    version = str(val).split("-", maxsplit=1)[0]
```
ES-Alexander:
```suggestion
                    version = str(val).rsplit("-", maxsplit=2)[0]
```
Splitting once from the left means beta images would only display the start numbers, as though the version is stable. It would also display just the first word in a custom tag. Factory _should_ be stable, but users can technically change it and I think it doesn't hurt to reflect which version it actually is.

Splitting from the right like this gets rid of the extra stuff and just returns the Git tag name (whatever it may be).

PR #3643, opened 2025-11-06
core: tools: linux2rest: Update to 0.8.1

patrickelectric:
Add jornal and kernel msgs to be accessible via zenoh


joaoantoniocardoso (approved):
🎉

Issue #3417, opened 2025-07-08
core: Move python services API to be zenoh queryables


patrickelectric:
Since we are moving the communication to zenoh, we should move the current endpoints to zenoh queryables

PR #3657, opened 2025-12-04
Add zenoh queryables to the kraken service

nicoschmdt:
Part of #3417

test image: nicoschmdt/blueos-core:queryables-kraken

## Summary by Sourcery

Integrate zenoh queryables into the Kraken service and expose key v2 API endpoints over zenoh alongside existing HTTP routes.

New Features:
- Expose Kraken manifest, container, extension, and jobs v2 endpoints as zenoh queryables via a new ZenohRouter integration.

Enhancements:
- Introduce a reusable zenoh_helper module that manages a shared zenoh session, maps FastAPI route paths to zenoh key expressions, and decorates routers to capture route metadata.
- Wire zenoh queryable routing into the Kraken FastAPI application with a dedicated zenoh router hierarchy and ensure the zenoh session is closed on application shutdown.

### DIFF
--- a/core/services/kraken/api/app.py
+++ b/core/services/kraken/api/app.py
@@ -1,4 +1,6 @@
+from contextlib import asynccontextmanager
 from os import path
+from typing import AsyncGenerator
 
 # Routers
 from api.v1.routers import extension_router_v1, index_router_v1
@@ -10,14 +12,24 @@ from api.v2.routers import (
     manifest_router_v2,
 )
 from commonwealth.utils.apis import GenericErrorHandlingRoute
+from commonwealth.utils.zenoh_helper import ZenohRouter, ZenohSession
+from config import SERVICE_NAME
 from fastapi import FastAPI
 from fastapi.responses import RedirectResponse
 from fastapi.staticfiles import StaticFiles
 from fastapi_versioning import VersionedFastAPI
 
+
+@asynccontextmanager
+async def lifespan(fastapi_app: FastAPI) -> AsyncGenerator[None, None]:  # pylint: disable=unused-argument
+    yield
+    zenoh_session.close()
+
+
 application = FastAPI(
     title="Kraken API",
     description="Kraken is the BlueOS service responsible for installing and managing extensions.",
+    lifespan=lifespan,
 )
 application.router.route_class = GenericErrorHandlingRoute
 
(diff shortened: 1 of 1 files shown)


core/libs/commonwealth/src/commonwealth/utils/zenoh_helper.py
```diff
@@ -0,0 +1,140 @@
...
+        if self._executor:
+            self._executor.shutdown()
+            self._executor = None
+
+    def zenoh_config(self) -> None:
+        configuration = {
+            "mode": "client",
+            "connect/endpoints": ["tcp/127.0.0.1:7447"],
+            "adminspace": {"enabled": True},
+            "metadata": {"name": "zenoh-queryables"},
```
patrickelectric: this should be the name of the service, otherwise all services will show as `zenoh-queryables` when fetching the metadata of each client, no ? (Check BlueOS zenoh network page) 
also, can we somehow use the same client of the log ? or share the clients ?
We already have on in commonwealth. I believe that's better to organize it first and share a single client for both, not sure if it's the best idea, but it _sounds_ better than having multiple clients per service.

core/services/kraken/api/app.py
```diff
@@ -33,7 +47,16 @@
...
 
-application = VersionedFastAPI(application, prefix_format="/v{major}.{minor}", enable_latest=True)
+application = VersionedFastAPI(application, prefix_format="/v{major}.{minor}", enable_latest=True, lifespan=lifespan)
+
+# Zenoh
+zenoh_router = ZenohRouter("kraken")
+zenoh_router.include_router(zenoh_container_router)
+zenoh_router.include_router(zenoh_extension_router)
+zenoh_router.include_router(zenoh_jobs_router)
+zenoh_router.include_router(zenoh_manifest_router)
```
patrickelectric: This `include_router` sounds a bit unnecessary, you could extract the router via something like...

```python
for route in app.router.routes:
    if isinstance(route, APIRoute):
        print(route.path)
```

nicoschmdt: But those are `ZenohRouter`s, they are not included in the fastapi routers. I had to declare one `ZenohRouter` in each file that had an `APIRoute`.

core/services/kraken/api/app.py
```diff
@@ -33,7 +44,12 @@
 application.include_router(jobs_router_v2)
 application.include_router(manifest_router_v2)
 
-application = VersionedFastAPI(application, prefix_format="/v{major}.{minor}", enable_latest=True)
+# Zenoh
+zenoh_session = ZenohSession()
+zenoh_router = ZenohRouter("kraken")
+zenoh_router.add_routes_to_zenoh(application)
```
patrickelectric, line 50: shouldn't have a start before here ?

patrickelectric: Would it make more sense to have start being called by the constructor and have the argument name in the constructor itself ?

PR #3658, opened 2025-12-05
core: Split main python runtime in 2 separate environments

joaomariolago:
* Split BlueOS core runtime into two uv-managed virtualenvs (primary `core/pyproject.toml` and secondary `core/python-venv2/
    pyproject.toml`) so services can upgrade their deps independently;
*  Extends `.hooks/pre-push` to deal with both environments, added **--fix**, **--fix-primary**, **--fix-secondary** options;
*  Relaxed the coverage gate for the secondary environment temporally to 65 % (default remains 75 % for the primary) due to individual services like versionchooser not meeting minimum criteria alone.

## Summary by Sourcery

Introduce a second Python runtime environment for selected services and wire it into the core image build and startup configuration.

New Features:
- Add a dedicated secondary Python virtual environment for services requiring newer dependencies while keeping the existing primary environment intact.

Bug Fixes:
- Add missing python-multipart dependency to the versionchooser service to satisfy its runtime requirements.

Enhancements:
- Update the Python libraries installation script to support installing into multiple virtual environments or system Python based on environment variables.
- Adjust core Docker image to build, sync, and export both primary and secondary Python virtual environments and expose them via environment variables and PATH configuration.
- Split core Python dependencies so selected services (kraken, versionchooser, commonwealth consumers) are moved into the new secondary environment.

## Summary by Sourcery

Introduce a secondary Python virtual environment for selected core services and update tooling, Docker image, and hooks to support running and checking code across both environments.

New Features:
- Add a dedicated secondary Python runtime workspace under core/python-venv2 for services requiring newer dependencies.

Bug Fixes:
- Add the missing python-multipart dependency to the versionchooser service to satisfy its runtime requirements.

Enhancements:
- Generalize the Python tools installation script to install into one or more virtual environments or the system Python based on configuration.
- Adjust the core Docker image to build, populate, and expose both primary and secondary Python virtual environments via environment variables and PATH.
- Split core Python dependencies so versionchooser (and related workspace members) are managed in the new secondary environment.
- Extend the Git pre-push hook with shared helpers and a python_checks library to run formatting, linting, type checking, and tests separately for primary and secondary environments with configurable coverage thresholds.
- Relax the coverage gate for the secondary environment to a lower threshold while keeping the default for the primary environment.

### DIFF
--- a/.hooks/lib/common.sh
+++ b/.hooks/lib/common.sh
@@ -22,7 +22,7 @@ require_commands() {
 }
 
 check_required_tools() {
-  require_commands python3 parallel uv poetry shellcheck
+  require_commands docker python3 parallel uv poetry shellcheck
 }
 
 run_nginx_validation() {


.hooks/lib/common.sh
```diff
@@ -0,0 +1,72 @@
...
+    fi
+}
+
+check_required_tools() {
+    require_command uv "uv is not installed. Aborting. Please install it and try again."
+    require_command poetry "poetry is not installed. Aborting. Please install it and try again."
+    require_command shellcheck "shellcheck is not installed. Aborting. Please install it and try again."
+    require_command parallel "parallel is not installed. Aborting. Please install it and try again."
+    require_command python3 "python3 is not installed. Aborting. Please install it and try again."
+}
```
patrickelectric:
```bash
require_commands() {
  for cmd in "$@"; do
    if ! command -v "$cmd" >/dev/null 2>&1; then
      printf 'Required tool not found: %s\n' "$cmd" >&2
      exit 1
    fi
  done
}

check_required_tools() {
  require_commands python3 parallel uv poetry shellcheck
}
```

.hooks/lib/common.sh
```diff
@@ -0,0 +1,72 @@
...
+        exit 1
+    fi
+}
+
+check_required_tools() {
+    require_command uv "uv is not installed. Aborting. Please install it and try again."
+    require_command poetry "poetry is not installed. Aborting. Please install it and try again."
+    require_command shellcheck "shellcheck is not installed. Aborting. Please install it and try again."
+    require_command parallel "parallel is not installed. Aborting. Please install it and try again."
+    require_command python3 "python3 is not installed. Aborting. Please install it and try again."
```
patrickelectric: missing docker requirement

.hooks/lib/get_secondary_scope.py
```diff
@@ -0,0 +1,75 @@
...
+    core_dir = args.core_dir.resolve()
+    pyproject = args.pyproject.resolve()
+    prefix = args.prefix
+
+    if not pyproject.exists():
+        return 0
+
+    document = load_toml(pyproject)
+    if not document:
+        return 0
```
patrickelectric:
```py
if not (document := load_toml(pyproject)):
    return 0
```
Can you explain why we failing in silence for such cases ? It's not clear.

core/start-blueos-core
```diff
@@ -114,7 +118,7 @@ PRIORITY_SERVICES=(
 
 SERVICES=(
     # This services are not prioritized because they are not fundamental for the vehicle to work
-    'kraken',0,"nice -19 $SERVICES_PATH/kraken/main.py"
+    'kraken',0,"nice -19 $BLUEOS_PYTHON_BIN_PRIMARY $SERVICES_PATH/kraken/main.py"
```
patrickelectric: Why only kraken is in primary ?

core/python-venv2/.python-version
```diff
@@ -0,0 +1 @@
+3.11
```
patrickelectric, line 1: Should we update python version as well ? Since it's the second env..

core/tools/install-python-libs.sh
```diff
@@ -8,4 +8,32 @@ TOOLS=(
...
-parallel --halt now,fail=1 '/home/pi/tools/{}/setup-python-libs.sh' ::: "${TOOLS[@]}"
+determine_target_envs() {
+    if [ -n "$BLUEOS_PYTHON_ENVS" ]; then
+        read -r -a TARGET_ENVS <<< "$BLUEOS_PYTHON_ENVS"
+    elif [ -n "$VIRTUAL_ENV" ]; then
+        TARGET_ENVS=("$VIRTUAL_ENV")
+    else
+        TARGET_ENVS=("")
+    fi
+}
```
patrickelectric: This can be simple replaced by:
```sh
TARGET_ENVS=(${BLUEOS_PYTHON_ENVS:-${VIRTUAL_ENV:-}})
```

core/tools/install-python-libs.sh
```diff
@@ -8,4 +8,32 @@ TOOLS=(
...
+    fi
+
+    parallel --halt now,fail=1 '/home/pi/tools/{}/setup-python-libs.sh' ::: "${TOOLS[@]}"
+}
+
+determine_target_envs
+
+for env_path in "${TARGET_ENVS[@]}"; do
+    install_for_env "$env_path"
+done
```
patrickelectric:
```sh
for env in "${TARGET_ENVS[@]}"; do
  [[ -n $env ]] && export VIRTUAL_ENV="$env" || unset VIRTUAL_ENV
  echo "Installing python tools for ${env:-system python}"
  parallel --halt now,fail=1 '/home/pi/tools/{}/setup-python-libs.sh' ::: "${TOOLS[@]}"
done
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+                filtered+=("$file")
+            fi
+        fi
+    done
+
+    printf '%s\n' "${filtered[@]}"
+}
+
+collect_mypy_targets() {
+    local scope="$1"
```
patrickelectric:
```
local scope="${1:?scope must be 'primary' or 'secondary'}"
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+    local -a all_projects=()
+    mapfile -t all_projects < <(git -C "$CORE_DIR" ls-files '*/pyproject.toml')
+
+    local -a filtered=()
+    local project
+    for project in "${all_projects[@]}"; do
```
patrickelectric:
```sh
git -C "$CORE_DIR" ls-files '*/pyproject.toml' |
while read -r project; do
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+        [[ -z "$project" ]] && continue
+        case "$project" in
+            libs/*|services/*)
+                ;;
+            *)
+                continue
+                ;;
+        esac
```
patrickelectric:
```
[[ $project == libs/* || $project == services/* ]] || continue
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+
+        if path_matches_any "$project" "${SECONDARY_SCOPE_PATHS[@]}"; then
+            if [ "$scope" = "secondary" ]; then
+                filtered+=("$(dirname "$project")")
+            fi
+        else
+            if [ "$scope" = "primary" ]; then
+                filtered+=("$(dirname "$project")")
+            fi
+        fi
```
patrickelectric, line 92:
```
if path_matches_any "$project" "${SECONDARY_SCOPE_PATHS[@]}"; then
  [[ $scope == secondary ]] && echo "$dir"
else
  [[ $scope == primary ]] && echo "$dir"
fi
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+    done
+
+    printf '%s\n' "${filtered[@]}"
```
patrickelectric: There is no need to have a filtered field, that makes more complex. If we need a list we can just echo our way around.

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+path_matches_any() {
+    local file="$1"
+    shift || true
+    local scope_path
+    for scope_path in "$@"; do
+        [[ -z "$scope_path" ]] && continue
+        if [[ "$file" == "$scope_path" ]] || [[ "$file" == "$scope_path/"* ]]; then
+            return 0
+        fi
+    done
```
patrickelectric, line 57: Perhaps..
```sh
for scope_path in "$@"; do
    [[ -z "$scope_path" ]] && continue
    [[ $file == "$scope_path" || $file == "$scope_path/"* ]] && return 0
done
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+    return 1
+}
+
+collect_python_files() {
+    local scope="$1"
```
patrickelectric:
```
local scope="${1:?scope must be 'primary' or 'secondary'}"
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+    local -a all_python_files=()
+    mapfile -t all_python_files < <(git -C "$CORE_DIR" ls-files '*.py')
+
+    local -a filtered=()
+    local file
+    for file in "${all_python_files[@]}"; do
```
patrickelectric:
```
git -C "$CORE_DIR" ls-files '*.py' |
while read -r file; do
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+        [[ -z "$file" ]] && continue
+        if path_matches_any "$file" "${SECONDARY_SCOPE_PATHS[@]}"; then
+            if [ "$scope" = "secondary" ]; then
+                filtered+=("$file")
+            fi
+        else
+            if [ "$scope" = "primary" ]; then
+                filtered+=("$file")
+            fi
+        fi
```
patrickelectric, line 74:
```
if path_matches_any "$file" "${SECONDARY_SCOPE_PATHS[@]}"; then
    [[ "$scope" == secondary ]] && echo "$file"
else
    [[ "$scope" == primary ]] && echo "$file"
fi
```

.hooks/lib/python_checks.sh
```diff
@@ -0,0 +1,199 @@
...
+    done
+
+    printf '%s\n' "${filtered[@]}"
```
patrickelectric: There is no need to have a filtered field, that makes more complex. If we need a list we can just echo our way around.

core/python-venv2/.python-version
joaomariolago: I think that we gonna update one step at time, first all libraries when moving services from primary venv to secondary venv and them python version when returning from secondary to primary. Otherwise diffs will be to large and could also increase image size.

.hooks/lib/common.sh
joaomariolago: Added!

.hooks/lib/common.sh
joaomariolago: It was not present before, but is a good enforcement, add in a separate commit

.hooks/lib/get_secondary_scope.py
joaomariolago: Changed a bit the logic, now it should fail in most of cases and announce to user, only in case secondary .toml file has no members it will instead of silent failing warn user and force disabling secondary environment checks.

core/start-blueos-core
joaomariolago: Nice catch, code from old commit when both versionchooser and kraken were being migrated. Already removed it.

patrickelectric (approved):

Issue #3267, opened 2025-04-14
core: frontend: vehicle-setup: reflect output pin function


ES-Alexander:
### Current behaviour

#3122 introduced some additional configuration options, but does so assuming every output pin is configured as some form of standard servo, which is unintuitive for those that have other limit cases / output types.

### Expected or desired behaviour

1. There should be a toggle switch for reversal state
2. Motor outputs should display `MOT_PWM_MIN/MAX` instead of the servo ones (especially since at the moment the servo ones are not even considered for motors)
3. Relay outputs should show the on/off state and `RELAYn_DEFAULT` rather than servo limits and trim
    - some example UI in #2766 
5. Relay outputs should also display the function as `Relay X` (instead of Disabled / GPIO, since the `SERVOn_FUNCTION parameter is ignored for relays)
    - ideally they should also be _selectable_ as "Relay X" in the dropdown, and set the corresponding `RELAYx_PIN` value, and the `SERVOn_FUNCTION` to GPIO

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

ES-Alexander:
Suggested interface:

![Image](https://github.com/user-attachments/assets/f1c0e8c3-a29e-48a2-8fbf-e47c2ee4a373)

All the relevant configuration has been merged into the flat page, but if we want to we could replicate (or expand upon) those interfaces in the channel function selector popup window.

PR #3668, opened 2025-12-08
Custom pwm output widgets

Williangalvani:
adds custom widgets/optiosn for:

regular servo
motors
actuators
lights
gpio
relays

Screenshot 2025-12-18 at 11 07 05
Screenshot 2025-12-18 at 11 08 04
Screenshot 2025-12-18 at 11 08 33
Screenshot 2025-12-18 at 11 08 54

helps #3267

## Summary by Sourcery

Introduce specialized editors for different servo function types in the parameter editor dialog to support motor- and GPIO-specific PWM configuration while keeping a generic range editor for other functions.

Enhancements:
- Refactor ServoFunctionEditorDialog to dynamically render a function-specific editor component based on the selected servo function option.
- Add a reusable ServoFunctionRangeEditor component for configuring generic servo PWM ranges via slider and inline parameter editors.
- Add a ServoFunctionMotorEditor with motor-specific PWM range handling that integrates MOT_PWM_TYPE, MOT_PWM_MIN, and MOT_PWM_MAX parameters and supports angle-based configurations.
- Introduce placeholder ServoFunctionActuatorEditor and ServoFunctionGpioEditor components to support future specialized UIs for actuators and GPIO outputs.

## Summary by Sourcery

Introduce specialized PWM output configuration editors for different servo functions and improve GPIO-based output handling in the PWM setup UI.

New Features:
- Add dynamic servo function editor dialog that selects function-specific editors for motors, GPIO, actuators, and lights while keeping a generic range editor for other outputs.
- Add dedicated motor configuration editor that integrates MOT PWM type, shared or per-motor PWM ranges, and reverse direction handling.
- Add GPIO configuration editor that maps servo outputs to GPIO-capable functions such as relays and leak detection, with dedicated relay and leak setup UIs.
- Add actuator and lights editors that expose relevant parameters (e.g., increment, joystick steps) alongside a common PWM range configuration slider.

Enhancements:
- Extract a reusable ServoFunctionRangeEditor component for visual PWM min/trim/max configuration that can be shared across different output types.
- Improve PWM setup table to present GPIO-assigned servo functions with user-friendly labels by resolving which parameter currently owns the GPIO pin.
- Initialize current board metadata in the PWM setup view to support board-specific servo-to-GPIO mapping.

### DIFF
--- a/core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
+++ b/core/frontend/src/components/parameter-editor/InlineParameterEditor.vue
@@ -207,10 +207,16 @@ export default Vue.extend({
     },
     param_value() {
       this.updateSelectedFlags()
-      if (this.last_sent_value === undefined) {
-        this.internal_new_value = this.param_value
-        this.internal_new_value_as_string = String(this.internal_new_value)
+      if (
+        this.last_sent_value !== undefined
+         && this.internal_new_value !== this.last_sent_value
+         && this.param_value === this.last_sent_value
+         && this.param_value === this.last_sent_value
+      ) {
+        return
       }
+      this.internal_new_value = this.param_value
+      this.internal_new_value_as_string = String(this.internal_new_value)
     },
   },
   mounted() {


core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
```diff
@@ -0,0 +1,137 @@
...
+  computed: {
+    motor_number(): number | undefined {
+      const function_name = this.param.options?.[this.param.value] as string
+      const motor_number = function_name.match(/\d+/)?.[0]
+      if (!motor_number) return undefined
+      return parseInt(motor_number, 10)
+    },
+    trim_param(): Parameter | undefined {
+      if (this.mot_pwm_type === PwmTypes.Normal) {
+        return autopilot.parameter(`MOT_${this.motor_number}_TRIM`)
```
ES-Alexander: This should be nothing, right? There's no trim control for Normal motors (IIRC that's why we needed a different motor type).

core/frontend/src/components/vehiclesetup/PwmSetup.vue
```diff
@@ -455,13 +482,25 @@ export default Vue.extend({
...
+    parameterToUserFriendlyText(param: Parameter): string {
+      if (param.value === -1) { // GPIO
+        const servo_number = parseInt(/\d+/g.exec(param.name)?.[0] ?? '0', 10)
+        const gpio = this.servo_to_gpio[servo_number]
+        const param_using_this_gpio = this.gpio_to_parameter[gpio]
+        if (param_using_this_gpio) {
+          const pretty_name = param_using_this_gpio.name.split('_')[0].toLowerCase().toTitle()
+          return `GPIO [${pretty_name}]`
+        }
+        return `Unknown GPIO ${gpio}`
```
ES-Alexander:
```suggestion
        return `GPIO ${gpio} [Unknown]`
```
I think consistent formatting would be nicer. Starting with GPIO also makes it easier to find in the list.

core/frontend/src/components/parameter-editor/ServoFunctionGpioEditor.vue
```diff
@@ -0,0 +1,182 @@
...
+        return this.servo_number
+      }
+      // We are assuming most boards follow the same GPIO numbering scheme as the Pixhawk1
+      // Aux channels start at 50, and Main channels start at 101
+
+      if (this.servo_number <= 8) {
+        return this.servo_number + 100
+      }
+      return this.servo_number - 9 + 50
+    },
```
ES-Alexander, line 82: Would it be possible to reuse the `servo_to_gpio` function from `PwmSetup.vue`?

core/frontend/src/components/parameter-editor/ServoFunctionGpioEditor.vue
```diff
@@ -0,0 +1,182 @@
...
+          return-object
+          hide-details
+          @change="onChange"
+        />
+      </v-col>
+    </v-row>
+    <relay-setup
+      v-if="function_type_using_this_gpio?.name?.includes('RELAY')"
+      :relay-parameter="function_type_using_this_gpio"
+    />
```
ES-Alexander, line 31: I think there should be a `leak-setup` here, that allows setting the "logic level when dry" (`LEAKn_LOGIC`).

We _could_ also add the `FS_LEAK_ENABLE` failsafe Action parameter, although I'm unsure whether it's better to just point the user to the failsafes page or something...

core/frontend/src/components/parameter-editor/ServoFunctionActuatorEditor.vue
```diff
@@ -0,0 +1,73 @@
...
+        <inline-parameter-editor
+          :label="actuator_inc_param?.name"
+          :param="actuator_inc_param"
+        />
+      </v-col>
+      <v-row>
+        <v-col cols="12">
+          <v-alert
+            v-if="!has_joystick_function_configured"
+            type="error"
```
ES-Alexander: I'm unsure whether this should be an error or just a warning, since someone may just want to control an Actuator programatically (e.g. with a Cockpit slider, or some Actions).

That said, I suppose _right now_ it's not possible to do that, so we could make it an error and update it once `DO_SET_ACTUATOR` support is added?

core/frontend/src/components/parameter-editor/ServoFunctionGpioEditor.vue
```diff
@@ -0,0 +1,182 @@
...
+        <v-select
+          v-model="selected_function_type"
+          :items="options"
+          item-text="text"
+          item-value="value"
+          return-object
+          hide-details
+          @change="onChange"
+        />
+      </v-col>
```
ES-Alexander, line 26: I think these should be in a combined column, to avoid the redundant function name:

Image

core/frontend/src/components/parameter-editor/RelaySetup.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+  <v-card class="mt-4 mb-4">
+    <v-card-text>
+      <v-row align="center">
+        <v-col cols="6">
+          <span class="text-h6">Relay {{ relay_number }}</span>
```
ES-Alexander: I think this is redundant, and should be left to the UI outside this component (see comment further down for suggested alternative).

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
```diff
@@ -0,0 +1,137 @@
+<template>
+  <div>
+    <v-row class="mb-4">
+      <v-col cols="6">
+        <span class="text-h6">PWM Type</span>
```
ES-Alexander, line 5:
```suggestion
        <span class="text-h6">Motor Control Signal (shared)</span>
```
Nice to make it clear that this can't be set per-motor, and I feel like the "PWM Type" part is redundant with the `MOT_PWM_TYPE` description over the dropdown.

Undecided whether "shared" or "all" is better.

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
```diff
@@ -0,0 +1,137 @@
...
+    </v-row>
+
+    <servo-function-range-editor
+      v-if="mot_pwm_type_is_supported"
+      :param="param"
+      :min-param="min_param"
+      :trim-param="trim_param"
+      :max-param="max_param"
+      title="Motor Configuration"
+      tooltip="Adjust the minimum, trim, and maximum PWM values for this motor"
```
ES-Alexander: Could we make these dynamic based on the selected type? So the title says "Motor Configuration (shared)" when Normal is selected, and just "Motor Configuration" when PWMAngle? And for the tooltip, end with "all motors" and skip the `, trim,` when Normal, or just end with "this motor" when PWMAngle?

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
```diff
@@ -0,0 +1,137 @@
...
+  Normal = 0,
+  OneShot = 1,
+  OneShot125 = 2,
+  Brushed = 3,
+  DShot150 = 4,
+  DShot300 = 5,
+  DShot600 = 6,
+  DShot1200 = 7,
+  PWMRange = 8,
+  PWMAngle = 9,
```
ES-Alexander, line 66: Do we expect users to be familiar with these options?

Just curious whether we want `PWMAngle` to be put second for Sub users, and possibly with some extra text to explain that it supports trim and independent PWM ranges...

core/frontend/src/components/parameter-editor/ServoFunctionLightsEditor.vue
```diff
@@ -0,0 +1,47 @@
...
+      <v-col cols="6">
+        <span class="text-h6">Joystick Steps</span>
+      </v-col>
+      <v-col cols="6">
+        <inline-parameter-editor
+          :label="steps_param?.name"
+          :param="steps_param"
+        />
+      </v-col>
+      <servo-function-range-editor
```
ES-Alexander, line 13: This seems to be only partially implemented (there are no linked servo channel variables?), and weird because it's on one side:
Image

Would also be nice if it showed ticks based on the joystick steps, and allowed testing with the normal trim tick (maybe even with percentage-based tick labels).

core/frontend/src/components/parameter-editor/ServoFunctionActuatorEditor.vue
```diff
@@ -0,0 +1,73 @@
...
+          >
+            <span>Joystick functions not configured for this actuator.<br />
+              Please configure a joystick button for this actuator using your GCS of choice.</span>
+          </v-alert>
+        </v-col>
+      </v-row>
+      <servo-function-range-editor
```
ES-Alexander, line 24: This is off to the side:

Image

Would be nice to be able to test it with a central thumb (although I suppose that's not possible for actuators at the moment, so just for the record). I'm unsure whether actuators support an adjustable "center" value, or if it's just (min+max)/2.

ES-Alexander (changes requested):
Niiice!! Will hopefully be some significant improvements for the UX :-)

Various interface suggestions, plus some code and functionality ones.
Some great steps towards being able to do something like [this table idea](https://github.com/bluerobotics/BlueOS/issues/3267#issuecomment-3653255744) in the longer term :-)

core/frontend/src/components/parameter-editor/ServoFunctionActuatorEditor.vue
Williangalvani: good catch! I'm making it a warning

core/frontend/src/components/parameter-editor/ServoFunctionGpioEditor.vue
Williangalvani: I'm not following. can you draw for me? 🫠

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
Williangalvani: nice. done.

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
Williangalvani: yeah I`m not sure of what to do here... I kinda want to hide the others.

core/frontend/src/components/parameter-editor/ServoFunctionLightsEditor.vue
Williangalvani: fixed the unlink params. I think the ticks and testing will need to wait

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
ES-Alexander: They're valid, so some people might want to use them, but agree it would be nice to hide them for at least normal BR vehicle users... Could maybe push the extra ones to pirate mode when the user is running standard BR firmware + frame combos?

core/frontend/src/components/parameter-editor/ServoFunctionGpioEditor.vue
Williangalvani: ping @ES-Alexander

core/frontend/src/components/parameter-editor/RelaySetup.vue
```diff
@@ -0,0 +1,90 @@
...
+            :on-value="1"
+            label="Invert Output"
+          />
+          <parameter-switch
+            v-if="default_parameter"
+            :parameter="default_parameter"
+            :off-value="0"
+            :on-value="1"
+            label="Startup State"
+          />
```
ES-Alexander, line 19:
```suggestion
          <parameter-switch
            v-if="default_parameter"
            :parameter="default_parameter"
            :off-value="0"
            :on-value="1"
            label="Startup State"
          />
          <parameter-switch
            v-if="inverted_parameter"
            :parameter="inverted_parameter"
            :off-value="0"
            :on-value="1"
            label="Invert Output"
          />
```
Input over output makes more sense I think, with how we read.

core/frontend/src/components/parameter-editor/RelaySetup.vue
```diff
@@ -0,0 +1,90 @@
...
+              v-tooltip="'Turn relay on'"
+              class="ma-4"
+              @click="setRelay(1)"
+            >
+              <v-icon small left>
+                mdi-power
+              </v-icon>
+              Relay On
+            </v-btn>
+          </v-row>
```
ES-Alexander, line 45:
```suggestion
          <v-row>
            <v-btn
              v-tooltip="'Turn relay on'"
              class="ma-4"
              @click="setRelay(1)"
            >
              <v-icon small left>
                mdi-power
              </v-icon>
              Relay On
            </v-btn>
          </v-row>
          <v-row>
            <v-btn
              v-tooltip="'Turn relay off'"
              class="ma-4"
              @click="setRelay(0)"
            >
              <v-icon small left>
                mdi-power-off
              </v-icon>
              Relay Off
            </v-btn>
          </v-row>
```
Similar idea with this one - you usually have to turn something on before you can turn it off.

core/frontend/src/components/parameter-editor/ServoFunctionMotorEditor.vue
```diff
@@ -0,0 +1,142 @@
...
+      const motor_number = function_name.match(/\d+/)?.[0]
+      if (!motor_number) return undefined
+      return parseInt(motor_number, 10)
+    },
+    shared_min_max_param(): boolean {
+      return this.mot_pwm_type === PwmTypes.Normal
+    },
+    trim_param(): Parameter | undefined {
+      if (this.mot_pwm_type === PwmTypes.PWMAngle) {
+        return autopilot.parameter(`MOT_${this.motor_number}_TRIM`)
```
ES-Alexander: This should be `SERVO{pin_number}_TRIM`. Currently the correct parameter is showing up (presumably as a fallback because `MOT_n_TRIM` params don't exist), but the card for it is disabled:

Image

core/frontend/src/components/parameter-editor/ServoFunctionRangeEditor.vue
```diff
@@ -0,0 +1,509 @@
...
+            v-show="thumbType !== 'trim' || trimParam"
+            :key="thumbType"
+            class="slider-thumb"
+            :class="{ active: activeThumb === index }"
+            :style="{ left: `${getThumbPosition(index)}%` }"
+            @mousedown.stop="startDragging(index)"
+          >
+            <div class="thumb-label">
+              {{ thumbType }}: {{ getThumbValue(index) }}
+            </div>
```
ES-Alexander, line 49: Not a blocker, but ideally this would be higher up, so the labels point to the top of the circles - they currently point to the middle of them:

Image

core/frontend/src/components/parameter-editor/ServoFunctionActuatorEditor.vue
```diff
@@ -0,0 +1,83 @@
...
+  },
+  computed: {
+    actuator_number(): number | undefined {
+      const option_name = this.param.options?.[this.param.value] as string
+      const actuator_number = option_name?.match(/\d+/)?.[0]
+      if (!actuator_number) return undefined
+      return parseInt(actuator_number, 10)
+    },
+    actuator_inc_param(): Parameter | undefined {
+      return autopilot.parameter(`ACTUATOR${this.actuator_number}_INC`)
```
ES-Alexander, line 60: This is nonsensical:

Image

(0.01µs as the increment would require 80000 steps to get from min to max), but that's actually an ArduSub issue. 

The unit of the implemented functionality seems to actually be proportion of range (e.g. `value=0.01` &rightarrow; 1% of range per step), but that makes more sense to fix in ArduSub than to try to temporarily mask in BlueOS.

ES-Alexander (changes requested):
Functionality generally seems ok (from some minimal testing), but I did find a few small interface issues. Nearly done I reckon! :-D

Didn't have time to check on my Pixhawk setup, so haven't checked leak detector.

PR #3677, opened 2025-12-10
core: libs: commonwealth: general: Add ignore field in delete_everything

patrickelectric:
## Summary by Sourcery

Add support for excluding specific paths when recursively deleting configuration data and use it to preserve critical commander configuration directories during settings reset.

Bug Fixes:
- Update commander settings reset to avoid deleting bootstrap and ardupilot-manager configuration directories by passing them as ignored paths.

Enhancements:
- Extend delete_everything utility to accept an ignore list of paths that are not removed during recursive deletion.

### DIFF
--- a/core/services/commander/main.py
+++ b/core/services/commander/main.py
@@ -3,7 +3,6 @@ import asyncio
 import json
 import logging
 import os
-import shutil
 import subprocess
 import time
 from enum import Enum
@@ -167,11 +166,13 @@ async def do_eeprom_update(i_know_what_i_am_doing: bool = False) -> Any:
 async def reset_settings(i_know_what_i_am_doing: bool = False) -> Any:
     check_what_i_am_doing(i_know_what_i_am_doing)
     # Be sure to not delete bootstrap to avoid going back to factory image
-    bootstrap_config_path = "/root/.config/bootstrap/startup.json"
-    temporary_location = "/tmp/bootstrap_startup.json"
-    shutil.copy2(bootstrap_config_path, temporary_location)
-    delete_everything(Path(appdirs.user_config_dir()))
-    shutil.copy2(temporary_location, bootstrap_config_path)
+    ignore = [
+        Path("/root/.config/ardupilot-manager"),
+        Path("/root/.config/bag-of-holding"),
+        Path("/root/.config/bootstrap"),
+        Path("/root/.config/kraken"),
+    ]
+    delete_everything(Path(appdirs.user_config_dir()), ignore=ignore)
 
 
 @app.post("/services/remove_log", status_code=status.HTTP_200_OK)


core/services/commander/main.py
```diff
@@ -167,11 +166,11 @@ async def do_eeprom_update(i_know_what_i_am_doing: bool = False) -> Any:
 async def reset_settings(i_know_what_i_am_doing: bool = False) -> Any:
     check_what_i_am_doing(i_know_what_i_am_doing)
     # Be sure to not delete bootstrap to avoid going back to factory image
-    bootstrap_config_path = "/root/.config/bootstrap/startup.json"
-    temporary_location = "/tmp/bootstrap_startup.json"
-    shutil.copy2(bootstrap_config_path, temporary_location)
-    delete_everything(Path(appdirs.user_config_dir()))
-    shutil.copy2(temporary_location, bootstrap_config_path)
+    ignore = [
+        Path("/root/.config/ardupilot-manager"),
```
ES-Alexander, line 170: Should we still delete the settings.json here? I do feel like the preferred board and MAVLink endpoints are BlueOS settings...

core/services/commander/main.py
```diff
@@ -167,11 +166,11 @@ async def do_eeprom_update(i_know_what_i_am_doing: bool = False) -> Any:
...
+        Path("/root/.config/bootstrap"),
+    ]
```
ES-Alexander, line 174: Should we avoid deleting kraken config? I'm not sure users expect to delete all their extensions when resetting their settings (which I believe is what happens if that config gets removed), but if that is indeed the intended behaviour it should definitely be specified in the warning dialog.

core/libs/commonwealth/src/commonwealth/utils/general.py
```diff
@@ -56,18 +56,40 @@ def get_host_os() -> HostOs:
     return HostOs.Other
 
 
-def delete_everything(path: Path) -> None:
+# pylint:disable=dangerous-default-value
+def delete_everything(path: Path, ignore: list[Path] = []) -> None:
```
joaomariolago:
```suggestion
def delete_everything(path: Path, ignore: list[Path] | None = None) -> None:
    ignore = [] if ignore is None else ignore
```

It is better to not use lists as default argument value since python will not reinitialize between calls and the same list will persist.

ES-Alexander (approved):
Nice!

Some important questions about the ignore list though.

core/libs/commonwealth/src/commonwealth/utils/general.py
patrickelectric: nice catch

ES-Alexander: Heh, sourcery suggested the same thing, although functionally speaking the ignore list is not mutated internally, or returned, so reusing the empty default is fine, so I can understand why Patrick opted to just disable the linter check instead 🤷‍♂️ 

I suppose the current approach may be very marginally more performant, and a little leaner to maintain, but may also introduce a little extra risk if someone decides to change how the function works in future such that they modify or return that list.

core/services/commander/main.py
patrickelectric: It's a kind of factory reset.. so it cleans mostly everything settings-wise

core/services/commander/main.py
patrickelectric: Updated.

patrickelectric:
@ES-Alexander and @joaomariolago, PR updated

core/services/commander/main.py
ES-Alexander: Yes - that's what I was saying should happen 😅

core/services/commander/main.py
```diff
@@ -167,11 +166,12 @@ async def do_eeprom_update(i_know_what_i_am_doing: bool = False) -> Any:
 async def reset_settings(i_know_what_i_am_doing: bool = False) -> Any:
     check_what_i_am_doing(i_know_what_i_am_doing)
     # Be sure to not delete bootstrap to avoid going back to factory image
-    bootstrap_config_path = "/root/.config/bootstrap/startup.json"
-    temporary_location = "/tmp/bootstrap_startup.json"
-    shutil.copy2(bootstrap_config_path, temporary_location)
-    delete_everything(Path(appdirs.user_config_dir()))
-    shutil.copy2(temporary_location, bootstrap_config_path)
+    ignore = [
+        Path("/root/.config/ardupilot-manager/firmware/logs"),
```
ES-Alexander:
```suggestion
        Path("/root/.config/ardupilot-manager/firmware/logs"),
        Path("/root/.config/ardupilot-manager/firmware/scripts"),
        Path("/root/.config/ardupilot-manager/firmware/terrain"),
        Path("/root/.config/ardupilot-manager/logs"),
```
It may be easier to just ignore the whole firmware folder (plus the (t)logs folder). I'm unsure whether we want to delete the current Navigator firmware and the like? They don't seem like BlueOS settings 🤷‍♂️

patrickelectric: I believe that we should remove them, and keep only the logs, scripts can be failing or wrong updated.. I prefer the clean it as much as possible.

ES-Alexander: > keep only the logs

You're currently only keeping bin logs. Tlogs seemingly get generated in the logs folder that's outside the firmware folder (or at least they used to? I have one there anyway).

> scripts can be failing or wrong updated..

I agree, but I feel like that's better suited to a genuine "factory reset" feature of the whole onboard computer, since users with custom scripts probably don't consider them to be _settings_ of BlueOS.

Also, will that break BlueBoat configs? I don't imagine "I reset my BlueOS settings and now my NavLight doesn't work" would make sense to many users, since in principle the autopilot's "internal" configuration and setup are independent of BlueOS's settings.

In my mind, "which flight controller board does BlueOS want to connect to?", and "which MAVLink router is BlueOS using?" are BlueOS settings (which are stored in `ardupilot-manager/settings.json`), but everything else related to the autopilot is not a BlueOS setting.

ES-Alexander: If you want a less complex path though, I think it would be a smaller evil to just ignore the whole `ardupilot-manager` directory (like you started with), and accept that there are technically a couple of BlueOS settings hiding in there 🤷‍♂️

patrickelectric: Updated to a more global approach including bag-of-holding and ardupilot-manager

ES-Alexander (approved):
Seems reasonable to me!

PR #3728, opened 2026-01-21
add zenoh support for POST, PUT and DELETE requests

nicoschmdt:
A complement of #3657.

Code used for testing the endpoints can be found [here](https://gist.github.com/nicoschmdt/495b5b0731b85d2671032b6afd7f7d08).

## Summary by Sourcery

Extend Zenoh integration to support additional HTTP methods and richer response handling across FastAPI routes.

New Features:
- Route Zenoh POST, PUT, and DELETE requests to FastAPI versioned endpoints alongside existing GET handling.
- Support publishing handler responses, including streaming responses, back over Zenoh via a response topic namespace.
- Parse and type-cast Zenoh payload/attachment parameters into typed FastAPI handler arguments, including support for complex types.

Enhancements:
- Introduce a route tree structure to match Zenoh samples to registered endpoints and HTTP methods using wildcard path segments.
- Generalize JSON serialization through a helper that robustly handles non-serializable data and errors.
- Refine Zenoh path cleaning and parameter matching to allow typed path parameters and more flexible routing.

### DIFF
--- a/core/libs/commonwealth/src/commonwealth/utils/tree.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/tree.py
@@ -1,0 +1,72 @@
+from typing import Any, Callable, Dict, List
+
+import zenoh
+
+
+class TreeNode:
+    def __init__(self, segment: str):
+        self.segment = segment
+        self.children: Dict[str, "TreeNode"] = {}
+        self.is_valid = False
+        self.methods: Dict[str, Callable[..., Any]] = {}
+
+    def add_child(self, child: "TreeNode") -> "TreeNode":
+        if child.segment in self.children:
+            return self.children[child.segment]
+        self.children[child.segment] = child
+        return child
+
+    def get_methods(self) -> Dict[str, Callable[..., Any]]:
+        return self.methods
+
+    def add_node(self, segments: List[str], method: str, func: Callable[..., Any]) -> None:
+        if self.segment != segments[0]:
+            return
+
+        if len(segments) == 1:
+            self.is_valid = True
+            self.methods[method] = func
+            return
+
+        child = self.add_child(TreeNode(segments[1]))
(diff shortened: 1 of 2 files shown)


core/libs/commonwealth/src/commonwealth/utils/tree.py
```diff
@@ -0,0 +1,72 @@
...
+        self.segment = segment
+        self.children: Dict[str, "TreeNode"] = {}
+        self.is_valid = False
+        self.methods: Dict[str, Callable[..., Any]] = {}
+
+    def add_child(self, child: "TreeNode") -> "TreeNode":
+        if child.segment in self.children:
+            return self.children[child.segment]
+        self.children[child.segment] = child
+        return child
```
patrickelectric, line 17:
```suggestion
        if child.segment not in self.children:
            self.children[child.segment] = child
        return self.children[child.segment]
```

core/libs/commonwealth/src/commonwealth/utils/tree.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get_methods(self) -> Dict[str, Callable[..., Any]]:
```
patrickelectric, line 19: I would recommend to use _ prefix for private properties and avoid using get prefix in functions. 

https://github.com/bluerobotics/software-guidelines/blob/master/guidelines/style.md

core/libs/commonwealth/src/commonwealth/utils/tree.py
```diff
@@ -0,0 +1,72 @@
+from typing import Any, Callable, Dict, List
+
+import zenoh
+
+
+class TreeNode:
+    def __init__(self, segment: str):
```
patrickelectric, line 7: can you add a description of how this class operates ? It's pretty complex just looking at the code.

PR #3760, opened 2026-02-04
frontend: remove unused ArduPilot parameter files from build

Williangalvani:
Only keep JSON files from the ArduPilot-Parameter-Repository, removing XML, RST, MD, and HTML documentation files that are not used by the frontend.

## Summary by Sourcery

Limit frontend build artifacts from the ArduPilot parameter repository to JSON-only assets to reduce bundle size.

Build:
- Add a Vite build hook that removes non-JSON ArduPilot parameter files from the built assets directory.
- Stop including ArduPilot-Parameter-Repository JSON files via the generic assetsInclude configuration, relying instead on the new cleanup step.

### DIFF
--- a/core/frontend/vite.config.js
+++ b/core/frontend/vite.config.js
@@ -77,9 +77,31 @@ export default defineConfig(({ command, mode }) => {
             next()
           })
         }
-      }
+      },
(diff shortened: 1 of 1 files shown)


patrickelectric:
just tested, the final assets folder does not contain such files, the dist folder should not be used to check them

PR #3790, opened 2026-02-11
Reboot vehicle only for ROVs if heartbeat lost 

patrickelectric:
## Summary by Sourcery

Limit automatic Ardupilot restarts on heartbeat loss to ROV (ArduSub) vehicles and introduce a typed MAVLink firmware classification.

New Features:
- Add a MavlinkFirmwareType enum to represent ArduPilot firmware families and wire it through vehicle type helpers.

Bug Fixes:
- Prevent non-ROV vehicles from being automatically rebooted on heartbeat loss by checking the firmware vehicle type before restarting Ardupilot.

Enhancements:
- Update vehicle type logic and helpers to use the new MavlinkFirmwareType enum instead of raw strings for firmware classification.

### DIFF
--- a/core/services/ardupilot_manager/autopilot_manager.py
+++ b/core/services/ardupilot_manager/autopilot_manager.py
@@ -190,6 +190,45 @@ class AutoPilotManager(metaclass=Singleton):
         # Serial or others that are not processes based
         return self.should_be_running
 
+    async def _detect_firmware_vehicle_type(self) -> None:
+        if self._firmware_vehicle_type is not None:
+            return
+        try:
+            vehicle_type = await self.vehicle_manager.get_vehicle_type()
+            self._firmware_vehicle_type = vehicle_type.mavlink_firmware_type()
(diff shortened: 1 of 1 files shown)


core/services/ardupilot_manager/autopilot_manager.py
```diff
@@ -188,6 +190,45 @@ def is_running(self) -> bool:
...
+        except Exception as error:
+            self._heartbeat_fail_count += 1
+            logger.warning(
+                f"heartbeat check failed ({self._heartbeat_fail_count}/{self._max_heartbeat_failures}): {error}"
+            )
+
+        if self._heartbeat_fail_count < self._max_heartbeat_failures:
+            return
+
+        logger.warning("Consecutive heartbeat failures threshold reached — restarting Ardupilot.")
```
Williangalvani, line 223: shouldnt this line be before the retart call? as is, this is tell users we are restarting even if running plane/copter

PR #3791, opened 2026-02-11
frontend: MainView: Limit update rate of model-viewer to 1rad

patrickelectric:
## Summary by Sourcery

Limit updates of the 3D model orientation in MainView to reduce unnecessary renders.

New Features:
- Introduce a dedicated Orientation type and lastOrientation state to track attitude data for the model viewer.

Enhancements:
- Add a computed attitudeMessage and watcher that only updates orientation when any axis changes by more than 1 radian.

### DIFF
--- a/core/frontend/src/views/MainView.vue
+++ b/core/frontend/src/views/MainView.vue
@@ -103,6 +103,12 @@ import mavlink_store_get from '@/utils/mavlink'
 import CPUUsage from '@/widgets/CpuPie.vue'
 import Networking from '@/widgets/Networking.vue'
 
+interface Orientation {
+  roll: number
+  pitch: number
+  yaw: number
+}
+
 interface AppItem {
   icon?: string
   title?: string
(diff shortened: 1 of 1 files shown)


core/frontend/src/views/MainView.vue
```diff
@@ -232,11 +239,23 @@ export default Vue.extend({
...
+      const { roll, pitch, yaw } = this.lastOrientation
+      return `${roll}rad ${-pitch}rad ${-yaw}rad`
+    },
+  },
+  watch: {
+    attitudeMessage(msg: Orientation | undefined): void {
+      if (!msg) return
+      const { roll, pitch, yaw } = msg
+      const last = this.lastOrientation
+      // Only update if any axis changed by more than 1 rad
```
rafaellehmkuhl: 1 rad equals 57 degrees. Maybe that should be degrees instead?

patrickelectric: my bad!

PR #3816, opened 2026-03-04
[1.4 backport] core: start-blueos-core: Add runtime tuning for improved stream performance

joaoantoniocardoso:
## Summary

Adds a `tune_performance()` function to `core/start-blueos-core` that applies runtime sysctl tuning on every container start. These changes were essential for improved WebRTC performance and stream stability.

### UDP buffer sizes (`wmem_max`, `rmem_max`, `wmem_default`, `rmem_default`, `netdev_max_backlog`)

From testing, the minimum UDP buffer size for stable streaming was actually **2.5 MB**, but we set it to **16 MB** instead to be future-proof (4K 60fps, 100 Mbps, etc). We can lower this if reviewers think it is too much.

Note that this potentially increases the RAM used by **all services that use UDP**, not just the streaming pipeline, since these are system-wide sysctl settings.

### Swappiness (`vm.swappiness = 10`)

This did **not** come from test data. It was chosen as a better configuration than the current default (60), since the team had already mentioned during a meeting that it would be good to reduce swap use. On SD cards, swap I/O is extremely slow and creates latency spikes, so preferring file cache reclamation over swapping should improve overall responsiveness.

### Real-time scheduler limits (`sched_rt_runtime_us = 950000`)

This is a **safe rationale**, not derived from test data. MCM uses many real-time (SCHED_RR/SCHED_FIFO) threads for streaming, and we do not want to let them starve the rest of the system. The setting reserves 5% of each CPU core's time (50 ms per second) for non-RT threads, preventing a scenario where aggressive RT scheduling locks out essential system processes.

## Summary by Sourcery

Introduce runtime system tuning in start-blueos-core to improve streaming performance and system responsiveness by adjusting kernel networking and scheduling parameters on container startup.

Enhancements:
- Apply UDP-related sysctl tuning at container start to increase buffer sizes and backlog for more stable high-bitrate streaming.
- Lower system swappiness to reduce reliance on slow swap storage and improve interactivity on constrained media.
- Configure real-time scheduler runtime limits to prevent RT threads from starving other system processes during intensive streaming workloads.

### DIFF
--- a/core/start-blueos-core
+++ b/core/start-blueos-core
@@ -220,6 +220,65 @@ prepare_cgroups() {
 
 prepare_cgroups
 
+# Runtime performance tuning (applied every container start via /sys/ bind mount)
+tune_performance() {
+  echo "Applying runtime performance tuning..."
+
+  # Network buffer tuning for high-bandwidth WebRTC streaming (55Mbps+)
+  # Maximum send buffer size per socket. The Pi's default was 180KB.
+  # WebRTC/SRTP sends ~55Mbps of UDP traffic. When the socket's send buffer is
+  # too small, the kernel drops outgoing packets during CPU load spikes. 16MB
+  # gives GStreamer/WebRTC headroom to queue bursts without drops. This is the
+  # single most impactful change for preventing video stutters.
+  echo $((16 * 1024 * 1024)) > /proc/sys/net/core/wmem_max 2>/dev/null || true
+  echo "wmem_max: $(cat /proc/sys/net/core/wmem_max 2>/dev/null || echo N/A)"
+
+  # Maximum receive buffer size per socket. Same logic but for incoming traffic
+  # (STUN/TURN negotiation, RTCP feedback, MAVLink). Less critical than wmem
+  # for outbound-heavy streaming, but prevents receive-side drops during
+  # inbound bursts.
+  echo $((16 * 1024 * 1024)) > /proc/sys/net/core/rmem_max 2>/dev/null || true
+  echo "rmem_max: $(cat /proc/sys/net/core/rmem_max 2>/dev/null || echo N/A)"
+
+  # Default buffer sizes for newly created sockets (when the application
+  # doesn't explicitly set SO_SNDBUF/SO_RCVBUF). The Pi default is ~180KB. If
+  # MCM or any BlueOS service opens a socket without tuning its buffers, it
+  # inherits this default. 1MB is a reasonable baseline that prevents casual
+  # sockets from starving under load.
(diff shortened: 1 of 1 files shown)


core/start-blueos-core
```diff
@@ -220,6 +220,60 @@ prepare_cgroups() {
...
+tune_performance() {
+  echo "Applying runtime performance tuning..."
+
+  # Network buffer tuning for high-bandwidth WebRTC streaming (55Mbps+)
+  # Maximum send buffer size per socket. The Pi's default was 180KB.
+  # WebRTC/SRTP sends ~55Mbps of UDP traffic. When the socket's send buffer is
+  # too small, the kernel drops outgoing packets during CPU load spikes. 16MB
+  # gives GStreamer/WebRTC headroom to queue bursts without drops. This is the
+  # single most impactful change for preventing video stutters.
+  echo 16777216 > /proc/sys/net/core/wmem_max 2>/dev/null
```
patrickelectric: `$((16 * 1024 * 1024))`

core/start-blueos-core
```diff
@@ -220,6 +220,60 @@ prepare_cgroups() {
...
+
+  # Maximum receive buffer size per socket. Same logic but for incoming traffic
+  # (STUN/TURN negotiation, RTCP feedback, MAVLink). Less critical than wmem
+  # for outbound-heavy streaming, but prevents receive-side drops during
+  # inbound bursts.
+  echo 16777216 > /proc/sys/net/core/rmem_max 2>/dev/null
```
patrickelectric: $((16 * 1024 * 1024))

core/start-blueos-core
```diff
@@ -220,6 +220,60 @@ prepare_cgroups() {
...
+
+  # Default buffer sizes for newly created sockets (when the application
+  # doesn't explicitly set SO_SNDBUF/SO_RCVBUF). The Pi default is ~180KB. If
+  # MCM or any BlueOS service opens a socket without tuning its buffers, it
+  # inherits this default. 1MB is a reasonable baseline that prevents casual
+  # sockets from starving under load.
+  echo 1048576  > /proc/sys/net/core/wmem_default 2>/dev/null
```
patrickelectric: `$((1024 * 1024))`

core/start-blueos-core
```diff
@@ -220,6 +220,60 @@ prepare_cgroups() {
...
+  echo 1048576  > /proc/sys/net/core/rmem_default 2>/dev/null
```
patrickelectric: `$((1024 * 1024))`

patrickelectric (approved):

Issue #3127, opened 2025-02-06
core: frontend: kraken: improve custom Extension installation process


ES-Alexander:
### Current behaviour

1. It is frequently the case that custom Extensions will be installed to test a development fix, in which case the Docker tag is the name of a repository branch. Since GitHub allows copying the name of the source branch from a pull request, it could serve as a convenient way to copy in the tag, but is currently hampered by the author name before the colon:
    
    <img width="1032" alt="Image" src="https://github.com/user-attachments/assets/33b79d05-1d0d-40e2-b4df-e70068c28f07" />
    which is not part of the Docker tag.
2. The save button is in the bottom left, which in inconsistent with the rest of the interface, and annoying to click on

### Expected or desired behaviour

Ideally:
- [ ] pasting a pull-request source branch name into the Docker tag field would automatically remove anything up to and including a colon, if there is a colon in the pasted text
- [ ] the save button would be in the bottom right
- [ ] pressing `Enter` (or `CTRL+Enter` / `CMD+Enter`) would submit the form, instead of needing to manually click the button

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

PR #3822, opened 2026-03-09
frontend: kraken: Improve custom installation extension

nicoschmdt:
fix: #3127

## Summary by Sourcery

Improve the extension creation modal UX for setting Docker tags and saving custom extensions.

New Features:
- Allow saving a new extension from the creation modal using the Ctrl+Enter keyboard shortcut.
- Automatically extract and populate the Docker tag from pasted values that include a repository-style prefix (e.g., image:tag) in the tag field.

### DIFF
--- a/core/frontend/src/components/kraken/modals/ExtensionCreationModal.vue
+++ b/core/frontend/src/components/kraken/modals/ExtensionCreationModal.vue
@@ -57,6 +57,7 @@
             v-model="new_extension.tag"
             label="Docker tag"
             :rules="[validate_tag]"
+            @paste="handleTagPaste"
           />
 
           <json-editor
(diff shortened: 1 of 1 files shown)


core/frontend/src/components/kraken/modals/ExtensionCreationModal.vue
```diff
@@ -303,6 +305,17 @@ export default Vue.extend({
         }, 2000) // Reset after 2 seconds
       }
     },
+    handleTagPaste(event: ClipboardEvent) {
+      const pastedText = event.clipboardData?.getData('text')?.trim()
+      if (!pastedText) return
+
+      const match = pastedText.match(/^[a-zA-Z0-9_-]+:(.+)$/)
+      if (match) {
```
patrickelectric: As the comment suggests, isn't it better to use lastIndexOf over match in this case ? We are not matching.. just extracting something after `:`
```
input.slice(input.lastIndexOf(':') + 1) 
```

PR #3827, opened 2026-03-13
Frontend: limit calls to /status to 0.5hz

Williangalvani:
I was seeing requests to /status at ~3hz.
I don't expect huge improvements of performance or bandwidth, but this will at the very least making look at the network tab of developer tools more... sane.

## Summary by Sourcery

Limit frontend backend status checks to avoid excessive /status requests.

Enhancements:
- Throttle backend status polling by skipping status checks if the backend was recently online.
- Track the last time the backend was confirmed online in the frontend store to support rate limiting of status checks.

### DIFF
--- a/core/frontend/src/store/frontend.ts
+++ b/core/frontend/src/store/frontend.ts
@@ -17,6 +17,8 @@ class FrontendStore extends VuexModule {
 
   backend_status_request = null as Promise<AxiosResponse> | null
 
+  last_backend_online_time: Date = new Date()
+
   backend_offline = false
 
   frontend_id = (() => {
@@ -32,6 +34,9 @@ class FrontendStore extends VuexModule {
 
   @Mutation
   setBackendOffline(offline: boolean): void {
+    if (!offline) {
+      this.last_backend_online_time = new Date()
+    }
     this.backend_offline = offline
   }
 }
--- a/core/frontend/src/utils/api.ts
+++ b/core/frontend/src/utils/api.ts
@@ -12,8 +12,14 @@ export const isBackendOffline = (error: any): boolean => {
   return false;
 }
 
+const minimum_backend_status_check_interval = 2000
+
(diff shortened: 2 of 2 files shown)


core/frontend/src/store/frontend.ts
```diff
@@ -32,6 +34,9 @@ class FrontendStore extends VuexModule {
 
   @Mutation
   setBackendOffline(offline: boolean): void {
+    if (!offline) {
```
ES-Alexander, line 37: Is it possible for this to be called even when the backend is already offline? If so, this should also check for the current backend state, to avoid falsely claiming the backend was online more recently than it actually was.

core/frontend/src/utils/api.ts
```diff
@@ -12,8 +12,14 @@ export const isBackendOffline = (error: any): boolean => {
   return false;
 }
 
+const minimum_backend_status_check_interval = 2000
```
ES-Alexander, line 15:
```suggestion
const minimum_backend_status_check_interval = 2000  // ms
```
Or in the variable name (like Sourcery suggested), if that's preferable 🤷‍♂️

core/frontend/src/store/frontend.ts
Williangalvani: I'm not sure I understand your question...

ES-Alexander: Nevermind, I think I misunderstood how this was working originally. If it is being set to online then it makes sense to update the last online time regardless of the current state 👍

core/frontend/src/utils/api.ts
patrickelectric: ping @Williangalvani

patrickelectric:
@Williangalvani can you rebase over master ?

Williangalvani:
This was already one (a bit differently) in https://github.com/bluerobotics/BlueOS/pull/4026

PR #3868, opened 2026-04-08
frontend: show snackbar notifications on parameter changes

Williangalvani:
changes snackbar to a stacking-alerters layout, showing up to 5 simultaneously.
Shows ALL parameter changes, but only when in developer mode

Screenshot 2026-04-08 at 15 47 47

## Summary by Sourcery

Introduce stacked alert notifications in the frontend and emit developer-mode messages for autopilot parameter changes once all parameters are loaded.

New Features:
- Display multiple alert messages concurrently in a centered stacked layout instead of a single snackbar.
- Emit info-level messages for parameter value changes when all parameters are loaded and developer mode is enabled.

Enhancements:
- Refine parameter loading status checks via an allParametersLoaded helper in ParameterFetcher.
- Extend the message manager with support for removing previously registered callbacks.
- Report parameter updates with old values only when the value actually changes.

### DIFF
--- a/core/frontend/src/components/app/Alerter.vue
+++ b/core/frontend/src/components/app/Alerter.vue
@@ -1,21 +1,24 @@
 <template>
-  <v-snackbar
-    v-model="show"
-    :timeout="timeout"
-  >
-    {{ message }}
-
-    <template #action="{ attrs }">
-      <v-btn
-        :color="color"
-        text
-        v-bind="attrs"
-        @click="show = false"
+  <div class="alerter-stack">
+    <v-slide-y-reverse-transition
+      group
+      leave-absolute
+    >
+      <v-alert
+        v-for="alert in alerts"
+        :key="alert.id"
+        :value="true"
+        :type="alertType(alert.level)"
+        class="alerter-item"
+        dense
+        dismissible
+        elevation="6"
+        @input="dismiss(alert.id)"
       >
-        Close
-      </v-btn>
-    </template>
-  </v-snackbar>
+        {{ alert.message }}
+      </v-alert>
+    </v-slide-y-reverse-transition>
+  </div>
 </template>
 
 <script lang="ts">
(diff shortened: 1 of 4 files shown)


core/frontend/src/components/app/Alerter.vue
```diff
@@ -1,56 +1,127 @@
...
+const MAX_VISIBLE = 5
+const DRAIN_INTERVAL_MS = 1000
+
+interface AlertEntry {
+  id: number
+  level: MessageLevel
+  message: string
+}
+
+let nextId = 0
```
patrickelectric, line 38: why not having it inside data ?

core/frontend/src/components/app/Alerter.vue
```diff
@@ -1,56 +1,127 @@
...
           return 'success'
         case MessageLevel.Error:
+        case MessageLevel.Critical:
           return 'error'
-        case MessageLevel.Info:
-          return 'info'
         case MessageLevel.Warning:
           return 'warning'
-        case MessageLevel.Critical:
-          return 'critical'
```
patrickelectric, line 47: Why critical moved to error ?

core/frontend/src/components/app/Alerter.vue
```diff
@@ -1,56 +1,127 @@
...
+        this.showAlert(this.queue.shift()!)
+      }
+      if (this.queue.length === 0) {
+        this.stopDrain()
+      }
+    },
+    startDrain() {
+      if (this.drainTimer) return
+      this.drainTimer = setInterval(() => {
+        const evictIdx = this.alerts.findIndex((a) => this.getTimeout(a.level) > 0)
```
patrickelectric, line 96: why not using `find` and use the object directly ?

core/frontend/src/components/app/Alerter.vue
```diff
@@ -1,56 +1,127 @@
...
+    },
+    promoteFromQueue() {
+      while (this.queue.length > 0 && this.alerts.length < MAX_VISIBLE) {
+        this.showAlert(this.queue.shift()!)
+      }
+      if (this.queue.length === 0) {
+        this.stopDrain()
+      }
+    },
+    startDrain() {
```
patrickelectric, line 93: I'm not following the logic here, dismiss removes the alert but it's not taking in consideration the timeout at all, so all notifications will only show for a second. Is that correct ?

patrickelectric (changes requested):
This pull request should be broken in multiple commits, there are a couple of refactorings in the middle of the new feature.

PR #3869, opened 2026-04-08
frontend: views: ExtensionManagerView: center extension manager tabs

### LINKED ISSUE #2942: core: frontend: extensions manager tabs aren't centered
### Current behaviour

The tabs are intended to be centred, but seem to be pushed over to the left by the settings cog button:
Screenshot 2024-10-03 at 3 13 40 PM

Not a functional problem, just looks weird.

The tabs should be centred within the card, such that the edge of the underline of the selected tab is aligned with the centre gap between the installed extensions.

When the space runs out I think the tabs should be cut off on the right by the settings cog, and able to scroll out from underneath it (which will be extra relevant once there's an extra tab for Bazaar).

### Prerequisites

nicoschmdt:
fix #2942 

before: 
image

zoomed-in:
image

after:
image

zoomed-in:
image

## Summary by Sourcery

Center the extension manager tab bar while keeping the settings button consistently positioned.

Enhancements:
- Adjust ExtensionManagerView toolbar layout so tabs fill available space and remain centered across viewport sizes.
- Pin the extension manager settings button to the right side of the toolbar on larger screens for a stable, predictable location.

### DIFF
--- a/core/frontend/src/views/ExtensionManagerView.vue
+++ b/core/frontend/src/views/ExtensionManagerView.vue
@@ -216,10 +216,17 @@
       :extension-name="selected_log_extension_name"
     />
     <v-toolbar>
-      <v-spacer />
+      <v-btn
+        icon
+        class="extension-toolbar-spacer"
+        aria-hidden="true"
+        tabindex="-1"
+      />
       <v-tabs
         v-model="tab"
         fixed-tabs
+        show-arrows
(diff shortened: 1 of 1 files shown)


ES-Alexander:
Seems to be better but still not quite fixed?

image

patrickelectric (approved):

PR #3883, opened 2026-04-17
frontend: kraken: prevent container-not-up icon from overlapping update button

nicoschmdt:
looks like this now:
image

## Summary by Sourcery

Indicate the container-not-running state on the installed extension card using a badge over the extension logo instead of an overlapping standalone icon.

Enhancements:
- Replace the absolute-positioned container-not-up icon with a warning badge over the extension logo to avoid overlapping other controls.
- Introduce a computed flag for the container-not-up state to centralize the visibility condition.
- Adjust badge styling to keep the warning icon compact and visually consistent with the card layout.

### DIFF
--- a/core/frontend/src/components/kraken/cards/InstalledExtensionCard.vue
+++ b/core/frontend/src/components/kraken/cards/InstalledExtensionCard.vue
@@ -8,28 +8,29 @@
         size="150"
       />
     </div>
-    <v-icon
-      v-if="extension.enabled && !container && !loading"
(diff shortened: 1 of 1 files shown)


ES-Alexander (changes requested):
From the picture, I think it would be better if the badge overlapped the logo more (e.g. the halfway points on the bottom and left of the badge should both be just within the logo rectangle) - it currently looks quite awkward (to me at least).

An example from Apple:

Image

ES-Alexander (approved):
Seems to be working well - thanks :-)

PR #3888, opened 2026-04-23
[backport] core: move setup_ssh() over to startup_blueos_update

joaoantoniocardoso:
This is a backport of https://github.com/bluerobotics/BlueOS/pull/3050 into 1.4

Reopened as #3898

## Summary by Sourcery

Set up SSH key generation and authorized_keys management during the BlueOS startup update process.

New Features:
- Automatically generate and persist an SSH key pair in the container config volume if none exists.
- Ensure the generated public key is added to the target user's authorized_keys file with correct ownership and permissions at startup.

Enhancements:
- Invoke SSH setup as part of the startup update script, with error logging that does not block the main startup flow.

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -5,6 +5,8 @@ import logging
 import os
 import re
 import time
+import subprocess
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


patrickelectric:
continues in #3898

PR #3897, opened 2026-04-24
[backport/1.4] longer hashes

patrickelectric:
This is a backport of https://github.com/bluerobotics/BlueOS/pull/3614 into 1.4

> @joaoantoniocardoso: .. I made this change locally while building the next branches, but forgot to send it.


patrickelectric:
I can't approve because I'm the author, but the original PR is from @joaoantoniocardoso and the commits are from @Williangalvani, so I would say that this is well approved.

PR #3898, opened 2026-04-24
[backport/1.4] core: move setup_ssh() over to startup_blueos_update

patrickelectric:
This is a backport of #3050

## Summary by Sourcery

New Features:
- Add SSH setup routine that generates and stores an SSH key pair in the container config volume and ensures the public key is present in the target user's authorized_keys on startup.

### DIFF
--- a/core/tools/blueos_startup_update/blueos_startup_update.py
+++ b/core/tools/blueos_startup_update/blueos_startup_update.py
@@ -5,6 +5,8 @@ import logging
 import os
 import re
 import time
+import subprocess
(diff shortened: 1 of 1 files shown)


patrickelectric:
I would hold this PR until we do the stable release for RadCam.

joaoantoniocardoso:
> I would hold this PR until we do the stable release for RadCam.

From what I can remember, this patch is necessary for the bookworm image to work correctly.

patrickelectric:
We can merge after the radcam release

PR #3905, opened 2026-04-28
core: tools: nginx: Add a 404 if extension is not registered

### LINKED ISSUE #3904: BlueOS should make it clear for external sources when accessing an extension page that is not ready
The use-case here is when someone from the outside (in this case Cockpit) is showing the page of an extension, but BlueOS' extension system is still loading and ends up showing the BlueOS interface, a sinking boat or nothing, all with a 2xx status code.

If the extension system is not yet ready and the content being accessed is not going to be that of the extension itself, an specific status code should be returned that makes it clear.

patrickelectric:
Fix #3904

## Summary by Sourcery

Bug Fixes:
- Ensure unknown /extensionv2/ routes return a 404 error rather than being handled by the single-page application.

### DIFF
--- a/core/tools/nginx/nginx.conf
+++ b/core/tools/nginx/nginx.conf
@@ -313,6 +313,19 @@ http {
         location ~ ^/redirect-port/(?<port>\d+) {
             return 301 $scheme://$host:$port;
         }
+
+        # Catch-all for unknown extensions: returns a real 404 instead of
+        # falling through to the Vue SPA.
+        location ^~ /extensionv2/ {
+            error_page 404 = @extension_not_found;
+            return 404;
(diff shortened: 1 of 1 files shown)


patrickelectric:
Tested and working:
image

Williangalvani (approved):
nice!

rafaellehmkuhl:
@patrickelectric  I believe this PR is a good defensive measure, but it doesn't fulfill the needs of #3904?

rafaellehmkuhl:
> @patrickelectric I believe this PR is a good defensive measure, but it doesn't fulfill the needs of #3904?

Just tested and it actually does.

Issue #3906, opened 2026-04-28
We need to turn off USB autosuspend to avoid problems in BlueOS


patrickelectric:

patrickelectric:
Fixed by #3912

patrickelectric:
Reopened, #3912 only helps.

PR #3912, opened 2026-04-30
Add usb-power rules to disable autosuspend

patrickelectric:
Helps #3906

## Summary by Sourcery

New Features:
- Add installation of usb-power udev rules to disable USB autosuspend on relevant devices.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -204,6 +204,7 @@ echo "Going to install BlueOS version ${VERSION}."
 
 echo "Downloading and installing udev rules."
 curl -fsSL $ROOT/install/udev/100.autopilot.rules -o /etc/udev/rules.d/100.autopilot.rules
+curl -fsSL $ROOT/install/udev/110.usb-power.rules -o /etc/udev/rules.d/110.usb-power.rules
 
 if [ -f /etc/dhcpcd.conf ]
 then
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
Not blockers, just things to reason/discuss:

1. What is the symptoms the autosuspend in action? 
2. Is there any verification we can do to test whether this is enough? 

And because it is a global rule:

3. would all devices be compatible with this?
4. Could this break compatibility with any device?
5. If it can break something, how could the user revert it? 

It could be useful to document these things.

patrickelectric:
> * would all devices be compatible with this?
> * Could this break compatibility with any device?
> * If it can break something, how could the user revert it?

This is done by the OS, I'm not sure if I follow your questions.
> Linux autosuspend is a Linux kernel feature (specifically a power management mechanism) rather than a hardware-driven device function. It automatically suspends idle USB devices to save power, though it can cause issues with peripherals like mice or Bluetooth adapters. 
---
> 1. What is the symptoms the autosuspend in action? 
> 2. Is there any verification we can do to test whether this is enough? 

The wifi interface stop answering, hotspot starts to have problems, external cell phone monden has problems, there is a list of complains that we received that by using autosuspend off it was fixed.

Williangalvani (approved):
I haven't tested, but it makes sense, looks correct, and we have reports of this helping with usb modem and wifi dongles.

Williangalvani:
do we care enough for adding this to blueos_startup_update, too ?

patrickelectric:
> do we care enough for adding this to blueos_startup_update, too ?

In the future, yes.

PR #3914, opened 2026-04-30
frontend: parameter-editor: ServoFunctionRangeEditor: Fix thumb position

### LINKED ISSUE #3269: bug: vehicle-setup: servo function limits cover slider handles
### Current behaviour
Clicking on a servo function brings up a dialog that allows setting the min/max/trim, but the displayed value annotations overlap the circle handles, which seems unpolished:
Image

### Desired solution
Move them up, and ideally change them to more of a callout design, with a little triangle pointing from the annotation background to the circle:
```
 __________
|____  ____|
     \/
     O
```

patrickelectric:
image

Fix #3269

## Summary by Sourcery

Adjust the servo function range editor thumb label positioning and styling for correct display above the slider handle.

Enhancements:
- Refine thumb label positioning relative to the slider to ensure it appears correctly above the handle.
- Improve thumb label styling with updated padding, shadow, and a pointer-like arrow indicator while disabling pointer events.

### DIFF
--- a/core/frontend/src/components/parameter-editor/ServoFunctionRangeEditor.vue
+++ b/core/frontend/src/components/parameter-editor/ServoFunctionRangeEditor.vue
@@ -494,16 +494,29 @@ export default Vue.extend({
 
 .thumb-label {
   position: absolute;
-  top: -25px;
+  bottom: calc(100% + 8px);
   left: 50%;
   transform: translateX(-50%);
   background-color: var(--v-primary-base);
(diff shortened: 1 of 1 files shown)


ES-Alexander (approved):
Checked and it seems good - thanks! :-)

There is a very slight gap between the triangle and the body of the label when clicking on the thumb, but not a high priority to fix 🤷‍♂️  
Image

PR #3926, opened 2026-05-05
refresh ethernet interfaces immediately after write actions

### LINKED ISSUE #3076: bug: ethernet changes spinner ends before we get updated network data
### Bug description

https://github.com/user-attachments/assets/0db01ed3-6711-44bb-a335-432a4ce1c080

### Steps to reproduce

change a network ip
wait,
See ip still there
wait
ip is gone

### Primary pain point(s)

it looks like the operation didnt go through

### Additional context

_No response_

### Prerequisites

- [x] I (lied that I) have checked to make sure that a similar request has not already been filed or fixed.

nicoschmdt:
fix #3076

## Summary by Sourcery

Refresh ethernet interfaces immediately after configuration changes and improve UI loading behavior for ethernet management.

Bug Fixes:
- Ensure ethernet interfaces are re-fetched after creating, deleting, or modifying addresses, DHCP servers, or triggering dynamic IP to keep the UI in sync with backend state.
- Prevent the ethernet interfaces list from getting stuck in an updating state by centralizing updating flag handling in the store and updater component.

Enhancements:
- Introduce a dedicated action to refresh available ethernet interfaces and reuse it across ethernet-related actions.
- Adjust ethernet manager UI conditions to show loading indicators while interfaces are updating and only show the 'no interfaces' message when none are available.
- Change the ethernet updater to perform an immediate initial fetch before starting the periodic refresh task and disable automatic start of the polling task.

Chores:
- Remove temporary caching on the backend endpoint that retrieves ethernet interfaces to always return current interface information.

### DIFF
--- a/core/frontend/src/components/ethernet/EthernetManager.vue
+++ b/core/frontend/src/components/ethernet/EthernetManager.vue
@@ -3,20 +3,20 @@
     elevation="1"
     width="400"
   >
-    <v-expansion-panels v-if="are_interfaces_available && !updating_interfaces">
+    <v-expansion-panels v-show="are_interfaces_available && !updating_interfaces">
       <interface-card
         v-for="(ethernet_interface, key) in available_interfaces"
         :key="key"
         :adapter="ethernet_interface"
       />
     </v-expansion-panels>
-    <v-container v-else-if="updating_interfaces">
+    <v-container v-if="updating_interfaces">
       <spinning-logo
         size="30%"
         subtitle="Fetching available ethernet interfaces..."
       />
     </v-container>
-    <v-container v-else>
+    <v-container v-else-if="!are_interfaces_available">
       <div>
         No ethernet interfaces available
       </div>
--- a/core/frontend/src/components/ethernet/EthernetUpdater.vue
+++ b/core/frontend/src/components/ethernet/EthernetUpdater.vue
@@ -12,11 +12,17 @@ export default Vue.extend({
   name: 'EthernetUpdater',
   data() {
     return {
-      fetch_available_interfaces_task: new OneMoreTime({ delay: 5000, disposeWith: this }),
+      fetch_available_interfaces_task: new OneMoreTime({ delay: 5000, disposeWith: this, autostart: false }),
     }
   },
-  mounted() {
-    this.fetch_available_interfaces_task.setAction(this.fetchAvailableEthernetInterfaces)
+  async mounted() {
+    try {
+      await this.fetchAvailableEthernetInterfaces()
+    } finally {
+      ethernet.setUpdatingInterfaces(false)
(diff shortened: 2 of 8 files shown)


nicoschmdt:
> ### 1. Correctness & Implementation Bugs
> * 1.2 [minor] `core/frontend/src/store/ethernet.ts:185-189` — `getAvailableEthernetInterfaces`'s `.catch` still commits `setInterfaces([])`, but `setInterfaces` no longer flips `updating_interfaces` to `false`. After this PR, a transient failure during the post-write `refreshInterfaces` will wipe the cached interface list, then the calling action's `.finally` immediately commits `setUpdatingInterfaces(false)` — so the user sees the panels disappear and "No ethernet interfaces available" flash up after a successful-but-slow write whose follow-up GET happens to time out. Consider having `refreshInterfaces` swallow the error without overwriting `available_interfaces`, or stop clearing the list inside `getAvailableEthernetInterfaces`'s catch.
> 
This behavior was already present in the code so this seems a little out of scope to me

joaoantoniocardoso (changes requested):
I believe we have other problems now?

https://github.com/user-attachments/assets/b852e21d-fc19-48c8-b982-35c645bd7f89

nicoschmdt:
@joaoantoniocardoso I believe I fixed this issue, could you check it again?

joaoantoniocardoso:
> @joaoantoniocardoso I believe I fixed this issue, could you check it again?

It got better, but there's still this old state being shown:

https://github.com/user-attachments/assets/6a39c2e8-e162-4e9f-affa-a494af8f2f9a

Perhaps a "loading" step could fill the gap?

nicoschmdt:
Well, the curious thing is that the loading states show for me... I'll investigate why this is happening :thinking:

joaoantoniocardoso:
> Well, the curious thing is that the loading states show for me... I'll investigate why this is happening 🤔

It could be a firefox thing.. also, there can be a difference between serving the frontend locally and serving on the pi. In this test, I think I was serving locally.

nicoschmdt:
Oh, I tested only in the pi. I'll check out both firefox and serving locally too then.

nicoschmdt:
@joaoantoniocardoso I am still unable to replicate the issue. I tested with firefox, zen-browser (firefox based) and helium-browser (chromium based) and had no problems. Both serving locally or accessing through the PI worked with no delays.

Just in case, are you testing with a Pi 4 too?

joaoantoniocardoso (approved):
Sorry, that must have been some weird thing here in my previous test. I've just retested both on Chrome and Firefox, Pi 4; all good:

https://github.com/user-attachments/assets/53fd2b5a-9206-4049-94c5-d9a94e719d8e

Thanks

PR #3949, opened 2026-06-12
Frontend: allow using outdated param sets

Williangalvani:
Screenshot 2026-06-12 at 14 15 39
Screenshot 2026-06-12 at 14 22 37

## Summary by Sourcery

Update vehicle parameter set selection to support older-but-compatible firmware versions and clearly distinguish outdated presets in the UI.

New Features:
- Allow using parameter sets built for older firmware versions when no exact version match exists.

Enhancements:
- Revamp parameter set buttons to show firmware version labels, highlight outdated sets with warning styling and tooltips, and order current sets before older ones.


ES-Alexander (commented):
Cool idea!

UI-wise, I think a dropdown with an apply button would make more sense, rather than an unbounded and growing number of buttons as new versions get released.

PR #3950, opened 2026-06-12
Allow choosing the SITL frame in the frontend

Williangalvani:
Screenshot 2026-06-12 at 16 25 35

## Summary by Sourcery

Add frontend and API support for configuring and reading the current SITL frame type for the autopilot simulator.

New Features:
- Expose the current SITL frame via a new backend GET endpoint returning the SITLFrame value.
- Introduce a SITL configuration panel in the frontend autopilot view to select and change the SITL frame when using a SITL board.
- Store the selected SITL frame in the frontend Vuex store and keep it in sync with the backend via a polling task.

Enhancements:
- Mirror the backend SITLFrame enum in the frontend types to provide typed options for SITL frame selection.


core/frontend/src/components/autopilot/AutopilotManagerUpdater.ts
```diff
@@ -114,6 +114,20 @@ export async function fetchFirmwareVehicleType(): Promise<void> {
...
+      url: `${autopilot.API_URL}/sitl_frame`,
+      timeout: 10000,
+    })
+    autopilot.setSitlFrame(response.data as SITLFrame)
+  } catch (error) {
+    autopilot.setSitlFrame(null)
+    notifier.pushBackError('AUTOPILOT_SITL_FRAME_FETCH_FAIL', error)
+  }
+}
+
```
joaoantoniocardoso, line 130: Any reason why not using the `return back_axios({/*...*/}).then(/*...*/).catch(/*...*/)` pattern used in the similar functions below?

joaoantoniocardoso (approved):

PR #3965, opened 2026-06-23
core: Update docker base to 0.2.6

joaoantoniocardoso:
This fixes a problem in MCM in which the fake streams would not render the time overlay correctly because of a missing font.


patrickelectric:
@joaoantoniocardoso should the commit message be `to` ?

joaoantoniocardoso:
> @joaoantoniocardoso should the commit message be `to` ?

ty; fixed!

patrickelectric (approved):

PR #3969, opened 2026-06-29
Frontend: power failsafe/setup updates

Williangalvani:
## Summary by Sourcery

Improve visibility and handling of unavailable power configuration parameters and reboot requirements, and add dependency-aware presentation for battery-related failsafes.

New Features:
- Show a warning alert and reboot action when parameter changes require an autopilot reboot.
- Indicate when power configuration parameters (battery monitor, sensor pins, multipliers, capacity, arming voltage) are unavailable for the current vehicle.
- Display failsafe cards as unavailable when their required dependency parameter indicates the feature is disabled, with explanatory messaging.
- Add dependency metadata for the low-battery failsafe based on the battery monitor setting.

Enhancements:
- Relax parameter label and editor components to support missing parameters with clearer tooltips and disabled styling.
- Update inline parameter editor to correctly flag the system as requiring a reboot when changing reboot-required parameters.
- Unify styling for disabled and unavailable failsafes with clear visual badges and hover behavior.

### DIFF
--- a/core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
@@ -1,8 +1,11 @@
 <template>
   <v-card
-    v-if="all_required_params_are_available"
+    v-if="failsafe_valid_for_current_vehicle"
     elevation="2"
-    :class="{ 'disabled-failsafe': is_disabled }"
+    :class="{
+      'disabled-failsafe': is_disabled && !dependency_unmet,
+      'unavailable-failsafe': dependency_unmet,
+    }"
     class="mb-4 mt-4 pa-4 d-flex flex-row  flex-grow-0 justify-left failsafe-card"
   >
     <div class="ma-4">
@@ -15,8 +18,17 @@
       <v-card-text>
         {{ failsafeDefinition.generalDescription }}
       </v-card-text>
+      <v-alert
+        v-if="dependency_unmet || !all_required_params_are_available"
+        type="warning"
+        text
+        dense
+        class="mx-4 mb-2"
+      >
+        {{ dependency_message }}
+      </v-alert>
       <div>
-        <div v-for="param in available_params" :key="param.name">
+        <div v-for="param in failsafeDefinition.params" :key="param.name">
           <v-row class="justify-right">
             <v-col :key="param.name" class="action-col" cols="7">
               <v-icon v-if="param.icon">
@@ -31,6 +43,13 @@
                 :disabled="is_disabled && param.name !== actionParamName"
                 :param="params[param.name]"
               />
+              <template v-else>
+                <v-text-field
+                  disabled
+                  class="caption"
+                  :value="`${param.name} not found`"
+                />
+              </template>
             </v-col>
           </v-row>
         </div>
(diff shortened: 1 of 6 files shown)


ES-Alexander:
Relevant issues:
1. Battery failsafe (still ?) disappears after you enable a battery sensor, before rebooting the autopilot
1. Long options make the dropdown wider than the edge of the card:
    Screenshot 2026-07-03 at 5 59 51 am

Unrelated issues:
1. The "Power" setup page should logically come before the failsafes page, since the failsafes only make sense for sensors that are already configured (and calibrated / determined as valid)
2. The "Lights" setup page can come after the failsafes page if relevant, since it's for independent output configuration
3. The "Lights" setup page is currently setting and looking for RCIN9/10 values instead of lights1/2
4. The "Camera Gimbal" page servo reversing toggle appears even though the gimbal is disabled
    Screenshot 2026-07-03 at 6 11 25 am
5. The disabled gimbal card provides instructions on how to enable a gimbal mount, but it could likely just present a button that enables doing it directly.

core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
```diff
@@ -65,27 +86,46 @@ export default Vue.extend({
...
-    params(): Record<string, Parameter> {
-      return autopilot_data.parameters
-        .filter((param) => this.failsafeDefinition.params.map((parameter) => parameter.name)
-          .includes(param.name))
-        .reduce((dict: Record<string, Parameter>, param: Parameter) => {
-          dict[param.name] = param
+    params(): Record<string, Parameter | null> {
+      return this.failsafeDefinition.params.reduce(
+        (dict: Record<string, Parameter | null>, param: ParamDefinitions) => {
+          dict[param.name] = autopilot_data.parameters.find((parameter) => parameter.name === param.name) ?? null
```
joaoantoniocardoso: The autopilot store has this same accessor exposed as a `parameter(name: string)`, we can use here and in the other places of the diff.

core/frontend/src/components/vehiclesetup/configuration/failsafes/types.ts
```diff
@@ -6,9 +8,17 @@ export interface ParamDefinitions {
...
+  message: string,
+}
+
 export interface FailsafeDefinition {
   name: string,
   generalDescription: string,
   params: ParamDefinitions[],
   image: string,
+  dependsOn?: FailsafeDependency,
+  supportedVehicles: FirmwareVehicleType[] | undefined,
```
joaoantoniocardoso:
```suggestion
  supportedVehicles?: FirmwareVehicleType[],
```

core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
```diff
@@ -65,27 +86,46 @@ export default Vue.extend({
...
+      // spurious notice; once params load, this re-evaluates to the truth.
+      const depParam = autopilot_data.parameters.find((p) => p.name === dep.paramName)
+      if (!depParam) {
+        return false
+      }
+      return depParam.value === dep.disabledValue
+    },
+    failsafe_valid_for_current_vehicle(): boolean {
+      return this.failsafeDefinition.supportedVehicles
+        ?.includes(autopilot_data.vehicle_type as FirmwareVehicleType) ?? true
```
joaoantoniocardoso: Some of the failsafe cards disappeared from the UI on this branch; it's a type mismatch:

```suggestion
      const supported = this.failsafeDefinition.supportedVehicles
      if (!supported) {
        return true
      }

      const { firmware_vehicle_type } = autopilot     # import autopilot from '@/store/autopilot_manager'
      if (firmware_vehicle_type == null) {
        return false
      }

      return supported.includes(firmware_vehicle_type)
```

joaoantoniocardoso (changes requested):

patrickelectric:
ping @joaoantoniocardoso

joaoantoniocardoso (approved):

PR #3975, opened 2026-07-06
core: Extract duplicated subprocess spawn/wait boilerplate into a shared helper

### LINKED ISSUE #3194: [core] Extract subprocess calls into a function
Since we are using more and more of this, with slightly different parameters, we would benefit from having it extracted into an ergonomic function to guarantee a coherent code to these critical parts of the system logic.

              I think we can extract this into a function and share this code with [`core/services/autopilot_manager/mavlink_proxy/AbstractRouter.py:AbstractRouter::start`](https://github.com/bluerobotics/BlueOS/blob/1d6851042c65aea0fc893a7741140358d36ee6a9/core/services/ardupilot_manager/mavlink_proxy/AbstractRouter.py#L96) and [`core/services/helper/main.py:ping`](https://github.com/bluerobotics/BlueOS/blob/1d6851042c65aea0fc893a7741140358d36ee6a9/core/services/helper/main.py#L634).

_Originally posted by @joaoantoniocardoso in https://github.com/bluerobotics/BlueOS/pull/3154#discussion_r1983753501_

rafael-graunke:
Fixes #3194

## Summary

Adds `run_subprocess()` to `commonwealth.utils.general` and replaces the
`asyncio.create_subprocess_exec` + `communicate()` boilerplate that was
copy-pasted across the codebase.

The issue mentions `AbstractRouter.start` and `helper/main.py:ping` as the
places to share this with. In practice the duplication is bigger than that:
9 call sites had the identical spawn/wait/decode pattern:

- `commonwealth/utils/DHCPDiscovery.py`
- `services/helper/main.py` (`ping`)
- `services/disk_usage/main.py` (x2: `du` scan, `disktest` speed test)
- `services/recorder_extractor/main.py` (x5: `mcap doctor`, `mcap recover`,
  `gst-discoverer-1.0`, `gst-play-1.0`, `mcap-foxglove-video-extract`)

`AbstractRouter.start` was deliberately not touched: it spawns a long-lived
router process, stores the handle on `self._subprocess` for `exit()` and
`start_house_keepers()` to use afterward, and only calls `communicate()` on
the failure path. That's a "supervise a daemon" lifecycle, not "run to
completion and collect output", so forcing it through the new helper would
mean either killing the daemon immediately or leaking the `Process` object
back out, which defeats the point of the abstraction.

## Bug fix included

While extracting `recorder_extractor/main.py`'s `build_thumbnail_bytes`, found
that the success path referenced an undefined `stdout` variable
(`if play_proc.returncode != 0 or not stdout:`, only `stdout_bytes` was ever
assigned). On success (`returncode == 0`) this raised `NameError`, since `or`
only short-circuits past it on the failure path. Fixed as part of the
extraction (now checks `stdout_bytes`), since the refactor touches that exact
line anyway.

## Testing

- Added unit tests for `run_subprocess` (returncode, stdout/stderr split, and
  the `merge_stderr` bytes-vs-`None` case), ran locally against this repo's
  `asyncio_mode=STRICT` pytest config, all pass.
- No existing test coverage for `disk_usage`, `recorder_extractor`,
  `helper/main.py`, or `DHCPDiscovery.py`. Verified behavior preservation by
  inspection: each diff only replaces the spawn+communicate block, every
  downstream `.decode(...)` call and parsing/log line is outside the diff,
  unchanged. No automated test harness exists for these 4 services to verify
  at runtime.

## Summary by Sourcery

Introduce a shared async subprocess helper and adopt it across services that previously duplicated spawn/wait logic.

New Features:
- Add run_subprocess utility to commonwealth.utils.general for running async commands with optional stderr merging.

Bug Fixes:
- Fix recorder_extractor thumbnail generation by checking the correct stdout variable when validating gst-play-1.0 output.

Enhancements:
- Refactor recorder_extractor, disk_usage, DHCPDiscovery, and helper ping to use the shared run_subprocess helper instead of inline asyncio subprocess boilerplate.

Tests:
- Add unit tests for run_subprocess covering stdout/stderr capture, non-zero return codes, and merged stderr behavior.

### DIFF
--- a/core/libs/commonwealth/src/commonwealth/utils/DHCPDiscovery.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/DHCPDiscovery.py
@@ -4,6 +4,7 @@ import re
 import sys
 from typing import List
 
+from commonwealth.utils.general import run_subprocess
 from loguru import logger
 
 
@@ -40,15 +41,11 @@ async def discover_dhcp_servers(iface: str, timeout: int = 15) -> List[str]:
 
         # Run nmap asynchronously
         logger.info(f"Running nmap command: {' '.join(cmd)}")
-        process = await asyncio.create_subprocess_exec(
-            *cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
-        )
+        returncode, stdout_bytes, stderr_bytes = await run_subprocess(cmd)
+        output = stdout_bytes.decode()
+        error = stderr_bytes.decode()
 
-        stdout, stderr = await process.communicate()
-        output = stdout.decode()
-        error = stderr.decode()
-
-        if process.returncode != 0:
+        if returncode != 0:
             logger.info(f"nmap output: {output}")
             logger.info(f"nmap error: {error}")
             raise DHCPDiscoveryError(f"nmap failed: {error}")
--- a/core/libs/commonwealth/src/commonwealth/utils/general.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/general.py
@@ -6,7 +6,7 @@ from dataclasses import asdict, dataclass
 from enum import Enum
 from functools import cache
 from pathlib import Path
-from typing import Any, AsyncGenerator
+from typing import Any, AsyncGenerator, List, Optional, Tuple
 
 import psutil
 from commonwealth.utils.commands import load_file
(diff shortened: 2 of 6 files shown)


rafael-graunke:
## Coverage gate failing (71.91%, need 75%), not what it looks like

Not that the refactored code is undertested. It's a side effect of how coverage.py scopes files.

`commonwealth/utils/general.py` had zero test coverage before this PR, but it also wasn't in the coverage denominator. Nothing in the primary suite imported it (its only consumers, `disk_usage`/`recorder_extractor`, have no tests; `helper` is secondary-suite scope). Coverage.py only reports files something actually imports during the run.

This PR's `test_general.py` (added to test `run_subprocess`) is the first thing to `import commonwealth.utils.general` in the primary suite. That flips ~195 previously-invisible statements (plus ~68 more in the transitively
imported `commands.py`) from "absent" to "mostly uncovered," dragging the total down. The code I actually wrote is fully tested; the regression is pre-existing untested code becoming visible for the first time.

Simplest fix: move `run_subprocess` out of `general.py` into its own module (e.g. `commonwealth/utils/subprocess_runner.py`). Then `test_general.py` goes away, the 9 call sites import from the new module instead, and nothing left imports `general.py`/`commands.py` in the primary suite, so both go back to
invisible, same as master. No coverage-gate discussion needed.

Alternative, if you'd rather have real coverage on that file regardless: I wrote a full test suite for the rest of both files locally (55 new tests, 94%/96% file coverage). Happy to bundle it in or split it into a follow-up PR.

Let me know which you'd prefer.

PR #3978, opened 2026-07-07
Frontend: Lights: fix detection and setup for ArduSub >= 4.5.6

Williangalvani:
## Summary by Sourcery

Update lights configuration and detection to support both legacy and native ArduSub lights handling based on firmware version.

New Features:
- Introduce centralized getters for lights servo function values in the autopilot manager store to abstract legacy vs. new ArduSub behavior.

Bug Fixes:
- Fix lights parameter detection and labeling for ArduSub firmware versions >= 4.5.6 by no longer hardcoding RCIN9/RCIN10 IDs.

Enhancements:
- Simplify lights UI titles by removing hardcoded RC input references.
- Unify vehicle type access through the autopilot manager store in various setup components.
- Align PWM setup and overview components with the new lights servo function resolution logic.

### DIFF
--- a/core/frontend/src/components/vehiclesetup/PwmSetup.vue
+++ b/core/frontend/src/components/vehiclesetup/PwmSetup.vue
@@ -154,7 +154,6 @@
 </template>
 
 <script lang="ts">
-import { lte as sem_ver_lte } from 'semver'
 import Vue from 'vue'
 
 import { fetchCurrentBoard } from '@/components/autopilot/AutopilotManagerUpdater'
@@ -202,8 +201,7 @@ const param_value_map = {
   },
 } as Dictionary<Dictionary<string>>
 
-// ArduSub <= 4.5.4 used RCIN9/RCIN10 for lights control; newer versions handle lights natively
-const LEGACY_SUB_LIGHTS_MAX_VERSION = '4.5.5'
+// Legacy ArduSub versions used RCIN9/RCIN10 for lights control; newer versions handle lights natively
 const legacy_sub_param_value_map: Dictionary<string> = {
   RCIN9: 'Lights 1',
   RCIN10: 'Lights 2',
@@ -237,8 +235,9 @@ export default Vue.extend({
   computed: {
     effective_param_value_map(): Dictionary<Dictionary<string>> {
       const map = { ...param_value_map }
-      const version = autopilot.firmware_info?.version
-      if (this.is_sub && version && sem_ver_lte(version, LEGACY_SUB_LIGHTS_MAX_VERSION)) {
+      // Apply legacy RCIN9/RCIN10 -> Lights labels only when the store resolves
+      // to the legacy ArduSub function numbers.
+      if (autopilot.lights1_servo_function !== SERVO_FUNCTION.LIGHTS1) {
         map.Submarine = { ...legacy_sub_param_value_map, ...map.Submarine }
       }
       return map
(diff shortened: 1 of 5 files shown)


core/frontend/src/store/autopilot_manager.ts
```diff
@@ -7,6 +8,12 @@ import {
   AutopilotEndpoint, FirmwareInfo, FirmwareVehicleType,
   FlightController, SerialEndpoint, SITLFrame,
 } from '@/types/autopilot'
+import { SERVO_FUNCTION as SUB_SERVO_FUNCTION } from '@/types/autopilot/parameter-sub-enums'
+
+// ArduSub <= 4.5.4 used RCIN9/RCIN10 for lights control; newer versions handle lights natively
+const LEGACY_SUB_LIGHTS_MAX_VERSION = '4.5.5'
```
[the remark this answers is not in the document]
Williangalvani: good bot

core/frontend/src/store/autopilot_manager.ts
```diff
@@ -39,6 +46,24 @@ class AutopilotManagerStore extends VuexModule {
 
   autopilot_serials: SerialEndpoint[] = []
 
+  get lights1_servo_function(): number {
+    const is_sub = this.firmware_vehicle_type === FirmwareVehicleType.ArduSub
```
ES-Alexander, line 51: At minimum, we need a comment as to why this check is necessary - e.g. "// All vehicle types include the values ArduSub overloaded for legacy lights support, so we need to filter by it".

That said, I don't think there's anything stopping other vehicle types from adding their own servo functions to the values Sub now uses for dedicated lights outputs, so I feel like a better approach is to push the vehicle type filtering higher up (e.g. don't show the cards/windows at all for non-Sub vehicles), and these functions can just assume Sub type is known and skip the filter?

core/frontend/src/components/vehiclesetup/PwmSetup.vue
```diff
@@ -203,7 +202,6 @@ const param_value_map = {
 } as Dictionary<Dictionary<string>>
 
 // ArduSub <= 4.5.4 used RCIN9/RCIN10 for lights control; newer versions handle lights natively
```
ES-Alexander:
```suggestion
// ArduSub <= 4.5.5 used RCIN9/RCIN10 for lights control; newer versions handle lights natively
```
This comment was incorrect to start with, but got missed in the reviews of the initial PR.

Either the version should be corrected, or (given the logic in this file does not explicitly check for versions) we could change to something like "Legacy ArduSub versions used RCIN9/RCIN10 ...".

ES-Alexander (changes requested):
Functionally working well - nice! :-)

Just a couple of comments to improve maintainability and avoid an edge-case.

core/frontend/src/store/autopilot_manager.ts
Williangalvani: > That said, I don't think there's anything stopping other vehicle types from adding their own servo functions to the values Sub now uses for dedicated lights outputs

Those are not vehicle-dependent. They are [a list shared by all vehicles](https://github.com/ArduPilot/ardupilot/blob/master/libraries/SRV_Channel/SRV_Channel.h#L213-L214)

ES-Alexander (approved):
LGTM! :-)

PR #3988, opened 2026-07-16
core:frontend: Update major tom together with BlueOS

joaomariolago:
> This PR removes major_tom from the default extensions list to avoid background work to be performed without user interaction as well as avoids the update process to think that extension is already installed while is being currently pulled in background

## Summary by Sourcery

Integrate automatic installation and updating of the optional Major Tom extension as part of BlueOS version changes and cloud actions, with dedicated progress reporting and centralized utility logic.

New Features:
- Add UI support to show dedicated progress for Major Tom installation and update operations in the version chooser.
- Introduce utility helpers to fetch latest Major Tom extension data, detect installed versions, decide whether to install or update, and perform the installation via the Kraken v2 API.

Enhancements:
- Trigger Major Tom installation or update when changing or pulling BlueOS versions without blocking the core update if Major Tom fails.
- Refactor version application logic in the version chooser into a reusable method used by both pull-and-set and direct set flows.
- Unify Major Tom extension identifiers and installation data fetching across the frontend, and update cloud tray installation to use the new shared utility and Kraken v2 extension API.

### DIFF
--- a/core/frontend/src/utils/major_tom.ts
+++ b/core/frontend/src/utils/major_tom.ts
@@ -1,0 +1,65 @@
+import axios from 'axios'
+
+import { fetchInstalledExtensions } from '@/components/kraken/KrakenManager'
+import { InstalledExtensionData } from '@/types/kraken'
+import back_axios from '@/utils/api'
+
+export const MAJOR_TOM_IDENTIFIER = 'blueos.major_tom'
+export const MAJOR_TOM_INSTALL_URL = 'https://blueos.cloud/major_tom/install/'
+
+const KRAKEN_API_V2_URL = '/kraken/v2.0'
+
+export type MajorTomAction = 'skip' | 'install' | 'update'
+
+export async function fetchMajorTomInstallData(): Promise<InstalledExtensionData> {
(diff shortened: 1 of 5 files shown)


nicoschmdt (approved):

core/frontend/src/utils/major_tom.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const MAJOR_TOM_IDENTIFIER = 'blueos.major_tom'
+export const MAJOR_TOM_INSTALL_URL = 'https://blueos.cloud/major_tom/install/'
+
+const KRAKEN_API_V2_URL = '/kraken/v2.0'
+
+export type MajorTomAction = 'skip' | 'install' | 'update'
+
+export async function fetchMajorTomInstallData(): Promise<InstalledExtensionData> {
+  const response = await axios.get(MAJOR_TOM_INSTALL_URL)
```
patrickelectric, line 15: It would be nice to have a tmeout here.
```suggestion
  const response = await axios.get(MAJOR_TOM_INSTALL_URL, { timeout: 10000 })
```

patrickelectric (approved):

PR #3990, opened 2026-07-16
migrate Beacon from Pykson to Pydantic

nicoschmdt:
## Summary by Sourcery

Migrate Beacon service configuration from the legacy settings/serialization system to Pydantic-based models and manager.

Enhancements:
- Replace Pykson-based settings models with Pydantic BaseModel and PydanticSettings implementations for Beacon configuration.
- Update configuration manager usage from Manager to PydanticManager in the Beacon service startup.
- Align settings migration logic and version handling with the new Pydantic settings infrastructure, including safer defaults and interface serialization.

### DIFF
--- a/core/services/beacon/settings.py
+++ b/core/services/beacon/settings.py
@@ -1,37 +1,29 @@
 import json
 import re
 import socket
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Literal, Optional
 
 import psutil
-from commonwealth.settings import settings
+from commonwealth.settings.settings import PydanticSettings
 from loguru import logger
-from pykson import (
-    IntegerField,
-    JsonObject,
-    ListField,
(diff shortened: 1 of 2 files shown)


core/services/beacon/settings.py
```diff
@@ -58,7 +50,7 @@ def get_ip_strs(self) -> List[str]:
                 except Exception as e:
                     logger.warning(f"unable to add ip {iface.address}: {e}")
             return addresses
-        matches = re.findall(r"ips\[(\d+)]", self.ip)
+        matches = re.findall(r"ips\[(\d+)]", self.ip or "")
```
joaoantoniocardoso: This is likely a behavioral change.

Do we _need_ it here? If there's a problem with the current behavior, consider moving it to another PR.

joaoantoniocardoso (approved):

PR #3995, opened 2026-07-16
Improve system logs

### LINKED ISSUE #2145: Add journal output when downloading logs
Kernel output

joaoantoniocardoso:
* Replaces #3456
* Closes #2145

The idea here is to dump what we need inside the downloaded zip file.

Note: I deliberately changed to the parent folder so we get the extensions logs as well.

Strategy:
* Moves the system logs download from being downloaded directly from the frontend using filebrowser to the backend.
* Before the backend calls the filebrowser api, it dumps valuable system information.
* To have the filebrowser on the backend, the frontend client lib was replicated on the backend.

## Summary by Sourcery

Route system log downloads through the backend, augmenting them with host diagnostics and a shared filebrowser client.

New Features:
- Introduce a backend /services/download_system_logs endpoint that streams a zip of system logs via filebrowser.
- Add asynchronous host diagnostics collection into the system logs directory prior to download, including journal, kernel, USB, and snapshot data.
- Provide a reusable Python filebrowser client library for interacting with the filebrowser API from backend services.

Enhancements:
- Update frontend log download actions to use the new commander API endpoint instead of direct filebrowser access.
- Align frontend commander API usage by consistently referencing commander.API_URL rather than a locally defined constant.

Build:
- Register the new filebrowser library in core and commander pyproject dependencies and workspace sources.

Documentation:
- Add README for the new Python filebrowser client library describing its purpose and relation to the existing frontend helpers.

### DIFF
--- a/core/frontend/src/components/health/SelfHealthTest.vue
+++ b/core/frontend/src/components/health/SelfHealthTest.vue
@@ -34,7 +34,7 @@
 <script lang="ts">
 import Vue from 'vue'
 
-import filebrowser from '@/libs/filebrowser'
+import commander from '@/store/commander'
 import * as VCU from '@/utils/version_chooser'
 
 export default Vue.extend({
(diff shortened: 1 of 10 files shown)


patrickelectric:
I'm not sure if I follow, we already have journal via mcap.

joaoantoniocardoso:
> I'm not sure if I follow, we already have journal via mcap.

Yes, this is a second-effort so we can easily have what we need from the system logs. Keep in mind that this patch will be ported to 1.4.4 (so we don't remain in the dark meanwhile).

patrickelectric (approved):
Since the code only runs when requested by the frontend, we should be safe to move to 1.4

PR #4011, opened 2026-07-22
fix: detect Raspberry Pi Compute Module 4/5 in get_cpu_type()

BrageGZP:
## Summary

`get_cpu_type()` in `core/libs/commonwealth/src/commonwealth/utils/general.py` identifies the host by grepping `/proc/cpuinfo` for the literal strings "Raspberry Pi 4" / "Raspberry Pi 5". Compute Modules report themselves as "Raspberry Pi Compute Module 4/5", which never matches, so the function returns `CpuType.Other` on CM4 and CM5 hardware.

Since Navigator detection (`NavigatorPi4`/`NavigatorPi5` in `flight_controller_detector/linux/navigator.py`) is gated on `get_cpu_type()`, the Navigator is never detected on Compute Module boards even when all required I²C devices are present — the Autopilot Firmware page only offers SITL.

## Root cause

Actual model lines from `/proc/cpuinfo` on affected hardware:

```
Model           : Raspberry Pi Compute Module 4 Rev 1.1
Model           : Raspberry Pi Compute Module 5 Rev 1.0
```

Neither contains the substring "Raspberry Pi 4" or "Raspberry Pi 5".

## Change

Extend the two string matches in `get_cpu_type()` to also accept the Compute Module model strings:

- "Raspberry Pi Compute Module 4" → `CpuType.PI4`
- "Raspberry Pi Compute Module 5" → `CpuType.PI5`

Also adds a regression test (`commonwealth/utils/tests/test_general.py`) covering the Pi 3/4/5, Compute Module 4/5, and unknown-board model strings.

No other behavior changes.

## Verification

Tested on physical hardware with a Blue Robotics Navigator on the 40-pin GPIO header:

- **CM5** (CM5016016) on the official Raspberry Pi CM5 IO Board, BlueOS pi5 image: with this change applied (via sed inside blueos-core), `get_cpu_type()` returns `CpuType.PI5`, the Navigator is detected as Navigator64, and ArduSub navigator64 (stable) boots and produces a MAVLink heartbeat. Navigator I²C devices confirmed at 0x0C/0x48/0x76 (bus 1) and 0x40 (bus 3).
- **CM4** (CM4004032) on the official Raspberry Pi CM4 IO Board, regular bookworm image (core 1.4.2): same result via `CpuType.PI4` after the change; Navigator I²C devices confirmed at 0x0C/0x48/0x76 (bus 1) and 0x40 (bus 4).

Without the change, both boards return `CpuType.Other` and only SITL is offered.

Additionally verified with a clean arm64 image built from this branch (no manual patching) on CM4: `get_cpu_type()` returns `CpuType.PI4`, the Navigator is detected, the startup update automatically writes the `[pi4]` peripheral block to `/boot/firmware/config.txt` (previously missing), and everything persists across reboots.

## Related reports

- https://discuss.bluerobotics.com/t/using-navigator-with-rpi-cm4-waveshare-carrier-board/23155 — same symptom on CM4 (unresolved in thread; the reporter's `/dev/i2c-*` output also matches the issue below)
- https://discuss.bluerobotics.com/t/pi-5-with-navigator/15903 — Pi 5 background

## Related issue (resolves automatically with this fix)

On Compute Modules the startup configuration that writes the UART/I²C/SPI blocks to `/boot/firmware/config.txt` was never written, because it is gated on the same CPU detection. Verified on CM4 hardware with a clean build of this branch: the peripheral block is now written automatically on startup and persists across reboots, so no separate fix is needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/libs/commonwealth/src/commonwealth/utils/general.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/general.py
@@ -36,9 +36,9 @@ def blueos_version() -> str:
 def get_cpu_type() -> CpuType:
     with open("/proc/cpuinfo", "r", encoding="utf-8") as f:
         for line in f:
-            if "Raspberry Pi 4" in line:
+            if "Raspberry Pi 4" in line or "Raspberry Pi Compute Module 4" in line:
                 return CpuType.PI4
-            if "Raspberry Pi 5" in line:
+            if "Raspberry Pi 5" in line or "Raspberry Pi Compute Module 5" in line:
                 return CpuType.PI5
             if "Raspberry Pi 3" in line:
                 return CpuType.PI3
--- a/core/libs/commonwealth/src/commonwealth/utils/tests/test_general.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/tests/test_general.py
@@ -1,8 +1,14 @@
+import os
+import subprocess
+import uuid
+from pathlib import Path
 from unittest.mock import mock_open, patch
 
 import pytest
+from pyfakefs.fake_filesystem import FakeFilesystem
 
-from ..general import CpuType, get_cpu_type
+from .. import general
+from ..general import CpuType, HostOs, get_cpu_type
 
 CPUINFO_TEMPLATE = """processor	: 0
 BogoMIPS	: 108.00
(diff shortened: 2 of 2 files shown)


patrickelectric:
Thanks for the contributions, but it appears that the python-tests are failing for some reason, can you rebase over master ?

BrageGZP:
The python-tests failure turned out to be a coverage-gate effect rather than a broken test: the new test_general.py imports commonwealth/utils/general.py for the first time, which pulled its previously-unmeasured lines into the coverage denominator. The second commit adds proper tests for the rest of that module (93% on general.py), and the suite now clears the gate in a CI-equivalent local run.

One check will remain red: check-ai-authorship. The commits carry a truthful Co-Authored-By: Claude trailer — I used Claude Code to implement the change, while the root-cause diagnosis and all hardware verification (CM4 and CM5, including clean builds of this branch) are my own. I've kept the attribution honest rather than stripping it to pass the check, so that call is yours to make — I'm happy to adapt to whatever your contribution policy requires, including re-authoring the change myself if AI-assisted implementation isn't accepted.

patrickelectric (approved):

joaoantoniocardoso:
We should backport this one _after_ the next 1.4 stable release.

PR #4019, opened 2026-07-24
Fix API root redirects behind Nginx

### LINKED ISSUE #3452: bug: VersionChooser and Kraken API root endpoint redirects to wrong URL path
### Bug description

When trying to access the following VersionChooser API endpoint `http://blueos-avahi.local/version-chooser/v1.0/` it redirects to `http://blueos-avahi.local/v1.0/docs` instead of `http://blueos-avahi.local:8081/v1.0/docs`. 

The result is the same when accessing the root endpoint for the Kraken API `http://blueos-avahi.local/kraken/v1.0/`.

It would be nice if we could redirect to the correct API docs URL for both the Kraken and VersionChooser API.

### Steps to reproduce

1. Access `http://blueos-avahi.local/version-chooser/v1.0/` or `http://blueos-avahi.local/kraken/v1.0/`

### Primary pain point(s)

_No response_

### Additional context

_No response_

### Prerequisites

lainx86:
## Summary
- Change the VersionChooser v1 API root redirect to use the relative `docs` URL
- Apply the same fix to the Kraken v1 and v2 API root redirects
- Add regression tests that verify the returned `Location` header

## Problem
The affected API root endpoints returned root-relative paths such as
`/v1.0/docs` and `/v2.0/docs`.

When accessed through their Nginx service paths, these redirects dropped the
`/version-chooser/` or `/kraken/` prefix. For example:

```text
/version-chooser/v1.0/
```

redirected to:

```text
/v1.0/docs
```

instead of preserving the VersionChooser service path.

## Solution
The affected endpoints now return a relative redirect:

```text
Location: docs
```

This allows the client to resolve the documentation URL relative to the
requested endpoint, preserving the Nginx service path.

The same change is also applied to Kraken v2 because it used the identical
root-relative redirect pattern.

## Testing
- Focused regression tests: 3 passed
- Full VersionChooser and Kraken test files: 12 passed
- isort: passed
- Black: passed
- Ruff: passed
- Pylint: 10.00/10
- mypy: passed
- `git diff --check`: passed

The repository-wide pre-push hook could not be completed because Poetry is
not available in my local environment.

Fixes #3452

### DIFF
--- a/core/services/kraken/api/v1/routers/index.py
+++ b/core/services/kraken/api/v1/routers/index.py
@@ -59,4 +59,4 @@ async def root() -> RedirectResponse:
     """
     Root endpoint for the Kraken API V1.
     """
-    return RedirectResponse(url="/v1.0/docs")
+    return RedirectResponse(url="docs")
--- a/core/services/kraken/api/v2/routers/index.py
+++ b/core/services/kraken/api/v2/routers/index.py
@@ -14,4 +14,4 @@ async def root() -> RedirectResponse:
     """
     Root endpoint for the Kraken API V2.
     """
-    return RedirectResponse(url="/v2.0/docs")
+    return RedirectResponse(url="docs")
--- a/core/services/kraken/test_kraken.py
+++ b/core/services/kraken/test_kraken.py
@@ -1,0 +1,28 @@
+import os
+import subprocess
+import sys
+from pathlib import Path
+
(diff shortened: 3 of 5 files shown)


patrickelectric (approved):

joaoantoniocardoso:
@patrickelectric should this be backported?

PR #4028, opened 2026-07-25
frontend: Use linux2rest network probe deltas for tray rates

### LINKED ISSUE #3587: Negative bandwidth (?)
Image

joaoantoniocardoso:
## Summary
- Derive tray upload/download speeds from linux2rest `received_B` / `transmitted_B` (sysinfo per-sample deltas) divided by the Sampler interval (5s), instead of re-differencing `total_*` across frontend polls.
- That removes negative Mbps when totals reset and avoids flicker/spikes against the ≥0.9 sample cache.
- NetworkCard now shows probe deltas alongside totals (same shape as linux2rest's own UI).
- Also fixes the `transmitted_B` type docstring (was copy-pasted as "bytes received"), swaps the swapped `mdi-cloud-upload`/`mdi-cloud-download` icons on the byte rows, and rewords labels to `Packets received/transmitted`.

Closes #3587

Note: The target is BlueOS 1.5 only.

## Test plan
- [ ] Against BlueOS with linux2rest 0.9.x: enable eth0 Networking widget under steady traffic
- [ ] Rate stays stable across 2s polls within a 5s sample window (no flicker to 0)
- [ ] Rate updates when the Sampler advances; never goes negative
- [ ] System Information → Network cards show `delta / total` for packets and bytes
- [ ] Byte-row icons match direction (download on received, upload on transmitted)

### DIFF
--- a/core/frontend/src/components/system-information/NetworkCard.vue
+++ b/core/frontend/src/components/system-information/NetworkCard.vue
@@ -62,9 +62,9 @@
           mdi-package-down
         </v-icon>
       </v-list-item-icon>
-      <v-list-item-subtitle>Received package:</v-list-item-subtitle>
+      <v-list-item-subtitle>Packets received:</v-list-item-subtitle>
       <v-list-item-subtitle class="text-right">
-        {{ network.total_packets_received }}
+        {{ network.packets_received }} / {{ network.total_packets_received }} total
       </v-list-item-subtitle>
     </v-list-item>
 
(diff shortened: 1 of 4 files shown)


joaoantoniocardoso:
Addressed from the latest review:
- Clamp at source in `networkProbeRateBps` (formatter stays pure)
- camelCase `bytesSinceLastProbe`
- Consistent NetworkCard labels (`Packets received/transmitted`)

Deferred / intentional:
- Hardcoded 5 s Sampler interval → follow-up (expose from linux2rest or derive from `unix_time_seconds`)
- First-sample `0.00Mbps` until the first non-zero delta window is intentional
- Disk-rate unification out of scope here

joaoantoniocardoso:
Superseded by #4048 (same changes, head moved to fork `joaoantoniocardoso/BlueOS-docker`).

joaoantoniocardoso:
Closed in favor of #4048.

PR #4030, opened 2026-07-25
frontend: Fix MAVLink2Rest orphan WebSocket reconnects

joaoantoniocardoso:
## Summary
- On intentional close, clear reconnect timeouts and detach `onclose` so dying sockets cannot resurrect.
- `setBaseUrl` closes the previous send socket before opening a new one (probe can call setBaseUrl more than once).

## Test plan
- [ ] Load BlueOS, confirm MAVLink heartbeats / params still work
- [ ] Reload / HMR: console should not accumulate reconnecting zombie WS over time
- [ ] Vehicle reconnect after autopilot restart still recovers within a few seconds

PR #4044, opened 2026-07-27
frontend: Fix MAVLink2Rest orphan WebSocket reconnects

joaoantoniocardoso:
## Summary
- On intentional close, clear reconnect timeouts and detach `onclose` so dying sockets cannot resurrect.
- `setBaseUrl` closes the previous send socket before opening a new one (probe can call setBaseUrl more than once).

## Test plan
- [ ] Load BlueOS, confirm MAVLink heartbeats / params still work
- [ ] Reload / HMR: console should not accumulate reconnecting zombie WS over time
- [ ] Vehicle reconnect after autopilot restart still recovers within a few seconds

---
Supersedes #4030 (recreated from fork `joaoantoniocardoso/BlueOS-docker` instead of same-repo head).

### DIFF
--- a/core/frontend/src/libs/MAVLink2Rest/Endpoint.ts
+++ b/core/frontend/src/libs/MAVLink2Rest/Endpoint.ts
@@ -10,7 +10,14 @@ export default class Endpoint {
 
   latestData: any = null
 
+  private intentionalClose = false
+
+  private reconnectTimeout: number | null = null
+
+  private url: string
+
   constructor(url: string) {
+    this.url = url
     this.socket = this.createSocket(url)
   }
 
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
Rebased; the image is now available on Docker Hub.

patrickelectric:
Can you explain how the original error can be replicated ?

joaoantoniocardoso:
> Can you explain how the original error can be replicated ?

Sorry for the lack of context.

This one addresses an extra mavlink2rest WebSocket connection being created for each tracked topic a few seconds after reloading BlueOS, ending up with 1 dead + 2 active ones. The damage is controlled, as it doesn't pile indefinitely.

To see it, open any reasonably quiet page of BlueOS, open the Network tab from the browser's developer tools, filter for connection type WebSocket and for HEARTBEAT (example). Reload the page, wait a few seconds, and inspect each connection by clicking on each of them. You'll see two active ones and one dead. With this patch, we'll only ever have 1 active in the same scenario.

Before:
image

After:
image

nicoschmdt (approved):

PR #4032, opened 2026-07-25
frontend: Stop NetworkSpeedTest reconnect interval multiply

joaoantoniocardoso:
## Summary
- Keep a single 200ms ping interval for the page lifetime.
- On disconnect, only recreate the WebSocket (do not nest another `setInterval`).

## Test plan
- [ ] Open Tools → Network test, leave the page open with a flaky WS: interval count stays at one
- [ ] Latency updates while connected; reconnect recovers without runaway timers
- [ ] Leave the page: interval cleared, socket closed

PR #4045, opened 2026-07-27
frontend: Stop NetworkSpeedTest reconnect interval multiply

joaoantoniocardoso:
## Summary
- Keep a single 200ms ping interval for the page lifetime.
- On disconnect, only recreate the WebSocket (do not nest another `setInterval`).

## Test plan
- [ ] Open Tools → Network test, leave the page open with a flaky WS: interval count stays at one
- [ ] Latency updates while connected; reconnect recovers without runaway timers
- [ ] Leave the page: interval cleared, socket closed

---
Supersedes #4032 (recreated from fork `joaoantoniocardoso/BlueOS-docker` instead of same-repo head).

### DIFF
--- a/core/frontend/src/components/speedtest/NetworkSpeedTest.vue
+++ b/core/frontend/src/components/speedtest/NetworkSpeedTest.vue
@@ -139,10 +139,21 @@ export default Vue.extend({
     this.websocket?.close()
   },
   mounted() {
-    this.openWebSocket()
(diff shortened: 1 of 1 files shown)


Williangalvani (approved):
we should mark the ping value as something else when connection is lost. right now it just freezes...
But that is unrelated-ish to this PR

PR #4033, opened 2026-07-25
frontend: Discard tray MAVLink listeners on destroy

joaoantoniocardoso:
## Summary
- HealthTray HEARTBEAT and GpsTrayMenu GPS listeners are stored with `markRaw` and `discard()`ed in `beforeDestroy`.
- Avoids stacking callbacks on shared MAVLink2Rest endpoints (and deep-observing `Endpoint.latestData` every frame).

## Test plan
- [ ] Heartbeat icon and GPS trays still update normally
- [ ] Remount / HMR of App tray: listener count does not grow (no duplicate HEARTBEAT handling)
- [ ] No regression in armed state / GPS fix display

PR #4046, opened 2026-07-27
frontend: Discard tray MAVLink listeners on destroy

joaoantoniocardoso:
## Summary
- HealthTray HEARTBEAT and GpsTrayMenu GPS listeners are stored with `markRaw` and `discard()`ed in `beforeDestroy`.
- Avoids stacking callbacks on shared MAVLink2Rest endpoints (and deep-observing `Endpoint.latestData` every frame).

## Test plan
- [ ] Heartbeat icon and GPS trays still update normally
- [ ] Remount / HMR of App tray: listener count does not grow (no duplicate HEARTBEAT handling)
- [ ] No regression in armed state / GPS fix display

---
Supersedes #4033 (recreated from fork `joaoantoniocardoso/BlueOS-docker` instead of same-repo head).

### DIFF
--- a/core/frontend/src/components/health/GpsTrayMenu.vue
+++ b/core/frontend/src/components/health/GpsTrayMenu.vue
@@ -78,9 +78,10 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue'
+import Vue, { markRaw } from 'vue'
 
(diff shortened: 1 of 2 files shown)


Williangalvani (approved):
is this more like a correctness fix? these never get re-mounted, do they?

joaomariolago (approved):

PR #4031, opened 2026-07-25
frontend: Clean up MAVLink refresh rates on leave (store + vehicle setup)

joaoantoniocardoso:
## Summary
- Allow `setMessageRefreshRate` to lower rates (not only raise), replacing the listener when the rate changes.
- Vehicle setup / MainView / compass: ratchet rates down on destroy, discard position listeners, clear CompassDisplay render interval + GSAP tweens.

## Test plan
- [ ] Visit Home (ATTITUDE 10 Hz) then leave: rate drops to 1 Hz
- [ ] Open compass configure, leave: render interval stops; IMU/ATTITUDE rates ratchet down
- [ ] Gyro calib mount/unmount does not leave 10 Hz streams forever
- [ ] Level horizon dialog open/close does not fight CompassDisplay ATTITUDE rate while both tabs are alive

PR #4047, opened 2026-07-27
frontend: Clean up MAVLink refresh rates on leave (store + vehicle setup)

joaoantoniocardoso:
## Summary
- Track per-message rate claims in the mavlink store and apply `max(claims)` (idle 1 Hz) so one consumer can leave without overriding another.
- Vehicle setup / MainView / compass: release rate claims on destroy, discard position listeners, clear CompassDisplay render interval + GSAP tweens.

## Test plan
- [ ] Visit Home (ATTITUDE 10 Hz) then leave: rate drops to 1 Hz
- [ ] Open compass configure, leave: render interval stops; IMU/ATTITUDE rates drop when no other claimant remains
- [ ] Gyro calib mount/unmount does not leave 10 Hz streams forever
- [ ] Level horizon dialog open/close does not fight CompassDisplay ATTITUDE rate while both tabs are alive
- [ ] Gyro calib + CompassDisplay overlapping: tearing down Gyro keeps IMU at 10 Hz while CompassDisplay is mounted

---
Supersedes #4031 (recreated from fork `joaoantoniocardoso/BlueOS-docker` instead of same-repo head).

### DIFF
--- a/core/frontend/src/components/vehiclesetup/PwmSetup.vue
+++ b/core/frontend/src/components/vehiclesetup/PwmSetup.vue
@@ -426,7 +426,7 @@ export default Vue.extend({
     this.motor_writer_interval = setInterval(this.write_motors, 100)
     fetchCurrentBoard()
 
-    mavlink.setMessageRefreshRate({ messageName: 'SERVO_OUTPUT_RAW', refreshRate: 10 })
+    mavlink.subscribeMessageRefreshRate({ messageName: 'SERVO_OUTPUT_RAW', refreshRate: 10 })
     this.desired_armed_state = this.is_armed
     this.installListeners()
     this.updateReversionValues()
@@ -434,7 +434,7 @@ export default Vue.extend({
   beforeDestroy() {
     clearInterval(this.motor_zeroer_interval)
     clearInterval(this.motor_writer_interval)
-    mavlink.setMessageRefreshRate({ messageName: 'SERVO_OUTPUT_RAW', refreshRate: 1 })
+    mavlink.unsubscribeMessageRefreshRate({ messageName: 'SERVO_OUTPUT_RAW', refreshRate: 10 })
     this.uninstallListeners()
   },
   methods: {
--- a/core/frontend/src/components/vehiclesetup/configuration/compass/AutoCoordinateDetector.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/compass/AutoCoordinateDetector.vue
@@ -61,9 +61,10 @@
 
 <script lang="ts">
 
-import { PropType } from 'vue'
+import { markRaw, PropType } from 'vue'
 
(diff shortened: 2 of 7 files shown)


Williangalvani:
> * 1.1 [minor] `core/frontend/src/store/mavlink.ts:34-49` — the store still has no reference counting: `beforeDestroy` in one component can lower a rate that another live component depends on. The PR papers over one instance in `LevelHorizonCalibration.vue`, but the same class of bug remains latent — for example, `GyroCalib.beforeDestroy` ratchets `RAW_IMU`/`SCALED_IMU2`/`SCALED_IMU3` down to 1 Hz, and those exact three messages are also requested by `CompassDisplay` at 10 Hz. If a caller ever has both mounted concurrently (Vuetify's `v-tabs-items` cache, a dashboard widget, etc.) the surviving component silently drops to 1 Hz. Not a blocker for this PR, but worth tracking — a real fix would keep a per-message subscriber map and pick `max(rate)`.

Could we do this instead? that would take some more code, but we wouldnt override other's rates

joaoantoniocardoso:
> > * 1.1 [minor] `core/frontend/src/store/mavlink.ts:34-49` — the store still has no reference counting: `beforeDestroy` in one component can lower a rate that another live component depends on. The PR papers over one instance in `LevelHorizonCalibration.vue`, but the same class of bug remains latent — for example, `GyroCalib.beforeDestroy` ratchets `RAW_IMU`/`SCALED_IMU2`/`SCALED_IMU3` down to 1 Hz, and those exact three messages are also requested by `CompassDisplay` at 10 Hz. If a caller ever has both mounted concurrently (Vuetify's `v-tabs-items` cache, a dashboard widget, etc.) the surviving component silently drops to 1 Hz. Not a blocker for this PR, but worth tracking — a real fix would keep a per-message subscriber map and pick `max(rate)`.
> 
> Could we do this instead? that would take some more code, but we wouldnt override other's rates

Done; that's indeed a better approach, thanks.

Williangalvani:
tested by checking the RAW_IMU message rate on mavlink-server and going to the gyro calib page.
RAW_IMU streams at 2hz by default (param), raises to 10hz while gyro page gets opened, and then, when it is unmounted, it decreases to 1hz.
I think this is slightly odd, but fine? 
Screenshot 2026-08-06 at 12 39 07

joaoantoniocardoso:
> tested by checking the RAW_IMU message rate on mavlink-server and going to the gyro calib page. RAW_IMU streams at 2hz by default (param), raises to 10hz while gyro page gets opened, and then, when it is unmounted, it decreases to 1hz. I think this is slightly odd, but fine? Screenshot 2026-08-06 at 12 39 07

Yes, my thought was "we should keep things calm and only raise the frequency when needed", but if being lazy is bad for some reason, we can keep the nominal at the maximum required frequency.

Williangalvani (approved):

PR #4029, opened 2026-07-25
frontend: Gate network and platform polling by FetchType

joaoantoniocardoso:
## Summary
- Refcount `SystemNetworkType` and `PlatformType` like other system-info polls (drop always-on network task; gate platform task).
- Subscribe from Networking widgets / Network tab / HealthTray / VehicleInfo / DevicePathHelper.
- Keep a stable `network_interface_names` list so App tray widgets do not rebuild on every speed update.
- Also switches App tray `:key` from index to `widget_name` (stable under drag reorder).

## Test plan
- [ ] Open BlueOS with HealthTray mounted: PlatformType still polls; CPU/Disk/etc. only when widgets subscribed
- [ ] Enable eth0 Networking widget: SystemNetworkType polls; disable widget: polling stops (if no other network subscribers)
- [ ] Tray widget list does not flicker/rebuild every 2s under steady traffic
- [ ] Rename/hotplug iface (or swap names at same count): tray updates to new interface names
- [ ] Vehicle setup DevicePathHelper / VehicleInfo still get platform info

PR #4049, opened 2026-07-27
frontend: Gate network and platform polling by FetchType

joaoantoniocardoso:
## Summary
- Refcount `SystemNetworkType` and `PlatformType` like other system-info polls (drop always-on network task; gate platform task).
- Subscribe from Networking widgets / Network tab / HealthTray / VehicleInfo / DevicePathHelper.
- Keep a stable `network_interface_names` list so App tray widgets do not rebuild on every speed update.
- Also switches App tray `:key` from index to `widget_name` (stable under drag reorder).

## Test plan
- [ ] Open BlueOS with HealthTray mounted: PlatformType still polls; CPU/Disk/etc. only when widgets subscribed
- [ ] Enable eth0 Networking widget: SystemNetworkType polls; disable widget: polling stops (if no other network subscribers)
- [ ] Tray widget list does not flicker/rebuild every 2s under steady traffic
- [ ] Rename/hotplug iface (or swap names at same count): tray updates to new interface names
- [ ] Vehicle setup DevicePathHelper / VehicleInfo still get platform info

---
Supersedes #4029 (recreated from fork `joaoantoniocardoso/BlueOS-docker` instead of same-repo head).

### DIFF
--- a/core/frontend/src/App.vue
+++ b/core/frontend/src/App.vue
@@ -53,8 +53,8 @@
         <draggable v-model="selected_widgets" class="d-flex align-center justify-center">
           <component
             :is="getWidget(widget_name).component"
-            v-for="(widget_name, i) in selected_widgets"
-            :key="i"
+            v-for="widget_name in selected_widgets"
+            :key="widget_name"
             v-bind="getWidget(widget_name).props"
             class="mr-2"
             ripple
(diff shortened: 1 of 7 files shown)


Williangalvani (approved):
nice!

PR #4036, opened 2026-07-27
fix: handle disappearing interfaces in Ping discovery

lainx86:
## Summary

- Avoid a `KeyError` when an interface disappears between the
  `net_if_stats()` and `net_if_addrs()` snapshots.
- Add regression coverage for inconsistent interface snapshots.
- Preserve the existing IPv4, loopback, and inactive-interface filtering.

## Problem

`list_ips()` collects interface status and address information using two
separate psutil calls. If an interface exists in the status snapshot but is
missing from the address snapshot, directly indexing the address mapping
raises a `KeyError`.

The exception can terminate the Ping sensor watcher task and stop subsequent
device discovery.

## Solution

Treat an interface missing from the address snapshot as having no addresses,
allowing discovery to continue for the remaining interfaces.

## Testing

- Ping regression tests: 2 passed
- Black: passed
- isort: passed
- Ruff: passed
- Pylint: 10.00/10
- mypy: passed
- `git diff --check`: passed

The complete pre-push hook could not run locally because Poetry is not
installed.

Fixes #3247

PR #4051, opened 2026-07-27
core: services: ping: Keep port watcher running after unexpected errors

joaoantoniocardoso:
## Summary

- Catch unexpected errors in the Ping `PortWatcher` loop so discovery keeps running.
- Add a regression test that a transient `add_ping360` failure does not stop watching.

Follow-up to #4036: that PR fixes the `list_ips()` KeyError, but any other uncaught exception in the watch loop still permanently kills sensor discovery because `sensor_manager()` is started with a bare `asyncio.create_task`.

## Test plan

- [x] `pytest core/services/ping/test_portwatcher.py`
- [ ] Confirm Ping service continues discovering after SoftAP/`uap0` churn on a vehicle

---
Supersedes #4040 (recreated from fork `joaoantoniocardoso/BlueOS-docker` instead of same-repo head).

### DIFF
--- a/core/services/ping/portwatcher.py
+++ b/core/services/ping/portwatcher.py
@@ -78,20 +78,23 @@ class PortWatcher:
         """Start watching for plugged/unplugged serial devices in the system."""
         # TODO: try https://pypi.org/project/inotify/
         while True:
-            ports = serial.tools.list_ports.comports()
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
@joaomariolago This is the one I mentioned needs a ping to be fully tested.

joaomariolago (approved):

PR #4055, opened 2026-07-28
Add beacon to venv2 updated deps

nicoschmdt:


nicoschmdt:
So, bumping `zeroconf` to 0.148.0 seems like a big change, should I do this in this PR?

joaoantoniocardoso:
> So, bumping `zeroconf` to 0.148.0 seems like a big change, should I do this in this PR?

No, we should play it safe and do that in another PR.

joaoantoniocardoso (approved):

PR #4069, opened 2026-07-31
feat: consume vehicle-provided parameter metadata

JackSkellet:
## Summary

This adds support for consuming parameter metadata published by the connected flight controller instead of relying exclusively on BlueOS's bundled ArduPilot metadata files.

The implementation:

- receives MAVLink `COMPONENT_METADATA` advertisements;
- downloads bounded parameter metadata documents over MAVFTP or HTTPS;
- validates the advertised CRC and document structure;
- exposes the active document from ArduPilot Manager with ETag support;
- merges vehicle-provided metadata over the nearest compatible bundled metadata in the frontend;
- refreshes the table when the vehicle advertises a new metadata generation;
- keeps bundled fallback selection within the same vehicle family;
- upgrades mavlink2rest from `t0.11.25` to `0.11.26`, adding `COMPONENT_METADATA` while preserving the 0.11 API.

## Why

BlueOS currently assumes that parameter names, descriptions, units, ranges and enum choices are fully described by a static metadata file bundled with the frontend. That does not work well for custom firmware or for parameter choices that depend on devices discovered at runtime.

This enables use cases such as:

- custom or development firmware supplying descriptions for parameters that do not exist in an upstream ArduPilot release;
- detected sensors being presented as meaningful enum choices without a BlueOS-specific override;
- per-lane EKF source-target selection, for example selecting different DVLs for `EK3_SRC1`, `EK3_SRC2` and `EK3_SRC3`;
- refreshing source-target choices when sensors are attached, removed or replaced;
- vehicle-specific parameter descriptions, units, defaults, ranges, bitmasks and enum labels;
- continuing to use bundled ArduPilot metadata when a vehicle does not publish component metadata.

The live metadata overlay has exact-name priority and also supports parameter templates. Invalid documents fail closed and do not replace the last valid snapshot.

## Parameter import reliability

The parameter loader is also made tolerant of parameter files from another firmware version or feature set:

- parameters absent from the connected vehicle are skipped instead of blocking the whole import;
- `PARAM_SET` uses the live parameter's MAVLink type;
- rejected values stop retrying after the existing bounded retry budget and are reported as skipped;
- retry timers and state are cleaned up when the dialog completes or closes.

This allows a useful subset of a parameter file to be applied without leaving the loader permanently failed because optional parameters are unavailable.

## mavlink2rest compatibility

`COMPONENT_METADATA` is not available through the previously pinned mavlink2rest release. Version `0.11.26` adds it through rust-mavlink `0.10.13` while preserving the existing 0.11 JSON API.

## Validation

- ArduPilot Manager parameter-metadata tests: 9 passed.
- Frontend parameter-metadata tests: 8 passed.
- Targeted ESLint and Stylelint: passed.
- Ruff and Mypy checks for the changed Python paths: passed.
- Official mavlink2rest `0.11.26` artifacts resolved for x86_64, ARMv7 and AArch64.
- Live ARMv7 Raspberry Pi 4 and AArch64 Raspberry Pi 5 testing: BlueOS startup/shutdown, metadata advertisement, MAVFTP retrieval, ETag refresh, fallback selection and parameter-table rendering passed.


patrickelectric:
@JackSkellet can you rebase the PR over master ?

JackSkellet:
@patrickelectric rebase has been completed.

core/tools/mavlink2rest/bootstrap.sh
```diff
@@ -3,7 +3,7 @@
 # Immediately exit on errors
 set -e
 
-VERSION="t0.11.25"
+VERSION="1.0.2"
```
patrickelectric: you should not update mavlink2rest since it breaks the api from 0.X to 1.X.

patrickelectric (changes requested):

core/tools/mavlink2rest/bootstrap.sh
JackSkellet: t0.11.25 does not contain the COMPONENT_METADATA message (ID 397) required by this implementation. That message becomes available in mavlink2rest 1.0.x, but I understand that upgrading breaks the existing API. Would you prefer an API-compatible 0.x backport of message 397, or an alternative BlueOS integration path?

patrickelectric: The best way to workaround that, is to..
Send a PR to https://github.com/mavlink/rust-mavlink/tree/0.10 (0.10 branch), with a patch commit like this one https://github.com/mavlink/rust-mavlink/commit/b0da5b213812f1b26983c54d43d04163ee86835f and a update version commit like this one https://github.com/mavlink/rust-mavlink/commit/b210832fd0e31bdd12e69530f99a713ba8540153. After that, I will create a release.
With that done, we can then create a 0.x-dev branch to release an API compatible update mavlink2rest release.

JackSkellet:
rust-mavlink 0.10 backport: https://github.com/mavlink/rust-mavlink/pull/527

Tested on BlueOS with mavlink2rest 0.11.25; parameter metadata loads successfully.

patrickelectric:
check: https://github.com/bluerobotics/BlueOS/pull/4089

JackSkellet:
Updated to mavlink2rest 0.11.26 and reverted the temporary 1.x compatibility handling, restoring the existing BlueOS MAVLink code paths. Parameter metadata now decodes correctly on a live BlueOS system, while the remaining message handling works as before.

core/frontend/src/types/autopilot/parameter-table.ts
```diff
@@ -80,37 +66,56 @@ function fromPX4toArduPilotParametersMetadata(parameters: PX4ParametersMetadata[
...
       console.info(`Using metadata override from ${json_metadata_override}`)
       return metadata
     } catch (error) {
-      console.debug(`Metadata override not present`)
+      console.debug('Metadata override not present')
     }
     await fetchFirmwareVehicleType() // required to populate autopilot.vehicle_type
     const jsons = ardupilotParamPaths
-    let folder = "Copter"
+    let folder = 'Copter'
```
ES-Alexander: Was there a reason for this change (and the others like it)?

We generally avoid cleaning up code that we aren't already touching for other reasons, outside of occasional dedicated cleanup commits.

ES-Alexander (changes requested):
Thanks for the contribution @JackSkellet! :D

From some internal discussion, MAVLink Server [includes MAVFTP support](https://github.com/bluerobotics/mavlink-server/pull/209), and using its API for that would be a preferred approach over re-implementing MAVFTP handling in the ardupilot manager service.

core/frontend/src/types/autopilot/parameter-table.ts
JackSkellet: Ah, yes might have missed those, i had been doing some other unrelated changes on this branch. and then once i made the split, some of those changes might have slipped through.

i can revert those if that is prefered, so we keep the changes to a minimum.

photo of said unrelated work:
image


As for the Mavlink Server.  i will switch this implementation to the API later today.

JackSkellet: i looked into the MAVLink Server API and tested it.
The API receives the complete Metadata file, but then discards it when TerminateSession returns an EOF NAK. So there might be some cleaup handilng error going on here, unless i am missing something.

so should i make another Pull request over there for this specific issue?

ES-Alexander: > The API receives the complete Metadata file, but then discards it when TerminateSession returns an EOF NAK. So there might be some cleaup handilng error going on here, unless i am missing something.

@patrickelectric thoughts on this^?

> i can revert those if that is prefered, so we keep the changes to a minimum.

Yes please. Minimal changes are easier to review, including after a PR has merged (e.g. when inspecting the commit history to investigate something that's gone wrong).

patrickelectric: Hi @JackSkellet can you describe exactly what you are doing for me to replicate your issue ? If you use the mavlink-server GUI, does it work ?

JackSkellet: @patrickelectric the GUI seems to have the same issue.
image

replicating it using the API like this
`curl --get \
  --data-urlencode 'path=/@META/general.json' \
  http://localhost:8080/v1/mavftp/1/1/download`
  
It returns HTTP 500 with {"error":"FTP NAK: End of file","code":6}. parameters.json succeeds, so the problem is specific to how the completed general.json transfer’s termination response is handled.

JackSkellet:
ok, newest commit has the API instead of the custom MAVFTP implementation. and the unrelated cleanup should now also be restored.

The API's HTTP 500 error is still an issue.

PR #4072, opened 2026-07-31
commander: serialize environment variables as a dict

nicoschmdt:
Without this commander throws a serialization error in the terminal


joaoantoniocardoso (approved):
Good catch. It was showing `Could not get environment variables: [object Object].` on the frontend as well.

PR #3278, opened 2025-04-15
core: Move to use UV as package manager

joaomariolago:
## Note
Recommended extra testing since its a major change

## Tests:

* PI3: Tested (Pixhawk1 + HD USB Camera)
* PI4: Tested (Navigator + HD USB Camera)
* PI5: Tested (Navigator)

## Summary by Sourcery

Migrate core project to use UV as the package manager and restructure project dependencies

New Features:
- Add pyproject.toml configuration for core services and libraries
- Introduce UV package management configuration

Enhancements:
- Consolidate project dependencies across services
- Standardize Python version to 3.11+

Build:
- Add core-level pyproject.toml with workspace configuration
- Configure Python version and dependency sources

Chores:
- Remove existing setup.py files
- Standardize project structure with pyproject.toml
- Update dependency management

## Summary by Sourcery

Migrate core project to use UV as the package manager, standardizing Python dependencies and project structure across services and libraries.

New Features:
- Add pyproject.toml configuration for core services and libraries
- Introduce UV package management configuration

Enhancements:
- Consolidate project dependencies across services
- Standardize Python version to 3.11+

Build:
- Add core-level pyproject.toml with workspace configuration
- Configure Python version and dependency sources

Chores:
- Remove existing setup.py files
- Standardize project structure with pyproject.toml
- Update dependency management

PR #4094, opened 2026-08-06
Harden 1.4 against frontend corruption + freeze/lock dependencies

### LINKED ISSUE #4091: Build process can succeed and ship a corrupted frontend
Live example: the `1.4.4-beta.14` frontend corrupted during the build process, breaking the Vehicle Setup page. The `1.4-dev` image is fine.

Below is an investigation made by an AI:

---

## Investigation

The `1.4.4-beta.14` tag build ([run 30998631087](https://github.com/bluerobotics/BlueOS/actions/runs/30998631087)) and the `1.4-dev` build ([run 30936402008](https://github.com/bluerobotics/BlueOS/actions/runs/30936402008)) are the **same commit**, `82b545cd1777a4e570f3dfe067a2793753ab8960`. Inside the docker `frontend-builder` stage, Vite prints each chunk's size from the in-memory bundle, and `vite-plugin-compression` then prints the same file's size read back **from disk**. Those two numbers disagree in exactly one place:

| | Vite (in memory) | compression plugin (from disk) |
|---|---|---|
| 1.4-dev, `VehicleSetupView.3ca466a3.js` | 1279.05 KiB, gzip 340.02 KiB | 1279.11 kb, gzip **339.19 kb** |
| beta.14, `VehicleSetupView.df0d3c8b.js` | 1279.05 KiB, gzip 340.02 KiB | 1279.11 kb, gzip **204.11 kb** |

Rollup produced a byte-identical chunk in both runs. By the time the file was read back off disk in the beta.14 build, it still had the same length but compressed 40% smaller, which is what NUL-interleaved UTF-16 text does.

That pins the corruption to a specific window: **after rollup generated the chunk, during the write to disk, before compression ran.** It rules out the bundler, the compression plugin, the Docker image assembly, the registry, and the device, all of which were downstream of a file that was already wrong.

## Supporting evidence

The same run's `frontend-tests` job built the same source on a GitHub runner with the same bun 1.0.3 and vite 3.2.7, and produced a clean chunk (1279.11 kb on disk, gzip 341.00 kb). So the input was fine; only the docker-stage build went bad.

I scanned the core deploy job of the 25 most recent successful runs with the same in-memory-versus-on-disk comparison. Exactly one asset in one build is flagged, the beta.14 `VehicleSetupView.df0d3c8b.js`. Every other build is clean, so this is a roughly 1-in-25 nondeterministic event, not a regression.

The core job log shows only a single `vite building for production` for all three architectures, because `frontend-builder` is pinned to `$BUILDPLATFORM`. One bad write therefore propagated into the amd64, arm64 and arm/v7 manifests, which is what you observed in the published image.

## How it most likely happened

This part is inference from the numbers rather than something the log states outright, but the arithmetic is very specific. The corrupt file is 1,309,804 bytes, which is exactly 2 × 654,902 UTF-16 code units, and 1,309,804 bytes is also (within the log's rounding) the length the *correct* file should have had: 1,309,747 bytes of code plus the ~57-byte `//# sourceMappingURL=VehicleSetupView.df0d3c8b.js.map` trailer.

So the write emitted the right number of bytes and filled them with the wrong memory: the raw UTF-16 backing store of another string that rollup was writing concurrently, `model-viewer.1654248c.js.map`, truncated when the destination was full. A length taken from one string and content memcpy'd from another is the signature of a use-after-free or GC-relocation race in the runtime's file-write fast path, triggered by rollup writing all chunks and source maps concurrently under `build.sourcemap: true`. Nothing in JS-land can produce UTF-16 bytes in a UTF-8 write, so this is below the bundler.

The runtime is `oven/bun:1.0.3-slim`, pinned in `core/Dockerfile`, which is from September 2023. Bun has a long, documented history of `fs` write corruption bugs fixed since then, including [oven-sh/bun#16823 "Vite build fails on 1.2.1"](https://github.com/oven-sh/bun/issues/16823), fixed by [PR #16856](https://github.com/oven-sh/bun/pull/16856), which added a Vite integration test. I did not find an issue matching this exact fingerprint, so I can't point at one bug number and say "that's the one."

---

joaoantoniocardoso:
This patch is an amalgamation of cherries from the following PRs, now targeting 1.4:
* #4092 - Use frozen frontend dependencies in CI (still OPEN against master)
* #4053 - Fix CI: update bun
* #3278 - core: Move to use UV as package manager
* #3427 - core: services: Disable package in pyproject

The other commits will be sent to `master` soon.

Closes #4091


.github/workflows/test-and-deploy.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       run: |
         sudo apt-get install --no-install-recommends --assume-yes shellcheck parallel
         python -m pip install --upgrade pip
-        pip install poetry
+        pip install poetry==2.4.1
```
patrickelectric, line 34: Can we skip the python part since we don't have problems with and we are closer to a stable release ? I would limit the scope of this PR to just deal with the frontend.

joaoantoniocardoso: We can, but if we need to apply a backport in the future, there's no guarantee we won't introduce a bug. 

I made sure we are pinning to the same dependency versions that are being used in the build today.

I'm far more comfortable pinning what we can before release.

joaoantoniocardoso: update: reran the dependency diff, got one drift, fixed now.

patrickelectric: The scope of this PR is quite big for a stable release.
It mixes lock of frontend and big changes in backend project management.
- Move projects to uv
- Move projects from setup.py to pyproject.
- Move projects to venv
- Remove yarn.lock ?? For what reason, even in master we have yarn.

Python projects are already dependency lock, if there is a place that we are not, we can just add such version.

joaoantoniocardoso: The non-split of those concerns into separate PRs is intentional. I want to address all *before* the stable release.

The diff can look scary, but almost all of it is just the lock files themselves, so it's not difficult to review carefully.

> Python projects are already dependency lock, if there is a place that we are not, we can just add such version.

This is exactly what we are doing here: the Python dependencies are not pinned in practice (we "lock" `packaging==20.4`, but the image is shipping `26.2`), and we have broken dependencies already, and nothing guarantees we won't have more while I'm writing this message:

image

(Yes, we don't use flask directly, it's brought transitively via connexion, but the point is: it is there and silently broken today. Which will be the next one?)

The practical effect on the built image is tight and we must work to make it under control, so we should spend effort in improving this rather than ignoring the problem.

You can run agents to inspect the differences before/after; we can adjust any detail if needed, but the dependency-locking before the release is a *must* alongside the anti-corruption measure.

Docker dive:

[dive-1.4-dev.json](https://github.com/user-attachments/files/30837567/dive-1.4-dev.json)

[dive-branch.json](https://github.com/user-attachments/files/30837479/dive-branch.json)

pip freezes:

[pip-1.4-dev.txt](https://github.com/user-attachments/files/30837486/pip-1.4-dev.txt)
[pip-branch.txt](https://github.com/user-attachments/files/30837489/pip-branch.txt)

thanks

patrickelectric (approved):
image

PR #4095, opened 2026-08-06
core: kraken: Fix extension uptime after clock sync

### LINKED ISSUE #3416: extension-manager: wrong extensions uptime
Image

Daniiiil1:
Closes #3416

## What changed

- derive running-container uptime from the host process's monotonic start time instead of Docker's wall-clock duration
- expose that raw duration as `uptime_seconds` in the Kraken API
- keep human-readable duration formatting in the frontend, where presentation belongs
- preserve Docker status suffixes such as `(healthy)`
- fall back to Docker's original status if process uptime cannot be read
- cover monotonic calculation, fallback behavior, and API model output with regression tests

## Root cause

Docker builds the human-readable `Status` field by subtracting `StartedAt` from the current wall clock. BlueOS can start extensions before the Raspberry Pi synchronizes its clock, so a later NTP correction makes long-running extensions appear older than the host.

BlueOS Core runs in the host PID namespace. Kraken can therefore derive each container process's start offset from boot and compare it with `time.monotonic()`, which is unaffected by wall-clock corrections.

## Architecture

Kraken returns the raw monotonic uptime as data. The extension card formats that duration for display and combines it with Docker's health suffix. This keeps presentation out of the backend while retaining the existing UI.

## Validation

- `pytest -q core/services/kraken/harbor/test_container.py` (3 passed)
- targeted ESLint and Stylelint on the frontend changes
- production frontend build (2,584 modules transformed; existing Sentry-auth warning only)
- Ruff 0.0.288, Black 22.3.0, isort 5.8, and mypy 1.4.1 on the changed backend files

### DIFF
--- a/core/frontend/src/components/kraken/cards/InstalledExtensionCard.vue
+++ b/core/frontend/src/components/kraken/cards/InstalledExtensionCard.vue
@@ -242,7 +242,7 @@ import SpinningLogo from '@/components/common/SpinningLogo.vue'
 import settings from '@/libs/settings'
 import system_information from '@/store/system-information'
 import { JSONValue } from '@/types/common'
-import { ExtensionData, InstalledExtensionData } from '@/types/kraken'
+import { ExtensionData, InstalledExtensionData, RunningContainer } from '@/types/kraken'
 import { Disk } from '@/types/system-information/system'
 import { prettifySize } from '@/utils/helper_functions'
 
@@ -264,9 +264,9 @@ export default Vue.extend({
       required: true,
     },
     container: {
-      type: Object as PropType<{status: string}>,
+      type: Object as PropType<RunningContainer>,
       required: false,
-      default: undefined as {status: string} | undefined,
+      default: undefined as RunningContainer | undefined,
     },
     extensionData: {
       type: Object as PropType<ExtensionData>,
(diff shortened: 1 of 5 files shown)


core/services/kraken/harbor/container.py
```diff
@@ -13,6 +14,62 @@
...
+        return f"Up {cls._human_duration(uptime_seconds)}{suffix}"
+
+    @classmethod
+    def _container_model(cls, container: DockerContainer, details: Dict[str, Any]) -> ContainerModel:
+        pid = details.get("State", {}).get("Pid", 0)
+        return ContainerModel(
+            name=container["Names"][0],
+            image=container["Image"],
+            image_id=container["ImageID"],
+            status=cls._status_with_monotonic_uptime(container["Status"], pid),
```
patrickelectric: human representations should be done in the frontend, not in the back.
We should just fix the time with the monotonic clock.

patrickelectric (changes requested):

core/frontend/src/components/kraken/cards/InstalledExtensionCard.vue
```diff
@@ -381,7 +381,29 @@ export default Vue.extend({
...
     getStatus(): string {
-      return this.container?.status ?? 'N/A'
+      if (!this.container) return 'N/A'
+      if (this.container.uptime_seconds == null) return this.container.status
+
+      const suffix_start = this.container.status.indexOf(' (')
+      const suffix = suffix_start >= 0 ? this.container.status.slice(suffix_start) : ''
+      return `Up ${this.humanDuration(this.container.uptime_seconds)}${suffix}`
+    },
+    humanDuration(duration_seconds: number): string {
```
patrickelectric, line 391: we use date-fns

core/services/kraken/harbor/container.py
```diff
@@ -31,7 +54,9 @@ async def kill_all_by_name(client: Docker, container_name: str) -> None:
 
     @staticmethod
     # pylint: disable=too-many-locals
-    async def _get_stats_from_containers(containers: List[DockerContainer]) -> Dict[str, ContainerUsageModel]:
+    async def _get_stats_from_containers(
+        containers: List[DockerContainer],
```
patrickelectric, line 58: why this style change ?

patrickelectric (changes requested):

PR #4099, opened 2026-08-07
1.4: harden frontend builds against corrupted assets

Daniiiil1:
## Summary
- backport Bun 1.3.14 and setup-bun v2 to the 1.4 release branch
- regenerate the 1.4 lockfile with Bun 1.3.14 and enforce frozen installs in CI and Docker
- validate generated text assets as strict UTF-8 and reject NUL bytes after every Vite build

This targets the exact observed failure in #4091: the corrupted JS chunk was NUL-interleaved and would now fail the build instead of being published. Bun 1.3.14 also matches current master, removing the old 1.0.3 runtime implicated in the write corruption.

Addresses #4091 for `1.4-dev`.

## Testing
- Bun 1.3.14 frozen install passed (917 packages)
- full frontend build passed (2,439 modules)
- generated-asset validation passed (85 assets)
- frontend ESLint passed
- validator ESLint passed
- workflow YAML parsed successfully


core/frontend/package.json
```diff
@@ -9,7 +9,8 @@
   ],
   "scripts": {
     "dev": "vite --host",
-    "build": "NODE_ENV=production vite build",
+    "build": "NODE_ENV=production vite build && bun run validate-build",
+    "validate-build": "bun scripts/validate-build.ts",
```
patrickelectric, line 13: This is not being used in CI, what is the point ?

joaoantoniocardoso:
Thanks for the submission, but the original issue is already closed.

PR #4100, opened 2026-08-07
Limit logo and vehicle images to 400px

### LINKED ISSUE #3987: Limit logo and vehicle image to max 400px size
To avoid size problems with big images

Daniiiil1:
## Summary

- add optional maximum dimensions and file sizes to the shared image picker
- compress uploads through `browser-image-compression` in a Web Worker while preserving aspect ratio
- cap vehicle and logo images at 400 px and 0.5 MB without enlarging smaller files

The 400 px ceiling follows #3987 and still gives substantial headroom over the 35 px rendered branding slot, including high-density displays. The file-size cap addresses the storage and transfer concern independently of resolution.

Closes #3987

## Validation

- targeted ESLint on `ImagePicker.vue` and `VehicleBanner.vue`
- targeted Stylelint on both changed Vue files
- `corepack yarn build` (2,585 modules)


core/frontend/src/components/app/ImagePicker.vue
```diff
@@ -224,11 +229,60 @@ export default Vue.extend({
         })
       this.loadImages()
     },
+    async resizeImage(file: File): Promise<Blob> {
```
patrickelectric: we probably should use `browser-image-compression` over doing all this.

patrickelectric (changes requested):

rafaellehmkuhl:
I would suggest investigating if 400px is the correct number, taking 4K screens into consideration. The way that I see, the main problem is not the resolution/dimensions, but the image size, aspect ratio, and the way it's being rendered.

PR #3250, opened 2025-04-03
Update delete log to use stream

patrickelectric:
Peek 2025-04-02 21-50

Fix #3179 

## Summary by Sourcery

Update log deletion process to use streaming approach for real-time progress tracking

New Features:
- Add a new streaming endpoint for log file deletion with detailed progress information

Enhancements:
- Implement a streaming deletion mechanism that provides real-time updates about file deletion progress
- Add progress tracking and detailed information during log file deletion

Issue #3179, opened 2025-02-26
bug: Timeout trying to delete system logs thought the interface


voorloopnul:
### Bug description

I have 3.8gb of system logs in my blueos, when I try to delete it, i get the following error:

![Image](https://github.com/user-attachments/assets/dbb0adba-face-4b6d-9a7a-f4dd8fdb7033)

I suspect the exactly amount of logs that will cause this issue is also related to the sd card speed.

### Steps to reproduce

1. Add a bunch of log files in the system logs folder
2. Go to the BlueOS Gear Icon
3. Click in "Remove logs"

### Prerequisites

- [x] I have checked to make sure that a similar request has not already been filed or fixed.

voorloopnul:
I also realized that something breaks after this event, as the gear icon no longer works:

![Image](https://github.com/user-attachments/assets/57b242b0-d659-4a63-ba5a-57db4306b28c)

PR #4123, opened 2026-08-11
[1.4] Stream MAVLink log removal and stop asking lsof file by file

### LINKED ISSUE #4105: Timeout when removing ~1GB of `MAVLink Log Files`
`BlueOS 1.4.4-beta.15`

When clicking the "REMOVE" button on the BlueOS settings (COG icon), the user has no feedback, and it can fail with `Error: timeout of 20000ms exceeded`.

This timeout error is clearly an implementation limitation: we should use a stream there, like how we are doing with the "System Log Files".

Image of the error:

Image

After reloading BlueOS, I can see it had almost finished:

Image

joaoantoniocardoso:
Fixes #4105 for 1.4

## After:

https://github.com/user-attachments/assets/6cb1bea2-b459-4007-bc64-6e487fef39f9

## Problem

Clearing ~1GB of MAVLink logs failed with `Error: timeout of 20000ms exceeded` and no progress feedback, while the deletion kept running on the vehicle. System logs already stream their deletion (#3250, backported in #3806), but the MAVLink path was never migrated.

Digging into why deletion is slow at all turned up a second, larger problem. Every file is passed to its own `lsof` call to check whether something still has it open, and each `lsof` invocation rescans every process in the system. Deletion therefore costs one process spawn per file and scales with file count, not with bytes removed.

Measured on a Raspberry Pi 4 with 300 files (900 MiB), one file deliberately held open:

| strategy | time | detected the open file |
| --- | --- | --- |
| `lsof` per file (before) | 41.8 s | yes |
| one `lsof` call with all paths | 0.33 s | yes |
| one `lsof -w +D` on the folder | 0.19 s | yes |
| the actual `unlink()` calls | 0.006 s | — |

At 3000 files the single `+D` call still takes 0.245 s, where per-file would take about seven minutes. This is also why `/var/logs/blueos` (123 files) sat right at the 20 s limit in #3179.

## Changes

- `commonwealth.utils.general`: take one `lsof -w -n -P -S 2 -F n +D <folder>` snapshot of open files per deletion and reuse it while recursing, in both `delete_everything` and `delete_everything_stream`. When `lsof` fails the snapshot is `None` and each file falls back to the previous individual check, so the "when in doubt, keep the file" behaviour is unchanged.
- `commander`: add `/services/remove_mavlink_log_stream`, mirroring `remove_log_stream`. The old endpoint stays for API compatibility.
- `SettingsMenu`: consume the stream with no timeout and show the existing progress UI, which now sits next to the MAVLink logs rather than under the service logs.

Two details worth flagging for review: searching a folder makes `lsof` exit with `1` whether or not it found open files, so the result is parsed from stdout and failure is detected from stderr; and `-w` silences warnings about file systems `lsof` cannot stat, which are emitted on any Docker host and would otherwise look like failures.

## Testing

New tests in `commonwealth/utils/tests/test_general.py` cover open files surviving deletion, exactly one process being spawned for a 20-file tree (sync and streaming), the fallback path when `lsof` is unavailable, and `open_files_under` itself.

On hardware (Raspberry Pi 4, 1.4 image), 973 MB across 302 MAVLink log files:

- before: 40.4 s on the blocking endpoint, past the 20 s frontend timeout
- streaming alone: 41.4 s total, first byte in 5 ms, 338 progress updates, no timeout
- streaming with the single `lsof` pass: **0.74 s**, 300 progress updates

Commander stayed responsive throughout: latency for concurrent requests during deletion went from 201 ms median (495 ms worst) to 15 ms. A file held open by another process, standing in for the log ArduPilot is actively writing, was preserved and never reported as deleted. The UI was exercised end to end against a vehicle: progress renders under MAVLink Logs, the folder size drops, and no error is raised.

## Notes

Targets `1.4-dev`. The equivalent change for `master` is ready and will follow as a separate PR.

### DIFF
--- a/core/frontend/src/components/app/SettingsMenu.vue
+++ b/core/frontend/src/components/app/SettingsMenu.vue
@@ -127,7 +127,7 @@
             <v-btn
               v-tooltip="'Frees up space on the SD card deleting MAVLink logs'"
               class="ma-2"
-              :disabled="disable_remove_mavlink"
+              :disabled="disable_remove_mavlink || deletion_in_progress"
               @click="remove_mavlink_log_files"
             >
               <v-icon left>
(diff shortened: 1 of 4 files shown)


core/frontend/src/components/app/SettingsMenu.vue
```diff
@@ -378,24 +378,53 @@ export default Vue.extend({
...
+                this.current_deletion_path = info.path.length > 30 ? `...${info.path.slice(-20)}` : info.path
+                this.current_deletion_size = info.size
+                this.current_deletion_total_size = total_deleted
+                this.current_deletion_status = info.success ? 'Deleting..' : 'Failed to delete file..'
+              } catch (e) {
+                console.error('Error parsing deletion info:', e)
+              }
+            }
+          },
+          cancelToken: this.deletion_log_abort_controller?.token,
```
Williangalvani, line 415: what is this for? it doesnt look like we are using it.
Also it seems to be deprecated, not that it means much in our codebase...

joaoantoniocardoso: It is used: the `show_dialog` watcher a few lines above calls `this.deletion_log_abort_controller?.cancel()` when the settings menu closes, so an in-flight deletion stream is aborted instead of keeping the request (and the progress state) alive after the dialog is gone.

That is the same wiring `remove_service_log_files` already has in this file; this commit copies it so MAVLink deletion behaves the same way.

On the deprecation: axios `CancelToken` is replaced by `AbortController` / `signal`. This 1.4 file still uses `CancelToken` for the existing system-log stream, so I kept the new path on the same API. Master already migrated that sibling to `AbortController` (and the master PR follows that). Happy to migrate both streams here in a follow-up if you would rather not leave the deprecated API in place.

Williangalvani: hah cool.
I was looking for a cancel button, but we have to close the dialog instead. I think that is fair.

Williangalvani (approved):

PR #4138, opened 2026-08-12
stop sensor getters from burning CPU on every HEARTBEAT

### LINKED ISSUE #3465: BlueOS frontend is using too much resources
In Firefox, it happens quicker than in Google Chrome, taking only a few seconds to reach 1GB of RAM, while also spinning processes at 100% CPU.

In Google Chrome, it starts at around 200MB and grows by ~100MB per minute.

Here's a screenshot after ~8 minutes without doing anything:

Image

nicoschmdt:
- `autopilot.ts` uses a map for the parameters and its search is now O(1)
- `ardupilot_sensors.ts` uses fixed param names to avoid full-table scans
- `HealthTrayMenu.vue` moved the sensor/heartbeat to the computed property; sensor status only updates if the menu is open and heartbeat mutations are only fired if they change the current value of `system_id`/ `autopilot_type` / `armed`

fix: #3465

### DIFF
--- a/core/frontend/src/components/health/HealthTrayMenu.vue
+++ b/core/frontend/src/components/health/HealthTrayMenu.vue
@@ -34,6 +34,7 @@
       mdi-lightning-bolt
     </v-icon>
     <v-menu
+      v-model="menu_open"
       :close-on-content-click="false"
       nudge-left="150"
       nudge-bottom="25"
@@ -49,7 +50,7 @@
         >
           <v-icon
             v-if="heartbeat_age() < time_limit_heartbeat"
-            v-tooltip="heartbeat_message()"
+            v-tooltip="heartbeat_message"
             class="px-1"
             :color="heartbeat_color()"
           >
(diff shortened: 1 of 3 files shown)


core/frontend/src/components/health/HealthTrayMenu.vue
```diff
@@ -232,20 +265,6 @@ export default Vue.extend({
...
-      const opacity = Math.max(0.4, 1.4 - this.heartbeat_age() / 1000)
-
-      return `rgba(${r}, ${g}, ${b}, ${opacity})`
-    },
-    heartbeat_message() : string {
-      if (ardupilot_sensors.all_sensors_calibrated) {
-        return 'MAVLink heartbeats arriving as expected'
-      }
-      return 'One or more sensors are not calibrated'
-    },
```
Williangalvani, line 248: I think made the heartbeat stop updating. it just stays there, at fixed opacity.
Stopping the autopilot also didnt cause a switch to red/broken

nicoschmdt: Ooh, I'll see to fix that

patrickelectric (approved):

PR #4150, opened 2026-08-13
frontend: Use image icons for video thumbnail stills

joaoantoniocardoso:
https://github.com/user-attachments/assets/47d9da21-5500-4aca-ae39-5f09a364b160

## Summary
- Replace the camera / play-pause thumbnail controls with `mdi-image` (single still) and `mdi-image-sync` (1 Hz refresh) so the play button is no longer mistaken for a live video stream.
- Latch the sync button while continuous fetch is on, and show `1 Hz` instead of `LIVE` in pirate mode.

Fixes https://github.com/bluerobotics/BlueOS/issues/4149

## Test plan
- [ ] Open Video, confirm the thumbnail bar shows an image icon and an image-sync icon (not camera / play).
- [ ] Fetch a single thumbnail; confirm one still appears and continuous fetch does not start.
- [ ] Start continuous thumbnails; confirm the sync button stays visually latched and frames update about once per second.
- [ ] Stop continuous thumbnails; confirm fetching stops and the latch clears.
- [ ] In pirate mode, confirm the overlay says `1 Hz` rather than `LIVE`.

### DIFF
--- a/core/frontend/src/components/video-manager/VideoThumbnail.vue
+++ b/core/frontend/src/components/video-manager/VideoThumbnail.vue
@@ -40,7 +40,7 @@
         class="thumbnail-overlay text-caption"
       >
         <template v-if="continuous_mode">
-          LIVE
+          1 Hz
           <template v-if="last_fetch_ms !== null">
             {{ last_fetch_ms }}ms
           </template>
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

ES-Alexander:
I think I'd be marginally more inclined to `image-plus` or `image-refresh` for the left (single image) icon, but as-is it's already much less likely the original confusion will occur, where a user thought the right button would show a video preview 👍

PR #4157, opened 2026-08-13
core: frontend: vehiclesetup: PwmSetup: Fix components highlight visibility

### LINKED ISSUE #1674: Vehicle (on pwm outputs tab) very hard to see in dark mode
The motor, for example, cannot be seen.

image

A lighter background would make help here.

patrickelectric:
Fix #1674

[Kooha-2026-09-03-18-11-54.webm](https://github.com/user-attachments/assets/1ca3741b-0ec8-49e9-86cb-83132f9d91a1)

### DIFF
--- a/core/frontend/src/components/vehiclesetup/PwmSetup.vue
+++ b/core/frontend/src/components/vehiclesetup/PwmSetup.vue
@@ -6,7 +6,7 @@
         sm="8"
       >
         <v-card>
-          <vehicle-viewer :highlight="highlight" :transparent="true" :autorotate="false" />
+          <vehicle-viewer :highlight="highlight" :autorotate="false" />
         </v-card>
         <v-card class="mt-3">
           <v-overlay :value="!has_focus">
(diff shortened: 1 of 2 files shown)


joaoantoniocardoso:
I'm unsure whether this tackles the issue: I can barely notice the difference in light mode, and it looks weird in dark mode:

image

image

Maybe it should be less transparent, instead? Or having a transparency slider and/or a background color luminosity if we can't find good values that work for everybody?

patrickelectric:
@joaoantoniocardoso dropped the grey card, model is opaque now. Check now ?

joaoantoniocardoso:
> @joaoantoniocardoso dropped the grey card, model is opaque now. Check now ?

I think that solves the problem of the vehicle visibility by creating another problem: now, when hovering over the motors list, we can't visually locate which motor is which in the frame.

patrickelectric:
@joaoantoniocardoso restored transparency for hover and raised the ghost alpha. Check now ?

joaoantoniocardoso:
It does look better, but it is still difficult to identify the motors, especially in dark mode. Maybe we can tint the motor a color with good contrast against a light or dark background, say, red?

patrickelectric:
Ok, will take a look

patrickelectric:
Rebasing over master

patrickelectric:
@joaoantoniocardoso check now

joaoantoniocardoso (approved):

PR #4175, opened 2026-08-18
frontend: store: Divide disk write rate by linux2rest sample window

joaoantoniocardoso:
## Summary
- The tray disk widget (`W: … MB/s`) treated each `available_space_B` jump as happening over the **2 s** frontend poll. linux2rest only refreshes that snapshot every **5 s**, so a ~2.1 MB/s recording showed as **~5.0–5.1 MB/s**, then stuck there because a later zero delta kept the inflated rate.
- Divide capacity deltas by the same 5 s Sampler window already used for network (`linux2restProbeRateBps`). Match disks by `mount_point` (many bind mounts share `name: /dev/root`). Drop unused `Disk.last_update`.
- No linux2rest API change. Self-hosted frontend against a vehicle writing ~2 MB/s showed **W: 2.0 MB/s**.

Helps https://github.com/bluerobotics/BlueOS/issues/4172: this PR doesn't solve it, but at least we get a correct reading from the widget
Related (not required for this PR): https://github.com/patrickelectric/linux2rest/issues/38

## Test plan
- [x] Self-host `core/frontend` with `BLUEOS_ADDRESS=http://<vehicle>/ yarn dev`
- [x] With a sustained ~16 Mbps recording (or any ~2 MB/s writer), tray shows **~2 MB/s**, not ~5 MB/s, and holds there across several 2 s polls
- [x] After stopping the writer, rate drops toward zero on the next non-zero capacity change
- [x] Disk fill still updates (`xx/yyGB (zz%)`)
- [x] Network tray rates still look sane (this PR reuses their helper, does not change that path)

### DIFF
--- a/core/frontend/src/store/system-information.ts
+++ b/core/frontend/src/store/system-information.ts
@@ -20,7 +20,7 @@ import {
   CPU, Disk, Info, Memory, Network, Process, System, Temperature,
 } from '@/types/system-information/system'
 import back_axios, { isBackendOffline } from '@/utils/api'
-import { networkProbeRateBps } from '@/utils/networking'
+import { linux2restProbeRateBps } from '@/utils/linux2rest'
 
 export enum FetchType {
     ModelType = 'model',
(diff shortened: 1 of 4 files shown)


patrickelectric (approved):

patrickelectric (approved):

patrickelectric:
Because 1 approve is not enough

PR #4180, opened 2026-08-18
bag_of_holding: keep JSON file in memory

### LINKED ISSUE #4177: Bag of holding struggles with fast operations
Historically, we always had problems on syncing Cockpit settings with BlueOS, using the bag of holdings. It struggles to keep up when several changes are requested in series, and ends up failing on several of them.

We have "solved" this problem on Cockpit's side by implementing some retry pipelines, but that just masks the problem. In the end the user still has to wait sometimes dozens of settings for the whole sync to be finished, depending on how fast/slow Bag of Holdings responds.

This was also bring up today [by an user in our forums](https://discuss.bluerobotics.com/t/strange-error-messages-in-blueos-after-lauching-cockpit/23486). As soon as he opens Cockpit, Bag of Holdings starts to struggle.

Image

nicoschmdt:
fix: #4177 

Now bag of holding stays in memory and is written to the file with an interval of 1.0s. Changes that arrive in that interval will get patched together, that way only the most recent version of the Bag of Holding is written in the file.

I also organized the code sections of the file to be easier to read and so that we avoid having the endpoint methods together with the other functions

### DIFF
--- a/core/services/bag_of_holding/main.py
+++ b/core/services/bag_of_holding/main.py
@@ -67,6 +67,7 @@ async def parse_nullable_body(payload: Any | None = Body(None)) -> Any:
 
 
 @app.post("/overwrite")
+@version(1, 0)
 async def overwrite_data(payload: dict[str, Any] = Body(...)) -> JSONResponse:
     logger.debug(f"Overwrite: {json.dumps(payload)}")
     write_db(payload)
--- a/core/services/bag_of_holding/main.py
+++ b/core/services/bag_of_holding/main.py
@@ -15,7 +15,6 @@ from fastapi import Path as FastPath
 from fastapi.responses import HTMLResponse, JSONResponse
 from fastapi_versioning import VersionedFastAPI, version
 from loguru import logger
-from pydantic import BaseModel
 from uvicorn import Config, Server
 
 SERVICE_NAME = "bag-of-holding"
@@ -33,11 +32,6 @@ app = FastAPI(
 )
 app.router.route_class = GenericErrorHandlingRoute
 logger.info(f"Starting Bag of Holding: {FILE_PATH}")
-
-
-class KeyValue(BaseModel):
-    key: str
-    value: Any
 
 
 def read_db() -> Any:
--- a/core/services/bag_of_holding/main.py
+++ b/core/services/bag_of_holding/main.py
@@ -2,8 +2,10 @@
 import asyncio
 import json
 import logging
+import os
 from pathlib import Path
-from typing import Any, Dict
+from types import SimpleNamespace
+from typing import Any
 
 import appdirs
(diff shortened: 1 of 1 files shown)


rafaellehmkuhl:
Daaaaamn nice one Nico!

joaoantoniocardoso (commented):
Nice one!

I have two main concerns for us to think about:

1. I foresee this code, in the future, turning into async. Once that happens, the current approach will race. We can use a mutex around the internal state (the in-memory db) to cover it.
2. The in-memory strategy works fine but imposes limitations on what we can handle. We should ask @rafaellehmkuhl about how they were using it, how they are using it today, and how they plan to be using it in the future. To keep in mind: there may be leftovers from previous iterations on users' bags bigger than what we expect.

Other than that, the app declaration location change and the version seem unrelated to the rest of the commit.

Thanks

patrickelectric (changes requested):
There are multiple things being done in the same commit.
- KeyValue was removed
- Code was moved around without logical change
- Missing @version was added
- The issue fix itself

I would recommend breaking in multiple commits.

nicoschmdt:
> Nice one!
> 
> I have two main concerns for us to think about:
> 
> 1. I foresee this code, in the future, turning into async. Once that happens, the current approach will race. We can use a mutex around the internal state (the in-memory db) to cover it.
> 2. The in-memory strategy works fine but imposes limitations on what we can handle. We should ask @rafaellehmkuhl about how they were using it, how they are using it today, and how they plan to be using it in the future. To keep in mind: there may be leftovers from previous iterations on users' bags bigger than what we expect.
> 
> Other than that, the app declaration location change and the version seem unrelated to the rest of the commit.
> 
> Thanks

Regarding the adding of mutexes, I think it's better if we add them when moving the service to async. I can work on that next, it just seems the addition would be too out of place in this PR :thinking:

joaoantoniocardoso:
> > Nice one!
> > I have two main concerns for us to think about:
> > 
> > 1. I foresee this code, in the future, turning into async. Once that happens, the current approach will race. We can use a mutex around the internal state (the in-memory db) to cover it.
> > 2. The in-memory strategy works fine but imposes limitations on what we can handle. We should ask @rafaellehmkuhl about how they were using it, how they are using it today, and how they plan to be using it in the future. To keep in mind: there may be leftovers from previous iterations on users' bags bigger than what we expect.
> > 
> > Other than that, the app declaration location change and the version seem unrelated to the rest of the commit.
> > Thanks
> 
> Regarding the adding of mutexes, I think it's better if we add them when moving the service to async. I can work on that next, it just seems the addition would be too out of place in this PR 🤔

That's fine if we manage not to lose track of it :)

patrickelectric (approved):

PR #4184, opened 2026-08-19
ardupilot-manager: return empty list on non-Linux flight controllers

### LINKED ISSUE #4168: bug: GET /ardupilot-manager/v1.0/serials returns HTTP 500 on non-Linux flight controllers
### Bug description

`GET /ardupilot-manager/v1.0/serials` returns **HTTP 500** on vehicles whose flight controller is **not** a Linux board (e.g. Pixhawk).

**Affects `1.4-dev`.** Reproduced on a Pixhawk1 DUT.

`AutopilotManager.get_serials` (`core/services/ardupilot_manager/autopilot_manager.py`) does:

```python
if "serials" not in self.configuration:
    assert isinstance(self._current_board, LinuxFlightController)
    return self._current_board.get_serials()
```

On Navigator that assert holds (`LinuxFlightController.get_serials` is implemented). On Pixhawk / other serial FC, `_current_board` is not a `LinuxFlightController`, the assert fires, and FastAPI returns 500. `LinuxFlightController.get_serials` itself is `NotImplementedError` on the base class.

The UI or any client that lists ArduPilot serial mappings then fails instead of getting `[]` or a board-typed empty/unsupported response.

### Steps to reproduce

1. BlueOS `1.4-dev` attached to a Pixhawk (or any non-Linux FC), no `serials` key saved in ardupilot-manager settings.

```bash
curl -sS -D - -o /dev/null "http://<blueos>/ardupilot-manager/v1.0/serials"
```

2. Observe: `HTTP/1.1 500`.
3. Contrast: same GET on Navigator returns 200 and the ttyS/ttyAMA mapping.

Expected: 200 with `[]`, or 404/501 “serial mapping not applicable to this board” — not an assertion 500.

### Primary pain point(s)

Serial-port configuration and any UI that loads `/serials` is broken on Pixhawk-class vehicles even though the rest of autopilot-manager is up.

### Additional context

- Suggested fix: if the board is not a `LinuxFlightController`, return `[]` (or a typed 4xx) instead of asserting.
- Live QA: Pixhawk1 DUT only; Navigator DUTs did not hit this.

### Prerequisites

Safe-bet:
## Description
Fixes #4168

`GET /ardupilot-manager/v1.0/serials` previously asserted that `self._current_board` was an instance of `LinuxFlightController`, which caused an HTTP 500 error when connected to non-Linux flight controllers (e.g. Pixhawk).

This PR checks if `self._current_board` is a `LinuxFlightController` and returns an empty list `[]` if it is not, allowing clients and the UI to handle non-Linux flight controllers gracefully.

### DIFF
--- a/core/services/ardupilot_manager/autopilot_manager.py
+++ b/core/services/ardupilot_manager/autopilot_manager.py
@@ -302,7 +302,8 @@ class AutoPilotManager(metaclass=Singleton):
         # The first column comes from https://ardupilot.org/dev/docs/sitl-serial-mapping.html
 
         if "serials" not in self.configuration:
-            assert isinstance(self._current_board, LinuxFlightController)
+            if not isinstance(self._current_board, LinuxFlightController):
+                return []
             return self._current_board.get_serials()
         serials = []
         for entry in self.configuration["serials"]:


patrickelectric:

joaoantoniocardoso (approved):
Tested, thanks!

PR #4185, opened 2026-08-19
kraken: raise ExtensionNotFound when restarting uninstalled extension

### LINKED ISSUE #4160: bug: Restarting an unknown extension returns HTTP 400 (not running) instead of 404 (not found)
### Bug description

`POST /kraken/v2.0/extension/{identifier}/restart` for an identifier that was **never installed** returns **HTTP 400** with `Extension {identifier} have no running versions` instead of **HTTP 404** `Extension {identifier} not found`.

Restart uses `Extension.from_running(identifier)`. That loads settings by identifier only (no tag). An empty install list is treated as “not running” (`ExtensionNotRunning` → 400), not “not found”. `ExtensionNotFound` (404) is raised only when **both** identifier and tag are supplied, which is why tagged install/uninstall 404 correctly for the same missing id.

This wiring landed in https://github.com/bluerobotics/BlueOS/pull/2641. The PR does not discuss 400 vs 404 for restart. The request is still rejected (nothing restarts); the status and message are wrong.

### Steps to reproduce

1. On BlueOS `master` or `1.4-dev`:

```bash
curl -sS -D - -X POST \
  "http://<blueos>/kraken/v2.0/extension/np.no.such.extension/restart"
```

2. Observe: HTTP 400, detail `Extension np.no.such.extension have no running versions`.
3. Contrast a tagged route for the same missing id, e.g. uninstall/install with a tag — those return 404 `Extension …:tag not found`.

### Primary pain point(s)

Clients and the UI can show “extension is not running” for an identifier that does not exist. Missing resource vs stopped resource cannot be distinguished on this route.

### Additional context

- Suggested fix: in `from_running`, if `installed` is empty raise `ExtensionNotFound`; if installed but none enabled, keep `ExtensionNotRunning`.
- Searched open issues; nothing tracks this 400-vs-404 case.

### Prerequisites

Safe-bet:
## Description
Fixes #4160

`POST /kraken/v2.0/extension/{identifier}/restart` for an extension identifier that was never installed previously returned HTTP 400 (`Extension {identifier} have no running versions`) instead of HTTP 404 (`Extension {identifier} not found`).

This PR updates `Extension.from_running(identifier)` to raise `ExtensionNotFound` (HTTP 404) when the `installed` list is empty, while preserving `ExtensionNotRunning` (HTTP 400) when the extension is installed but disabled.

### DIFF
--- a/core/services/kraken/extension/extension.py
+++ b/core/services/kraken/extension/extension.py
@@ -389,6 +389,8 @@ class Extension:
     @classmethod
     async def from_running(cls, identifier: str) -> "Extension":
         installed: List[Extension] = cast(List[Extension], await cls.from_settings(identifier))
+        if not installed:
+            raise ExtensionNotFound(f"Extension {identifier} not found")
 
         enabled = [ext for ext in installed if ext.source.enabled]
         if not enabled:


patrickelectric:

joaoantoniocardoso (approved):
Tested, thanks!

PR #4186, opened 2026-08-19
kraken: check container exists before streaming logs to return 404

### LINKED ISSUE #4161: bug: Kraken container log endpoint returns HTTP 200 when the container does not exist
### Bug description

`GET /kraken/v2.0/container/{container_name}/log` always starts as **HTTP 200**, even when `{container_name}` is not a real container.

The handler returns a `StreamingResponse` before the log generator runs. Docker lookup (and the 404) only happen on the first yield, after headers are already sent. Contrast `GET /kraken/v2.0/container/{container_name}/details`, which awaits first and correctly returns 404.

HTTP clients and monitors that only look at status treat a missing container as success. Any 404 that does appear is a JSON fragment in the stream body, not the HTTP status. The default path also sends heartbeats (`streamer(..., heartbeats=0.1)`), so a client can see 200 + heartbeat JSON with no error at all.

This path landed in https://github.com/bluerobotics/BlueOS/pull/2640. Continuous logs in https://github.com/bluerobotics/BlueOS/pull/2852 kept the same pattern. Current `master` only adds `DockerCtx(timeout=0)` and a historical-logs helper; neither checks existence before streaming.

### Steps to reproduce

1. On BlueOS `master` or `1.4-dev`:

```bash
curl -sS -D - -o /tmp/log.body \
  "http://<blueos>/kraken/v2.0/container/np_no_such_container/log"
```

2. Observe: status line is `HTTP/1.1 200 OK`. Body is empty or heartbeat JSON, not an HTTP 404.
3. Contrast:

```bash
curl -sS -o /dev/null -w '%{http_code}\n' \
  "http://<blueos>/kraken/v2.0/container/np_no_such_container/details"
```

That one returns `404`.

### Primary pain point(s)

Scripts, probes, and anything that keys off HTTP status cannot tell “no such container” from “logs are empty.” The UI log modal uses Zenoh, so this is easy to miss in the browser.

### Additional context

- Suggested fix: resolve the container (same check as `/details`) *before* constructing `StreamingResponse`.
- Searched open issues; nothing tracks this HTTP-status case. Closed https://github.com/bluerobotics/BlueOS/issues/2267 is about live logs in the UI, not HTTP status on a missing container.

### Prerequisites

Safe-bet:
## Description
Fixes #4161

`GET /kraken/v2.0/container/{container_name}/log` previously constructed and returned a `StreamingResponse` without first checking if the container existed. As a result, requests for non-existent containers returned HTTP 200 OK headers followed by an empty stream or heartbeat fragments.

This PR awaits `ContainerManager.get_running_container_by_name(container_name)` before returning `StreamingResponse`, properly raising `ContainerNotFound` and returning HTTP 404 when the container does not exist.

### DIFF
--- a/core/services/kraken/api/v2/routers/container.py
+++ b/core/services/kraken/api/v2/routers/container.py
@@ -55,6 +55,7 @@ async def fetch_log_by_container_name(container_name: str, timeout: Optional[int
     Fetch logs of a given container.
     If timeout is provided, the stream will be closed after no log line is received for the given timeout.
     """
+    await ContainerManager.get_running_container_by_name(container_name)
     stream = ContainerManager.get_container_log_by_name(container_name)
 
     if timeout is not None:


patrickelectric:

joaoantoniocardoso (approved):
Tested, thanks!

PR #4192, opened 2026-08-20
core: frontend: ServoFunctionRangeEditor: Fix slider click target

patrickelectric:
Clicking the empty area around the servo range was changing trim instead of requiring a click on the track.


joaoantoniocardoso:
It shows up, but the handles are not draggable anymore, meaning that you can only change the value by precisely clicking on the ball:

Image

patrickelectric:
@joaoantoniocardoso fixed

joaoantoniocardoso (approved):
All good now, ty

PR #4206, opened 2026-08-20
core: frontend: vehiclesetup: FailsafeCard: Fix Low Battery DISABLED

### LINKED ISSUE #4205: bug: Low Battery failsafe card never shows DISABLED
### Bug description

The Low Battery failsafe card never gets the DISABLED badge. Excess Internal Temperature does, when Action is Disabled.

`FailsafeCard.findControlParam()` only matches Enable, `*_ENABLE`, or `replacementTitle === 'Action'`. Low Battery uses "Action when battery is low" / "Action when battery is critical" (`BATT_FS_LOW_ACT`, `BATT_FS_CRT_ACT`), so `is_disabled` stays false. Temperature works because `FS_TEMP_ENABLE` matches `_ENABLE`.

Using the action param as the control is also wrong. From https://github.com/ArduPilot/ParameterRepository :

- Sub-4.5: `BATT_FS_LOW_ACT` / `BATT_FS_CRT_ACT` is `0: None`, `2: Disarm`, `3: Enter surface mode`
- Sub-4.8 and Copter-4.8: 0 is Warn only
- There is no Disabled in those enums

The actual disable is the thresholds. `BATT_LOW_VOLT` / `BATT_CRT_VOLT`: "Set to 0 to disable." `BATT_LOW_MAH` / `BATT_CRT_MAH`: "Set to 0 to disable battery remaining failsafe." `BATT_LOW_TIMER=0` also "disables low voltage errors." `BATT_MONITOR=0` is the only Disabled, and that one already maps to UNAVAILABLE.

When a card is DISABLED, every editor except the control/action one is locked. Temperature: you can still flip Action, not the max temp. UNAVAILABLE does not lock editors.

The badge should follow the thresholds for both layers, including MAH even if we still do not show them (there is a TODO for that). Do not rename the action titles to Action.

### Steps to reproduce

1. Open Vehicle Setup -> Failsafes
2. Set Excess Internal Temperature Action to Disabled (badge shows)
3. Set `BATT_LOW_VOLT` and `BATT_CRT_VOLT` to 0 (and MAH if present)

### Primary pain point(s)

Looks like Low Battery is still armed when it is not.

### Prerequisites

patrickelectric:
findControlParam only matches Enable, *_ENABLE, or title Action, so Low Battery never shows DISABLED.

Fix #4205

### DIFF
--- a/core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
@@ -41,7 +41,7 @@
                 v-if="params[param.name] != null"
                 :key="failsafeDefinition.name"
                 :auto-set="true"
-                :disabled="is_disabled && !control_param_names.includes(param.name)"
+                :disabled="is_disabled && !dependency_unmet && !control_param_names.includes(param.name)"
                 :param="params[param.name] ?? undefined"
               />
               <template v-else>
--- a/core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
+++ b/core/frontend/src/components/vehiclesetup/configuration/failsafes/FailsafeCard.vue
@@ -41,7 +41,7 @@
                 v-if="params[param.name] != null"
                 :key="failsafeDefinition.name"
                 :auto-set="true"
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
Nice!

Image

PR #4209, opened 2026-08-21
core: commonwealth: Preserve response models in OpenAPI

Michael-F-Bryan:
`PrettyJSONResponse` inherits directly from Starlette's base `Response`. FastAPI therefore treats it as an arbitrary response when generating OpenAPI and describes successful responses as strings, even when a path operation explicitly declares a structured `response_model`. The runtime response remains valid JSON, leaving the published contract out of sync with the service.

This changes `PrettyJSONResponse` to inherit from `JSONResponse` while retaining its existing pretty-printing behaviour. FastAPI can now emit each path operation's declared response model in the generated OpenAPI schema. A regression test covers that contract.

### DIFF
--- a/core/libs/commonwealth/src/commonwealth/utils/apis.py
+++ b/core/libs/commonwealth/src/commonwealth/utils/apis.py
@@ -3,12 +3,12 @@ from typing import Any, Callable, Coroutine, Dict, Optional
 
 from commonwealth.utils.logs import stack_trace_message
 from fastapi import HTTPException, Request, Response, status
+from fastapi.responses import JSONResponse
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan:
I just checked CI and this is the error:

```
---------- coverage: platform linux, python 3.11.16-final-0 ----------
Coverage HTML written to dir htmlcov

FAIL Required test coverage of 73% not reached. Total coverage: 72.66%
```

I don't think the CI failure is related to this PR. It's a one-line change with about 20 lines of test code.

patrickelectric (approved):

PR #4216, opened 2026-08-21
core: libs: commonwealth: Cover remaining apis helpers

patrickelectric:
The OpenAPI test imported apis.py and dropped master under the coverage gate.

### DIFF
--- a/core/pyproject.toml
+++ b/core/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
 [dependency-groups]
 dev = [
   "coverage==7.5.1",
+  "httpx==0.28.1",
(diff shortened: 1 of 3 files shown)


core/pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
 [dependency-groups]
 dev = [
   "coverage==7.5.1",
+  "httpx==0.28.1",
```
joaoantoniocardoso, line 31: For reference: `The starlette.testclient module requires the httpx package to be installed.`

joaoantoniocardoso (approved):

PR #4218, opened 2026-08-21
core: services: wifi: Handle unknown ssid on remove

### LINKED ISSUE #4159: bug: Forgetting an unknown Wi-Fi SSID returns HTTP 200 instead of 400
### Bug description

`POST /wifi-manager/v1.0/remove?ssid=...` returns **HTTP 200** when the SSID is not in the saved-network list. The handler still has a `StopIteration` → 400 path (`Network '{ssid}' not saved.`), but that exception is never raised.

`WifiManager.remove_network(ssid)` filters saved networks with a list comprehension and loops over matches. An unknown SSID yields an empty list, the loop does nothing, and the route logs “Successfully removed”. The original `next(filter(...))` lookup that *did* raise `StopIteration` was replaced in https://github.com/bluerobotics/BlueOS/pull/1899 and later moved into `WifiManager` by https://github.com/bluerobotics/BlueOS/pull/3018. The `except StopIteration` in `core/services/wifi/main.py` has been dead since then.

Operators and API clients cannot tell “network forgotten” from “SSID never existed / typo / already removed.” The UI can show success while saved networks are unchanged.

### Steps to reproduce

1. On BlueOS `master` or `1.4-dev`, with any saved-network list (including empty):

```bash
curl -sS -D - -o /dev/null -X POST \
  "http://<blueos>/wifi-manager/v1.0/remove?ssid=__no_such_ssid__"
```

2. Observe: `HTTP/1.1 200 OK`.
3. `GET /wifi-manager/v1.0/saved` is unchanged.

Expected: HTTP 400 with `Network '__no_such_ssid__' not saved.` (the still-present handler).

### Primary pain point(s)

Forgetting a mistyped or already-removed network looks like success. There is no way for a client to distinguish a real remove from a no-op.

### Additional context

- Suggested fix: if `match_networks` is empty after lookup, raise 400 (or 404) from `remove_network(ssid)` (or from the route) so the existing handler is reachable again.
- Searched open issues; nothing tracks this HTTP-status case. Adjacent but different: https://github.com/bluerobotics/BlueOS/issues/1901 (BSSID rework), closed https://github.com/bluerobotics/BlueOS/issues/3582 (forget failed to persist on Bookworm).

### Prerequisites

patrickelectric:
The remove route is async, so StopIteration never reached the 400 handler.

Fix #4159

### DIFF
--- a/core/services/wifi/main.py
+++ b/core/services/wifi/main.py
@@ -91,11 +91,11 @@ async def connect(credentials: WifiCredentials, hidden: bool = False) -> Any:
 async def remove(ssid: str) -> Any:
     assert wifi_manager is not None
     logger.info(f"Processing remove request for SSID: {ssid}")
-    try:
-        await wifi_manager.remove_network(ssid)
-    except StopIteration as error:
+    saved_networks = await wifi_manager.get_saved_wifi_network()
+    if not any(network.ssid == ssid for network in saved_networks):
         logger.info(f"Network '{ssid}' is unknown.")
-        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=f"Network '{ssid}' not saved.") from error
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=f"Network '{ssid}' not saved.")
+    await wifi_manager.remove_network(ssid)
     logger.info(f"Successfully removed '{ssid}'.")
 
 


patrickelectric:
1.1: that's the first commit. 400 is raised in the route now

nicoschmdt (approved):

PR #4221, opened 2026-08-21
core: services: ardupilot_manager: Skip extra MAVLinkRouter TCP server

### LINKED ISSUE #3431: mavlink-routerd only supports one TCP server, but we don't prevent it
Mavlink-routerd only supports one TCP server, but we don't prevent it.

Yeah, this is an edge case, but people trying to customize it might face this.

Affects: `BlueOS-1.4.2`

---

### How to reproduce:

I've added a new TCP Server endpoint as `0.0.0.0:5778`, using MAVLinkRouter:

Image

But I couldn't connect to it, so I checked the terminal... The command line seems correct, but the mavlink-routerd seems to only spawn a TCP server to the last one:

Image

---

@ES-Alexander do you know if we have this limitation documented?

patrickelectric:
mavlink-routerd only listens on the last --tcp-port.

Fix #3431

### DIFF
--- a/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
+++ b/core/services/ardupilot_manager/mavlink_proxy/MAVLinkRouter.py
@@ -2,6 +2,7 @@ import re
 import subprocess
 from typing import Optional
 
+from loguru import logger
 from mavlink_proxy.AbstractRouter import AbstractRouter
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
Note that this does not fix the issue because MAVLink Router is only used in `1.4-dev`.

PR #4222, opened 2026-08-21
bag_of_holding: move persistence to async

nicoschmdt:
as discussed in https://github.com/bluerobotics/BlueOS/pull/4180#issuecomment-5345388064

### DIFF
--- a/core/services/bag_of_holding/main.py
+++ b/core/services/bag_of_holding/main.py
@@ -4,7 +4,6 @@ import json
 import logging
 import os
 from pathlib import Path
-from types import SimpleNamespace
 from typing import Any
 
 import appdirs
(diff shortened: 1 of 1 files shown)


core/services/bag_of_holding/main.py
```diff
@@ -43,40 +42,42 @@ def read_db() -> dict[str, Any]:
...
+            # Cleared with the snapshot: a mutation is either in this write or re-arms the next one
+            db_dirty.clear()
+            data = json.dumps(bag_of_holding_db)
+        # write_db reuses one temp file, so overlapping writers interleave into a corrupt database
+        async with disk_lock:
+            await asyncio.to_thread(write_db, data)
     except Exception:
         logger.exception("Failed to persist database")
-        _pending.handle = asyncio.get_running_loop().call_later(FLUSH_INTERVAL, _flush)
+        db_dirty.set()
```
patrickelectric, line 72: just to be sure.. how are we supposed to know that the write operation failed ?

patrickelectric: The pint here is that: It failed to write in disk, we still return success.

patrickelectric: We should put the write_db logic inside flush, to raise any issues when writing to disk.

PR #4232, opened 2026-08-21
[1.4] Fix Navigator not being detected on the first boot of a Bookworm install

### LINKED ISSUE #4230: Pi4, Bookworm doesn't detect Navigator on first boot
Reproduced on `1.4.4-beta.20`:

[blueos-system-logs-20260821-212719.zip](https://github.com/user-attachments/files/31321947/blueos-system-logs-20260821-212719.zip)

The overlay is definitely being wrongly handled:

```bash
pi@blueos:~ $ cat /boot/firmware/config.txt
# For more options and information see                                                                                             
# http://rptl.io/configtxt                                                                                                         
# Some settings may impact device functionality. See link above for details                                                        
                                                                                                                                   
# Uncomment some or all of these to enable the optional hardware interfaces                                                        
#dtparam=i2c_arm=on                                                                                                                
#dtparam=i2s=on                                                                                                                    
#dtparam=spi=on                                                                                                                    
                                                                                                                                   
# Enable audio (loads snd_bcm2835)                                                                                                 
dtparam=audio=on                                                                                                                   
                                                                                                                                   
# Additional overlays and parameters are documented                                                                                
# /boot/firmware/overlays/README                                                                                                   
                                                                                                                                   
# Automatically load overlays for detected cameras                                                                                 
camera_auto_detect=1                                                                                                               
                                                                                                                                   
# Automatically load overlays for detected DSI displays                                                                            
display_auto_detect=1                                                                                                              
                                                                                                                                   
# Automatically load initramfs files, if found                                                                                     
auto_initramfs=1                                                                                                                   
                                                                                                                                   
# Enable DRM VC4 V3D driver                                                                                                        
dtoverlay=vc4-kms-v3d                                                                                                              
max_framebuffers=2                                                                                                                 
                                                                                                                                   
# Don't have the firmware create an initial video= setting in cmdline.txt.                                                         
# Use the kernel's default instead.                                                                                                
disable_fw_kms_setup=1                                                                                                             
                                                                                                                                   
# Disable compensation for displays with overscan                                                                                  
disable_overscan=1                                                                                                                 
                                                                                                                                   
# Run as fast as firmware / board allows                                                                                           
arm_boost=1                                                                                                                        
                                                                                                                                   
[cm4]                                                                                                                              
# Enable host mode on the 2711 built-in XHCI USB controller.                                                                       
# This line should be removed if the legacy DWC2 controller is required                                                            
# (e.g. for USB device mode) or if USB support is not required.                                                                    
otg_mode=1                                                                                                                         
                                                                                                                                   
[cm5]

[all]

[pi4]
dtoverlay=dwc2,dr_mode=otg

[pi4]

[pi4]

[pi4]

[pi4]

pi@blueos:~ $ cat /boot/config.txt 
DO NOT EDIT THIS FILE

The file you are looking for has moved to /boot/firmware/config.txt
[pi4]
dtoverlay=dwc2,dr_mode=otg
gpio=37=op,pd,dl
gpio=11,24,25=op,pu,dh
dtoverlay=spi1-3cs
dtoverlay=spi0-led
dtparam=spi=on
dtoverlay=i2c6,pins_22_23,baudrate=400000
dtoverlay=i2c4,pins_6_7,baudrate=1000000
dtparam=i2c_arm_baudrate=1000000
dtoverlay=i2c1
dtparam=i2c_vc=on
dtoverlay=uart5
dtoverlay=uart4
dtoverlay=uart3
dtoverlay=uart1
enable_uart=1
```

joaoantoniocardoso:
Fixes the Pi4 Bookworm first-boot Navigator detection failure reported in https://github.com/bluerobotics/BlueOS/issues/4230, the underlying reason the startup patches could not recover from it on any board, and the damage it left on devices that already ran the affected release.

## Two independent bugs

**The install scripts wrote to the wrong file.** Bookworm moved the boot partition to `/boot/firmware` and left plain text stubs at `/boot/config.txt` and `/boot/cmdline.txt`. `install/boards/bcm_27xx.sh` hardcoded the old paths, so a Pi4 installed over Bookworm got its whole Navigator overlay set appended to a file the firmware never reads. Only those two are real files: `/boot/overlays` is a symlink into the boot partition, so the compiled overlay did land there, and `1.4.4-beta.20` ships `overlays/spi0-led.dtbo` next to a stock `config.txt` with nothing to load it. `bcm_2712.sh` (Pi5) already used `/boot/firmware` for its writes, which is why only Pi4 was reported, but it too reached the overlays directory through that symlink, which no package owns. `bcm_28xx.sh` had the same problem as the Pi4 script. All three now resolve the boot partition the same way.

**The startup patch that should have repaired that was a no-op on its first run.** `boot_config_get_or_append_section()` appended a missing board section as a single `"\n[pi4]"` list entry. The section lookup uses `re.match`, which anchors at offset 0 and therefore never matches a string starting with a newline, so every call appended another header. The start index was also set one past the end of the list, which made the `section_start < i < section_end` guard in the conflict filter unsatisfiable, so each configuration line was deleted immediately after being inserted.

A first pass over a stock `config.txt` produced 30 empty `[pi4]` sections and zero overlays. The overlays only landed on a later boot, once the saved file had been split back into lines, and since `BOOT_LOOP_DETECTOR` skips the boot right after a reboot, that took three boots. This part was never Pi4-specific; Pi5 was only saved by its install script writing the right file.

## Repairing devices that already ran the affected release

Every device that booted `1.4.4-beta.20` carries roughly 30 stray `[pi4]` headers. Two consequences, both fixed here:

- Only the first section is ever configured, so the strays stay forever, and any configuration the user put under one of them silently stops applying. `boot_config_merge_duplicated_sections()` now folds them into a single section on the next boot, keeping every user line in the order the firmware applies it, and leaves a `config.txt` with one section byte-identical so it does not ask for a needless reboot. Order matters because a duplicate header reopens the same filter rather than starting a section only the first of which is read: hoisting the strays above the survivor's own lines would have flipped the effective value of any key set under both headers, on exactly the devices this repairs.
- Re-running `install.sh` on such a device *removed* its Navigator configuration and reported success. `line_number=$(grep -n "\[pi4\]" ... )` returns one line per match, so `$line_number` became multi-line, every insertion `sed` failed with `unknown command`, and the deletion loop had already stripped the overlays. It now takes the first match only.

## Making the failures loud

Both halves of the original bug shared a failure mode: writing to the wrong place still exits 0.

- The boot-partition probe refuses to guess. `/boot/firmware/config.txt` wins; `/boot/config.txt` is used only when it is a real config and not the Bookworm stub (identified by `DO NOT EDIT THIS FILE` being its whole first line); otherwise the script prints why and exits 1. A Pi5 is newer than Bullseye, so `bcm_2712.sh` accepts nothing but `/boot/firmware`: boot files at `/boot` on that board are never its boot partition. Previously an unmounted boot partition silently configured a stub, and for the Pi5 that was a regression in failure mode, since hardcoding `/boot/firmware` at least made `sed` fail.
- `bcm_27xx.sh` and `bcm_28xx.sh` now `set -e`, which `bcm_2712.sh` already had. Every command in both was audited and both were run end to end against fake roots to confirm nothing that legitimately fails now aborts an install.
- The runtime repair path refuses the same stub. `blueos_startup_update.py` located `config.txt`/`cmdline.txt` with a first-match `find`, so on a Bookworm host whose boot partition was not mounted it settled on the `/boot` stub, wrote the whole board section into a file the firmware never reads, rebooted, and from then on found its own work in place and converged into a permanently unconfigured state with no error. It now checks the same `DO NOT EDIT THIS FILE` marker the install scripts key on and, on the stub, logs and skips the config patches through the existing `config_file is None` guards rather than writing nowhere. The marker has to be the file's whole first line, so a real `config.txt` that carries the phrase in a comment keeps being patched: refusing one would leave a correctly mounted board with no overlays, no cgroups and no OTG ethernet while still reporting success. It reads that line with `run_command` rather than `load_file`, because this runs before the `run_command_is_working()` check and `load_file` raises when the host connection is not up yet, which would take every patch of the boot down with it, including the ones this PR is not about.

## Other bugs found on the way

- `update_dwc2()` picked its section as `"pi4" if get_cpu_type() == CpuType.PI4 else "pi5"`, so every board that is not a Pi4 got a `[pi5]` section, and a board filter the firmware does not recognise is applied rather than ignored. Nothing reached a Pi3 in the field, since `main()` only runs this patch for a Pi4 and a Pi5, but a board check that once went wrong is exactly why its sibling `revert_update_dwc2()` exists. It now handles Pi4 and Pi5 explicitly and refuses anything else.
- `boot_cmdline_add_modules()` used a falsy index check, so when `modules-load=` was the first word of `cmdline.txt` the merged line was never written and a second, dead `modules-load=` was appended instead.
- Configuration values were interpolated into regexes unescaped. The Pi5 list holds both `dtoverlay=i2c3-pi5,baudrate=400000` and `dtoverlay=i2c3-pi5.baudrate=400000`, where the `.` matches the comma variant, so keeping both spellings depended on undocumented list order.
- `boot_config_remove_section()` on an absent section appended the header, deleted it again, and reported a change, requesting a reboot for a section that never existed.

## Guardrails

`core/tools/blueos_startup_update/test_blueos_startup_update.py` (54 tests) runs the patches against the authentic stock `config.txt` and `cmdline.txt` of both the Bullseye and the Bookworm base images, for every board we configure, and pins:

- the resulting board section matches exactly what the image install script writes, parsed out of the script rather than duplicated, so the runtime patch and the image build cannot drift apart
- one pass is enough, and a second pass changes nothing, so the reboot the patches request settles
- each patch individually reports whether it needs a restart, so a patch that rewrites a boot file and stays silent fails the suite — that is precisely the reported symptom
- conflicting and misplaced configuration is replaced while `# custom` protected lines survive
- a Pi3 never receives Navigator or dwc2 configuration, and the Pi3 cleanup leaves only sections its firmware recognises
- a device carrying stray board sections converges to one, keeps every user line, and does not loop
- the boot-partition probe of every board script, executed under real `bash` against four fake-root layouts, resolves correctly and exits non-zero when it cannot tell, down to the Pi5 script refusing a `/boot` layout
- the runtime patcher refuses the Bookworm stub the install scripts refuse, for both `config.txt` and `cmdline.txt`, and still accepts a real Bullseye `/boot` and a mounted `config.txt` whose comments happen to carry the marker
- `main()` itself resolves both boot files through that guard, so reverting the one line that wires it in fails the suite instead of quietly restoring the original bug
- merging strays keeps the user's configuration in the order the firmware reads it, and keeps it inside the section the other patches look in, for the two-header case a single boot on the bad release produces
- no install script reaches a boot file without `BOOT_PATH`, anywhere, outside the probe that resolves it

Every guard was checked by reverting the fix it protects and confirming the suite goes red, including the shell ones.

The image build now asserts that the boot partition of the image it just produced really carries the board configuration: in `config.txt` the board section and the Navigator overlays, and in `cmdline.txt` the memory cgroup docker needs and the removal of the serial console that would hold the autopilot port. `cmdline.txt` is reached through a path of its own and was lost the same way — the shipped `1.4.4-beta.20` Bookworm image still carries stock `console=serial0,115200` and no cgroup parameters. Run against that image, this check fails on every line it looks for, which is the bug sitting in a released artifact.

That check only ever runs upstream, so `test_image_build_checks_what_the_install_scripts_write` ties each line it looks for back to the install script that writes it. A check that stopped matching the scripts fails the test suite instead of going unnoticed until a release.

## Out of scope

Three problems found while working on this, none of them fixed here:

- https://github.com/bluerobotics/BlueOS/issues/4234 — CM5 and CM3 get no board configuration at all and the install still exits 0, while a CM4 is configured at image build time but never receives any runtime patch, because `get_cpu_type()` does not recognise it.
- https://github.com/bluerobotics/BlueOS/issues/4235 — the shared file helpers cannot report failure, so a read-only boot partition makes the patches request a reboot for work they did not do, on every other boot, forever.
- `deploy/pimod/blueos.Pifile` runs `touch /boot/ssh`, which on Bookworm lands on the rootfs stub directory rather than the boot partition.

## Test plan

Four combinations to cover: Pi4 Bullseye, Pi4 Bookworm, Pi5 Bookworm and Pi3 Bullseye. Three images cover them, and all three were built from this branch: `arm-v7 bullseye` for the Pi3 and the Pi4, `arm-v7 bookworm` for the Pi4, and `arm64-v8 bookworm` for the Pi5. They came from `266584936`, a temporary commit that opened the image build up to pull requests while writing nothing to any registry, and which differs from the current head only in that build plumbing. It has since been dropped, so the branch is the fix alone.

Each artifact was read before burning: its FAT boot partition was opened with `mtools` and checked against expectations parsed out of the install scripts. The same check reports 5 problems against `1.4.4-beta.20`.

The Pi5 probe was narrowed to `/boot/firmware` after those images were built, so the probe of all three board scripts was run again, read only, on four live boards: two Pi5 Bookworm, a Pi4 Bookworm and a Pi4 Bullseye. Each resolved the boot partition its own board actually uses, `/boot/config.txt` is a stub on all three Bookworm boards, and `bcm_2712.sh` refused the Pi4 Bullseye layout.

Every image carries `blueos-core:1.4-dev`, so the boot files come from the fixed install scripts while the patches running on the first boot are the old ones. That is what makes a first boot a test of the install scripts on their own, and it held: on both Bookworm boards the first boot asked to restart for `dns`, `noIPV6`, `swap`, `wpa` and `networkmanager`, and for no boot configuration patch, because there was nothing left to fix. The fixed patches were then deployed to each board and run twice.

### Pi4 Bullseye — Pi 4B Rev 1.2, `arm-v7 bullseye pi4`

- [x] `BOOT_PATH` resolved to `/boot` and the configuration went into the `[pi4]` section the Bullseye image already ships, leaving `[cm4]`, `[all]`, `[pi4]`, `[all]` — no second board section, and the two stock `[all]` sections left alone
- [x] First boot: the 10 overlay lines are active, `cgroup_enable=memory` is set, `console=serial` is gone, and `i2c-4`, `i2c-6`, `spidev0.0`, `spidev1.0-1.2` and `ttyAMA0-3` are all present
- [x] `Detected Linux board: NavigatorPi4`, `Using Navigator flight-controller`
- [x] The fixed patches converge on the image as burned: no change to `config.txt` or `cmdline.txt` and no restart requested, on either of two runs
- [x] Repair on hardware: 31 stray `[pi4]` headers with user configuration under one of the strays collapsed to a single section with that line inside it, and a second run changed nothing and asked for no reboot

### Pi4 Bookworm — Pi 4B Rev 1.5, `arm-v7 bookworm`

- [x] `BOOT_PATH` resolved to `/boot/firmware`, `spi0-led.dtbo` was compiled into `/boot/firmware/overlays`, and `[pi4]` carries the whole set
- [x] First boot: `i2c-4`, `i2c-6`, `spidev0.0`, `spidev1.0-1.2` and `ttyAMA3/4/5` present, `cgroup_enable=memory` set, `console=serial` gone
- [x] `Detected Linux board: NavigatorPi4`, `Using Navigator flight-controller`
- [x] The fixed patches converge: no change to `config.txt` or `cmdline.txt` and no restart requested, on either of two runs
- [x] With `/boot/firmware` unmounted, `bcm_27xx.sh` printed why and exited 1, leaving the `/boot` stub untouched, and that status carries: `configure_board.sh` ends on the board script's pipeline and `install.sh` runs under `set -e`, so the install stops there rather than reporting success

### Pi5 Bookworm — Pi 5B Rev 1.0, `arm64-v8 bookworm pi5`

- [x] `BOOT_PATH` resolved to `/boot/firmware`, the overlay was compiled into `/boot/firmware/overlays`, and `[pi5]` carries the Pi5 set with both `i2c3-pi5` spellings
- [x] First boot: `i2c-3`, `i2c-6`, `spidev0.0`, `spidev1.0-1.2` and `ttyAMA0/2/3/4/10` present, `cgroup_enable=memory` set, `console=serial` gone
- [x] `Detected Linux board: NavigatorPi5`
- [x] The fixed patches converge: no change and no restart requested, on either of two runs

### Pi3 Bullseye — Pi 3B, `arm-v7 bullseye pi4`

This board booted the image three times and came all the way up, but it never got a DHCP lease on `eth0` and lost its hotspot a minute in to https://github.com/bluerobotics/BlueOS/issues/4233, so it stayed unreachable and was read back from its SD card rather than over ssh.

- [x] The cleanup left `[all]`, `[all]`: the `[cm4]` and `[pi4]` sections the Pi4 image ships are gone, and no `i2c`, `spi`, `uart` or `dwc2` overlay line survives anywhere in `config.txt`
- [x] `modules-load=dwc2,g_ether` was taken back out of `cmdline.txt`
- [x] It converged in one pass: the first boot applied `revert_update_dwc2` and `clean_config_pi3` and rebooted once, and the boot after that applied nothing and asked for no restart
- [x] Replaying the fixed patches over that card's own boot files, as a Pi3 on Bullseye, converges on the first pass and produces a `config.txt` byte-identical to the one the shipped patches left on the board

`bcm_28xx.sh` cannot be reached through an image: `configure_board.sh` picks the board script from the running model, and the Bullseye image is built on a Pi4, so a Pi 0-3 only runs that script when `install.sh` is executed on it directly. The suite covers it by running it under real `bash` against fake roots.


install/boards/bcm_2712.sh
patrickelectric: isn't this supposed to use BOOT_PATH ?

patrickelectric: in bookworm the overlays are in `/boot/firmware/overlays`

patrickelectric: We should also update the test for it

patrickelectric (changes requested):

joaoantoniocardoso:
Using the Raspbian image built from this branch (from the HACK commit, now ditched, but artifacts still available [here](https://github.com/bluerobotics/BlueOS/actions/runs/32556405739?pr=4232)), I burned the SD cards for each device and booted them up.

Aside from what the LLM verified by accessing them via SSH, I manually and independently tested and verified myself:
  - [x] Pi3 bullseye (BlueOS-raspberry-linux-arm-v7-bullseye-pi4.zip) w/ no board -> cmdline.txt and config.txt all good.
  - [x] Pi4 bullseye (BlueOS-raspberry-linux-arm-v7-bullseye-pi4.zip) w/ Navigator  -> Navigator recognized, cmdline.txt and config.txt all good.
  - [x] Pi4 bookworm, (BlueOS-raspberry-linux-arm-v7-bookworm.zip) w/ Navigator -> Navigator recognized, cmdline.txt and config.txt all good.
  - [x] Pi5 bookworm, (BlueOS-raspberry-linux-arm64-v8-bookworm-pi5.zip) w/ Navigator -> Navigator recognized, cmdline.txt and config.txt all good.

Functionally, I believe this is ready. @patrickelectric can you check whether we are missing anything?

joaoantoniocardoso:
ping @patrickelectric

install/boards/bcm_2712.sh
```diff
@@ -8,8 +8,18 @@ VERSION="${VERSION:-master}"
...
-CMDLINE_FILE=/boot/firmware/cmdline.txt
-CONFIG_FILE=/boot/firmware/config.txt
+# Bookworm moved the boot partition to /boot/firmware and left plain text stubs behind at /boot,
+# so writing to /boot there configures nothing.
+if [ -f /boot/firmware/config.txt ]; then
+    BOOT_PATH=/boot/firmware
+elif [ -f /boot/config.txt ] && ! grep -q "DO NOT EDIT THIS FILE" /boot/config.txt; then
+    BOOT_PATH=/boot
+else
+    echo "No boot partition found at /boot/firmware or /boot, is it mounted?" >&2
```
patrickelectric: Is this correct ? should we error if DO NOT EDIT THIS FILE exist in /config ?

install/boards/bcm_27xx.sh
```diff
@@ -6,8 +6,18 @@ VERSION="${VERSION:-master}"
...
+# Bookworm moved the boot partition to /boot/firmware and left plain text stubs behind at /boot,
+# so writing to /boot there configures nothing.
+if [ -f /boot/firmware/config.txt ]; then
+    BOOT_PATH=/boot/firmware
+elif [ -f /boot/config.txt ] && ! grep -q "DO NOT EDIT THIS FILE" /boot/config.txt; then
+    BOOT_PATH=/boot
+else
+    echo "No boot partition found at /boot/firmware or /boot, is it mounted?" >&2
+    exit 1
+fi
```
patrickelectric, line 21: the bcm27 of the pi5 does not support bullseye.

install/boards/bcm_2712.sh
```diff
@@ -28,7 +28,7 @@ echo "- compile spi0 device tree overlay."
 DTS_PATH="$ROOT/install/overlays"
 DTS_NAME="spi0-led"
 curl -fsSL -o /tmp/$DTS_NAME $DTS_PATH/$DTS_NAME.dts
-dtc -@ -Hepapr -I dts -O dtb -o /boot/overlays/$DTS_NAME.dtbo /tmp/$DTS_NAME
+dtc -@ -Hepapr -I dts -O dtb -o "$BOOT_PATH/overlays/$DTS_NAME.dtbo" /tmp/$DTS_NAME
```
patrickelectric, line 30: shoudnt this be in c7e3d6634334eec852d9889ce6623476d13d4abd ?

install/boards/bcm_27xx.sh
```diff
@@ -26,7 +26,7 @@ echo "- compile spi0 device tree overlay."
 DTS_PATH="$ROOT/install/overlays"
 DTS_NAME="spi0-led"
 curl -fsSL -o /tmp/$DTS_NAME $DTS_PATH/$DTS_NAME.dts
-dtc -@ -Hepapr -I dts -O dtb -o /boot/overlays/$DTS_NAME.dtbo /tmp/$DTS_NAME
+dtc -@ -Hepapr -I dts -O dtb -o "$BOOT_PATH/overlays/$DTS_NAME.dtbo" /tmp/$DTS_NAME
```
patrickelectric, line 32: shouldnt this be in c7e3d6634334eec852d9889ce6623476d13d4abd ?

core/tools/blueos_startup_update/test_blueos_startup_update.py
```diff
@@ -691,6 +693,36 @@ def test_install_scripts_reach_boot_files_only_through_boot_path(script_name: st
     assert not hardcoded, f"{script_name} reaches boot files without BOOT_PATH: {hardcoded}"
 
 
+def test_image_build_checks_what_the_install_scripts_write() -> None:
```
patrickelectric, line 723: this should be squashed in a67566604cfd94667d05c38a3ec5db61064bcdc9

patrickelectric (changes requested):

install/boards/bcm_27xx.sh
joaoantoniocardoso: uh, a danger dead code here. ty!

install/boards/bcm_2712.sh
joaoantoniocardoso: It is: we don't ever want to write to a no-longer-used config file.

Wording was adjusted.

.github/workflows/test-and-deploy.yml
```diff
@@ -312,8 +326,27 @@ jobs:
...
+            MISSING+=("cmdline.txt: console=serial is still there")
+          fi
           sudo umount /mnt/piboot
           sudo kpartx -d deploy/pimod/blueos.img
+          if [ ${#MISSING[@]} -ne 0 ]; then
+            echo "Missing from the boot partition of the ${{ matrix.os }} ${{ matrix.platform }} image built on ${{ matrix.runner }}:"
+            printf '  %s\n' "${MISSING[@]}"
+            echo "The board install script did not configure the image's boot partition."
+            exit 1
+          fi
```
patrickelectric, line 349: just to be sure.. have you tested this in the pimod ci machine ?

joaoantoniocardoso: Yes, I changed the CI file temporarily to build the images for this PR, then burned sdcards and tested on pi3 bullseye, pi4 bullseye, pi4 bookworm and pi5 bookworm. 

The CI run and its logs from that commit are here: https://github.com/bluerobotics/BlueOS/actions/runs/32556405739?pr=4232

After that I dropped the commit changing the CI.

patrickelectric (approved):
LGTM

Issue #4236, opened 2026-08-22
A board with no DHCP lease is reachable only at 192.168.2.2, the address every board falls back to


joaoantoniocardoso:
When `eth0` gets no DHCP lease, the only address left on a BlueOS device is the static fallback `192.168.2.2/24`, which every other BlueOS device also falls back to. On a bench with more than one vehicle, that fallback is what makes the device unreachable rather than what saves it:

- Two boards in this state on the same switch both answer for `192.168.2.2`, so neither can be reached reliably and there is no way to tell which one replied.
- Reaching one at all needs a dedicated topside interface configured into `192.168.2.0/24`, so with several boards on one router you end up unplugging boards, or running one cable per board, to talk to any of them.
- The hostname does not disambiguate either: every board is `blueos`, and with three of them on the router here `blueos.local` resolved to a different board depending on which resolver answered — `getent hosts` said one address and `avahi-resolve` said another, at the same moment.

Combined with #4233, where the hotspot is destroyed about a minute after boot on the `wpa_supplicant` handler, a board in this state has **no reachable address at all**: no DHCP address, no access point, and a fallback address shared with its neighbours.

That is how it showed up here. A Pi3 looked like a board that would not boot: 20 minutes of power cycling, never appearing on the network. It had in fact booted correctly three times and run BlueOS for about 12 minutes each time, with the version chooser online and every service up. The only way to find that out was to pull the SD card and read the logs on another machine.

## DHCP itself is not broken

Worth stating plainly, because the symptom invites the wrong conclusion. Moving the same Bullseye Pi4 from a segment with no DHCP server onto a router leases immediately:

```
eth0: carrier acquired
eth0: soliciting a DHCP lease
eth0: offered 192.168.0.88 from 192.168.0.1        # 1 s after carrier
eth0: leased 192.168.0.88 for 14400 seconds       # 5 s after carrier
eth0: adding default route via 192.168.0.1
```

On that boot `cable-guy` logs a single timeout and **zero** config mismatches, so the retry loop below is specific to having no lease. What this issue is about is the state a board is left in when nothing answers DHCP — which is the normal direct-tether topside setup as much as it is a wrong cable or a dead switch port.

## What the logs show when there is no lease

Pi3, from its SD card, across three boots:

- 107 × `dhclient timed out for interface eth0`
- 0 × `Got new IP ... from DHCP for eth0`
[...]

PR #4238, opened 2026-08-22
[1.4] Fix WiFi hotspot dying after a second enable_hotspot

### LINKED ISSUE #4233: WiFi hotspot dies ~1 minute after boot on Bullseye (Pi3 and Pi4), leaving the vehicle unreachable
## Problem

On a board using the `wpa_supplicant` wifi handler (Bullseye), the WiFi hotspot comes up normally at boot and then dies about a minute later and never comes back. Confirmed on a **Pi3B** and a **Pi4B**, so this is not specific to one radio.

When the vehicle has no ethernet link and no reachable saved WiFi network — the out-of-the-box situation — this leaves it with no access path at all: no hotspot, no `wlan0`, no `eth0`. The machine is up and perfectly healthy, it is just invisible. The only way out is a power cycle, and from the operator's side it is indistinguishable from a vehicle that failed to boot.

Nothing reports the failure. `hostapd` keeps running, `dnsmasq` is restarted on the dead interface and reports success, and `GET /v1.0/hotspot` keeps returning `true` while no access point is on the air.

## Impact

- A vehicle reachable only over the hotspot becomes unreachable about a minute after every boot.
- The symptom looks like a boot failure or a hang: no UI, no SSH, no mDNS. Users power cycle repeatedly trying to get it to "finally boot".
- When ethernet happens to work, the failure is silent — the hotspot is already gone before anyone needs it, and they find out in the field.

## Why it happens

`enable_hotspot()` warns that the hotspot is already running and then starts it anyway — there is no `return` after the warning:

```python
if self.hotspot.is_running():
    logger.warning("Hotspot already running. No need to enable it again.")
await self.hotspot.start()
```

`Hotspot.start()` then calls `_create_virtual_interface()`, which **deletes and recreates `uap0` unconditionally**, and afterwards skips restarting `hostapd` because `is_running()` only checks whether its subprocess is alive:

```python
self._create_virtual_interface()
if not self.is_running():
    self._subprocess = subprocess.Popen(self.command_list(), ...)
```

So `hostapd` is left bound to a netdev that no longer exists, and the freshly created `uap0` has nobody beaconing on it. `is_running()` still says `True`, which is why every status report stays green.

The trigger is automatic. `auto_reconnect` fires 60 s after `wlan0` is first seen disconnected and, with `smart_hotspot_enabled` defaulting to on, calls `enable_hotspot()`. A vehicle with no saved WiFi network is permanently "disconnected", so this happens on every boot.

## Pi4B / Bullseye, unattended, on a clean boot

```
19:51:51  INFO     Hotspot:start - Starting hotspot.                      uap0 UP,LOWER_UP
19:51:59  DEBUG    WifiManager:start_hotspot_watchdog - Starting hotspot watchdog.
19:52:59  DEBUG    WifiManager:auto_reconnect - Starting smart-hotspot.
19:52:59  WARNING  WifiManager:enable_hotspot - Hotspot already running. No need to enable it again.
19:52:59  INFO     Hotspot:start - Starting hotspot.                      uap0 DOWN,NO-CARRIER
```

Exactly 60 s, with nobody touching the device, and `hostapd` still alive afterwards. It also reproduces on demand, which makes it easy to confirm on any board:

| step | `uap0` | `GET /v1.0/hotspot` | hostapd |
| --- | --- | --- | --- |
| `POST /v1.0/hotspot?enable=false` | DOWN | `false` | gone |
| `POST /v1.0/hotspot?enable=true` | **UP, LOWER_UP** | `true` | running |
| `POST /v1.0/hotspot?enable=true` again | **DOWN, NO-CARRIER** | `true` | running |

Stop-then-start recovers it, until the next boot.

## Pi3B / Bullseye

Same sequence, plus the radio itself fails the teardown:

```
[215.548075] ieee80211 phy0: brcmf_cfg80211_stop_ap: SET SSID error (-52)
[215.548389] ieee80211 phy0: brcmf_cfg80211_stop_ap: BRCMF_C_DOWN error -52
[215.548682] ieee80211 phy0: brcmf_cfg80211_stop_ap: setting AP mode failed -52
[215.551528] ieee80211 phy0: brcmf_set_mpc: fail to set mpc
[216.099981] ieee80211 phy0: brcmf_cfg80211_get_channel: chanspec failed (-52)
             dhcpcd[573]: uap0: carrier lost
```

The `brcmfmac` errors are the only Pi3-specific part, and they are not what kills the access point. On the Bullseye Pi4 the radio stays clean and the access point is equally dead, and on Bookworm the same teardown failure appears and the access point survives it (below).

## Bookworm is not affected

Checked on a Pi4B and a Pi5B on Bookworm, both of which log `wifi_handlers.wpa_supplicant.WifiManager can work: False` and select the NetworkManager handler. After 1.5 hours of uptime `uap0` is `UP,LOWER_UP` with carrier 1 on both, and none of `Watchdog activated`, `Starting smart-hotspot`, `already running` or `virtual access point interface` appears anywhere in their logs.

Neither trigger does anything to them:

- Enabling the smart hotspot and waiting 80 s — longer than the 30 s NetworkManager loop and the 60 s Bullseye one — produces no hotspot activity at all, because that watchdog checks `if not await self.hotspot_is_running()` before calling `enable_hotspot()`.
- Calling `POST /v1.0/hotspot?enable=true` on a running hotspot restarts it cleanly: `uap0: AP-ENABLED`, `Successfully started create_ap`, carrier still 1. `NetworkManagerWifi.enable_hotspot()` terminates the existing `create_ap` before spawning a new one, so nothing is left bound to the old interface.

That second test also produces `ieee80211 phy0: brcmf_cfg80211_stop_ap: setting AP mode failed -52` on both boards — the same firmware complaint as the Pi3 — and the access point comes back anyway. So this is a property of the `wpa_supplicant` handler keeping a `hostapd` that nobody restarts, not of any board or radio.

## How it was found

Chasing a Pi3 that "only booted on the second try". `kern.log` recorded four boots: two never got ethernet carrier and lost the hotspot at 188 s and 215 s, and both ended with no shutdown messages at all — hard power cuts by the operator, one after 2 minutes and one after 15. The system was fully booted and running BlueOS the whole time.

It happened again later on a freshly imaged Pi3, which cost 20 minutes of power cycling before its SD card was read on another machine: three clean boots, BlueOS fully up each time, unreachable because the hotspot was gone and `eth0` had no DHCP lease (see #4236).

## Environment

- Raspberry Pi 3B Rev 1.2, Bullseye, kernel 5.10.33-v7+, brcmfmac BCM43430/1 firmware 7.45.98, BlueOS 1.4.4-beta.20 and 1.4.4-beta.21
- Raspberry Pi 4B Rev 1.2, Bullseye, kernel 5.10.33-v7l+, brcmfmac BCM4345/6 firmware 7.45.241, BlueOS 1.4.4-beta.21
- Not reproduced on Raspberry Pi 4B Rev 1.5 or Pi 5B Rev 1.0 on Bookworm, which select the NetworkManager handler
- The code path is present on both `master` and `1.4-dev`

## Steps to reproduce

Either wait:

1. Boot a Bullseye board with no reachable saved WiFi network.
2. Connect to the BlueOS hotspot — it works.
3. About a minute later the hotspot disappears and never returns. With ethernet also unplugged, the vehicle is unreachable until a power cycle.

Or force it:

1. With the hotspot running, `POST /v1.0/hotspot?enable=true`.
2. `ip -br link show uap0` goes to `DOWN,NO-CARRIER` while `hostapd` stays alive and the API still reports the hotspot as enabled.

joaoantoniocardoso:
~~**Note**: Because of #4241 (addressed by #4240), this PR will fail the CI in the coverage step until that gets merged and this one rebased.~~ -> That was merged, and this one got rebased.

---

Fixes https://github.com/bluerobotics/BlueOS/issues/4233 on the wpa_supplicant (Bullseye) path only. NetworkManager / Bookworm is unchanged.

`enable_hotspot()` warned that the hotspot was already running and then started it anyway. `Hotspot.start()` always deletes and recreates `uap0`, then skipped respawning hostapd because `is_running()` only checked whether the subprocess was still alive. hostapd stayed bound to a netdev that no longer existed, `GET /v1.0/hotspot` kept returning `true`, and `auto_reconnect` did this automatically ~60s after boot whenever `wlan0` was disconnected (the default for a vehicle with no saved WiFi).

This PR:

- Returns early from `enable_hotspot()` when the AP is actually up, so a second enable (including smart-hotspot) is a no-op
- Kills leftover hostapd before recreating `uap0` if `start()` does run, so a dead AP can recover
- Treats a down `uap0` as not running, so the API and the hotspot watchdog stop lying

DHCP / `192.168.2.2` (#4236) is intentionally out of scope.

## Test plan

Live on Pi 4B Rev 1.2 Bullseye. Unpatched baseline first, then the patched files in the running core.

- [x] Unpatched: with the hotspot already up, `POST /v1.0/hotspot?enable=true` left `uap0` `DOWN,NO-CARRIER`, hostapd still alive, `GET /hotspot` still `true`
- [x] Patched: the same second `enable=true` (twice) is a no-op (~29ms); `uap0` stayed `UP,LOWER_UP` carrier 1; log `Hotspot already running. No need to enable it again.`
- [x] `POST /v1.0/hotspot?enable=false` then `enable=true` still brings the AP back
- [x] Catalog `toggle_hotspot` mutating smoke from this machine as station: `passed=7 failed=0`, lease `192.168.42.133`, L3 ping `192.168.42.1` ok
- [x] Second `enable=true` while this machine was associated: still `UP,LOWER_UP`, 3/3 pings, 0% loss
- [x] Reboot with smart-hotspot on and no saved WiFi: `auto_reconnect` fired at 60s (`Watchdog activated` / `Starting smart-hotspot` / `already running`) and did **not** call `Hotspot.start()` again. `uap0` stayed `UP,LOWER_UP` carrier 1 through 3 min uptime; this machine held lease `192.168.42.133` and pinged `192.168.42.1` across that window.
- [x] `core/services/wifi/test_hotspot.py` (5 tests) covers the no-op, leftover-hostapd kill, and `is_running()` on a down `uap0`

### DIFF
--- a/core/services/wifi/test_hotspot.py
+++ b/core/services/wifi/test_hotspot.py
@@ -13,7 +13,8 @@ for _name in ("exceptions", "settings", "typedefs"):
 
 sys.modules["pyroute2"] = MagicMock()
 sys.modules["settings"] = MagicMock()
-for _pkg in (
+# Restore after import so other collected tests still see real commonwealth (pytest-xdist).
+_STUBBED = (
     "commonwealth",
     "commonwealth.utils",
     "commonwealth.utils.DHCPServerManager",
@@ -22,13 +23,21 @@ for _pkg in (
     "commonwealth.settings.manager",
     "commonwealth.settings.settings",
     "fastapi",
-):
-    sys.modules[_pkg] = MagicMock()
+)
+_previous = {name: sys.modules.get(name) for name in _STUBBED}
+for name in _STUBBED:
+    sys.modules[name] = MagicMock()
 
 import pytest
 
 from wifi_handlers.wpa_supplicant.Hotspot import HotspotManager
 from wifi_handlers.wpa_supplicant.WifiManager import WifiManager
+
+for name, previous in _previous.items():
+    if previous is None:
+        del sys.modules[name]
+    else:
+        sys.modules[name] = previous
 
 
 def _bare_hotspot() -> HotspotManager:
--- a/core/services/wifi/test_hotspot.py
+++ b/core/services/wifi/test_hotspot.py
@@ -1,0 +1,116 @@
+import sys
+from ipaddress import IPv4Address
+from pathlib import Path
+from typing import Any, List
+from unittest.mock import AsyncMock, MagicMock, patch
+
+_WIFI_DIR = str(Path(__file__).resolve().parent)
+sys.path.insert(0, _WIFI_DIR)
+for _name in ("exceptions", "settings", "typedefs"):
+    _mod = sys.modules.get(_name)
+    if _mod is not None and not str(getattr(_mod, "__file__", "")).startswith(_WIFI_DIR):
+        del sys.modules[_name]
+
+sys.modules["pyroute2"] = MagicMock()
+sys.modules["settings"] = MagicMock()
+for _pkg in (
+    "commonwealth",
+    "commonwealth.utils",
+    "commonwealth.utils.DHCPServerManager",
+    "commonwealth.utils.general",
+    "commonwealth.settings",
+    "commonwealth.settings.manager",
+    "commonwealth.settings.settings",
+    "fastapi",
+):
+    sys.modules[_pkg] = MagicMock()
+
+import pytest
+
+from wifi_handlers.wpa_supplicant.Hotspot import HotspotManager
+from wifi_handlers.wpa_supplicant.WifiManager import WifiManager
+
+
+def _bare_hotspot() -> HotspotManager:
+    hotspot = HotspotManager.__new__(HotspotManager)
+    hotspot.supports_hotspot = True
+    hotspot._ap_interface_name = "uap0"
+    hotspot._subprocess = None
+    hotspot._dhcp_server = None
+    hotspot._ipv4_gateway = IPv4Address("192.168.42.1")
+    return hotspot
+
+
+def _alive_process() -> MagicMock:
+    process = MagicMock()
+    process.poll.return_value = None
+    process.returncode = None
+    return process
+
+
+def test_is_running_is_false_when_uap0_is_down() -> None:
+    hotspot = _bare_hotspot()
+    hotspot._subprocess = _alive_process()
+    down = MagicMock(isup=False)
+    with patch("wifi_handlers.wpa_supplicant.Hotspot.psutil.net_if_stats", return_value={"uap0": down}):
+        assert hotspot.is_running() is False
+
+
+def test_is_running_is_true_when_hostapd_alive_and_uap0_up() -> None:
+    hotspot = _bare_hotspot()
+    hotspot._subprocess = _alive_process()
+    up = MagicMock(isup=True)
+    with patch("wifi_handlers.wpa_supplicant.Hotspot.psutil.net_if_stats", return_value={"uap0": up}):
+        assert hotspot.is_running() is True
+
+
+def test_stop_kills_hostapd_even_if_uap0_is_down() -> None:
+    hotspot = _bare_hotspot()
+    process = _alive_process()
+    hotspot._subprocess = process
+    hotspot._dhcp_server = MagicMock()
+    with patch("wifi_handlers.wpa_supplicant.Hotspot.psutil.net_if_stats", return_value={}):
+        hotspot.stop()
+    process.kill.assert_called_once()
+    hotspot._dhcp_server.stop.assert_called_once()
+
+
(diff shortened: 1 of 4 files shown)


joaoantoniocardoso:
ping @bluerobotics/blueos-team

patrickelectric (approved):

patrickelectric:
nice work

PR #4240, opened 2026-08-22
[1.4] Measure all production Python in the coverage gate

### LINKED ISSUE #4241: Python coverage setup is wrong
I've seen the following pattern multiple times, both in `1.4` and `master`:
* Someone submits a PR that includes commits adding tests.
* Coverage unexpectedly decreases.
* We lower the coverage gate number, and move on.

This is fundamentally wrong/dishonest.

joaoantoniocardoso:
Closes #4241

# Summary
- Coverage now measures every production module under `bootstrap`, `core/libs`, `core/services`, and `core/tools`, including services that tests never import.
- The old 75% gate only saw files a test imported, because coverage.py skips directories without `__init__.py` unless `include_namespace_packages` is on. That is why adding tests for wifi or `blueos_startup_update.py` dropped the number.
- `fail_under` is 15.5% (measured 15.57% of the full tree, tests omitted). Adding tests for previously unmeasured code raises the percentage instead of tanking it.
- CI prints the per-file table and uploads the HTML report even when python-tests fail.

## Test plan
- [x] `python-tests` on this PR reaches ~15.57% and passes the 15.5% gate
- [x] Coverage HTML artifact lists wifi, kraken, commander, helper, and `blueos_startup_update.py` (at 0% until they have tests)
- [x] A failed python-tests run still uploads the `coverage` artifact
- [x] Existing 90 pytest tests still pass

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -38,10 +38,12 @@ jobs:
         .hooks/pre-push
 
     - name: Upload coverage
+      if: always()
       uses: actions/upload-artifact@v4
       with:
         name: coverage
         path: htmlcov
+        if-no-files-found: ignore
 
 
   frontend-tests:
--- a/.hooks/pre-push
+++ b/.hooks/pre-push
@@ -74,7 +74,6 @@ for arg in "$@"; do
 done
 
 echo "Running pre push hook!"
-repository_path=$(git rev-parse --show-toplevel)
 tag_name=$(git tag --points-at=HEAD | head -n 1)
 
 [ -n "$tag_name" ] && {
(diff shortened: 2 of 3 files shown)


joaoantoniocardoso:
> ## Automated PR Review
>  ...
>  1.1 [major] pyproject.toml — include_namespace_packages = true is placed under
>  ...
>  8.1 [nit] .github/workflows/test-and-deploy.yml — if: always() combined with if-no-files-found:
>  ...
> _Generated by PR Review Bot. This is advisory, a human reviewer must still approve._

1.1 is incorrect for coverage 7.5.1 (the version we pin). `include_namespace_packages` is a `[report]` setting there (`report:include_namespace_packages` in config.py, and the 7.5.1 docs). This `pyproject.toml` loads it as True. CI on this PR reports wifi/kraken/commander/helper/`blueos_startup_update.py` at 0% and 15.56% total, so namespace-package discovery is working. Moving the key to `[run]` would be ignored by 7.5.1.

8.1: the upload-on-failure behavior is documented in the commit message; not adding an inline comment.

patrickelectric (approved):

PR #4242, opened 2026-08-22
[1.4] Make host file helpers report failure instead of silently patching

### LINKED ISSUE #4235: Boot configuration patches can silently do nothing, or wipe config.txt, and still ask for a reboot
## Problem

`core/tools/blueos_startup_update/blueos_startup_update.py` patches the host's `/boot/config.txt` and
`/boot/cmdline.txt` on every boot, over ssh, through helpers that report success no matter what happened. Every
failure mode below ends with the script asking for a reboot for work it did not do.

### A failed read is indistinguishable from an empty file

`load_file()` in `core/libs/commonwealth/commonwealth/utils/commands.py` shells out with `check=False` and
returns stdout:

```python
def load_file(file_name: str) -> str:
    command = f'cat "{file_name}"'
    return run_command(command, False).stdout
```

If the `cat` fails — boot partition not mounted, ssh to the host down, wrong path — the caller gets `""`. The
patches then treat that as a `config.txt` with no lines, add their own configuration to it, and save. The file
that reaches the boot partition contains *only* the Navigator section: the distribution's configuration and
everything the user added are gone. There is a `.bak`, but it is on a device that may no longer boot.

### `locate_file` cannot report "not found"

It is annotated `Optional[str]` but never returns `None`:

```python
def locate_file(candidates: List[str]) -> Optional[str]:
    # first match will return
    command = f"find {' '.join(candidates)} -type f -print -quit"
    return run_command(command, False, log_output=False).stdout.strip()
```

`find` printing nothing yields `""`. So the `if config_file is None:` / `if cmdline_file is None:` guards in
the startup script never fire, and instead of aborting cleanly the patches proceed with an empty path.

### A failed write is logged and forgotten

`save_file()` calls `upload_file(..., check=False)`, which logs `Failed to upload file` and returns the failed
result; nobody looks at it. The patch function returns `True` regardless, which is the script's signal for
"applied, reboot to take effect".

## Impact

The realistic trigger is a boot partition mounted read-only, which is what a FAT filesystem does after the
unclean shutdowns a vehicle sees routinely. On such a device:

1. the patches read `config.txt`, decide it needs the Navigator overlays,
2. the write silently fails,
3. the script requests a reboot,
4. `BOOT_LOOP_DETECTOR` suppresses the patches for exactly one boot, then they run again.

The result is a device that reboots itself every other boot, forever, with nothing in the UI to say why. The
same swallowed-failure path can instead leave a `config.txt` containing nothing but the Navigator section, on a
device whose boot files were simply unreadable at the moment the patch ran.

## Environment

By inspection on `1.4-dev` and `master`. The helpers are shared, so anything else calling `load_file` /
`save_file` / `locate_file` inherits the same behaviour.

joaoantoniocardoso:
Follow-up to #4232. Addresses #4235.

The shared host-file helpers used by `blueos_startup_update` could not report failure. A failed `cat` looked like an empty file (so the next successful write could replace `config.txt` / `cmdline.txt` with only the patch), `locate_file` never returned `None` (so the existing `is None` guards never fired), and `save_file` ignored a failed upload then returned to a caller that asked for a reboot.

## What changed

- `load_file` raises `HostFileError` when `cat` fails. A real empty file (`returncode == 0`) is still `""`.
- `locate_file` returns `None` when `find` prints nothing. It still accepts a match when `find` exits 1 because an earlier candidate was missing (Bullseye: `/boot/firmware/config.txt` does not exist, `/boot/config.txt` does).
- `save_file` raises `HostFileError` when the `scp` or the `sudo mv` onto the destination fails. A failed backup is logged and is not treated as success of the write.
- `revert_update_dwc2` and `clean_config_pi3` skip when the boot file was not found, matching the other boot patches.
- A failed patch is logged and skipped; patches that already returned `True` still request a reboot. The previous list comprehension aborted on the first `HostFileError`, so a `startup.json` rewrite could lose its restart.
- `get_host_os()` returns `HostOs.Other` if the host `/etc/os-release` cannot be read, which is what an empty read used to do. `load_file` stays strict. That keeps `wifi.can_work` from taking down the wifi API when SSH is not up yet.

A patch that raises no longer returns `True`, so `BOOT_LOOP_DETECTOR` is not created for work that did not happen.

## Test plan

Local: tests in `core/libs/commonwealth/commonwealth/utils/tests/test_commands.py` and `test_general.py`. Reverting each of the three helper fixes turns the matching test red. `locate_file` is also pinned for find exit 1 with a printed path (Bullseye). `get_host_os` is pinned to `Other` when `load_file` raises.

On hardware, the patched `commands.py` was copied into `blueos-core` on three boards, exercised, then restored. Boot partitions were remounted read-only for the save test and remounted read-write afterwards. `config.txt` / `cmdline.txt` hashes were unchanged.

| Board | Image | Locate (real) | Missing locate | Missing load | RO save |
|---|---|---|---|---|---|
| Pi 4B Rev 1.5 (`192.168.0.177`) | Bookworm, `/boot/firmware` | `/boot/firmware/config.txt` | shipped `''` → patched `None` | shipped swallowed → `HostFileError` | `HostFileError` on `mv`, file unchanged |
| Pi 5B Rev 1.0 (`192.168.0.124`) | Bookworm, `/boot/firmware` | `/boot/firmware/config.txt` | shipped `''` → patched `None` | shipped swallowed → `HostFileError` | `HostFileError` on `mv`, file unchanged |
| Pi 4B Rev 1.2 (`192.168.0.88`) | Bullseye, `/boot` | `/boot/config.txt` | shipped `''` → patched `None` | shipped swallowed → `HostFileError` | `HostFileError` on `mv`, file unchanged |

On the read-only boot partition, `scp` to `/tmp` still succeeds and `sudo mv` onto the FAT is what fails. That matches the investigation of #4235: the silent every-other-boot reboot is the ignored-`scp` path; a remounted-RO boot raises (previously `CalledProcessError`, now `HostFileError`) and does not reboot. Either way the patches no longer claim success.

- [x] Pi4 Bookworm
- [x] Pi5 Bookworm
- [x] Pi4 Bullseye

### DIFF
--- a/core/libs/commonwealth/commonwealth/utils/commands.py
+++ b/core/libs/commonwealth/commonwealth/utils/commands.py
@@ -10,6 +10,10 @@ class KeyNotFound(Exception):
     """Raised when the SSH key is not found."""
 
 
+class HostFileError(Exception):
+    """Raised when a host file cannot be read or written."""
+
+
 def run_command_with_password(command: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
     # attempt to run the command with sshpass
     # used as a fallback if the ssh key is not found
@@ -131,8 +135,10 @@ def upload_file_with_ssh_key(source: str, destination: str, check: bool = True) 
 
 
 def load_file(file_name: str) -> str:
-    command = f'cat "{file_name}"'
-    return run_command(command, False).stdout
+    result = run_command(f'cat "{file_name}"', False)
+    if result.returncode != 0:
+        raise HostFileError(f"Failed to read {file_name}: {result.stderr}")
+    return result.stdout
 
 
 def upload_file(file_content: str, destination: str, check: bool = True) -> "subprocess.CompletedProcess['str']":
@@ -147,22 +153,30 @@ def upload_file(file_content: str, destination: str, check: bool = True) -> "sub
         logger.warning("SSH key not found, falling back to password authentication")
         ret = upload_file_with_password(temp_file_in_container, temp_file_in_host, check)
     logger.debug(ret)
-    if ret.returncode == 0:
-        run_command(f"sudo mv {temp_file_in_host} {destination}")
-    else:
+    if ret.returncode != 0:
         logger.error(f"Failed to upload file: {ret.stderr}")
+        return ret
+    moved = run_command(f"sudo mv {temp_file_in_host} {destination}", False)
+    if moved.returncode != 0:
+        logger.error(f"Failed to install uploaded file at {destination}: {moved.stderr}")
+        return moved
     return ret
 
 
 def locate_file(candidates: List[str]) -> Optional[str]:
     # first match will return
     command = f"find {' '.join(candidates)} -type f -print -quit"
-    return run_command(command, False, log_output=False).stdout.strip()
+    found = run_command(command, False, log_output=False).stdout.strip()
+    return found or None
 
 
 def save_file(file_name: str, file_content: str, backup_identifier: str, ensure_newline: bool = True) -> None:
     if ensure_newline and not file_content.endswith("\n"):
         file_content += "\n"
     command = f'sudo cp "{file_name}" "{file_name}.{backup_identifier}.bak"'
-    run_command(command, False)
-    upload_file(file_content, file_name, False)
+    backup = run_command(command, False)
+    if backup.returncode != 0:
+        logger.warning(f"Failed to backup {file_name}: {backup.stderr}")
+    result = upload_file(file_content, file_name, False)
+    if result.returncode != 0:
+        raise HostFileError(f"Failed to save {file_name}: {result.stderr}")
--- a/core/libs/commonwealth/commonwealth/utils/general.py
+++ b/core/libs/commonwealth/commonwealth/utils/general.py
@@ -11,7 +11,7 @@ from typing import Any, AsyncGenerator
 import psutil
 from loguru import logger
 
-from commonwealth.utils.commands import load_file
+from commonwealth.utils.commands import HostFileError, load_file
 from commonwealth.utils.decorators import temporary_cache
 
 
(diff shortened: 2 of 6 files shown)


core/libs/commonwealth/commonwealth/utils/commands.py
```diff
@@ -147,22 +153,30 @@ def upload_file(file_content: str, destination: str, check: bool = True) -> "sub
...
+        return moved
     return ret
 
 
 def locate_file(candidates: List[str]) -> Optional[str]:
     # first match will return
     command = f"find {' '.join(candidates)} -type f -print -quit"
-    return run_command(command, False, log_output=False).stdout.strip()
+    found = run_command(command, False, log_output=False).stdout.strip()
+    return found or None
```
patrickelectric, line 170: can you explain this ?

joaoantoniocardoso: Yes. The callers check `is None` for the semantics of _no-match_, and this is doing the `""` to `None` conversion they expect.

patrickelectric: how is that possible with `run_command` return type ?

joaoantoniocardoso: I believe it's what's chained there

patrickelectric: I'm not following, in which situation `found` that has type of `subprocess.CompletedProcess['str']` ([definition](https://github.com/python/cpython/blob/e5ed2ad9be8e0014a3bf4dc9f89c4ad2695500cd/Lib/subprocess.py#L591)) will be "false" ? This looks that will always return found and the `or None` does not make sense.

joaoantoniocardoso: In our use, `CompletedProcess.stdout` is `''` in a _no-match_ scenario, and the function was returning `''` when its callers expected `None`.

patrickelectric: ooooh, I see the stdout.strip now 🤦

patrickelectric:
Let's wait for 1.4.4 release.

patrickelectric (approved):

PR #3996, opened 2026-07-17
core: services: cable_guy: api: manager: Remove subnet from deleted IP

patrickelectric:
Delete the orphaned route on IP removal and scrub it from saved settings for good.

## Summary by Sourcery

Clean up orphaned subnet routes when removing IP addresses and avoid persisting kernel-managed connected routes.

Bug Fixes:
- Remove subnet routes that outlive their deleted IP addresses and could hijack traffic.
- Ensure kernel-managed connected routes are not adopted as persistent routes in saved interface settings.

PR #4252, opened 2026-08-24
[1.4] core: services: cable_guy: api: manager: Remove subnet from deleted IP

patrickelectric:
Delete the orphaned route on IP removal and skip kernel connected routes.
Cherry-pick of #3996


joaoantoniocardoso (approved):
tested

Issue #4267, opened 2026-08-25
bug: GET extension details by identifier returns 200 [] for an unknown id


joaoantoniocardoso:
### Bug description

`GET /kraken/v2.0/extension/{identifier}/details` for an identifier that was never installed returns **HTTP 200** with an empty list `[]` instead of **HTTP 404**.

The handler loads settings by identifier only (no tag) and returns whatever `from_settings` found. An empty install list is serialized as success. Tagged details for the same missing id (`GET /kraken/v2.0/extension/{identifier}/{tag}/details`) return HTTP 404 `Extension {identifier}:{tag} not found` because `from_settings(identifier, tag)` raises `ExtensionNotFound` when both are supplied.

### Steps to reproduce

On BlueOS `1.4-dev` or `master`:

```bash
curl -sS -D - \
  "http://<blueos>/kraken/v2.0/extension/np.no.such.extension/details"
```

Observe: HTTP 200, body `[]`.

Contrast:

```bash
curl -sS -D - \
  "http://<blueos>/kraken/v2.0/extension/np.no.such.extension/sometag/details"
```

Observe: HTTP 404.

Confirmed live on `1.4.4-beta.23`. Same `fetch_by_identifier` body is on `1.4-dev` and `master`.

### Primary pain point(s)

Clients cannot tell a missing extension from an installed extension that has no versions in settings.

joaoantoniocardoso:
Closed by #4270

PR #4283, opened 2026-08-25
[1.4] kraken: Return 404 when listing tags of an unknown extension

### LINKED ISSUE #4277: bug: GET manifest tags for an unknown extension returns 200 [] instead of 404
### Bug description

`GET /kraken/v2.0/manifest/tags/{extension_identifier}` for an identifier that is not in the catalog returns **HTTP 200** with an empty list `[]` instead of **HTTP 404**.

`fetch_extension_versions` returns `[]` when `fetch_extension` finds nothing. The handler serializes that as success. The same empty-list path is used by `GET /kraken/v2.0/manifest/tags/{manifest_identifier}/{extension_identifier}/` when the extension is missing from a known manifest (unknown manifest already 404s via `ManifestNotFound`).

Contrast identifier-only installed details (`GET /kraken/v2.0/extension/{identifier}/details`), which has the same empty-list pattern and is tracked in https://github.com/bluerobotics/BlueOS/issues/4267.

**Affects `master` and `1.4-dev`.**

### Steps to reproduce

On BlueOS `master` or `1.4-dev`:

```bash
curl -sS -D - \
  "http://<blueos>/kraken/v2.0/manifest/tags/np.no.such.extension"
```

Observe: HTTP 200, body `[]`.

Contrast a missing manifest source:

```bash
curl -sS -D - \
  "http://<blueos>/kraken/v2.0/manifest/np.no.such.manifest/details?data=false"
```

That one returns HTTP 404.

Confirmed in source: `fetch_extension_versions` returns `[]` when the extension is missing; the tags handlers do not raise `ExtensionNotFound`.

### Primary pain point(s)

Clients cannot tell "this identifier is not in the catalog" from "this extension has no tags".

### Prerequisites

joaoantoniocardoso:
Fixes #4277.

`GET /kraken/v2.0/manifest/tags/{extension_identifier}` for an identifier that is not in the catalog returned HTTP 200 with body `[]`. The same empty-list path is used by `GET /kraken/v2.0/manifest/tags/{manifest_identifier}/{extension_identifier}/` when the extension is missing from a known manifest. Unknown manifest sources already 404 via `ManifestNotFound`.

`fetch_extension_versions` returned `[]` when `fetch_extension` found nothing. The tags handlers serialized that as success. Clients could not tell a missing catalog identifier from an extension that has no tags.

This raises `ExtensionNotFound` in the two tags handlers when `fetch_extension` returns `None`. An extension that exists but has an empty `versions` dict, or `stable=true` with only prereleases, still returns 200 `[]`. The existing `manifest_to_http_exception` decorator now maps `ExtensionNotFound` to 404 alongside `ManifestNotFound`.

The handlers look up the catalog entry once and derive tags from it (`versions_from_entry`). They do not call `fetch_extension` a second time. `fetch_latest_extension_version` uses the same helper on its already-fetched entry. `fetch_consolidated` is not cached at the merge layer, so a second lookup was a full rebuild of the enabled-manifest list.

Independent of #4267 / #4270 (unknown installed details). That PR touches `fetch_by_identifier` on the extension router.

## Test plan

On `192.168.0.124` (`1.4.4-beta.23`), patched `api/v2/routers/manifest.py` and `manifest/manifest.py` via container copy + `docker restart blueos-core`.

Status-code split:

- [x] Before patch: unknown `GET /manifest/tags/{id}` → 200 `[]`
- [x] Before patch: unknown `GET /manifest/tags/{factory}/{id}/` → 200 `[]`
- [x] After patch: unknown `GET /manifest/tags/{id}` → 404 `Extension np.no.such.extension not found`
- [x] After patch: unknown `GET /manifest/tags/{factory}/{id}/` → 404
- [x] Known `GET /manifest/tags/bluerobotics.cockpit` → 200, 16 tags
- [x] Known `?stable=true` → 200 `["1.18.2","1.18.1","1.18.0"]`
- [x] Unknown manifest details → 404 (unchanged)

Full Kraken lifecycle with `bluerobotics.cockpit:v1.18.2`. `major_tom` left installed. 123/123 checks passed after the single-lookup change, including never-installed `POST /{id}/install?stable=true`.

- [x] v1 `POST /extension/install` of a never-installed id → 200 pull stream, not 404
- [x] Cockpit listed, enabled, container running; `major_tom` still present
- [x] v2 details by identifier and by identifier+tag → 200
- [x] v1/v2 container list, stats, and log streams
- [x] v2 and v1 restart of the running extension → 202, container comes back
- [x] v2 disable → 204, container gone, restart → 400; v2 enable → 204, container back
- [x] v1 disable → 200; v1 enable of the installed id → 200
- [x] v1 uninstall → 200
- [x] v2 `POST /extension/{id}/{tag}/install` of a never-installed id → 200, not 404
- [x] Reinstall while already running → 200, not 404
- [x] v2 `POST /{id}/install?stable=true` of a never-installed id → 200, not 404
- [x] DUT restored to only `major_tom`; management address still pings

### DIFF
--- a/core/services/kraken/api/v2/routers/manifest.py
+++ b/core/services/kraken/api/v2/routers/manifest.py
@@ -85,11 +85,10 @@ async def fetch_ext_tags_from_manifest(
     """
     Get all tags for a given extension identifier using one manifest as source.
     """
-    if not await manifest_manager.fetch_extension(extension_identifier, manifest_identifier):
+    ext = await manifest_manager.fetch_extension(extension_identifier, manifest_identifier)
+    if not ext:
         raise ExtensionNotFound(f"Extension {extension_identifier} not found")
-    versions = await manifest_manager.fetch_extension_versions(extension_identifier, stable, manifest_identifier)
-
-    return [str(version) for version in versions]
+    return [str(version) for version in manifest_manager.versions_from_entry(ext, stable)]
 
 
 @manifest_router_v2.get("/tags/{extension_identifier}", status_code=status.HTTP_200_OK)
@@ -98,11 +97,10 @@ async def fetch_ext_tags_from_consolidated(extension_identifier: str, stable: bo
     """
     Get all tags for a given extension identifier using consolidated manifest as source.
     """
-    if not await manifest_manager.fetch_extension(extension_identifier):
+    ext = await manifest_manager.fetch_extension(extension_identifier)
+    if not ext:
         raise ExtensionNotFound(f"Extension {extension_identifier} not found")
-    versions = await manifest_manager.fetch_extension_versions(extension_identifier, stable)
-
-    return [str(version) for version in versions]
+    return [str(version) for version in manifest_manager.versions_from_entry(ext, stable)]
 
 
 @manifest_router_v2.post("/", status_code=status.HTTP_201_CREATED)
--- a/core/services/kraken/api/v2/routers/manifest.py
+++ b/core/services/kraken/api/v2/routers/manifest.py
@@ -5,6 +5,7 @@ from fastapi import APIRouter, HTTPException, status
 from fastapi.responses import Response
 from fastapi_versioning import versioned_api_route
 
+from extension.exceptions import ExtensionNotFound
 from manifest import ManifestManager
 from manifest.exceptions import (
     ManifestDataFetchFailed,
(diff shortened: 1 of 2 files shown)


nicoschmdt (approved):

patrickelectric (commented):
@joaomariolago can you take a look in this PR ?

joaomariolago (approved):
Strictly speaking, `ExtensionNotFound` is reserved for extensions themselves rather than `RepositoryEntry`. It doesn't hurt to reuse the exception here, but in the future, separating it into something like `ExtensionEntryNotFound` may be more convenient and could help avoid circular imports from the manifest module. Other than that changes are good!

PR #4298, opened 2026-08-26
[1.4] frontend: Stop create from installing a placeholder identifier

### LINKED ISSUE #4014: "Ghost" extension
The steps to reproduce don't fit on the main user journey -> normal priority.

How I managed to get it:
1. Installed RadCam Manager from the store using the UI
2. Clicked "edit", copied the content using the UI button
3. Disabled the extension
4. Added a new extension, changed the Docker URL and tag, and clicked the "CREATE" UI button. After this, only 1 extension was showing on the extension manager UI page.
5. Kraken, however, spawned both extensions in the backend, showing 2 "Radcam manager" on the left menu, and `docker ps` would show both. I tried rebooting, I tried killing the running extension containers, nothing would recover it.
6. Then, I uninstalled the only extension available on the extensions manager UI page.
7. To my surprise, the originally installed extension from the store was still alive, and stored in the settings, but hidden from the UI:

Image

Image

Image

Image

Kraken logs:

[ghost_extension_logs_kraken.zip](https://github.com/user-attachments/files/30279317/ghost_extension_logs_kraken.zip)

joaoantoniocardoso:
Fixes #4014.

Create Extension prefills `yourorganization.yourextension`, which already passes identifier validation. Changing only Docker URL and tag, then CREATE, installs a second extension under that placeholder id. Kraken starts both. Installed cards were keyed by `docker`, so two rows that share an image repo collapse to one card. Uninstalling the visible store entry leaves the extra one running.

On `192.168.0.147` (`1.4-dev` on `eth0`), issue kraken logs plus a live API repro showed `yourorganization.yourextension:<copied tag>` sitting next to the original identifier.

This PR:

- Starts the identifier field empty (example stays as placeholder text). CREATE with only docker/tag fails validation.
- Keys installed cards by `identifier` instead of `docker`.

`createOrUpdateExtension` still installs `this.edited_extension`. The modal's `new_extension` is that same object on a fresh Create (`data()` assigns the prop; `watch.extension` is not `immediate`), so form edits already land on the parent. Switching to the emit payload was a no-op and was dropped.

Same create-dialog defaults and `:key="extension.docker"` are on `master`.

## Test plan

Vite frontend (`BLUEOS_ADDRESS=http://192.168.0.147/`) against that DUT, after dropping the payload switch.

Before patch (DUT UI + v1 install):

- [x] Create dialog identifier is `yourorganization.yourextension`
- [x] Store-like install + CREATE with that default identifier + different tag → both in `GET /kraken/v2.0/extension/`
- [x] Uninstall store identifier → placeholder leftover still in settings and started by the watchdog

After patch:

- [x] Create dialog identifier is empty
- [x] Issue steps: disabled store-like `bluerobotics.radcam-manager` + CREATE with name/docker/tag only → validation error; no `yourorganization.yourextension`
- [x] CREATE with identifier `testharness.ghostfix` (alpine 3.19) while store-like alpine 3.20 is present → settings get `testharness.ghostfix`, both cards visible (same `docker`, different identifier)
- [x] Uninstall both test vehicles → only `blueos.major_tom` remains

## Skipped

- Full Kraken HTTP lifecycle: backend routes are unchanged.
- Installed cards still do not print `identifier`; two same-named custom cards can look identical.
- Typing the example identifier is still allowed.

### DIFF
--- a/core/frontend/src/views/ExtensionManagerView.vue
+++ b/core/frontend/src/views/ExtensionManagerView.vue
@@ -120,7 +120,7 @@
         <div class="installed-extensions-container">
           <InstalledExtensionCard
             v-for="extension in installed_extensions"
-            :key="extension.docker"
+            :key="extension.identifier"
             :extension="extension"
             :loading="extension.loading"
             :metrics="metricsFor(extension)"
--- a/core/frontend/src/views/ExtensionManagerView.vue
+++ b/core/frontend/src/views/ExtensionManagerView.vue
@@ -389,7 +389,7 @@ export default Vue.extend({
     },
     openCreationDialog() : void {
       this.edited_extension = {
-        identifier: 'yourorganization.yourextension',
+        identifier: '',
         name: '',
         docker: '',
         enabled: true,
--- a/core/frontend/src/components/kraken/modals/ExtensionCreationModal.vue
+++ b/core/frontend/src/components/kraken/modals/ExtensionCreationModal.vue
@@ -35,6 +35,7 @@
           <v-text-field
             v-model="new_extension.identifier"
             label="Extension Identifier"
+            placeholder="yourorganization.yourextension"
             :disabled="is_editing"
             :rules="[validate_identifier]"
           />


core/frontend/src/views/ExtensionManagerView.vue
```diff
@@ -120,7 +120,7 @@
         <div class="installed-extensions-container">
           <InstalledExtensionCard
             v-for="extension in installed_extensions"
-            :key="extension.docker"
+            :key="extension.identifier"
```
patrickelectric, line 123: just to be sure.. how is this related to the placeholder ?

joaoantoniocardoso: Sorry, I'm splitting it into two different commits, as they are different parts of the fix but not exactly related. The docker key allows for the frontend to hide the "ghost" installed extensions, switching to identifier solves that part.

patrickelectric (approved):

PR #4331, opened 2026-08-26
do not try to fetch remote tags if there is no internet

### LINKED ISSUE #1788: version chooser: we shouldn't try to fetch remote tags if there is no internet
this error is somehwat criptical for users:

image

nicoschmdt:
fix: #1788

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionChooser.vue
+++ b/core/frontend/src/components/version-chooser/VersionChooser.vue
@@ -44,12 +44,13 @@
           @pull-and-apply="pullAndSetVersion"
         />
         <v-alert
-          v-if="available_versions.error"
+          v-if="remote_alert_message"
+
           color="red lighten-2"
           dark
         >
           <div class="d-flex align-center justify-space-between">
-            <span>{{ available_versions.error }}</span>
+            <span>{{ remote_alert_message }}</span>
             <v-btn
               color="warning"
               dark
@@ -118,12 +119,13 @@
           />
         </v-form>
         <v-alert
-          v-if="available_versions.error"
+          v-if="remote_alert_message"
+
           color="red lighten-2"
           dark
         >
           <div class="d-flex align-center justify-space-between">
-            <span>{{ available_versions.error }}</span>
+            <span>{{ remote_alert_message }}</span>
             <v-btn
               color="warning"
               dark
@@ -345,16 +347,32 @@ export default Vue.extend({
     inputFileRequiredMessage(): string {
       return 'File is required'
     },
-    has_internet(): boolean {
-      return helper.has_internet !== InternetConnectionState.OFFLINE
+    internet_state(): InternetConnectionState {
+      return helper.has_internet
+    },
+    is_offline(): boolean {
+      return this.internet_state === InternetConnectionState.OFFLINE
+    },
+    can_fetch_remote(): boolean {
+      return this.internet_state === InternetConnectionState.ONLINE
+        || this.internet_state === InternetConnectionState.LIMITED
+    },
+    remote_alert_message(): string | null {
+      if (this.is_offline) {
+        return "No internet connection available, can't fetch remote images"
+      }
+      return this.available_versions.error
     },
   },
   watch: {
-    has_internet(value: boolean) {
-      if (value) {
-        this.loadAvailableVersions()
-      } else {
+    is_offline(offline: boolean) {
+      if (offline) {
         this.resetToNoInternetAvailable()
+      }
+    },
+    can_fetch_remote(can_fetch: boolean) {
+      if (can_fetch) {
+        this.loadAvailableVersions()
       }
     },
   },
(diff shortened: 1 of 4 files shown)


core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -323,18 +323,40 @@ export default Vue.extend({
...
+    },
+    is_offline(): boolean {
+      return this.internet_state === InternetConnectionState.OFFLINE
+    },
+    can_fetch_remote(): boolean {
+      return this.internet_state === InternetConnectionState.ONLINE
+        || this.internet_state === InternetConnectionState.LIMITED
+    },
+    offline_error_message(): string {
+      return "No internet connection available, can't fetch remote images"
```
patrickelectric: Two things here.. We are comparing a string of the front with a string of the backend ? We should avoid this.. Also.. This string is not in the codebase at all, where is this string coming from ?

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -323,18 +323,40 @@ export default Vue.extend({
...
-      return helper.has_internet !== InternetConnectionState.OFFLINE
+    internet_state(): InternetConnectionState {
+      return helper.has_internet
+    },
+    is_offline(): boolean {
+      return this.internet_state === InternetConnectionState.OFFLINE
+    },
+    can_fetch_remote(): boolean {
+      return this.internet_state === InternetConnectionState.ONLINE
+        || this.internet_state === InternetConnectionState.LIMITED
```
patrickelectric, line 358: is this `!is_offline` ?

patrickelectric (changes requested):

core/frontend/src/components/version-chooser/VersionChooser.vue
nicoschmdt: no because if internet_state is unknown then it'd would consider online and try to fetch the remotes

core/services/versionchooser/utils/chooser.py
```diff
@@ -375,6 +376,10 @@ async def set_remote_versions(
             output["error"], online_tags = await TagFetcher().fetch_remote_tags(
                 repository, [image["tag"] for image in output["local"]]
             )
+        except (aiohttp.ClientConnectionError, OSError) as error:
+            logger.warning(f"Skipping remote tags, DockerHub unreachable: {error}")
+            online_tags = []
+            output["error"] = NO_INTERNET_ERROR
         except Exception as error:
             logger.critical(f"error fetching online tags: {error}")
```
patrickelectric, line 396: Now with timeout, the error as string will be empty. https://stackoverflow.com/questions/52257101/why-asyncio-raises-timeouterror-without-any-message

we should have a catch for it.

core/services/versionchooser/utils/dockerhub.py
```diff
@@ -12,6 +12,10 @@
 import aiohttp
 from loguru import logger
 
+# Unreachable DockerHub used to hang on aiohttp's 5-minute default timeout.
+DOCKERHUB_TIMEOUT = aiohttp.ClientTimeout(total=60, connect=10)
+NO_INTERNET_ERROR = "Unable to reach DockerHub, can't fetch remote images"
```
patrickelectric: over having a variable for an error message that is being used in a single place, can we have the string directly where it is used ?

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -457,10 +476,13 @@ export default Vue.extend({
         })
     },
     async loadAvailableVersions() {
-      if (!this.has_internet) {
+      if (this.internet_state === InternetConnectionState.OFFLINE) {
```
patrickelectric: can't we use is_offline ?

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -457,10 +476,13 @@ export default Vue.extend({
...
         this.resetToNoInternetAvailable()
         return
       }
+      if (this.internet_state === InternetConnectionState.UNKNOWN) {
```
patrickelectric: should we have a computed here ? we have for the others.

core/frontend/src/components/version-chooser/VersionChooser.vue
```diff
@@ -323,18 +323,37 @@ export default Vue.extend({
...
+    can_fetch_remote(): boolean {
+      return this.internet_state === InternetConnectionState.ONLINE
+        || this.internet_state === InternetConnectionState.LIMITED
+    },
+    offline_error_message(): string {
+      return "No internet connection available, can't fetch remote images"
+    },
+    remote_alert_message(): string | null {
+      if (this.is_offline) {
+        return this.offline_error_message
```
patrickelectric: Can we have the message here directly since it's not being used in other places ?

patrickelectric (changes requested):

patrickelectric (approved):
nice

PR #4338, opened 2026-08-27
[1.4] Fix extension update wiping the running version when pull fails

### LINKED ISSUE #1643: Kraken: (new) issue updating extensions:
image

joaoantoniocardoso:
## Summary
- Treat a failed Docker image pull as an install/update failure instead of success, so Kraken no longer uninstalls the running tag and leaves a version with no local image.
- Run extension `update()` and v2 tagged/`from_latest` install atomically (default). A failed pull restores a copy of prior settings for that tag (including same-tag reinstall) and re-enables the previously running sibling, without `images.delete` of a possible alias.
- Inspect `docker:tag` for what `start()` actually runs. Only Docker 404 means the image is absent. Any other inspect/daemon error refuses the local-image fallback and rolls back — it does not report success.
- Reject a retag of the running sibling's image as the requested version on the failed-pull path. If the sibling image cannot be inspected (including Docker 404), refuse the local-image fallback. After a clean pull, do not apply that sibling-Id guard.
- A pull stream is not success unless it includes `Downloaded newer image` or `Image is up to date`. Disconnect or empty completion rolls back instead of purging the running tag.
- Retag from the catalog platform digest when that ref is already local. Do not scan leftover images of the same repo.
- If the pull fails but a usable local `docker:tag` already exists and is not the sibling's image, keep that tag and do not purge other tags.
- Yield the pull-stream error line before failing, log local-image fallback at warning, and cap start-attempt backoff with `min` instead of `max` (was ~600s after the first failure).

Closes #1643

## Test plan
- [x] Reproduced on DUT `192.168.0.147` (1.4-dev): `update_to_version` Cockpit `v1.18.2` → `v1.19.0-beta.9` with Docker Hub blocked returned HTTP 200 with an empty body, replaced settings with the new tag, deleted the old image, and the watchdog then failed with `No such image`.
- [x] Radius **43 passed / 0 failed** on every `install()` / `update()` HTTP caller plus watchdog `start()`, overlay confirmed (do not `docker restart blueos-core` on this DUT). Includes:
  - pull/digest/404 helper table in-process
  - v1/v2 custom install pull-fail (fragment + reason, not listed)
  - same-tag update while running
  - v1 `update_to_version` Hub blocked, target not local (settings + image + container stay)
  - watchdog `start()` with image deleted and Hub blocked (`Image not found locally` + `Failed to start`, no `No such image`; retry ~15s with `min` backoff)
  - local-image update keeps the old image and disabled old tag
  - v2 PUT local tag
  - wrong local tag (retag of the running image) does not look like a successful update; alias docker tag is not untagged
  - leftover unique digest-only image of the same repo is **not** promoted; running tag is restored
  - catalog platform digest without `:tag` is retagged and update succeeds
  - v2 tagged and `from_latest` install never-installed, Hub blocked
  - v2 PUT not-local restores the running sibling
- [x] Review DUT checks after sibling-404 and pull success-status: untag running `docker:A`, alias `docker:B`, Hub blocked `update_to_version` → `Failed to pull`, A stays enabled. Empty pull stream (`Pulling fs layer` only) → `pull finished without a success status`, A kept, B not enabled. **14/14**.
- [x] `journey_http --extension-lifecycle --allow-mutating --identifier bluerobotics.cockpit --tag v1.18.2 --assert-unknown`: **181 passed, 0 failed**; DUT restored to baseline (`blueos.major_tom` only).

### DIFF
--- a/core/services/kraken/extension/extension.py
+++ b/core/services/kraken/extension/extension.py
@@ -146,29 +146,18 @@ class Extension:
         finally:
             cls.unlock(container_name)
 
-    async def _image_is_available_locally(self) -> bool:
-        try:
-            image_ref = f"{self.source.docker}:{self.tag}" + (f"@{self.digest}" if self.digest else "")
-            async with DockerCtx() as client:
-                await client.images.inspect(image_ref)
-                return True
-        except DockerError:
-            return False
-
-    async def install(  # pylint: disable=too-many-branches
-        self, clear_remaining_tags: bool = True, atomic: bool = False
-    ) -> AsyncGenerator[bytes, None]:
-        logger.info(f"Installing extension {self.identifier}:{self.tag}")
-
-        # First we should make sure no other tag is running
-        running_ext = None
+    async def _disable_running_extension(self) -> Optional["Extension"]:
+        """Disable any currently running extension with the same identifier."""
         try:
             running_ext = await self.from_running(self.identifier)
             if running_ext:
                 await running_ext.disable()
+            return running_ext
         except (ExtensionNotRunning, ExtensionNotFound):
-            pass
+            return None
 
+    def _create_extension_settings(self) -> ExtensionSettings:
+        """Create and save extension settings."""
         new_extension = ExtensionSettings(
             identifier=self.identifier,
             name=self.source.name,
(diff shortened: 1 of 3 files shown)


core/services/kraken/extension/extension.py
```diff
@@ -180,25 +169,57 @@ async def install(  # pylint: disable=too-many-branches
...
+        return new_extension
+
+    def _prepare_docker_auth(self) -> Optional[str]:
+        """Prepare Docker authentication string from source auth credentials."""
+        if self.source.auth is None:
+            return None
+        docker_auth = f"{self.source.auth.username}:{self.source.auth.password}"
+        return base64.b64encode(docker_auth.encode("utf-8")).decode("utf-8")
 
+    async def _image_is_available_locally(self) -> bool:
```
patrickelectric, line 181: can we avoid moving this function around ? it just add unnecessary diff

core/services/kraken/extension/extension.py
```diff
@@ -180,25 +169,57 @@ async def install(  # pylint: disable=too-many-branches
...
-                async for line in client.images.pull(
-                    tag, repo=self.source.docker, tag=self.tag, auth=docker_auth, stream=True
-                ):
-                    # TODO - Plug Error detection from docker image here
-                    yield json.dumps(line).encode("utf-8")
-                # Make sure to add correct tag if a digest was used since docker messes up the tag
-                if self.digest:
-                    await client.images.tag(tag, f"{self.source.docker}:{self.tag}")
+        # First we should make sure no other tag is running
+        running_ext = await self._disable_running_extension()
```
patrickelectric, line 214: we are doing nothing with this variable, is this expected ?

core/services/kraken/extension/extension.py
```diff
@@ -168,6 +168,7 @@ def _create_extension_settings(self) -> ExtensionSettings:
             user_permissions=self.source.user_permissions,
         )
         # Save in settings first, if the image fails to install it will try to fetch after in main kraken check loop
+        # Atomic failure rolls this entry back and re-enables the previously running sibling.
```
patrickelectric, line 171: should this be in anoter commit ?

core/services/kraken/extension/extension.py
```diff
@@ -178,27 +179,95 @@ def _prepare_docker_auth(self) -> Optional[str]:
...
+                return None
+            raise
+
+    async def _ensure_tagged_local_image(self, client: Any, sibling_image_id: Optional[str] = None) -> bool:
+        # start() and a successful pull run docker:tag. Catalog platform digests are not
+        # stored in RepoDigests after `docker pull repo:tag` (that records the index
+        # digest), so a digest match against the catalog cannot be required.
+        # sibling_image_id is only for the failed-pull fallback: a retag of the running
+        # sibling onto the new name is not the requested version. Do not pass it after a
+        # clean pull -- two tags can share an image Id (aliases) and still be the pull.
```
patrickelectric, line 197: Image

This is not friendly at all.
It took so much time to understand, and I'm still not sure.

---

We get an image from DockerHub, their "catalog" ID does not match with what we receive, is that right ?

sibling_image_id is the image that is used as base from a failed attempt ? Or is the image that has different "repo:tag" ?

core/services/kraken/extension/extension.py
```diff
@@ -178,27 +179,95 @@ def _prepare_docker_auth(self) -> Optional[str]:
...
+            if prior_settings is not None:
+                self._save_settings(prior_settings)
+            elif atomic:
+                self._save_settings()
+            else:
+                await self.set_enabled(False)
+        except Exception as rollback_error:
+            logger.warning(f"Failed to roll back {self.identifier}:{self.tag} after pull failure: {rollback_error}")
+        if not running_ext:
+            return
```
patrickelectric, line 237: should we have a log message here ?

patrickelectric (changes requested):

Issue #1695, opened 2023-05-23
bug: No Extensions available.


phamquandung:
### Bug description

No Extensions available. Make sure the vehicle has internet access and try again.

### Steps to reproduce

1. Access: http://192.168.2.2/tools/extensions-manager
2. The error pop-up: No Extensions available. Make sure the vehicle has internet access and try again.

I check the internet, all work file. I tried on BlueOS version 1.1.0-beta-21 and 18. And also for Remote Versions: error fetching online tags: Cannot connect to host auth.docker.io:443 ssl:default [Temporary failure in name resolution]
![error](https://github.com/bluerobotics/BlueOS-docker/assets/28943896/d7bebf7f-9a07-45d4-88f5-d8a82c675f79)

### Prerequisites

- [X] I have checked to make sure that a similar request has not already been filed or fixed.

patrickelectric:
Hi @phamquandung,

Thanks for reporting the issue, is this a consistent problem ? Is it working fine now ?

phamquandung:
> Hi @phamquandung,
> 
> Thanks for reporting the issue, is this a consistent problem ? Is it working fine now ?
Hello @patrickelectric . It's still there. Please check
![error](https://github.com/bluerobotics/BlueOS-docker/assets/28943896/96540316-d6f7-400c-be6b-d1020206c45e)

patrickelectric:
Please connect your vehicle to wifi, check the wifi icon.
https://docs.bluerobotics.com/ardusub-zola/software/onboard/BlueOS-1.1/getting-started/#connect-wifi

patrickelectric:
It appears that Docker can run only with ipv6 in some cases.

```
getent hosts auth.docker.io hub.docker.com

2a06:98c1:3104::6812:2bb2 auth.docker.io.cdn.cloudflare.net auth.docker.io
2606:4700:4403::ac40:904e auth.docker.io.cdn.cloudflare.net auth.docker.io
```

This appears to be one of the problems that results in:

<img width="900" height="156" alt="Image" src="https://github.com/user-attachments/assets/c3dacec3-3cfe-4f77-a83c-0aa1696295a7" />

patrickelectric:
Fixed the fetch but not the download, working on it:

```
The operation failed: error while pulling image: DockerError(500, 'Get "https://registry-1.docker.io/v2/": dial tcp [2600:1f18:2148:bc01:e939:a0be:6929:ed84]:443: connect: network is unreachable')
```

PR #4342, opened 2026-08-27
core: services: versionchooser: dockerhub: Handle DNS failures

patrickelectric:
Related to #1695

### DIFF
--- a/core/services/versionchooser/utils/dockerhub.py
+++ b/core/services/versionchooser/utils/dockerhub.py
@@ -4,9 +4,13 @@ Responsible for interacting with dockerhub
 adapted from https://github.com/al4/docker-registry-list
 """
 
+import asyncio
 import platform
+import socket
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from functools import partial
+from typing import Any, List, Optional, Tuple
 from warnings import warn
 
 import aiohttp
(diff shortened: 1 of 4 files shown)


joaoantoniocardoso:
I'm reviewing it, but before start.. the targeted issue doesn't seem to be about versionchooser, but .. kraken?

@patrickelectric is that the right issue?

patrickelectric:
Well, is more a related than a fix. Will update de body. And work on that later.

joaoantoniocardoso:
> Well, is more a related than a fix. Will update de body. And work on that later.

Right, it's the same class of problem: DNS and IP version.

So, looking at this patch, my main concern is that we are assuming a strategy that will have a short life in the codebase: vehicles in IPv6-only networks, like boats using a cellular connection, will break and we will have to figure out a better solution.

Today, the _right_ path I can see is to try both AF_INET and AF_INET6 concurrently and let the race decide the destiny. We could also bias this strategy with an input from our knowledge of the network from cable_guy's routes API, looking at the default route.

Also, architectural-wise, I'd say that we should outsource the part of the code that interfaces with docker to a shared library between kraken and versionchooser (but it'd be a significantly bigger effort).

patrickelectric:
> Today, the _right_ path I can see is to try both AF_INET and AF_INET6 concurrently and let the race decide the destiny.

Ok, I'll do that.

> We could also bias this strategy with an input from our knowledge of the network from cable_guy's routes API, looking at the default route.

I would prefer to not touch in in cable_guy in this PR, we could.. but I prefer to raise that as a consequent issue over exploding the scope of this one.



> Also, architectural-wise, I'd say that we should outsource the part of the code that interfaces with docker to a shared library between kraken and versionchooser (but it'd be a significantly bigger effort).

This code does not change the current architecture, is the same one, the big diff is only moving the indentation level and some tests being done in different places. But logic/architecture are still the same. I would raise that also in another issue if necessary.

core/services/versionchooser/utils/dockerhub.py
```diff
@@ -31,6 +41,108 @@ def get_current_arch() -> str:
...
+        raise socket.gaierror(socket.EAI_NONAME, f"Name or service not known: {host}")
+
+    async def close(self) -> None:
+        pass
+
+
+def _session() -> aiohttp.ClientSession:
+    # First successful family wins when AAAA lookups stall.
+    return aiohttp.ClientSession(connector=aiohttp.TCPConnector(resolver=_FamilyRaceResolver()))
+
```
joaoantoniocardoso, line 124: Almost that, but instead of racing at the DNS, race at the connection, e.g.:

```Python
_DNS_FAMILY_TIMEOUT_S = 2.0


class _FamilyRaceResolver(aiohttp.resolver.ThreadedResolver):
    """Resolve AF_INET and AF_INET6 in parallel. Keep every family that answers."""

    async def resolve(
        self, host: str, port: int = 0, family: int = socket.AF_UNSPEC
    ) -> List[Any]:
        if family != socket.AF_UNSPEC:
            return await super().resolve(host, port, family)

        last_error: Optional[BaseException] = None

        async def lookup(fam: int) -> List[Any]:
            nonlocal last_error
            try:
                return await asyncio.wait_for(
                    super(_FamilyRaceResolver, self).resolve(host, port, fam),
                    timeout=_DNS_FAMILY_TIMEOUT_S,
                )
            except OSError as error:
                last_error = error
                return []

        v4, v6 = await asyncio.gather(lookup(socket.AF_INET), lookup(socket.AF_INET6))
        hosts = [*v4, *v6]
        if hosts:
            return hosts
        if last_error is not None:
            raise last_error
        raise socket.gaierror(socket.EAI_NONAME, f"Name or service not known: {host}")


def _session() -> aiohttp.ClientSession:
    return aiohttp.ClientSession(
        connector=aiohttp.TCPConnector(
            resolver=_FamilyRaceResolver(),
            happy_eyeballs_delay=0.25,
        )
    )

```

patrickelectric:
Updated.

core/services/versionchooser/utils/dockerhub.py
patrickelectric: Updated.

patrickelectric:
haaa, CI is not happy now.

core/services/versionchooser/utils/dockerhub.py
patrickelectric: fixed

joaoantoniocardoso (approved):

PR #4347, opened 2026-08-28
[1.4] core: services: versionchooser: dockerhub: Handle DNS failures

patrickelectric:
Hub lookups stall when DNS fails.

Cherry-pick of #4342
Related to #1695

### DIFF
--- a/core/frontend/src/components/version-chooser/VersionChooser.vue
+++ b/core/frontend/src/components/version-chooser/VersionChooser.vue
@@ -45,11 +45,22 @@
         />
         <v-alert
           v-if="available_versions.error"
-
           color="red lighten-2"
           dark
         >
-          {{ available_versions.error }}
+          <div class="d-flex align-center justify-space-between">
+            <span>{{ available_versions.error }}</span>
+            <v-btn
+              color="warning"
+              dark
+              small
+              :loading="loading_images"
+              :disabled="loading_images"
+              @click="loadAvailableVersions"
+            >
+              Retry
+            </v-btn>
+          </div>
         </v-alert>
       </v-card>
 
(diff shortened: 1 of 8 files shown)


patrickelectric:
python-tests is red on `happy_eyeballs_delay`. 1.4 is still on aiohttp 3.7.4, that kwarg does not exist there. Master got 3.13.2 in #4074.

patrickelectric:
Updating aiohttp on 1.4 is not straightforward, versionchooser still uses connexion 2.14.2 and aiohttp-jinja2. Master moved off that before #4074.

patrickelectric:
Dropped `happy_eyeballs_delay`.

joaoantoniocardoso (changes requested):
> Dropped `happy_eyeballs_delay`.

We are lacking HE on this version, substantially decreasing the value of the proposed change as a solution to the problem.

Alternatively, we can add the `aiohappyeyeballs==2.7.1` dependency and wire it ourselves, like:

```python
class _HappyEyeballsConnector(aiohttp.TCPConnector):
    async def _create_direct_connection(  # type: ignore[override]
        self,
        req: Any,
        traces: Any,
        timeout: Any,
        *,
        client_error: Any = aiohttp.ClientConnectorError,
    ) -> Any:
        sslcontext = self._get_ssl_context(req)
        host = req.url.raw_host
        port = req.port
        hosts = await self._resolve_host(host, port, traces=traces)
        addr_infos = []
        for hinfo in hosts:
            if hinfo["family"] == socket.AF_INET6:
                addr: Any = (hinfo["host"], hinfo["port"], 0, 0)
            else:
                addr = (hinfo["host"], hinfo["port"])
            addr_infos.append(
                (hinfo["family"], socket.SOCK_STREAM, socket.IPPROTO_TCP, "", addr)
            )
        try:
            sock = await aiohappyeyeballs.start_connection(
                addr_infos,
                happy_eyeballs_delay=0.25,
                interleave=1,
            )
        except OSError as error:
            raise client_error(req.connection_key, error) from error
        try:
            return await self._wrap_create_connection(
                self._factory,
                sock=sock,
                timeout=timeout,
                ssl=sslcontext,
                server_hostname=host if sslcontext else None,
                req=req,
                client_error=client_error,
            )
        except Exception:
            sock.close()
            raise


def _session() -> aiohttp.ClientSession:
    return aiohttp.ClientSession(
        connector=_HappyEyeballsConnector(resolver=_FamilyRaceResolver()),
    )
```

What do you think?

patrickelectric:
@joaoantoniocardoso check now

patrickelectric:
python-tests is red, missing aiohappyeyeballs in the test env.

patrickelectric:
fixed

joaoantoniocardoso (approved):

PR #4348, opened 2026-08-29
core: frontend: wifi: Freeze network list on hover

### LINKED ISSUE #637: core: frontend: wifi: Do not update the wifi list if the mouse hovers over it
Maybe we should add an update button ?
Issue: It's a bit annoying when the WiFi network list refreshes when you're going to click on one, and then it moves. Not a huge deal but when you have a lot of networks nearby it gets worse.

patrickelectric:
The scan replaces the list every 20s and the row jumps under the pointer.
Fix #637


patrickelectric:
As discussed, we should have a small banner saying that the update is in hold. Maybe add a manual reload wifi list

PR #4350, opened 2026-08-29
core: frontend: FirmwareManager: Only flash the selected board

### LINKED ISSUE #3324: core: frontend: Remove board selection combobox from firmware update
### Current behaviour

At the moment is confuse that we can do a firmware update for a board that is not selected

We should only flash the selected board

### Prerequisites

patrickelectric:
The board combobox in pirate mode lets us flash a board that is not the selected one.
Fix #3324

old

image

new

image

### DIFF
--- a/core/frontend/src/components/autopilot/FirmwareManager.vue
+++ b/core/frontend/src/components/autopilot/FirmwareManager.vue
@@ -107,11 +107,10 @@
       </div>
       <div id="update-options">
         <v-select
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):

patrickelectric:
can't cherry-pick this onto 1.4-dev, FirmwareManager still uses the pirate-mode board select.

Issue #2755, opened 2024-06-26
MAVLink Distance option was turned off


patrickelectric:
The user has a BlueOS heavy and a Ping1D.

patrickelectric:
Update to 1.3.0-beta.6

patrickelectric:
https://discord.com/channels/1135646343776456765/1135646344569168027/1255638972097499188

joaomariolago:
#4353 Helps with this issue but is not the complete solution.

PR #4353, opened 2026-08-29
core: frontend: ping1d: Start MAVLink Distances off

patrickelectric:
The switch defaulted on before the backend status arrived.
Helps #2755

### DIFF
--- a/core/frontend/src/components/ping/ping1d.vue
+++ b/core/frontend/src/components/ping/ping1d.vue
@@ -79,7 +79,7 @@ export default Vue.extend({
   data() {
     return {
       expand: false,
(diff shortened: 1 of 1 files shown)


joaomariolago (commented):
The changes look fine, but I'm not sure they address #2755. The user mentioned that the switch turned off after updating, which gives the impression that it was enabled before the update and was disabled by some other mechanism, rather than this being simply a frontend issue.

We can merge this change, but I don't think we should mark #2755 as resolved.

patrickelectric:
@joaomariolago changed to helps

joaomariolago (approved):

PR #4359, opened 2026-08-31
core: frontend: ParamSets: Remount loader when selecting a paramset

### LINKED ISSUE #3136: loading parameter sets on configure->parameters multiple times can cause issues
The UI scrolls up but the parameters are not applied, plus we get this beauty:

Image

patrickelectric:
v-if on the object is always true. leftover writing/error survive the next click.

Fix #3136


joaoantoniocardoso (approved):
tested

PR #4365, opened 2026-09-02
Allow users to change the system ID in the Vehicle Setup component

### LINKED ISSUE #3604: Unified way to change vehicle ID
As a follow-up from #3591, we need a way for the users to change the vehicle ID (`MAV_SYSTEM_ID` in BlueOS) from the UI.

Since we are using an environment variable, to propagate it throughout services and extensions, a reboot of all containers will be required.

nukelet:
We would like to provide a unified way for users to configure the system ID for a vehicle. As it currently stands, there are two sources of truth for the vehicle ID that need to be kept in sync:

- The `MAV_SYSID` parameter in the Ardupilot firmware
- The `MAV_SYSTEM_ID` environment variable, which is used to determine the vehicle ID in `mavlink-server`/`mavlink2rest`

This PR aims to provide an interface in the UI for users to change the vehicle ID:

- Introduce a new `autopilot_manager` endpoint, `/system_id`, to set the value for `MAV_SYSTEM_ID` through a POST request, making the change persistent across reboots by storing the variable in the `environment` field of `bootstrap/startup.json`
- Add a simple component to the Vehicle Setup tab that allows users to modify the Vehicle ID to their desired value

Note that (currently) it is necessary to restart the BlueOS container in order to have the `MAV_SYSTEM_ID` environment variable propagated to all services (there is a button in the Vehicle ID setup UI for restarting the container).

Depends on #4364
Fixes #3604

### DIFF
--- a/core/frontend/src/components/vehiclesetup/overview/SystemId.vue
+++ b/core/frontend/src/components/vehiclesetup/overview/SystemId.vue
@@ -1,0 +1,121 @@
+<template>
+  <v-card class="pa-1" elevation="2">
+    <v-card-title class="d-flex align-center">
+      Configure MAVLink system ID
+    </v-card-title>
+    <v-card-text>
+      Note that it is necessary to restart the BlueOS core container in order for the vehicle ID change to fully
+      take place.
+    </v-card-text>
+    <v-card-text>
+      <inline-parameter-editor
+        v-if="mav_sysid"
+        :param="mav_sysid"
+        :label="'MAVLink system ID'"
+        @change="pending_value = $event"
+        @form-valid-change="is_form_valid = $event"
+      />
+      <v-alert
+        v-if="save_error"
+        type="error"
+        dense
+        text
+        class="mt-3 mb-0"
+      >
+        Failed to set system ID: {{ save_error }}
+      </v-alert>
+    </v-card-text>
+    <v-card-actions class="justify-end">
+      <v-btn
+        v-tooltip="'Restart the BlueOS core container'"
+        :loading="restarting_core"
+        :disabled="restarting_core || saving"
+        @click="restartCore"
+      >
+        <v-icon left color="orange">
+          mdi-folder-refresh
+        </v-icon>
+        Restart Core
+      </v-btn>
+      <v-btn
+        color="primary"
+        :loading="saving"
+        :disabled="!can_save || saving"
+        @click="save"
+      >
+        Save
+      </v-btn>
+    </v-card-actions>
+  </v-card>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+
+import InlineParameterEditor from '@/components/parameter-editor/InlineParameterEditor.vue'
+import Notifier from '@/libs/notifier'
+import autopilot from '@/store/autopilot'
+import Parameter from '@/types/autopilot/parameter'
+import { autopilot_service } from '@/types/frontend_services'
+import back_axios from '@/utils/api'
+
+const notifier = new Notifier(autopilot_service)
+
+export default Vue.extend({
+  name: 'SystemId',
+  components: {
+    InlineParameterEditor,
+  },
+  data() {
+    return {
+      pending_value: undefined as number | undefined,
+      is_form_valid: true,
+      restarting_core: false,
+      saving: false,
+      save_error: undefined as string | undefined,
+    }
+  },
+  computed: {
+    mav_sysid(): Parameter | undefined {
+      return autopilot.parameter('MAV_SYSID')
+    },
+    has_pending_change(): boolean {
+      return this.mav_sysid !== undefined
+        && this.pending_value !== undefined
+        && this.pending_value !== this.mav_sysid.value
+    },
+    can_save(): boolean {
+      return this.has_pending_change && this.is_form_valid
+    },
+  },
+  methods: {
+    async save(): Promise<void> {
+      if (!this.mav_sysid || this.pending_value === undefined) {
+        return
+      }
+
+      this.save_error = undefined
+      this.saving = true
+      await back_axios({
+        method: 'post',
+        url: `/autopilot-manager/v1.0/system_id?value=${this.pending_value}`,
+      }).catch((error) => {
+        this.save_error = error.response?.data?.detail ?? error.response?.data ?? error.message
+        notifier.pushBackError('SET_MAV_SYSTEM_ID', error)
+      }).finally(() => {
+        this.saving = false
+      })
+    },
+    async restartCore(): Promise<void> {
+      this.restarting_core = true
+      await back_axios({
+        method: 'post',
+        url: '/version-chooser/v1.0/version/restart',
+      }).finally(() => {
+        // Give the backend a bit to go down, then reload so the user reconnects to the fresh core
+        setTimeout(() => window.location.reload(), 15000)
+      })
+    },
+  },
+})
+</script>
--- a/core/frontend/src/components/vehiclesetup/overview/SystemId.vue
+++ b/core/frontend/src/components/vehiclesetup/overview/SystemId.vue
@@ -28,8 +28,7 @@
     <v-card-actions class="justify-end">
       <v-btn
         v-tooltip="'Restart the BlueOS core container'"
-        :loading="restarting_core"
-        :disabled="restarting_core || saving"
+        :disabled="restarting || saving"
         @click="restartCore"
       >
         <v-icon left color="orange">
@@ -40,12 +39,18 @@
       <v-btn
         color="primary"
         :loading="saving"
-        :disabled="!can_save || saving"
+        :disabled="!can_save || saving || restarting"
         @click="save"
       >
         Save
       </v-btn>
     </v-card-actions>
+    <v-container v-if="restarting">
+      <p class="text-md-center">
+        Core container is restarting, please wait.
+      </p>
+      <spinning-logo size="20%" />
+    </v-container>
   </v-card>
 </template>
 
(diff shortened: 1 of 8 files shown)


joaoantoniocardoso:
Cool!

I didn't test it yet, but I have one point: if we move the parameter set part to the backend, we allow non-UI clients to complete the full change-mavlink-id journey. If so, I'd adjust the endpoint name to be more generic.

Thanks

nukelet:
Yeah, that's a good idea, I'll move the `PARAM_SET` bits to the backend and rename the endpoint to something like `/system_id`. I'll also address some of the comments from the bot review. Thanks for the suggestion!

core/services/ardupilot_manager/api/v1/routers/index.py
```diff
@@ -294,3 +295,34 @@ async def restore_default_firmware(board_name: Optional[str] = None) -> Any:
...
+
+            # make sure to remove MAV_SYSTEM_ID if it is already defined in
+            # bootstrap/startup.json
+            environment = [v for v in environment if not v.startswith("MAV_SYSTEM_ID=")]
+            environment.append(f"MAV_SYSTEM_ID={value}")
+            settings["core"]["environment"] = environment
+
+            startup_settings.seek(0)
+            startup_settings.write(json.dumps(settings, indent=2))
+            startup_settings.truncate()
```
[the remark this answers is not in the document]
nukelet: i disagree with the bot here, if we fail to write to `startup.json` then we have bigger problems to worry about. it's definitely not the weakest link here

core/services/ardupilot_manager/api/v1/routers/index.py
```diff
@@ -294,3 +295,34 @@ async def restore_default_firmware(board_name: Optional[str] = None) -> Any:
 @index_to_http_exception
 async def available_boards() -> Any:
     return await autopilot.available_boards(True)
+
+
+@index_router_v1.post("/system_id", summary="Set the vehicle's system id.")
```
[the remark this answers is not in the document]
nukelet: updated the PR description

nukelet:
I moved the system ID change operation entirely into the backend and renamed the endpoint to `/system_id`. You can poke at it with curl:

```bash
curl -X POST http://<blueos-addr>/autopilot-manager/v1.0/system_id\?value\=42 -v
```

This was a bit more painful to figure out than I anticipated since there isn't a clean way for the onboard computer to know that the `MAV_SYSID` parameter was updated successfully in the firmware. The frontend (system id 255) gets to hear about it immediately because Ardupilot will (following the MAVLink spec) acknowledge a `PARAM_SET` operation by broadcasting a `PARAM_VALUE` message containing the new value of the updated parameter; however, since the onboard computer (which emits the `PARAM_SET`) has the same system id as Ardupilot (who broadcasts the `PARAM_VALUE`) and mavlink-server doesn't allow loopback messages, BlueOS never gets the `PARAM_VALUE` confirmation (this took me an embarassing amount of time to figure out).

core/libs/commonwealth/src/commonwealth/mavlink_comm/VehicleManager.py
```diff
@@ -137,3 +154,34 @@ async def vehicle_is_safe(self) -> bool:
...
+        # [1] https://mavlink.io/en/messages/common.html#PARAM_SET
+
+        # hack around this limitation by waiting for self.mavlink2rest to
+        # timeout and trigger system ID detection, then set the new target
+        # system value
+        start_time = time.time()
+        timeout = 30.0
+        while time.time() - start_time < timeout:
+            if await self.is_heart_beating():
+                continue
```
patrickelectric, line 182: not sure if I follow, isn't this the same than waiting 30s ? also.. would be nice to put the reference where in the system reflects in this 30s value. 

Image

core/services/ardupilot_manager/api/v1/routers/index.py
```diff
@@ -294,3 +295,34 @@ async def restore_default_firmware(board_name: Optional[str] = None) -> Any:
...
+@index_to_http_exception
+async def set_system_id(value: int = Query(ge=1, le=255)) -> Any:
+    if not autopilot.current_board:
+        message = "No board running, system ID is unavailable"
+        logger.warning(message)
+        return PlainTextResponse(message, status_code=503)
+
+    try:
```
patrickelectric, line 308: IIRC try here is unnecessary since you are using the index_to_http_exception decorator

core/frontend/src/components/vehiclesetup/overview/SystemId.vue
```diff
@@ -0,0 +1,121 @@
...
+      })
+    },
+    async restartCore(): Promise<void> {
+      this.restarting_core = true
+      await back_axios({
+        method: 'post',
+        url: '/version-chooser/v1.0/version/restart',
+      }).finally(() => {
+        // Give the backend a bit to go down, then reload so the user reconnects to the fresh core
+        setTimeout(() => window.location.reload(), 15000)
```
patrickelectric, line 116: we should not guess, take a look in PowerMenu: waitForBackendToBeOnline

core/services/ardupilot_manager/api/v1/routers/index.py
```diff
@@ -294,3 +295,34 @@ async def restore_default_firmware(board_name: Optional[str] = None) -> Any:
...
+        await autopilot.vehicle_manager.set_system_id(value)
+
+        if autopilot.vehicle_manager.target_system != value:
```
patrickelectric, line 311: wait, isn't this supposed to be done already in set_system_id ?

core/services/ardupilot_manager/api/v1/routers/index.py
```diff
@@ -294,3 +295,34 @@ async def restore_default_firmware(board_name: Optional[str] = None) -> Any:
...
+            return PlainTextResponse("Failed to set system ID", status_code=500)
+
+        with open(autopilot.settings.startup_settings_file, "r+", encoding="utf-8") as startup_settings:
```
patrickelectric, line 314: it would be better to have it encapsulated by a class or something, maybe the settings itself, over opening and manipulating a settings file inside an api endpoint.

patrickelectric: also.. maybe not even in autopilot.settings, but maybe autopilot.bootstrap.settings, with that you can create the abstraction in autopilot.bootstrap; also, if we are going to have shared logic of bootstrap here and with other services, does it worth to have a commonwealth place for it ?

patrickelectric (changes requested):
It's in a good direction, let's continue

core/libs/commonwealth/src/commonwealth/mavlink_comm/VehicleManager.py
nukelet: yes, this should have been a `break` instead of `continue`. definitely needed more coffee when wrapping up these changes

nukelet:
@patrickelectric, I believe I addressed all the points from your review. Notable changes:

- Frontend now shows a spinner + waits for the backend to come back online instead of guessing a timeout
- Refactored the cursed validation logic to confirm the system id change by waiting for `get_most_recent_vehicle_id()` to match the value we attempted to configure. This check is (obviously) much faster now, taking only 2 or 3 seconds
- Created a small `EnvironmentManager` class in `commonwealth/utils` with only a `set_variable` method (doing essentially what was being done in the `/system_id` endpoint before, e.g. writing the variable to `bootstrap/startup.json`). I'm keeping it small for now so as to satisfy only the needs of its single user (`VehicleManager.set_system_id`)

nukelet:
btw, I left a bunch of FIXUP commits to make it easier to iterate on this round of chagnes, but I intend to squash them later

patrickelectric (commented):
The PR looks much better. I did a review on the surface, will do a better one later today.

core/libs/commonwealth/src/commonwealth/mavlink_comm/VehicleManager.py
```diff
@@ -137,3 +155,43 @@ async def vehicle_is_safe(self) -> bool:
...
+
+        # hack around this limitation by listening to the vehicle heartbeat
+        # messages and waiting for the source system id to change to the value
+        # we just configured. ardupilot sends heartbeats at 1Hz so 10s is a
+        # reasonable timeout
+        start_time = time.time()
+        timeout = 10.0
+        while time.time() - start_time < timeout:
+            if await self.mavlink2rest.get_most_recent_vehicle_id() == value:
+                self.set_target_system(value)
```
patrickelectric, line 187: just to be sure.. we did that in the start of the function, why do we need to do that again ?

patrickelectric: Also, if the target_system is defined in the start of the function, and the vehicle did not update the system_id, the later if will check the previously defined value and not the new one.

patrickelectric (changes requested):

core/libs/commonwealth/src/commonwealth/mavlink_comm/VehicleManager.py
nukelet: > we did that in the start of the function, why do we need to do that again ?

we do it at the start because `self.target_system` is initially set to 1, so if the vehicle already has a system id other than 1 we will end up sending the `PARAM_SET` message to the wrong target system unless we update `self.target_system` first.

changing it inside the loop on the other hand means we confirmed the system id change was successful and that we can go ahead and update `self.target_system`

> Also, if the target_system is defined in the start of the function, and the vehicle did not update the system_id, the later if will check the previously defined value and not the new one.

not sure i follow, the point is precisely that if we didn't update `self.target_system` within the `while` loop then the autopilot firmware didn't change `MAV_SYSID` at all?

patrickelectric: brain lapse

patrickelectric (approved):

rmackay9:
Thanks very much for this!

PR #4378, opened 2026-09-03
[1.4] do not try to fetch remote tags if there is no internet

### LINKED ISSUE #1788: version chooser: we shouldn't try to fetch remote tags if there is no internet

patrickelectric:
Unreachable DockerHub used to hang on aiohttp's 5-minute default timeout.

Cherry-pick of #4331
Fix #1788


joaoantoniocardoso (approved):
tested

PR #4390, opened 2026-09-03
core: frontend: App: Add battery information widget

### LINKED ISSUE #4389: Add heartbeat information as header widgets in BlueOS interface
Add cpu temperature, battery current and battery voltage.

patrickelectric:
Battery voltage and current as a top bar widget.
image

Fix #4389

### DIFF
--- a/core/frontend/src/widgets/Cpu.vue
+++ b/core/frontend/src/widgets/Cpu.vue
@@ -65,13 +65,8 @@ export default Vue.extend({
       return cpus.map((cpu) => cpu.frequency)
     },
     cpu_temperature(): string {
-      const temperature_sensors = system_information.system?.temperature
-      const main_sensor = temperature_sensors?.find(
-        (sensor) => sensor.name.toLowerCase().includes('cpu')
-          || sensor.name.toLowerCase().includes('soc')
-          || sensor.name.toLowerCase().includes('ccd'),
-      )
-      return main_sensor ? main_sensor.temperature.toFixed(1) : 'Loading..'
+      const temperature = system_information.cpu_temperature
+      return temperature === undefined ? 'Loading..' : temperature.toFixed(1)
     },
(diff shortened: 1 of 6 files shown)


ES-Alexander (commented):
1. Current usage is also a battery property, right? Can we just vertically stack the voltage + current values (like they are in the Cockpit battery widget), instead of having a separate widget for current? Would save some space and make the widget look less weird.
2. Given CPU temperature is already included in the CPU widget, do we need a new widget just to pull it out by itself? That seems unnecessary 🤷‍♂️ 
    - Note that there's already a warning icon that appears at the right side of the header bar if the RPi is throttling from CPU temperature, so there is already dedicated space for it when it's causing problems

patrickelectric:
> Can we just vertically stack the voltage + current values

done

> do we need a new widget just to pull it out by itself?

no

core/frontend/src/widgets/Cpu.vue
```diff
@@ -65,13 +65,8 @@ export default Vue.extend({
...
     cpu_temperature(): string {
-      const temperature_sensors = system_information.system?.temperature
-      const main_sensor = temperature_sensors?.find(
-        (sensor) => sensor.name.toLowerCase().includes('cpu')
-          || sensor.name.toLowerCase().includes('soc')
-          || sensor.name.toLowerCase().includes('ccd'),
-      )
-      return main_sensor ? main_sensor.temperature.toFixed(1) : 'Loading..'
+      const temperature = system_information.cpu_temperature
+      return temperature === undefined ? 'Loading..' : temperature.toFixed(1)
```
joaoantoniocardoso, line 69: BTW, isn't the "Loading..." text too big for the 1~3-digit space we have?

patrickelectric: its fine 
image

joaoantoniocardoso (approved):

PR #4414, opened 2026-09-04
Warn user when DNS is not working

### LINKED ISSUE #690: Core: Version-chooser: Temporary failure in name resolution
We're seeing this somewhat randomly. Last report was just after a fresh install.

nicoschmdt:
fix: #690 

web icon changes to dns and has a tooltip informing the user that a DNS problem was identified. if the user opens it the DNS page will be the main one.

image


core/services/helper/main.py
```diff
@@ -140,6 +146,7 @@ class SimpleHttpResponse(BaseModel):
     as_json: Optional[Union[List[Any], Dict[Any, Any]]]
     error: Optional[str]
     timeout: bool
+    dns_error: bool = False
```
patrickelectric: Well, since we have the WebsiteError, why not use it here over multiple binaries ? Since they are exclusive

core/frontend/src/types/helper.ts
```diff
@@ -24,6 +24,13 @@ export interface Service {
     metadata?: ServiceMetadata
 }
 
+// Mirrors the WebsiteError enum from
+// core/services/helper/main.py
```
patrickelectric: I think that this comments is unnecessary since the name of this file is `types/helper.ts` and this is how all the other files work.

core/frontend/src/components/app/InternetTrayMenu.vue
```diff
@@ -56,7 +57,13 @@ export default Vue.extend({
     show_menu: false,
   }),
   computed: {
+    dns_failure(): boolean {
+      return helper.dns_failure
```
patrickelectric: why we have this if everything else of this file uses helper directly ?

core/frontend/src/components/app/NetworkInterfaceMenu.vue
```diff
@@ -47,15 +47,27 @@ export default Vue.extend({
...
+  props: {
+    initialPage: {
+      type: String,
+      default: '',
+    },
+  },
   data() {
     return {
-      page_selected: null as string | null,
+      page_selected: null as number | null,
```
patrickelectric: I think this is wrong:
https://v2.vuetifyjs.com/en/api/v-tabs/#props

the `pages` value is string.

patrickelectric: The problem is a missing `:value="page.value"` in v-tab-for, with this page_selected will be a string and the mounted logic is unnecessary.

core/frontend/src/components/app/InternetTrayMenu.vue
```diff
@@ -71,6 +78,9 @@ export default Vue.extend({
       }
     },
     icon(): string {
+      if (this.dns_failure) {
+        return 'mdi-dns'
```
patrickelectric, line 82: This icon by itself is not as a "warning" compared with `mdi-web-off` and `mdi-web-minus`, it may be better for us to have a 'warning' color in the icon itself.

patrickelectric (changes requested):

patrickelectric (approved):

PR #4430, opened 2026-09-11
Block merging PRs with fixup tags

patrickelectric:


rafaellehmkuhl:
If you want to take a look we have added [a similar but broader solution](https://github.com/rafaellehmkuhl/cockpit-public/blob/master/.github/workflows/commit-finality.yml) in Cockpit.

joaoantoniocardoso (approved):

PR #4431, opened 2026-09-11
core: services: ardupilot_manager: Migrate settings to commonwealth's PydanticManager

### LINKED ISSUE #3901: Autopilot manager is not using the settings backend

EpicEric:
Closes #3901

### DIFF
--- a/core/libs/commonwealth/src/commonwealth/settings/managers/pydantic_manager.py
+++ b/core/libs/commonwealth/src/commonwealth/settings/managers/pydantic_manager.py
@@ -1,19 +1,21 @@
 import pathlib
 import re
-from typing import Any, Optional, Type
+from typing import Any, Generic, Optional, TypeVar
 
 import appdirs
 from commonwealth.settings.bases.pydantic_base import PydanticSettings, _save_lock
 from loguru import logger
 
+T = TypeVar("T", bound=PydanticSettings)
 
-class PydanticManager:
+
+class PydanticManager(Generic[T]):
     SETTINGS_NAME_PREFIX = "settings-"
 
     def __init__(
         self,
         project_name: str,
-        settings_type: Type[PydanticSettings],
+        settings_type: type[T],
         config_folder: Optional[pathlib.Path] = None,
         load: bool = True,
     ) -> None:
@@ -28,7 +30,7 @@ class PydanticManager:
         )
         self.config_folder.mkdir(parents=True, exist_ok=True)
         self.settings_type = settings_type
-        self._settings: Optional[PydanticSettings] = None
+        self._settings: Optional[T] = None
         logger.debug(
             f"Starting {project_name} settings with {settings_type.__name__}, configuration path: {config_folder}"
         )
@@ -36,7 +38,7 @@ class PydanticManager:
             self.load()
 
     @property
-    def settings(self) -> Any:
+    def settings(self) -> T:
         """Getter point for settings
 
         Returns:
@@ -44,11 +46,13 @@ class PydanticManager:
         """
         if not self._settings:
             self.load()
+        if not self._settings:
+            raise RuntimeError("PydanticManager.load() failed")
 
         return self._settings
 
     @settings.setter
-    def settings(self, value: Any) -> None:
+    def settings(self, value: T) -> None:
         """Setter point for settings. Save settings for every change
 
         Args:
@@ -73,7 +77,7 @@ class PydanticManager:
         return self.config_folder.joinpath(f"{PydanticManager.SETTINGS_NAME_PREFIX}{version}.json")
 
     @staticmethod
-    def load_from_file(settings_type: Type[PydanticSettings], file_path: pathlib.Path) -> Any:
+    def load_from_file(settings_type: type[T], file_path: pathlib.Path) -> Any:
         """Load settings from a generic location and settings type
 
         Args:
--- a/core/libs/commonwealth/src/commonwealth/settings/tests/test_manager_pydantic.py
+++ b/core/libs/commonwealth/src/commonwealth/settings/tests/test_manager_pydantic.py
@@ -69,64 +69,64 @@ def test_basic_settings_save_load() -> None:
     config_path = pathlib.Path(temporary_folder)
 
     # Test v1 save
-    settings_manager = PydanticManager("ManagerTest", SettingsV1, config_path)
-    settings_manager.settings.version_1_variable = 2022
-    settings_manager.save()
+    settings_manager_v1 = PydanticManager("ManagerTest", SettingsV1, config_path)
+    settings_manager_v1.settings.version_1_variable = 2022
+    settings_manager_v1.save()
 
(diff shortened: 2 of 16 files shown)


EpicEric:
The only thing I haven't tested yet is if serial interfaces are properly migrated with the rest of the settings. I intend to test it once I have the hardware to do so.

EpicEric:
Tested in a VM with SITL and fake UDP serials, the settings get propagated correctly from `settings.json` to `settings-1.json`.

core/libs/commonwealth/src/commonwealth/settings/managers/pydantic_manager.py
```diff
@@ -28,27 +30,29 @@ def __init__(
...
-    def settings(self) -> Any:
+    def settings(self) -> T:
         """Getter point for settings
 
         Returns:
             [Type[PydanticSettings]]: The settings defined in the constructor
         """
         if not self._settings:
             self.load()
+        if not self._settings:
```
patrickelectric: There are two different things happening in this commit, the type T change and also the re-check of settings, they should be splitted in different commits.

core/libs/commonwealth/src/commonwealth/settings/tests/test_manager_pydantic.py
```diff
@@ -69,64 +69,64 @@ def test_basic_settings_save_load() -> None:
     config_path = pathlib.Path(temporary_folder)
 
     # Test v1 save
-    settings_manager = PydanticManager("ManagerTest", SettingsV1, config_path)
-    settings_manager.settings.version_1_variable = 2022
-    settings_manager.save()
+    settings_manager_v1 = PydanticManager("ManagerTest", SettingsV1, config_path)
```
patrickelectric, line 72: this changes are also unrelated to the commit.

core/services/beacon/main.py
```diff
@@ -139,7 +139,7 @@ def get_vehicle_name(self) -> str:
         return self.manager.settings.vehicle_name or "BlueROV2"
 
     def create_async_service_infos(
-        self, interface: str, service_name: str, domain_name: str, ip: str
+        self, interface: Interface, service_name: str, domain_name: str, ip: str
```
patrickelectric, line 142: wait, how CI was not catching this ?

core/services/ardupilot_manager/api/v1/routers/index.py
```diff
@@ -199,7 +198,7 @@ async def install_firmware_from_file(
     parameters: Optional[Parameters] = None,
 ) -> Any:
     try:
-        custom_firmware = Path.joinpath(autopilot.settings.firmware_folder, "custom_firmware")
+        custom_firmware = autopilot.settings.firmware_folder / "custom_firmware"
```
patrickelectric: There is a bunch of things happening at once in this commit.

core/services/ardupilot_manager/autopilot_manager.py
```diff
@@ -259,6 +252,10 @@ async def auto_restart_ardupilot(self) -> None:
     async def start_mavlink_manager_watchdog(self) -> None:
         await self.mavlink_manager.auto_restart_router()
 
+    @property
+    def settings(self) -> SettingsV1:
```
patrickelectric, line 256: Just to be sure, who is calling this function ?

core/services/ardupilot_manager/autopilot_manager.py
```diff
@@ -300,19 +297,12 @@ def get_serials(self) -> List[Serial]:
...
-                return []
+        serials = self._settings_manager.settings.serials
+        if serials:
+            return serials
+        if isinstance(self._current_board, LinuxFlightController):
             return self._current_board.get_serials()
-        serials = []
-        for entry in self.configuration["serials"]:
-            try:
-                serials.append(Serial(port=entry["port"], endpoint=entry["endpoint"]))
```
patrickelectric, line 311: we can remove the `Serial` import

core/services/ardupilot_manager/settings.py
```diff
@@ -1,116 +1,121 @@
...
+from pydantic import field_validator
+from typedefs import FlightController, Serial, SITLFrame
+from typing_extensions import override
+
+
+# pylint: disable=too-many-branches
+def migrate_from_old_settings(target_settings: "SettingsV1") -> None:
+    path = Path(appdirs.user_config_dir(SERVICE_NAME)) / "settings.json"
+    if not path.exists():
+        return
```
patrickelectric: We need a warning/error message here

core/services/ardupilot_manager/settings.py
```diff
@@ -1,116 +1,121 @@
...
-    defaults_folder = Path.joinpath(blueos_files_folder, "ardupilot-manager/default")
-    sitl_frame = SITLFrame.UNDEFINED
-    preferred_router: Optional[str] = None
+        data["VERSION"] = SettingsV1.STATIC_VERSION
 
-    def __init__(self) -> None:
-        self.root: Dict[str, Union[int, Dict[str, Any]]] = {"version": 0, "content": {}}
+    @override
+    def on_settings_created(self, _: Path) -> None:
+        migrate_from_old_settings(self)
```
patrickelectric, line 111: This will probably not work since `on_settings_created` is only called when the path does not exist, and from what I could understand, the settings path is the same between the old settings and the new one. Am I wrong ?

```py
if not file_path.exists():
    # We call this to allow users to initialize the settings from another source if needed
    self.on_settings_created(file_path)
```

patrickelectric (changes requested):
good cleanup

core/services/ardupilot_manager/settings.py
EpicEric: The path changes, the previous one is `settings.json` and the new one is `settings-1.json` (as per the PydanticManager code)

core/services/ardupilot_manager/settings.py
EpicEric: I'll use a warning, since the old settings would be missing if the user's first version is 1.5.0

core/services/ardupilot_manager/autopilot_manager.py
EpicEric: The type is still used on `AutoPilotManager.update_serials()`

core/services/ardupilot_manager/autopilot_manager.py
EpicEric: I can't find any callers, removing

core/services/beacon/main.py
EpicEric: Since the settings weren't generic, the types were being casted - sometimes to the wrong type, as was the case for Interface. The interface is only used here for logging, so the typechecker didn't complain about str.

core/services/ardupilot_manager/autopilot_manager.py
EpicEric: Thanks to mypy, I realized that this is called by the install_firmware_from_file route

Issue #4434, opened 2026-09-12
Improve frontend performance


patrickelectric:
Hear me out, BlueOS frontend works without performance issue, but sometimes I have performance issues!!!

<img width="869" height="768" alt="Image" src="https://github.com/user-attachments/assets/6f9900bd-c922-49a7-ba92-7f0655e81a70" />

So, I did a test for all pages, 5 min each.

- MAVLink Inspector is 40.34% of one core while idle
  - 60 fps redraw with nothing changing (Fixed in https://github.com/bluerobotics/mavlink-server/pull/231)
  - iframe to mavlink-server, FireAnimationFrame
- Vehicle Setup is 16.30%
  - `/mavlink-camera-manager/v4l` at delay 1000, 455 KB per response, 1.53 GB/hour
  - Video Streams uses delay 5000 for the same endpoint and is 6.55%
- Zenoh Inspector is 11.09%
  - `declareSubscriber('**')` on mount, 1140 msg/s
  - peak 61.5%
- `/system-information/platform` grows with uptime
  - 706 KB to 13.2 MB per 5 min window in under 2 hours
  - raspberry.events.list never trims
  - 85-91% of bytes on 18 of 22 pages
- Network Test heap peaks at 159.2 MB
  - 100 MB ArrayBuffer allocated on mount, no test started
- OneMoreTime pollers leak
  - `data: () => ({ disposeWith: this })` in Pings.vue and MainView.vue
  - after a second visit to `/vehicle/pings`, `/system-information/serial` doubled from 12.4 to 23.8 req/min
- NMEA Injector is 3.53% (shell only)
  - 14 of 22 pages stay within 2x of that

Tested in: [heads/master-0-g65b2affb](https://github.com/bluerobotics/BlueOS/tree/65b2affb)

Here are some data:

<img width="1863" height="1420" alt="Image" src="https://github.com/user-attachments/assets/4aaff222-cfea-473a-bff9-8215edcbd6e9" />
<img width="1863" height="1420" alt="Image" src="https://github.com/user-attachments/assets/782a1454-8cb6-44e8-be4b-7d1cfd175c04" />
<img width="1862" height="1420" alt="Image" src="https://github.com/user-attachments/assets/34f54784-ad94-49fc-b6ec-0fa0a85aed2d" />
<img width="1863" height="1420" alt="Image" src="https://github.com/user-attachments/assets/6f6ac7c7-2f31-4704-aed9-5117b80d9348" />
[...]

PR #4435, opened 2026-09-12
core: tools: mavlink_server: Update to 0.11.0

patrickelectric:
[changelog](https://github.com/bluerobotics/mavlink-server/releases/tag/0.11.0)

Helps #4434


joaoantoniocardoso (approved):

joaoantoniocardoso:
We should probably update MS on next stable as well.

PR #4453, opened 2026-09-14
core: frontend: AboutThisSystem: Show host OS name

### LINKED ISSUE #4452: bug: /tools/system-information reports the wrong OS on NixOS
### Bug description

When running BlueOS on NixOS, the "About" tab of the "System Information" incorrectly reports "Debian GNU/Linux 12" as the operating system. 

Image

### Steps to reproduce

1. Install blueos-core on a NixOS machine (in my case, a virtual machine)
2. Access the System Information page and access the About tab

### Primary pain point(s)

Very minor visual bug.

### Additional context

_No response_

### Prerequisites

patrickelectric:
linux2rest reads the container `/etc/os-release`, so About showed Debian on NixOS hosts.

Host pretty name comes from the host's `/etc/os-release` (not the container): prefer a local host path (`/etc/os-release.host` if mounted, or `/proc/1/root/etc/os-release` when the PID namespace is shared and the content differs from the container), then fall back to SSH-backed `load_file`.

Fix #4452


patrickelectric:
It appears that some versions of myself are out of control.

PR #4462, opened 2026-09-15
core: services: ping: asyncify serial interface calls

### LINKED ISSUE #4458: Ping service does not answer when probing baudrates
When the service starts and checks the device's communication baud rate, it fails to respond to endpoint requests.

EpicEric:
Closes #4458


core/services/ping/pingdriver.py
```diff
@@ -65,11 +78,8 @@ async def start(self) -> None:
         if self.port is None:
             raise NoUDPPortAssignedToPingDriver("PingDriver attempted to stash with no UDP port.")
 
-        self.baud = self.detect_highest_baud()
-        # Do a ping connection to set the baudrate
-        PingDevice().connect_serial(self.ping.port.device, self.baud)
-        set_low_latency(self.ping.port)
```
patrickelectric: since we are not using `port` anymore here in this functions, and the checks moved to _open_bridge, we should remove the checks from start.

core/services/ping/pingdriver.py
```diff
@@ -57,6 +58,18 @@ def detect_highest_baud(self) -> Baudrate:
         logger.info(f"Highest baudrate detected: {last_valid_baud}")
         return last_valid_baud
 
+    def _open_bridge(self) -> Bridge:
+        if self.ping.port is None:
```
patrickelectric: Since the tests are now being done here and not in the start function, we should remove from `start`.

core/services/ping/pingdriver.py
```diff
@@ -65,11 +78,8 @@ async def start(self) -> None:
         if self.port is None:
             raise NoUDPPortAssignedToPingDriver("PingDriver attempted to stash with no UDP port.")
 
-        self.baud = self.detect_highest_baud()
-        # Do a ping connection to set the baudrate
-        PingDevice().connect_serial(self.ping.port.device, self.baud)
-        set_low_latency(self.ping.port)
-        self.bridge = Bridge(self.ping.port, self.baud, "0.0.0.0", 0, self.port, automatic_disconnect=False)
+        self.baud = await asyncio.to_thread(self.detect_highest_baud)
+        self.bridge = await asyncio.to_thread(self._open_bridge)
```
patrickelectric: can't we just have:

```py
    def open_bridge() -> Bridge:
        PingDevice().connect_serial(self.ping.port.device, self.baud)
        set_low_latency(self.ping.port)
        return Bridge(self.ping.port, self.baud, "0.0.0.0", 0, self.port, automatic_disconnect=False)

    self.bridge = await asyncio.to_thread(open_bridge)
```

the tests are already here in start and being duplicated. The function is also only being used here as well.

patrickelectric: This will also reduce the diff in ping1d_driver and logically will not change much

patrickelectric (changes requested):

core/services/ping/pingdriver.py
EpicEric: Mypy (rightfully) complains, as the inferred types don't encode the check that's been performed already. It might be the case where the principle of "parse, don't validate" would be a better fit.

core/services/ping/pingdriver.py
```diff
@@ -57,6 +64,11 @@ def detect_highest_baud(self) -> Baudrate:
         logger.info(f"Highest baudrate detected: {last_valid_baud}")
         return last_valid_baud
 
+    def open_bridge(self, config: BridgeConfig) -> Bridge:
```
patrickelectric: self is unnecessary here

core/services/ping/pingdriver.py
```diff
@@ -7,19 +9,27 @@
...
         self.ping = ping
         self.port = port
         self.bridge: Optional[Bridge] = None
+        self.bridge_config: Optional[BridgeConfig] = None
         self.ping.driver = self
-        self.baud: Optional[Baudrate] = None
         self.driver_status = DriverStatus(udp_port=port, mavlink_driver_enabled=False)
 
-    def detect_highest_baud(self) -> Baudrate:
+    def detect_highest_baud(self, port: SysFS) -> Baudrate:
```
patrickelectric: self is not being used here anymore

patrickelectric (changes requested):
The commits should also be squashed since the latest commit is modifying code from the first commit.

patrickelectric (approved):

PR #4467, opened 2026-09-16
core:frontend:InstalledExtensionCard: Add disable warning

joaomariolago:
* Warn users that some features may be unavailable if disabling  major tom extension

image


patrickelectric (approved):

patrickelectric:
@joaomariolago can you fix the frontend error ?

PR #4468, opened 2026-09-16
frontend: version: show/link to release notes/docs

### LINKED ISSUE #1645: core: frontend: version-chooser: show/link to release notes / docs
### Current behaviour

Version Chooser notifies when new versions are available, but doesn't specify what new features or bug fixes are included, so the main reason for updating is the idea of "staying up to date", which isn't very exciting.

It would be cool if updates could show the release notes/changelog between the current version and the new one being considered, or at least the new features in that particular release. Ideally this would be provided as a few major points with a link to the full release notes, but if that's not feasible it would still be nice to have the release notes either embedded or linked to.

It would also be nice if installed versions could include a link to the docs for that version. This would be extra accessible if the docs came embedded for local hosting in each version (or were made accessible via an extension with the same version numbering as BlueOS), but a hyperlink to the relevant online docs for the active version would still be an improvement.

### Prerequisites

nicoschmdt:
fix: #1645

Local versions:
> image

Remote versions:
> image

use the flag `-w` when checking the VersionCard file to ignore the indentation changes


patrickelectric (changes requested):
The render is different here.. also.. why are we dealing with markdown over using `marked` ?

Image

Also, can we just show documentation for semver + master ?

Image

nicoschmdt:
> The render is different here.. also.. why are we dealing with markdown over using `marked` ?

`marked` has deprecated it's sanitize function due to security reasons. 

The options were to add a dependency (DOMPurify), keep sanitizing by hand, or stop dealing with markdown at all. I went with the last one: we now ask GitHub for `body_html` and render that, which is the same sanitized HTML the release page serves, so `marked` is no longer used here at all.

Regarding the difference of render in the content: we're just getting what was published for the releases of the stables and betas. The content for `1.5.0-beta.20` is the same as we have in GitHub Releases, the difference it's because recent releases have more organized sections.

core/frontend/src/components/version-chooser/VersionCard.vue
```diff
@@ -1,133 +1,162 @@
...
+            class="text-caption text--secondary ma-0"
+            v-text="image.repository"
+          />
+        </div>
+      </div>
+      <v-spacer />
+      <div class="d-flex flex-wrap justify-center align-center my-2">
+        <v-btn
+          v-if="showDocsLink"
+          v-tooltip="'Documentation for this version'"
```
ES-Alexander:
```suggestion
          v-tooltip="'Change log for this version'"
```
Documentation is something else. It would be nice to link to that too if possible (e.g. `BlueOS 1.4.*` --> `https://blueos.cloud/docs/1.4/`, and `.../docs/latest/` for dev/beta releases), but maybe it's better/simpler for us to fix that upstream in the BlueOS release text instead of as a dynamic guess here.

core/frontend/src/components/version-chooser/VersionCard.vue
```diff
@@ -1,133 +1,162 @@
...
+          icon
+          :href="docsUrl"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          <v-icon>mdi-book-open-variant</v-icon>
```
ES-Alexander: I think `mdi-list-box`(`-outline`?) or `mdi-creation` works better, since this isn't a docs link.

We should also maybe have some text in the button, saying "What's new?"

core/frontend/src/utils/version_chooser.ts
```diff
@@ -129,6 +129,84 @@ function getLatestVersion(versions_query: VersionsQuery, current_version: Versio
...
+  }
+}
+
+/**
+ * Only the official BlueOS image is covered: forks have their own releases, and tags that are
+ * not versions (`master`, `factory`, branch builds) have no release to point at, which GitHub
+ * answers with a 404. Notes are cosmetic, so an unreachable GitHub reads as "no notes".
+ * Answers are shared between every card asking for the same version, and misses are forgotten
+ * so a vehicle that comes online later can still get the notes.
+ * @param tag - Docker tag of the image, e.g. `1.4.5` or `1.5.0-beta.42`
```
ES-Alexander, line 168: Ideally the 'full changelog' link can be replaced with a comparison link to the version the user is actually on, instead of just the release before the listed one. Understandably that only works if the user is on a SemVer tag, and would need the versions swapped around if the current version is newer than the one being considered.

Not a blocker, just a nice idea.

Even better would be something that can semantically combine multiple sets of release notes, or at least stack them (for relevant versions), but that's likely quite a bit more complex to implement.

core/frontend/src/components/version-chooser/VersionCard.vue
```diff
@@ -1,133 +1,162 @@
...
-        v-text="'Download and Apply'"
-      />
     </div>
+    <v-expansion-panels
+      v-if="showReleaseNotes"
+      flat
+    >
+      <v-expansion-panel>
+        <v-expansion-panel-header class="px-2">
+          What's new in {{ image.tag }}
```
ES-Alexander, line 149: I think this is better triggered as a "What's new?" button to the left of the the other buttons (except docs, which can be the leftmost button). Maybe also worth adding a preceding icon of `mdi-list-box`(`-outline`?) or `mdi-creation`?

Once expanded I think we should hide the "What's new?" button, and add an upwards arrow button at the far right of the buttons list, so it's like the whole list item is expanded / can be shrunk back down - e.g.

1. `(o) Release title!!             (docs) [What's New?] [Delete] [Apply]`
2. `(o) Release title!!                       (docs) [Delete] [Apply] (^)`
    `  Frontend ...`

and 

1. `(🌏) Release title!!        (docs) [What's New?] [Download and apply]`
2. `(🌏) Release title!!                  (docs) [Download and apply] (^)`
    `  Frontend ...`

core/frontend/src/components/version-chooser/ReleaseNotes.vue
```diff
@@ -0,0 +1,76 @@
...
+        class="release-notes"
+        v-html="notes.html"
+      />
+      <!-- eslint-enable vue/no-v-html -->
+      <a
+        :href="notes.url"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        Full release notes
```
ES-Alexander, line 22: This seems weird given the notes there will be exactly the same as in the BlueOS interface. Maybe "Release page" would work better, but best might be to just remove this line entirely?

ES-Alexander (changes requested):
Cool to see this getting worked on! :D

The documentation buttons don't seem to be appearing for remote versions.

Screenshot 2026-09-19 at 8 54 25 am

core/frontend/src/utils/version_chooser.ts
nicoschmdt: I do think this is a nice idea, but how should we handle huge changelogs? A user going from 1.4.* to 1.5.* would just show a huge block of text to them.

PR #4482, opened 2026-09-18
Fix the order and the scope of the board section in config.txt

### LINKED ISSUE #4477: config.txt: the board section is written in reverse order
On a fresh image, the `[pi4]` / `[pi5]` section of `config.txt` is upside down. The order we want is not the order the device gets.

The section starts with the `dwc2` line and ends with `enable_uart=1`, which is the exact opposite of what we ask for.

This is not only ugly. The order of these lines decides what the firmware does with them, so a reversed section can leave hardware unconfigured.

Reported by @rmackay9, whose `[pi5]` section is fully reversed: https://github.com/bluerobotics/BlueOS/issues/3804#issuecomment-5727472848

### LINKED ISSUE #4478: config.txt: a wrong board section is never repaired
BlueOS checks `config.txt` on every boot, but it only adds the lines that are missing. It never fixes the order of the lines that are already there.

So a device that shipped with a bad board section keeps that bad section forever. Updating BlueOS does not help. The user has to edit the file by hand.

Users also report that the file keeps changing on every reboot, so it never settles.

Reported by @rmackay9: "I found that something was changing the file on each reboot which was quite annoying." https://github.com/bluerobotics/BlueOS/issues/3804#issuecomment-5727472848

### LINKED ISSUE #4479: config.txt: dtparam lines are written after dtoverlay lines
Inside our board section, the `dtparam` lines come after the `dtoverlay` lines.

The firmware does not read a `dtparam` on its own. It first looks at the overlay that is in scope. Only when that overlay does not know the name does the setting reach the board itself. So a `dtparam` placed after a `dtoverlay` can be sent to the wrong place and do nothing.

Symptom: I2C or SPI does not work, or a camera is not detected, even though the line is present in the file. Moving the same line above the `dtoverlay` lines fixes it.

Reported by @rmackay9, who got his IMX477 camera working only after moving `dtparam=i2c_arm=on` up by hand: https://github.com/bluerobotics/BlueOS/issues/3804#issuecomment-5727472848

### LINKED ISSUE #4480: config.txt: our dtparam lines inherit the overlay scope of the distribution
A section header like `[pi5]` does not end an overlay scope. It only selects which board the lines apply to.

The Raspberry Pi OS image loads its own overlay near the top of `config.txt`. That overlay is still in scope when the firmware reaches our board section further down. So our `dtparam` lines are offered to that overlay first, and only reach the board if the overlay does not know the name.

This means our settings can land on the wrong target even when the order inside our own section is correct. Today it works by luck: the names we use do not clash with the names that overlay uses. Any change in the distribution image can break it.

Symptom: I2C, SPI or UART settings are present in the file but have no effect.

### LINKED ISSUE #4481: config.txt: an unknown board is treated as a Pi5
When BlueOS writes the USB gadget setting, it asks one question: is this a Pi4? If the answer is no, it assumes a Pi5.

So any board that is neither a Pi4 nor a Pi5 gets a `[pi5]` section. The firmware does not ignore a section it does not recognise, it applies it. Pi5 pin settings then reach a board they were never meant for.

Symptom: wrong pin behaviour on a board we do not support, with nothing in the logs to explain it.

This affects master. The 1.4 line is not affected.

joaoantoniocardoso:
One list now describes the board section, and both writers use it: the install script at image build, and the startup patch at every boot.

How each issue is solved:

- Closes #4477 — the install script inserts each line below the one before it, instead of all of them in the same place.
- Closes #4478 — the startup patch rewrites the whole section instead of only adding what is missing, so a bad section is repaired. Lines the user added are kept, and a line the user protected wins over ours.
- Closes #4479 — the list declares every `dtparam` before every `dtoverlay`.
- Closes #4480 — the section starts by ending the overlay scope, so our `dtparam` lines reach the board. We skip this when nothing above us opened a scope, to keep HAT detection working, and we never remove a scope end the user wrote. The Pi3 script gets the same line, so all boards are treated the same way.
- Closes #4481 — one map answers both "is this board supported" and "which section name", so an unknown board is refused instead of being taken for a Pi5.

Tested on three devices: Pi5 bookworm, Pi4 bullseye, Pi4 bookworm. On each one the section is repaired, the hardware is the same after reboot, and a second run changes nothing.

A backport to 1.4 is ready and will follow.


install/boards/bcm_2712.sh
```diff
@@ -55,22 +55,22 @@ echo "Line number of [pi5] tag: $line_number"
 
 
 for STRING in \
-    "enable_uart=1" \
+    "dtparam=i2c_arm=on" \
+    "dtparam=spi=on" \
```
patrickelectric, line 63: can we have a comment in the code explaning that the order is important ?

install/boards/bcm_27xx.sh
```diff
@@ -58,22 +58,22 @@ echo "Line number of [pi4] tag: $line_number"
 
 
 for STRING in \
-    "enable_uart=1" \
+    "dtparam=i2c_vc=on" \
```
patrickelectric, line 65: comment plos

patrickelectric: also.. bcm_28 ?

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -260,44 +260,44 @@ def navigator_managed_entries(cpu_type: CpuType) -> List[Tuple[str, str]]:
     # Keep in sync with install/boards/bcm_27xx.sh (Pi4) and bcm_2712.sh (Pi5).
     if cpu_type == CpuType.PI4:
         return [
-            ("enable_uart=1", "^enable_uart=.*"),
+            ("dtparam=i2c_vc=on", "^dtparam=i2c_vc=.*"),
+            ("dtparam=i2c_arm_baudrate=1000000", "^dtparam=i2c_arm_baudrate.*"),
+            ("dtparam=spi=on", "^dtparam=spi=.*"),
```
patrickelectric, line 264: a comment about the order would be nice.

patrickelectric (changes requested):

patrickelectric (approved):

PR #4486, opened 2026-09-18
Close the overlay scope after the board section in config.txt

### LINKED ISSUE #4484: config.txt: a dtparam written below our block goes to our USB overlay
Our block of Navigator lines ends with an overlay. On a Pi4 and a Pi5 that overlay is `dwc2`. On a Pi3 it is `uart1`.

An overlay stays in scope until a line closes it. A section header does not close it. So every `dtparam` line written below our block is offered to that overlay first, and only reaches the board if the overlay does not know the name.

Most names are unknown to the overlay, so most settings still work. A name the overlay does know is taken by the overlay, and the board never sees it.

Symptom on a Pi4: a user adds `dtparam=dr_mode=host` to `config.txt`. The line was meant for the board, but our USB overlay knows that name, so the overlay switches to host mode. The USB gadget disappears and the network over the USB cable stops working.

### LINKED ISSUE #4485: config.txt: the empty dtoverlay= line of the user is removed
An empty `dtoverlay=` line closes the overlay above it. A user writes that line to stop their own overlay from taking the `dtparam` lines that come after it.

We rewrite the board section on every boot. We read every empty `dtoverlay=` line in that section as one of ours and remove it, including the one the user wrote. The other lines of the user are kept, so the file looks almost the same and the missing line is easy to miss.

The overlay of the user is then still open. The `dtparam` lines below it go to that overlay instead of to the board.

Symptom: a user writes the lines the Raspberry Pi documentation asks for, reboots, and the setting does not work. The line that closed the overlay is gone from the file.

joaoantoniocardoso:
Follow-up to #4482. That change opened the overlay scope before our block. This one closes it after our block, so the section no longer changes the meaning of the lines that come after it.

How each issue is solved:

- Closes #4484 — the block now ends with an empty `dtoverlay=`, so a `dtparam` written below it reaches the board instead of our USB overlay. Both writers do it: the install script at image build, and the startup patch at every boot. The Pi3 script gets the same line, so all boards are treated the same way.
- Closes #4485 — the startup patch no longer claims every empty `dtoverlay=` in the section. A copy the user wrote stays where the user put it. A copy that closes nothing is dropped, which reclaims the copy we wrote on an earlier boot and keeps the file settled after one run.

One smaller fix rides along: the guard that decides whether to open the block now reads a line the way the firmware does, so a directive written with extra spaces is no longer invisible to it.

Tested on three devices: Pi5 bookworm, Pi4 bullseye, Pi4 bookworm. On each one the section converges in one run, a line the user protected survives once, and after a reboot the I2C, SPI and UART devices and the USB gadget are all still there. On both Pi4s the empty `dtoverlay=` the user wrote is removed by master and kept by this branch.

A backport to 1.4 will follow.


core/tools/blueos_startup_update/test_boot_config_order.py
```diff
@@ -133,23 +133,87 @@ def test_board_section_keeps_the_hat_overlay_loadable(distribution: Distribution
...
+    [
+        ("dtoverlay=uart1", True),
+        ("dtoverlay=uart1 # a comment", True),
+        ("dtoverlay=", False),
+        ("dtoverlay= # a comment", False),
+        # the firmware trims the line before it reads the directive
+        ("   dtoverlay=   ", False),
+        ("\tdtoverlay=uart1", True),
+        # the firmware accepts a space around the equal sign
+        ("dtoverlay = uart1", True),
```
Williangalvani, line 204: [this](https://ubuntu.com/hardware/docs/boards/how-to/special_hardware/rpi-config-txt/#edit-the-raspberry-pi-boot-configuration) says that is not supported

core/tools/blueos_startup_update/test_boot_config_order.py
```diff
@@ -133,23 +133,87 @@ def test_board_section_keeps_the_hat_overlay_loadable(distribution: Distribution
...
+    ), "a line written below the board section would go to the last overlay of the board"
+
+
+@pytest.mark.parametrize(
+    "line, overlay_is_open",
+    [
+        ("dtoverlay=uart1", True),
+        ("dtoverlay=uart1 # a comment", True),
+        ("dtoverlay=", False),
+        ("dtoverlay= # a comment", False),
```
Williangalvani: apparently [comments need to be at the start of the file](https://ubuntu.com/hardware/docs/boards/how-to/special_hardware/rpi-config-txt/#edit-the-raspberry-pi-boot-configuration:~:text=Comments%20may%20only%20start%20at%20the%20beginning%20of%20a%20line%2C%20not%20trailing%20a%20property)

core/tools/blueos_startup_update/blueos_startup_update.py
```diff
@@ -220,6 +220,17 @@ def boot_config_line_is_protected(line: str) -> bool:
     return bool(re.match(f"^.*#.*{CONFIG_USER_PROTECTION_WORD}.*$", line, regex_flags))
 
 
+def boot_config_overlay_is_open(lines: List[str]) -> bool:
+    overlay_is_open = False
+    for line in lines:
+        # the firmware trims the line and drops the comment before it reads the directive
+        directive = re.sub(r"\s*#.*$", "", line.strip())
+        overlay = re.match(r"^dtoverlay\s*=\s*(.*)", directive, re.IGNORECASE)
```
Williangalvani: [Spaces are not permitted (or more precisely are not ignored) either side of the “=” in a key=value setting and hence, should not be used](https://ubuntu.com/hardware/docs/boards/how-to/special_hardware/rpi-config-txt/#edit-the-raspberry-pi-boot-configuration:~:text=Spaces%20are%20not%20permitted%20(or%20more%20precisely%20are%20not%20ignored)%20either%20side%20of%20the%20%E2%80%9C%3D%E2%80%9D%20in%20a%20key%3Dvalue%20setting%20and%20hence%2C%20should%20not%20be%20used)

Williangalvani (changes requested):
Found a couple of issues. apparently there's some misinformation online.
I got this answer while trying to find an authoritative source for this info...

Image

joaoantoniocardoso:
Thanks @Williangalvani!

After reading your comments and reviewing the source code, I noticed that our "#custom" feature is partially broken. It works for `dtparam`, but not for `dtoverlay` lines: [issue here](https://github.com/bluerobotics/BlueOS/issues/4488).

I narrowed the code to what they support in practice today, but we should rework it to comply with their specs (probably following what Ubuntu documents), not the Raspberry Pi's current implementation.

PR #4491, opened 2026-09-22
Stop downgrading the Bullseye GPU firmware

### LINKED ISSUE #4490: bug: Bullseye images wont boot on CM4
### Bug description

today's master bullseye image. Bookworm armv7 seens to work fine.

Image

### Steps to reproduce

flash .img into emmc

### Primary pain point(s)

I'd like it to work

### Additional context

_No response_

### Prerequisites

joaoantoniocardoso:
## Summary

The Bullseye image ships `start4.elf` from 20 January 2022. The install script then runs `rpi-update 1340be4`, which replaces that file with the one from 30 April 2021. A current CM4 rejects the older file, so the Bullseye image does not boot. This removes that `rpi-update`. Bookworm never ran it.

Closes #4490.

The last commit builds the Raspberry images for this pull request and must be dropped before merging. It is `266584936d` from August, with the core image tag set to `master` because that is the branch this pull request is based on.

## Test plan

- [ ] The Bullseye image from this pull request boots on the CM4 that shows "This board requires newer software" today
- [ ] The same image still boots on an older Pi 4
- [ ] Drop the `HACK` commit before merging


patrickelectric:
just to be sure.. is this not valid anymore https://github.com/raspberrypi/linux/issues/2981 ?

joaoantoniocardoso:
> just to be sure.. is this not valid anymore [raspberrypi/linux#2981](https://github.com/raspberrypi/linux/issues/2981) ?

Yeah, that's the point: at the time of that commit, that was an _upgrade_ to latest, today, that's a _downgrade_.

